PR #2, opened 2023-10-25
Chore/4391/devtooling

pete-watters:
This PR adds some changes to the mono-repo:
- adds a new `eslint-config` package
- adds a new `tsconfig-config` package
- sets up basic Git actions to run jobs
- sets up `husky` and `commitlint` to enforce standards locally
- creates and updates documentation

I did not publish the new packages to `npm` yet. The next step I see is to either:
- add `Nx` and setup automatic releases
- publish manually for now and move `panda` + design tokens in then the other libs

I found this good example setup I have mostly been taking inspiration from / copying: 
- https://github.com/muravjev/monorepo/
- https://github.com/muravjev/configs/tree/main 

He is also using some other things we could consider:
- [changesets](https://github.com/muravjev/configs/blob/main/docs/tools/changesets.md)
- [syncpack](https://github.com/muravjev/configs/blob/main/packages/syncpack/README.md)


fbwoolf (approved):
Looks like a good start @pete-watters 👍

TODO.md
```diff
@@ -0,0 +1,3 @@
+- add package and setup for dependancy cruiser?
+- lint-filename: https://ls-lint.org/?
+- unused-exports: https://www.npmjs.com/package/ts-unused-exports ?
```
kyranjamie, line 3: Should we just use issues?

package.json
```diff
@@ -1,11 +1,40 @@
...
   },
   "keywords": [],
   "author": "",
-  "license": "ISC"
+  "license": "MIT",
+  "devDependencies": {
+    "@changesets/cli": "^2.26.2",
+    "@commitlint/cli": "^18.0.0",
+    "@commitlint/config-conventional": "^18.0.0",
+    "@leather-wallet/eslint-config": "workspace:^",
```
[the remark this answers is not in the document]
kyranjamie: do we want this or `workspace:*`? Thought I saw that in the nx vids

.github/workflows/code-checks.yml
```diff
@@ -0,0 +1,42 @@
...
+
+  lint-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Lint commit message
+        uses: wagoid/commitlint-github-action@v4
```
kyranjamie, line 42: If naming the same, I'd omit `name` field altogether as it's helpful to see the cmd

kyranjamie (commented):
Looks great @pete-watters, awesome start. 

Do we need to repeat the LICENSE files?

.npmrc
```diff
@@ -0,0 +1,2 @@
+  engine-strict = true
+  auto-install-peers = true
```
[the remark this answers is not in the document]
pete-watters: Not 100% needed - works without so I can remove if you think we don't need to add it? 

I just added it based on a monorepo template I was following and it seemed useful

TODO.md
pete-watters: Yes,  I can delete this and put it in an issue, thanks. I wasn't sure if we want to do this stuff, it's probably more app specific and not for the monorepo config

.github/workflows/code-checks.yml
pete-watters: They should be named differently, I'll fix thanks

package.json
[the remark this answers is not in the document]
pete-watters: You are right, there he declares them as `*`. I can change it but I'm not 100% what's best to use. 

This is the default used by `pnpm`. 

On the `pnpm` docs they give [this example](https://pnpm.io/workspaces):
 
Screenshot 2023-11-01 at 08 23 23

On that [Nx tutorial](https://dev.to/nx/setup-a-monorepo-with-pnpm-workspaces-and-speed-it-up-with-nx-1eem) they use `*` but I amn't sure if it gives us benefit
Screenshot 2023-11-01 at 08 22 58

Any ideas what is best for us? 🤔

pete-watters:
> Looks great @pete-watters, awesome start.
> 
> Do we need to repeat the LICENSE files?

We don't need to repeat them but I thought if we are publishing packages independently it's nice to have them bundled with all they need, including license.

I can easily remove them though to trim things down

PR #12, opened 2023-11-10
feat: Move from pnpm workspace to npm+nx workspace. Add mobile-wallet package

edgarkhanzadian:
As discussed with @pete-watters and @kyranjamie, after having issues with pnpm we decided to move to a npm+nx workspace.

This PR also includes basic initialization of an expo project in mobile-wallet package.

To run mobile app:
1. `npm install` in the root directory
2. `npx nx ios @leather-wallet/mobile-wallet` in the root directory, or `cd projects/mobile-wallet` and then `npm run ios`. You might need to install some prerequisites like XCode on your machine.

To run linters:
`npm run lint-projects:all` to lint all projects in the monorepo (currently only mobile-wallet implements lint, format and typecheck scripts though).


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-pnpm lint-staged --quiet
+npm run lint-projects:affected
```
pete-watters, line 4: 👍

package.json
```diff
@@ -3,38 +3,25 @@
...
-    "install:fresh": "pnpm install:clean && pnpm i",
-    "install:nuke": "rm -rf pnpm-lock.yaml && pnpm install:fresh",
-    "lint:eslint": "eslint .",
-    "lint:eslint:fix": "pnpm lint:eslint --fix",
-    "lint:prettier": "prettier --check \"*.{ts,tsx,js,json}\"",
-    "lint:prettier:fix": "prettier --write \"*.{ts,tsx,js,json}\" *.js",
-    "lint:types": "tsc --noEmit",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "lint-projects:all": "nx run-many -t format,lint,typecheck --parallel=1",
```
pete-watters, line 7: This is running each projects individual linting now right? Hence we don't need to install the linting packages at the top level

package.json
```diff
@@ -3,38 +3,25 @@
...
+    "nx": "^17.1.1"
   },
   "engines": {
     "node": ">=20.3.0",
-    "pnpm": ">=7.10.0",
-    "npm": "please-use-pnpm",
-    "yarn": "please-use-pnpm"
-  },
-  "packageManager": "pnpm@8.7.5"
+    "npm": ">=10"
```
pete-watters, line 25: Maybe we could keep the enforcement here but for `npm` instead of `pnpm`? Not a big deal really

pete-watters:
Nice work @edgarkhanzadian 👍 I'm happy to change to `npm` if this helps Expo. Was there a reason to go with `npm` over `yarn`?

package.json
```diff
@@ -3,38 +3,25 @@
...
+    "prepare": "husky install",
+    "lint-projects:all": "nx run-many -t format,lint,typecheck --parallel=1",
+    "lint-projects:affected": "npx nx affected -t format,lint,typecheck --parallel=1 --base=dev",
+    "commitlint": "commitlint"
   },
   "keywords": [],
   "author": "",
   "license": "MIT",
+  "workspaces": [
+    "packages/*"
```
pete-watters, line 15: I missed this earlier but I think that the mobile app should live under `apps`

So we will have:
- `apps/*`
- `packages/*`

There is different conventions calling things `packages` or `libs`. I was considering renaming them as `libs` as then this [workflow tool ](https://kevinkreuzer.medium.com/how-to-release-npm-libraries-with-nx-461f7671a96f)could work for us

edgarkhanzadian: Great point! Moved to apps/ folder

package.json
edgarkhanzadian: Done!

package.json
edgarkhanzadian: Yeap! We might want to keep packages smaller to have more use out of nx caching in the future (e.g. like splitting up api folder into multiple api folders)

apps/mobile-wallet/app/[...missing].tsx
```diff
@@ -0,0 +1,41 @@
...
+    fontSize: 20,
+    fontWeight: 'bold',
+  },
+  link: {
+    marginTop: 15,
+    paddingVertical: 15,
+  },
+  linkText: {
+    fontSize: 14,
+    color: '#2e78b7',
```
Comment (pete-watters) [inline apps/mobile-wallet/app/[...missing].tsx:39]: Maybe we can have tokens for colours from the start? (not a big deal for now / this PR)

docs/core/MONOREPO.md
```diff
@@ -30,15 +30,18 @@ The purpose of this monorepo is to host various **[npm packages ↗](https://doc
       "install:nuke": "rm -rf pnpm-lock.yaml && pnpm install:fresh"
```
pete-watters, line 30: We need to clear out more of the `pnpm` stuff like this line `rm -rf pnpm-lock.yaml && pnpm install:fresh`. I can help with that

docs/tools/HUSKY.md
```diff
@@ -18,46 +18,19 @@ It runs on staged files via `lint-staged`. As this is a clean project it is quit
...
-    "prepare": "husky install"
-    ...
-  }
-  ```
-
 - Add `pre-commit` hook for linting and formatting indexed files using [lint-staged](./lint-staged.md):
 
   ```sh
-  pnpm husky add .husky/pre-commit 'pnpm lint-staged --quiet'
+  npx husky add .husky/pre-commit 'run-command-on-pre-commit'
```
pete-watters, line 24: Is this working and only running on `staged` files? I set it up for staged only so it's not too slow running on all files

pete-watters (approved):
This LGTM. A couple of things to clean up but I can help with that and no blockers. Nice job!

We need to test auto-deploy of packages on release / revisit the release

docs/tools/HUSKY.md
edgarkhanzadian: I've just changed it to a more general term, like `run-command-on-pre-commit` is any command you want to add to a pre-commit hook.

docs/core/MONOREPO.md
edgarkhanzadian: Whoops! Will get to it now

apps/mobile-wallet/app/[...missing].tsx
Comment (edgarkhanzadian) [reply apps/mobile-wallet/app/[...missing].tsx:39]: I can address in a separate PR, i expect adding the tokens for colors, typography, sizes, etc. might be a bit too much for this PR.

Comment (pete-watters) [reply apps/mobile-wallet/app/[...missing].tsx:39]: makes sense, just mentioning now 👍

docs/tools/HUSKY.md
pete-watters: OK cool. I think the `lint-staged` package is removed now. That helps it only consider staged files. I can check it and re-added if needed, don't worry!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          cache: "npm"
+      - run: npm ci
+      - uses: nrwl/nx-set-shas@v3
+      # This line is needed for nx affected to work when CI is running on a PR
+      - run: git branch --track dev origin/dev
+
+      - run: npm run lint-projects:affected
```
kyranjamie, line 33: Doesn't bundling all the jobs into one means if one fails, you need to open the job to see which one?

kyranjamie: Also, while we might want to only run CI on staged files locally and at commit time? Is this even faster, given that all the CI jobs can be parallelised?

apps/mobile-wallet/app/modal.tsx
```diff
@@ -0,0 +1,36 @@
+import { Platform, StyleSheet } from 'react-native';
+
+import { StatusBar } from 'expo-status-bar';
+
+import EditScreenInfo from '../components/EditScreenInfo';
+import { Text, View } from '../components/Themed';
+
+export default function ModalScreen() {
```
kyranjamie, line 8: As with extension wallet, let's avoid default imports in mobile wallet too

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          fetch-depth: 0
+      # Cache node_modules
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          cache: "npm"
+      - run: npm ci
+      - uses: nrwl/nx-set-shas@v3
+      # This line is needed for nx affected to work when CI is running on a PR
```
pete-watters, line 30: @edgarkhanzadian - I've just realised this causes fails when PR's are merged to `dev`.

I'll try add a restriction to not run on `dev` / `main` but that's not really a good solution. Any ideas here? 

I'm thinking maybe we shouldn't run this job in CI and have individual jobs

Screenshot 2023-11-29 at 07 01 25

PR #20, opened 2023-11-20
feat(crypto): add bitcoin package with a small subset of utils and constants

edgarkhanzadian:
Create a separate package for extension's bitcoin folder: https://github.com/leather-wallet/extension/tree/dev/src/shared/crypto/bitcoin

Use a small subset of [constants](https://github.com/leather-wallet/extension/blob/dev/src/shared/constants.ts) and [utils](https://github.com/leather-wallet/extension/tree/dev/src/shared/utils). Those should be moved to this repo in full in another PR

### DIFF
--- a/packages/bitcoin/src/bip322/bip322-utils.ts
+++ b/packages/bitcoin/src/bip322/bip322-utils.ts
@@ -1,6 +1,3 @@
-// Compatible package that isn't tiny-secp256k1
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
 import ecc from '@bitcoinerlab/secp256k1';
 import { PaymentTypes } from '@btckit/types';
 import { isString } from '@leather-wallet/utils';
@@ -64,8 +61,7 @@ function tapTweakHash(pubKey: Buffer, h: Buffer | undefined): Buffer {
 }
 
 export function tweakSigner(signer: bitcoin.Signer, opts: any = {}): bitcoin.Signer {
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
+  // @ts-expect-error privateKey exists on signer
   let privateKey: Uint8Array | undefined = signer.privateKey!;
   if (!privateKey) {
     throw new Error('Private key is required for tweaking signer!');
--- a/packages/bitcoin/src/bip322/bip322-utils.ts
+++ b/packages/bitcoin/src/bip322/bip322-utils.ts
@@ -1,0 +1,88 @@
+// Compatible package that isn't tiny-secp256k1
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import ecc from '@bitcoinerlab/secp256k1';
+import { PaymentTypes } from '@btckit/types';
+import { isString } from '@leather-wallet/utils';
+import { sha256 } from '@noble/hashes/sha256';
+import { hexToBytes, utf8ToBytes } from '@stacks/common';
+import * as bitcoin from 'bitcoinjs-lib';
+import ECPairFactory from 'ecpair';
+import { encode } from 'varuint-bitcoin';
+
+import { toXOnly } from '../bitcoin.utils';
+
+const bip322MessageTag = 'BIP0322-signed-message';
+
+const ECPair = ECPairFactory(ecc);
+bitcoin.initEccLib(ecc);
+
+export function ecPairFromPrivateKey(key: Uint8Array) {
+  return ECPair.fromPrivateKey(Buffer.from(key));
+}
+
+// See tagged hashes section of BIP-340
+// https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#design
+const messageTagHash = Uint8Array.from([
+  ...sha256(utf8ToBytes(bip322MessageTag)),
+  ...sha256(utf8ToBytes(bip322MessageTag)),
+]);
+
+export function hashBip322Message(message: Uint8Array | string) {
+  return sha256(
+    Uint8Array.from([...messageTagHash, ...(isString(message) ? utf8ToBytes(message) : message)])
+  );
+}
+
+export const bip322TransactionToSignValues = {
+  prevoutHash: hexToBytes('0000000000000000000000000000000000000000000000000000000000000000'),
+  prevoutIndex: 0xffffffff,
+  sequence: 0,
+};
+
+function encodeVarString(b: Buffer) {
+  return Buffer.concat([encode(b.byteLength), b]);
+}
+
+const supportedMessageSigningPaymentTypes: PaymentTypes[] = ['p2wpkh', 'p2tr'];
+
+export function isSupportedMessageSigningPaymentType(paymentType: string) {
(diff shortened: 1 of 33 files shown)


kyranjamie (approved):
Great start @edgarkhanzadian 💯 

Can you describe the role of constants and how we'll use it? What it should / shouldn't contain? If a constant value is only used in one package, where does it belong?

packages/bitcoin/src/bip322/bip322-utils.ts
```diff
@@ -0,0 +1,88 @@
+// Compatible package that isn't tiny-secp256k1
```
pete-watters, line 1: While we are moving these files, it seems like a good time to break them up. We shouldn't have huge `utils.ts` files, functions should be separated into their own files

packages/bitcoin/src/bip322/sign-message-bip322.ts
```diff
@@ -0,0 +1,54 @@
+import * as btc from '@scure/btc-signer';
+import { hexToBytes } from '@stacks/common';
+
+import { hashBip322Message } from './bip322-utils';
+
+// TODO: Complete this fn
```
pete-watters, line 6: Lets add some context when moving stuff like this.

In the `extension` repo, in the history we can see the PR this was added and the ticket number:
Screenshot 2023-11-20 at 09 22 11

When migrating stuff like this I think it's a good idea to update the comments with some of this information so we can keep the context in the new repo

packages/bitcoin/src/p2wpkh-address-gen.ts
```diff
@@ -0,0 +1,68 @@
...
+import { HDKey } from '@scure/bip32';
+import * as btc from '@scure/btc-signer';
+
+import { getBtcSignerLibNetworkConfigByMode } from './bitcoin.network';
+import {
+  BitcoinAccount,
+  deriveAddressIndexZeroFromAccount,
+  getBitcoinCoinTypeIndexByNetwork,
+} from './bitcoin.utils';
+// import { BitcoinNetworkModes } from '@shared/constants';
```
pete-watters, line 11: Lets remove this commented code

packages/bitcoin/src/p2wsh-p2sh-address-gen.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as bitcoin from 'bitcoinjs-lib';
+
+export async function deriveBtcBip49SeedFromMnemonic(mnemonic: string) {
+  return mnemonicToSeed(mnemonic);
+}
+
+export function deriveRootBtcKeychain(seed: Uint8Array) {
+  return HDKey.fromMasterSeed(seed);
+}
+
```
pete-watters, line 16: I think we should delete this code / unused exports. Now we are moving unused things to a new package to not be used more efficiently

packages/bitcoin/src/p2wsh-p2sh-address-gen.spec.ts
```diff
@@ -0,0 +1,182 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
pete-watters, line 1: Maybe we can get rid of some of these `eslint-disable-next-line`  /`@typescript-eslint/ban-ts-comment` ? 

It would be nice to start off the new packages with better quality

packages/constants/index.ts
```diff
@@ -0,0 +1,8 @@
+// Partially imported constants from extension repo
```
pete-watters, line 1: Maybe we need a shared configuration package for things like this? `@leather/config`

packages/utils/README.md
```diff
@@ -0,0 +1 @@
+# @leather-wallet/bitcoin
```
pete-watters, line 1: Can we add some docs as we go?

pete-watters (approved):
Nice work @edgarkhanzadian , I left some suggestions along the way. 

I wonder if we need a `bitcoin` package right away? I thought we would make a `crypto` package with `bitcoin` as a folder under there? 

It's not a big deal but I wonder if soon things will grow a lot if we add everything at the top level `packages`. 

Another suggestion is to rename `constants` as `config` as then we could add more different things there but we can see how things go. 

Once we get this merged and the packages published, lets also remove the code from the `extension` and test it

packages/bitcoin/src/p2wsh-p2sh-address-gen.ts
edgarkhanzadian: This seems to be used in tests tho: https://github.com/leather-wallet/mono/pull/20/files/b6c2a00e5ed088777c633b3e9406ab00c442052a#diff-0a9a7f7a5ac8ac518739538ca3867bb29f32881b8cd82345d3f0b53c5426ace1R118

pete-watters: I'm talking about `deriveBtcPayment` below with the `ts-unused-exports:disable-next-line`

edgarkhanzadian:
@pete-watters thank you a lot for suggestions!

This packages are going to morph into bigger or smaller versions of themselves as we go further into the weeds. I've tried to use the structure mentioned in this issue https://github.com/leather-wallet/extension/issues/4297#issuecomment-1809861163, but as we actually start using these packages we might want to refactor these.

I'd say the same goes for constants vs config stuff, i agree with you as we might want to use that package for more stuff other than constants, let's try to move more code in here and understand how everything should work together. Wdyt?

@kyranjamie thank you!

I was expecting to have there anything that is a constant basically, but probably that is too broad of a term. We might want to separate that package into smaller packages. The same goes for `utils`.

> If a constant value is only used in one package, where does it belong?

Well i'd say it depends on what type of constant that is. If it's some kind of a `HEADER_PADDING = "60px"`, then this constant is very specific to the component it is used in. If it's "IS_FEATURE_FLAG_ENABLED" and it's only used in one file, i'd say we should keep it in the shared package as it might be used in other files / components down the road

packages/bitcoin/src/bip322/sign-message-bip322.ts
edgarkhanzadian: I tried reading through the history but still don't understand much context on that part. @kyranjamie can you, please, help us our on this one? (tagging you because of git blame :D )

edgarkhanzadian:
@pete-watters sounds good! We can do that way

packages/bitcoin/src/bip322/sign-message-bip322.ts
kyranjamie: This comment relates to issues ran into between the two different btc libraries. I understand Pete is just asking you to reference this in the commit msg?

kyranjamie:
In diagram I meant to indicate that @leather/bitcoin depends on @leather/core, a child of sorts.

If anything, I wonder if utils and constants should merge into a single "core" package, though not against trying the separated approach as a first try.

fbwoolf (approved):

packages/bitcoin/src/bip322/sign-message-bip322.ts
pete-watters: Yeah I just think we should add the ticket number in the comment so we can trace in case we ever need to

PR #23, opened 2023-11-24
chore: add token package, fix prettier, trial nx-release

pete-watters:


packages/tokens/src/index.ts
```diff
@@ -0,0 +1,8 @@
+import { breakpoints } from './breakpoints';
+import { colors } from './colors';
+import { keyframes } from './keyframes';
+import { semanticTokens } from './semantic-tokens';
+import { textStyles } from './typography';
+import { tokens } from './tokens';
+
+export { breakpoints, colors,  keyframes, textStyles, tokens, semanticTokens } ;
```
kyranjamie, line 8:
```suggestion
export { breakpoints } from './breakpoints';
export { colors } from './colors';
export { keyframes } from './keyframes';
export { semanticTokens } from './semantic-tokens';
export { textStyles } from './typography';
export { tokens } from './tokens';
```

Can export in one line

packages/tokens/src/tokens.ts
```diff
@@ -0,0 +1,59 @@
...
+  },
+  icons: {
+    'icon.xs': { value: '12px', description: '12px' },
+    'icon.sm': { value: '16px', description: '16px' },
+    'icon.md': { value: '24px', description: '24px' },
+    'icon.lg': { value: '30px', description: '30px' },
+    'icon.xl': { value: '36px', description: '36px' },
+  },
+  transition: { value: 'all 0.2s cubic-bezier(0.23, 1, 0.32, 1)' },
+}
```
kyranjamie, line 59: We can improve typing of these using `as const`. We should do this on all objects

packages/tokens/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+  {
+    "extends": ["@leather-wallet/tsconfig-config/tsconfig.base.json"],
+    "include": ["**/*", ".*.ts"],
+    "exclude": []
+  }
```
kyranjamie, line 5: image

packages/tokens/src/index.ts
pete-watters: Thanks! 🤯

package.json
```diff
@@ -16,15 +16,53 @@
     "apps/*"
   ],
   "devDependencies": {
+    "@babel/core": "^7.14.5",
+    "@babel/preset-react": "^7.14.5",
     "@commitlint/cli": "^18.0.0",
     "@commitlint/config-conventional": "^18.0.0",
+    "@leather-wallet/eslint-config": "workspace:^",
```
pete-watters, line 23: I was testing out `Nx` react library generators to create the `tokens` package but it adds a lot of extras here we don't really want / need. 

I'll rework this and add the package manually which is actually easier than using the generator and then cleaning up afterwards.

We can consider creating our own generator in the future if we need to but I think am updated `README` with Copy+paste style tips would suffice

pete-watters:
I'm going to close this off as per: https://github.com/leather-wallet/mono/pull/23/files#r1407523553 

I'll take the good work done and create a cleaner PR

PR #31, opened 2023-11-30
feat(auth): Add Argon2 to Android and iOS native modules with Swift and Kotlin

edgarkhanzadian:
- Add [argon2kt](https://github.com/lambdapioneer/argon2kt) for Android
- Add [Argon2Swift](https://github.com/tmthecoder/Argon2Swift) for iOS
- Add JS wrapper for both packages at `apps/mobile-wallet/utils/argon2-util.ts` and types for native modules at `apps/mobile-wallet/types/nativeModules.d.ts`


kyranjamie:
Thanks @edgarkhanzadian. Is any of this auto-generated/able to be excluded, or do we just have to bundle it all like this building a RN app?

edgarkhanzadian:
@kyranjamie some files are auto-generated but those are required for building the application. As we are now integrating native modules we also need to have basic obj-c and java configurations for the mobile application in the repo (we can't rely on a stock Expo app anymore).

You can review this PR on a commit by commit basis. First one is a result of running `npx expo prebuild --clean` that generates the base for a custom app build. The commits after that are updating native modules configs and add more native modules' files.

fbwoolf (approved):

PR #40, opened 2024-01-17
fix: component bg hover color, ref #4805

alter-eggo:
Ref [#4805](https://github.com/leather-wallet/extension/issues/4805)

### DIFF
--- a/packages/tokens/src/semantic-tokens.ts
+++ b/packages/tokens/src/semantic-tokens.ts
@@ -55,7 +55,7 @@ export const semanticTokens = {
         value: { base: '{colors.lightModeBrown.4}', _dark: '{colors.darkModeBrown.5}' },
       },
       'component-background-hover': {
-        value: { base: '#9486771A', _dark: '#9486771A' },
+        value: { base: 'rgba(148, 134, 119, 0.1)', _dark: 'rgba(148, 134, 119, 0.1)' },
       },
       'component-background-default': {
         value: { base: '{colors.lightModeBrown.3}', _dark: '{colors.darkModeBrown.3}' },


packages/tokens/src/semantic-tokens.ts
```diff
@@ -55,7 +55,7 @@ export const semanticTokens = {
         value: { base: '{colors.lightModeBrown.4}', _dark: '{colors.darkModeBrown.5}' },
       },
       'component-background-hover': {
-        value: { base: '#9486771A', _dark: '#9486771A' },
+        value: { base: 'rgba(148, 134, 119, 0.1)', _dark: 'rgba(148, 134, 119, 0.1)' },
```
pete-watters, line 58: I know we had `hex` here before but we need to add this as a proper colour to `colors.ts` then include that colour here.

Is this coming from some design? Can we fund out from them what the name of it should be? 

We should add them like the others in `colors`:
```
  lightModeStacks: { value: '#5546FF' },
  darkModeStacks: { value: '#7F80FF' },
  overlay: { value: 'rgba(0,0,0,0.4)' },
```

alter-eggo: hm, yeah, agree here
https://github.com/leather-wallet/extension/issues/4805 link to issue 🤔

pete-watters (approved):
I will approve this to unblock https://github.com/leather-wallet/extension/pull/4755. 

In case you want to merge it please go ahead and we can release a new token package. 

It could be easier to just wait until we have a name for the colour and do it once.

packages/tokens/src/semantic-tokens.ts
pete-watters: We had a call about this and I will fix this. I'll:
- name `rgba(148, 134, 119, 0.1)` as `componentBackgroundHover` in `colors` 
- get rid of the unused `ink` tokens in there
- merge this and publish the new colours package

pete-watters:
@alter-eggo I merged and I will change the name as part of https://github.com/leather-wallet/mono/issues/41 then publish a new package

PR #49, opened 2024-02-06
fix: component bg hover color

alter-eggo:
see title


fbwoolf (approved):

pete-watters:
👍 there are no docs here but we need to manually increment the version in `package.json` for packages for now. 

I'll improve that in the future, for now I can release this

PR #51, opened 2024-02-12
refactor: semantic tokens

fbwoolf:
This PR coordinates semantic tokens for red, blue, yellow, and green. Step one is to release this here > then I will install in the extension and replace the ones I removed here.

Please let me know if there are other changes to make here before I merge/release.

https://github.com/leather-wallet/extension/issues/4831

### DIFF
--- a/packages/tokens/src/semantic-tokens.ts
+++ b/packages/tokens/src/semantic-tokens.ts
@@ -1,6 +1,6 @@
 export const semanticTokens = {
   colors: {
-    // Primitive colours defined as semantic tokens to match Radix
+    // Primitive colors defined as semantic tokens to match Radix
     ink: {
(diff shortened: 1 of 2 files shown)


packages/tokens/src/semantic-tokens.ts
```diff
@@ -1,6 +1,6 @@
 export const semanticTokens = {
   colors: {
-    // Primitive colours defined as semantic tokens to match Radix
+    // Primitive colors defined as semantic tokens to match Radix
```
pete-watters, line 3: 🇺🇸 🇺🇸 🇺🇸

pete-watters (approved):
👍 LGTM

kyranjamie:
Would also be good to remove the redDarkMode/redLightMode stuff too in favour of a linear scale

PR #55, opened 2024-02-15
feat: reset to pnpm workspace

edgarkhanzadian:
- Copy packages added on `dev` branch to this branch.
- Keep pnpm workspaces
- Make Expo work with pnpm
- Use syncpack to 
- - To keep all local package dependencies to `workspace:*`
- - To use only exact versions of packages
- Fix CI, run linters and formatters recursively in each package


kyranjamie:
Thanks @edgarkhanzadian. So does this have all the latest package changes as well? If so LGTM

edgarkhanzadian:
@kyranjamie yeap! Except the argon2 module, but i think i am going to keep it separate for some time until we actually need it. So that we don't deal with building native modules for some time

kyranjamie (approved):
Can we rename the earlier commits of Pete's to correct format seeing as we're starting fresh. 

Do we need to force push this to main?

kyranjamie:
Also, how about calling it `apps/mobile` rather than mobile wallet?

edgarkhanzadian:
You mean so that it conforms to conventional commits?

Sure, no preference there, let's change it to mobile

apps/mobile-wallet/src/components/useClientOnlyValue.web.ts
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+// `useEffect` is not invoked during server rendering, meaning
+// we can use this to determine if we're on the server or not.
```
kyranjamie, line 4: This comes with expo? Kinda odd to see server rendering stuff 🤔

pete-watters:
> Can we rename the earlier commits of Pete's to correct format seeing as we're starting fresh.
> 
> Do we need to force push this to main?

I tried squashing mine just now but I ended up in a rebase mess when I started hitting Edgars new commits. 

I suggest we:
- squash this PR into one commit
- open it to `dev` instead of `main`

Then we can do a proper release from `dev` to `main`.

pete-watters (approved):
Thanks for doing this @edgarkhanzadian . It looks good to me. 

Next step is figuring out the best way to merge this. 

Once its on `dev` we can test out the auto release of npm packages

apps/mobile-wallet/app.json
```diff
@@ -0,0 +1,36 @@
+{
+  "expo": {
+    "name": "mobile-wallet",
+    "slug": "mobile-wallet",
```
fbwoolf, line 4: Should `leather` be in the naming here?

apps/mobile-wallet/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "live-mobile",
```
fbwoolf, line 2: This name should prob be different too?

fbwoolf (approved):
Looks good @edgarkhanzadian, there are just some naming inconsistencies I think we should get right in the json files.

edgarkhanzadian:
I'm afraid merging it to dev might introduce weird bugs and inconsistencies. Maybe we can rename the current dev branch to old_dev and rename this one into dev?

kyranjamie:
It's early days in monorepo land, so suggest we just start clean and force push the changes (making a new branch for interim safety)?

PR #57, opened 2024-02-19
feat: add argon2 module for android and ios and run expo prebuild

edgarkhanzadian:
We shouldn't merge this until we start the work with argon2, as this will add native modules to the expo mobile app.
We should try to keep it as up to date as possible


markmhendrickson:
@edgarkhanzadian is this PR still relevant for merging or best to close?

PR #58, opened 2024-02-19
refactor: semantic token colors

fbwoolf:
This PR updates our semantic-tokens w/ our new color coordination based on this issue:
https://github.com/leather-wallet/extension/issues/4831

Here, I removed `colors` bc we decided to use the values directly in the semantic-tokens based on use-case.

I mentioned on the issue we should coordinate border tokens as well. For now, I updated the ones using our previous `error` and `warning` token colors.

Once this is published, I'll coordinate changes in the extension using the ref'd issue.


pete-watters (commented):
👍 we will need to replace the remaining usages of `ink` in the extension once taking in this new package.

Regarding the mono-repo in general, I can see a failing `build-ios` job for this branch but really that shouldn't even run at all here

kyranjamie (approved):

edgarkhanzadian (changes requested):
Blocking this PR until dev is fixed

edgarkhanzadian:
Rebased to dev and cherry-picked the commit

PR #65, opened 2024-02-23
fix: precommit hooks. Use pnpm filtering instead of lint-staged

### LINKED ISSUE #62: Fix precommit hooks

edgarkhanzadian:
Closes #62
Some issues i had with lint-staged:
- It would use eslintrc located in the root of the monorepo, not the one that is located in each individual package
- We are unable to run typecheck only on staged files with tsconfig, known issue: https://github.com/microsoft/TypeScript/issues/27379

So i'm taking another approach here, run lint, format, typecheck recursively and in parallel only in those packages that have changes compared to the last commit (and on packages that are dependent on packages with changes)


.husky/pre-commit
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-pnpm lint-staged --quiet
+LATEST_HASH=$(git rev-parse --verify HEAD)
+
+# Run commands only in packages with changed files compared to previous commit and on their dependants in parallel
+pnpm -r --filter="...[$LATEST_HASH]" --parallel "/^format$|^lint$|^typecheck$/"
```
kyranjamie, line 7: Nice, this is faster? 

I'm a little skeptical of pre-commit jobs. I've tried using them in the past and it gets annoying having a delay post-commit and I end up --no-verifying relying on CI anyway

kyranjamie (approved):

.husky/pre-commit
edgarkhanzadian: Not a big fan of it too but it's nice to keep the option for now. We can remove it down the road if it turns out too big a of a pain.

running with parallel recursive pnpm is faster right now, might become a bit slower than lint-staged down the road. But on the plus side, we can run tsc, which imho usually gives more info than eslint.

PR #66, opened 2024-02-24
fix: token border colors

fbwoolf:
Fixes/updates some changes needed to remove current hardcoded values in the extension.

### DIFF
--- a/packages/tokens/src/colors.ts
+++ b/packages/tokens/src/colors.ts
@@ -1,0 +1,3 @@
+export const colors = {
+  overlay: { value: 'rgba(0,0,0,0.4)' },
+};
--- a/packages/tokens/src/tokens.ts
+++ b/packages/tokens/src/tokens.ts
@@ -1,8 +1,27 @@
+import { colors } from './colors';
+
 export const tokens = {
   animations: {
-    spin: {
-      value: 'spin 1s linear infinite',
-    },
+    spin: { value: 'spin 1s linear infinite' },
+  },
+  borders: {
+    action: { value: '1px solid {colors.ink.action-primary-default}' },
+    active: { value: '2px solid {colors.ink.border-default}' },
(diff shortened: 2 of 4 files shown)


packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,4 @@
+export const colors = {
+  current: { value: 'currentColor' },
+  overlay: { value: 'rgba(0,0,0,0.4)' },
```
fbwoolf, line 3: I accidentally removed these in my previous update.

packages/tokens/src/tokens.ts
```diff
@@ -1,8 +1,27 @@
+import { colors } from './colors';
+
 export const tokens = {
   animations: {
-    spin: {
-      value: 'spin 1s linear infinite',
-    },
+    spin: { value: 'spin 1s linear infinite' },
+  },
+  borders: {
```
fbwoolf, line 7: Some reorg in this file, but mostly catching some token color updates that were broken for borders with the change from `accent` to `ink`.

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,4 @@
+export const colors = {
+  current: { value: 'currentColor' },
```
kyranjamie, line 2: Saw this before but not a fan tbh. Why not just use currentColor?

kyranjamie (approved):

pete-watters (approved):

packages/tokens/src/colors.ts
pete-watters: I tracked this back to a PR I did [here](https://github.com/leather-wallet/extension/pull/4602/files) and it was already set as `current: { value: 'currentColor' }` then, I'm not sure why it's like that. 

I think we can remove it if `currentColor` will still work as that is used quite a lot in SVGs.

PR #68, opened 2024-02-27
Feat: add design system

edgarkhanzadian:


apps/mobile/src/theme/textVariants.ts
```diff
@@ -0,0 +1,104 @@
...
+  lineHeight: 20,
+};
+const caption02 = {
+  fontFamily: 'DiatypeRegular',
+  fontSize: 13,
+  fontStyle: 'normal',
+  fontWeight: 400,
+  lineHeight: 16,
+};
+
```
kyranjamie, line 89: These values we should be able to pull directly from tokens right?

edgarkhanzadian: yea but currently everything is in rem and web specific (rem is non-existant in react native). I would need to create base values for these, and then we would modify those for web and react native.

edgarkhanzadian: i've got this design system working, now moving it to a separate package

pete-watters: We have another task to fix the typography just created here: https://github.com/leather-wallet/extension/issues/4982 

As part of that lets alter it to make it more React Native friendly and use `px` over `rem` if it means we can share it

kyranjamie: I'm sure there are conversion tool etc we can use where it's just a case of looping props etc. 

Might end up with some funky theme definitions, but worth it if coming from a single source of truth

packages/constants/package.json
```diff
@@ -34,5 +33,6 @@
     "type": "git",
     "url": "git@github.com:leather-wallet/mono.git",
     "directory": "packages/constants"
-  }
+  },
+  "type": "module"
```
kyranjamie, line 37: Why at the bottom? This is a really important prop to see no?

kyranjamie (commented):
Great start @edgarkhanzadian

Comments/questions:
- Can we stick to kebab case naming for consistency 🙏🏼 
- Just to keep the import path cleaner, wouldn't `@leather-wallet/ui/web` be a nicer name?
- Thoughts on having web at root level and native on `/native`? In the same way it's `react` and `react-native`?
- Should the UI tool we use for web/native be in this package? Web doesn't have `Text` or `View`, so here the colocation doesn't really work. Does Panda work if we export `Box` via elsewhere?
- `theme-web/global` has extension specific code that we don't want on website for example, so that should remain in extension only.

packages/constants/package.json
edgarkhanzadian: > Can we stick to kebab case naming for consistency 🙏🏼

Oh, yeap!

> Just to keep the import path cleaner, wouldn't @leather-wallet/ui/web be a nicer name?

+1 

> Thoughts on having web at root level and native on /native? In the same way it's react and react-native?

I would say it's better to keep it as `ui/web` and `ui/native` to be explicit

> Should the UI tool we use for web/native be in this package? Web doesn't have Text or View, so here the colocation doesn't really work. Does Panda work if we export Box via elsewhere?

Well i'd say that we would have instances when some components are very one platform specific anyways, so i'd say we should be probably fine for these basic components. I think having panda + restyle + storybook in one package is helpful as it keeps all building blocks close to each other. How can we approach it another way? 

> `theme-web/global` has extension specific code that we don't want on website for example, so that should remain in extension only.

Sure, later on we can separate it from theme-web and create a separate theme-extension.

> Why at the bottom? This is a really important prop to see no?

Syncpack orders all json keys alphabetically unless we explicitly set the order of that key with `sortFirst` in .syncpackrc. I'll add the type there now

kyranjamie: Thanks for comments. Sounds good 👍🏼 

> Sure, later on we can separate it from theme-web and create a separate theme-extension.

Only other comment is re this. If we have styles/logic that only impact the extension, imo they should live in extension not mono repo. For me the web/native terminology refers to environment, not product.

packages/tsconfig-config/tsconfig.base.json
```diff
@@ -20,8 +20,11 @@
     "resolveJsonModule": true,
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true
-  },
-  "include": ["*"],
-  "exclude": [""]
+    "forceConsistentCasingInFileNames": true,
```
kyranjamie, line 23: Much work to also add `nouncheckedindexedaccess` prop?

packages/ui/src/panda.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineGlobalStyles } from '@pandacss/dev';
+
+import { pandaConfig } from './panda.config.base';
+
+// Use dummy globalCss without any extension specific styles
+const dummyGlobalCss = defineGlobalStyles({});
+
+export default pandaConfig({ globalCss: dummyGlobalCss });
```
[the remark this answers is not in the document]
kyranjamie: I started trying to make a package of the panda preset https://www.npmjs.com/package/@leather-wallet/panda-preset but didn't quite finish as that was when I started dealing with esm problems

packages/tokens/src/semantic-tokens.ts
```diff
@@ -1,119 +1,55 @@
+import { Palette, baseColors, darkColors } from './colors';
+
+function createColorObjForKey<T extends keyof Palette>(key: T) {
+  return { value: { base: baseColors[key], _dark: darkColors[key] } };
+}
```
kyranjamie, line 5: 👍🏼

packages/tokens/src/colors.ts
```diff
@@ -1,3 +1,132 @@
...
+
+  'green.text-primary': '#1A3124',
+  'green.action-primary-default': '#00753A',
+  'green.border': '#B2DFC8',
+  'green.background-secondary': '#99D8B9',
+  'green.background-primary': '#E6F5ED',
+
+  invert: '#34312A',
+  stacks: '#5546FF',
+} as const satisfies Palette;
```
kyranjamie, line 89: Wouldn't it be less code to have one object of keys that defines both light & dark mode?

packages/ui/src/public/html/debug.html
```diff
@@ -0,0 +1,36 @@
...
+            Password<br />
+            <input name="password" type="text" />
+          </label>
+        </p>
+        <button>Decrypt</button>
+      </fieldset>
+    </form>
+    <script src="browser-polyfill.js"></script>
+  </body>
+</html>
```
kyranjamie, line 36: None of the HTML should be taken from the extension, let's remove

packages/ui/src/theme-native/textVariants.ts
```diff
@@ -0,0 +1,104 @@
...
+  fontWeight: 400,
+  lineHeight: 20,
+};
+const caption02 = {
+  fontFamily: 'DiatypeRegular',
+  fontSize: 13,
+  fontStyle: 'normal',
+  fontWeight: 400,
+  lineHeight: 16,
+};
```
kyranjamie, line 88: Can't we build these definitions from the core tokens? This is duplicate?

packages/ui/src/theme-native/textVariants.ts
```diff
@@ -0,0 +1,104 @@
+const heading01 = {
```
kyranjamie, line 1: filename*

packages/ui/tsconfig.web.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../tsconfig-config/tsconfig.base",
+  "compilerOptions": {
+    "module": "es2022",
+    "moduleResolution": "Bundler",
+    "allowJs": false,
+    "checkJs": false,
+    "jsx": "react-native",
```
kyranjamie, line 8: wrong jsx value

tsconfig.json
```diff
@@ -1,3 +1 @@
-{
-  "extends": "../../packages/tsconfig-config/tsconfig.base"
-}
+{}
```
kyranjamie, line 1: Pretty sure we need the root tsconfig for editor support?

packages/ui/tsconfig.native.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../tsconfig-config/tsconfig.base",
+  "compilerOptions": {
+    "module": "es2022",
+    "moduleResolution": "Bundler",
```
kyranjamie, line 5: These props aren't needed because they're defined in the base

packages/tsconfig-config/tsconfig.base.json
```diff
@@ -20,8 +20,11 @@
...
+    "outDir": "./dist",
```
kyranjamie, line 24: Wonder if outDir should be something defined in each package?

kyranjamie (commented):
Great work Edgar.

Can remove the audit report and html. 

This solves the main concern of colocation and quite happy with the result. Perhaps I'll make some changes to the web set up when working on it, but nothing that changes the general approach.

My pref would be not to have the `.web` extension and import the same as shown in [RN docs](https://reactnative.dev/docs/platform-specific-code#native-specific-extensions-ie-sharing-code-with-nodejs-and-web) but that's really only because I think `*.web` is ugly 🙃.

Are you sure removing the root tsconfig doesn't cause problems? Why not import from the base? I needed this to get eslint to work iirc.

kyranjamie:
Isn't it possible for the files to have `*.web.ts` but for them to be bundled and packaged as `@leather-wallet/ui`? No biggie either way

packages/tokens/src/typography.ts
```diff
@@ -1,101 +1,246 @@
...
+  mobile: {
+    firaCode: 'Fira Code',
+    diatype: 'DiatypeRegular',
+    marchePro: 'MarcheSuperPro',
+  },
 };
 
-const diatype = 'Diatype';
+// we might want to split up mobile into ios and android in the future.
+type Platform = 'mobile' | 'extension';
```
kyranjamie, line 15:
```suggestion
type Platform = 'mobile' | 'web';
```

packages/tokens/src/typography.ts
```diff
@@ -1,101 +1,246 @@
...
+  function transformWeight<T extends number>(weight: T) {
+    switch (platform) {
+      case 'web':
+        return weight;
+      case 'mobile':
+        return weight.toString();
+    }
+  }
+
+  const display01 = {
```
fbwoolf, line 49: Should we take on this issue with these updates? https://github.com/leather-wallet/mono/issues/72

packages/tokens/src/semantic-tokens.ts
```diff
@@ -1,119 +1,55 @@
+import { Palette, colorThemes } from './colors';
+
+function createColorObjForKey<T extends keyof Palette>(key: T) {
```
fbwoolf, line 3: Nice improvement here. 👍

packages/ui/src/components/button/button.native.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <View style={{ alignItems: 'center', justifyContent: 'center', flex: 1 }}>
+        <Story />
+      </View>
+    ),
+  ],
+};
+
+export default meta;
+
+// good example of argtypes: https://github.com/storybookjs/react-native/blob/next/examples/expo-example/components/ControlExamples/ControlExample/ControlExample.stories.tsx
```
fbwoolf, line 21: Does this comment need to stay?

packages/ui/src/components/text/text.native.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <View style={{ alignItems: 'center', justifyContent: 'center', flex: 1 }}>
+        <Story />
+      </View>
+    ),
+  ],
+};
+
+export default meta;
+
+// good example of argtypes: https://github.com/storybookjs/react-native/blob/next/examples/expo-example/components/ControlExamples/ControlExample/ControlExample.stories.tsx
```
fbwoolf, line 21: Same comment.

packages/ui/src/theme-web/colors.ts
```diff
@@ -0,0 +1,8 @@
+import { defineTokens } from '@pandacss/dev';
+
+// ts-unused-exports:disable-next-line
+export const colors = defineTokens.colors({
+  // TODO: Move to mono repo
+  current: { value: 'currentColor' },
+  overlay: { value: 'rgba(0,0,0,0.4)' },
```
fbwoolf, line 7: I think these can be removed bc in tokens pkg now?

packages/ui/src/theme-web/tokens.ts
```diff
@@ -0,0 +1,23 @@
+import { tokens as leatherTokens } from '@leather-wallet/tokens';
+import { defineTokens } from '@pandacss/dev';
+
+import { colors } from './colors';
+
+// ts-unused-exports:disable-next-line
+export const tokens = defineTokens({
+  ...leatherTokens,
+  // TODO: Update in mono repo
```
fbwoolf, line 9: I believe these are in the tokens pkg too now?

packages/ui/src/theme-web/colors.ts
pete-watters: We have `overlay` in tokens but not `current`. I don't think that needs to be in `tokens` though?

PR #70, opened 2024-02-27
fix: make expo work by making babel and metro files cjs

edgarkhanzadian:
Currently mobile app fails when we use babel and metro files as esmodules.

For now, use metro and babel as common js files


kyranjamie (approved):
Sorry about that, didn't test.

PR #71, opened 2024-02-27
Refactor: Typescript ESM settings

kyranjamie:
In https://github.com/leather-wallet/mono/pull/67 several packages were migrated to ES Modules. This PR follows up on that ensuring Typescript is configured to work with modules.

[Recommend reading this guide](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c), you need to set `module` and `moduleResolution` to `nodenext`. Note that imports now require `.js` endings.

@edgarkhanzadian `app/mobile` was extending from a duplicate root `tsconfig.json` I've removed. I've set it to `commonjs` as modules breaks it, but maybe you can look into getting it to work when removing, 

```diff
-    "module": "commonjs",
-    "moduleResolution": "node",
```

so it inherits the es module set up from the base config?


edgarkhanzadian:
Added use of `moduleResolution: bundler` with `module: es2022`. It should allow us to not use .js extensions in imports and reject cjs in pure esm modules.
Ref: https://www.typescriptlang.org/tsconfig#moduleResolution
https://www.typescriptlang.org/tsconfig#module

edgarkhanzadian (approved):

PR #73, opened 2024-03-05
chore: add new token sizes

pete-watters:
This adds some new size tokens for the containers branch. 

I had to hack my `pnpm` and `Node` version into package.json so I could even commit the change, maybe we are too strict on versions there?

I was getting this error from the pre-commit:
```
➜  tokens git:(chore/4370/design-system-containers) ✗ git commit src/tokens.ts -m "chore: add new token sizes"
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your pnpm version is incompatible with "/Users/petew/src/monorepo/mono".

Expected version: >=8.15.3
Got: 8.7.5

This is happening because the package's manifest has an engines.pnpm field specified.
To fix this issue, install the required pnpm version globally.

To install the latest version of pnpm, run "pnpm i -g pnpm".
To check your pnpm version, run "pnpm -v".

Your Node version is incompatible with "/Users/petew/src/monorepo/mono".

Expected version: >=20.11.0
Got: v20.9.0

This is happening because the package's manifest has an engines.node field specified.
To fix this issue, install the required Node version.
husky - pre-commit hook exited with code 1 (error)
```


packages/tokens/src/tokens.ts
```diff
@@ -24,14 +24,28 @@ export const tokens = {
     lg: { value: '12px' },
   },
   sizes: {
-    centeredPageFullWidth: { value: '500px' },
```
pete-watters, line 27: Removing `centeredPageFullWidth` is a breaking change so I need to update to `0.0.10` in the containers branch

kyranjamie (approved):

PR #76, opened 2024-03-08
fix(tokens): bundle pkg

kyranjamie:
Already published, not sure on status of auto-publish?

This PR builds the token package, needed for use outside monorepo. [Adds some configs having read this guide.](https://adamcoster.com/blog/pnpm-config)


pete-watters:
I'm not sure what's happening with the CI here. 

Auto publish is working but on `release`. So once this is merged if you manually create a release it will then release all the updated packages

pete-watters:
> I see. Any reason not to release from main like we do on extension?

We do release from `main`. Once this gets merged to `dev` we then create a release of the newest stuff on `dev` and that gets merged to `main` and auto-releases the packages, creating release notes etc. 

We can add improvements if we want to automate it better. Packages should probably auto release once merged to `dev`

packages/bitcoin/package.json
```diff
@@ -4,7 +4,6 @@
   "version": "0.0.1",
   "type": "module",
   "author": "Leather.io contact@leather.io",
-  "type": "module",
```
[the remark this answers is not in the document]
kyranjamie: it was duplicated, see L5

packages/tokens/package.json
```diff
@@ -1,10 +1,12 @@
 {
   "name": "@leather-wallet/tokens",
   "description": "Shared design tokens",
-  "version": "0.0.10",
+  "version": "0.0.13",
   "author": "Leather.io contact@leather.io",
   "scripts": {
-    "typecheck": "tsc --noEmit -p ./tsconfig.json"
+    "build": "tsc",
```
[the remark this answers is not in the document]
kyranjamie: done

fbwoolf (approved):

alter-eggo (approved):

packages/ui/package.json
```diff
@@ -6,13 +6,13 @@
     "build-storybook:web": "storybook build -c ./src/.storybook-web",
     "lint": "eslint .",
     "lint:fix": "pnpm lint --fix",
-    "prepare": "panda codegen -c ./panda.config.ts",
+    "panda": "panda codegen -c ./panda.config.ts",
```
pete-watters, line 9: Good idea renaming this 👍

packages/ui/src/theme-web/typography.ts
```diff
@@ -1,7 +1,7 @@
-import { getExtensionTextVariants } from '@leather-wallet/tokens';
+import { getWebTextVariants } from '@leather-wallet/tokens';
```
pete-watters, line 1: This is going to be a breaking change in the extension so we should release and update it soon

packages/ui/package.json
kyranjamie: Not just rename, `prepare` default runs on `pnpm i` which creates a problem on install as it requires there to be a build first

pete-watters: WOW, OK, good catch then!

PR #78, opened 2024-03-11
feat: move basics from extension

edgarkhanzadian:
Move more constants, utilities and models from extension to monorepo packages.
We need these later on for features that we would move from extension to monorepo.


kyranjamie (commented):
Great work @edgarkhanzadian 

Only comments relate to code I think shouldn't belong here, as it's specific to a given app

fbwoolf:
@kyranjamie it seems like we use`model` and `type` interchangeably in our naming, should we decide on a better convention there so it is more clear why one or the other is being used? I'm not sure I've ever fully understood why one is used over the other?

kyranjamie:
>  So you suggest moving types from models package to queries package?

The `interface`s that represent the API response could well live in a `@leather/queries` package yeah. 

One example of model vs types. Maybe we query from 3 different APIs information about what an inscription is:

```typescript
// In @leather/queries
interface ApiOneInscriptionResp {
  txid: string; index: number; offset: number;
} 
interface ApiTwoInscriptionResp {
  id: string; satPosition: number;
} 
```

Two APIs with different schemas. But in our apps, we want to have a single representation of an inscription that we can map values too, so we have a consistent API no matter the data source.

So we have an inscription _model_

```typescript
interface Inscription {
  txid: string; 
  index: number; 
  offset: number;
}
```

it just happens to be the same as `ApiOneInscriptionResp` but for the sake of separation of concerns, I'd still even declare twice, and we need to "property mapping" for that API.

kyranjamie (approved):
Great work @edgarkhanzadian

PR #79, opened 2024-03-12
feat: move basic bitcoin queries from extension

edgarkhanzadian:
Closes part of leather-wallet/issues#49

Move basic bitcoin queries from extension to monorepo.


packages/bitcoin/src/bitcoin.utils.ts
```diff
@@ -223,3 +224,8 @@ export function getTaprootAddress({ index, keychain, network }: GetTaprootAddres
 
   return payment.address;
 }
+
+export function mnemonicToRootNode(secretKey: string) {
+  const seed = mnemonicToSeedSync(secretKey);
+  return HDKey.fromMasterSeed(seed);
+}
```
kyranjamie, line 231: This is a kind of generic function that's needed for stacks too. Where else might it live?

packages/query/src/leather-query-provider.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+export function useCurrentNetworkState() {
+  const currentNetwork = useLeatherNetwork();
+
+  return useMemo(() => {
+    const isTestnet = currentNetwork.chain.stacks.chainId === ChainID.Testnet;
+    const mode = (isTestnet ? 'testnet' : 'mainnet') as NetworkModes;
+    return { ...currentNetwork, isTestnet, mode };
+  }, [currentNetwork]);
+}
```
kyranjamie, line 48: What if in the wallet, we need to query mainnet and testnet at the same time? I think there are one or two places where we need to do this. Maybe we should have different hooks, or the hooks accept a param? 🤔

packages/query/src/bitcoin/balance/btc-native-segwit-balance.hooks.ts
```diff
@@ -0,0 +1,10 @@
+import { useMemo } from 'react';
+
+import { createBitcoinCryptoCurrencyAssetTypeWrapper } from '../address/address.utils';
+import { useGetBitcoinBalanceByAddress } from './btc-balance.hooks';
+
+// Balance is derived from a single query in address reuse mode
+export function useNativeSegwitBalance(address: string) {
+  const balance = useGetBitcoinBalanceByAddress(address);
+  return useMemo(() => createBitcoinCryptoCurrencyAssetTypeWrapper(balance), [balance]);
+}
```
kyranjamie, line 10: I think a hard rule for this package should be **never exporting wrapped queries**. 

This has caused problems within the extension. This should be a query with a `select`, so we can still access the internals of the query.

packages/query/src/bitcoin/bitcoin-client.ts
```diff
@@ -0,0 +1,292 @@
...
+    block_time: number;
+  };
+  value: number;
+}
+
+export interface UtxoWithDerivationPath extends UtxoResponseItem {
+  derivationPath: string;
+}
+
+interface BestinslotInscription {
```
fbwoolf, line 27: At some point we prob need to think better on where models/types live. Rn we have both a `models` and `types` package, and we have them living in individual packages like this. It will be hard to know where to find them?

fbwoolf: Good question, @kyranjamie what is the intention there vs the models pkg?

kyranjamie: In this example, the types describe the response of the best in slot API, as such believe they should remain co-located by the query.

Types/interfaces are language level constructs. I wouldn't group them all together, in the same way I wouldn't recommend grouping all React hooks together. It's not helpful for developers. If best in slot change their API, why should I have to dig through multiple packages to update them? When would I want to consume the bestinslot API separately from the query?

I don't think it's necessary to have both `models` (domain models) and `types` (Typescript types) packages in the definitions used here. `Money` is an `interface` that represents our domain modelling of money, therefore it should live in models. We don't really use class-based models, but if we did, they'd live there too. 

The types package in it's existing form is a package for external developers to easily add ts typings to their global object. Maybe we should call it something else, but the idea is it's extremely lightweight by nature of it being types-only. 

I've made a [database here documenting our packages, and their responsibilities](https://www.notion.so/trustmachines/ea186ca503cd4820bdc3ceb888930906?v=af67dc8559a94f69aa586465cbab4275&pvs=4). It's very important we get the design of these packages right first time, as changing them will be hard work. Adding packages should be deliberate and well-justified.

packages/query/src/bitcoin/bitcoin-client.ts
```diff
@@ -0,0 +1,292 @@
...
+}
+
+class BestinslotApi {
+  url = 'https://api.bestinslot.xyz/v3';
+  private defaultOptions = {
+    headers: {
+      'x-api-key': `${process.env.BESTINSLOT_API_KEY}`,
+    },
+  };
+  constructor(public configuration: Configuration) {}
```
kyranjamie, line 103: I think we should move away from the pattern of wrapping APIs in big classes. 

It isn't beneficial vs flat functions, and the only reason we do this is because that's how the auto-generated API from Hiro looks.

kyranjamie (approved):

PR #81, opened 2024-03-13
fix: primary action hover in dark mode

fabric-8:
Use non-transparent colour for action-primary-hover in dark mode to fix broken hover states


pete-watters:
Yes, @fabric-8 👏

PR #82, opened 2024-03-15
feat: move stacks queries from extension

edgarkhanzadian:
Move stacks queries from src/app/query of extension to query package of monorepo


packages/query/package.json
```diff
@@ -22,23 +22,32 @@
   },
   "bugs": "https://github.com/leather-wallet/mono/issues",
   "dependencies": {
+    "@fungible-systems/zone-file": "2.0.0",
```
fbwoolf, line 25: If this is being installed just for `parseZoneFile`, it might be we could just add that functionality directly and get rid of this? We appear to only use it in `bns.utils.ts`.

packages/query/src/common/remote-config/remote-config.query.ts
```diff
@@ -0,0 +1,153 @@
+import { createMoney, isUndefined } from '@leather-wallet/utils';
```
fbwoolf, line 1: I believe you will need to update this with the new `runesEnabled` config change.

fbwoolf (approved):

fbwoolf:
@edgarkhanzadian I saw this open still, just a heads up that I'll be making quite a few changes to the Stacks queries in the extension with the new models package updates I made. If this gets merged, I can update them after.

PR #87, opened 2024-03-22
chore: update token package version

pete-watters:
Yesterday, I incremented the `token` package version to `11` based on the previous version.

I then did a release `@leather-wallet/tokens@0.0.11` and published the new version. 

After doing that and going to NPM I see we already release version `0.0.13` 2 weeks ago and my latest release was still labelled as `0.0.13`. 

Screenshot 2024-03-22 at 06 15 54

Creating this manual request now but we need a better way so that the `package.json` version is auto updated in each package OR PR's are blocked without an updated package version

### DIFF
--- a/packages/tokens/package.json
+++ b/packages/tokens/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@leather-wallet/tokens",
   "description": "Shared design tokens",
-  "version": "0.0.11",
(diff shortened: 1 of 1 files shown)


pete-watters:
I'm just merging this as I need to install a new version of the  tokens package

fbwoolf (approved):

PR #90, opened 2024-04-04
fix: more granular non-interactive colors

fabric-8:
Adds `ink.text-non-interactive` and `ink.component-background-non-interactive` for better control over non-interactive element styling. Currently, it can get confusing how to exactly apply non-interactive, especially with buttons that feature a solid fill.

Once merged, `non-interactive` occurences in extension need to be updated to use the more specific color tokens. And once that update occured, the current `non-interactive` can be entirely removed from the monorepo.


fbwoolf (commented):
I think you can remove `non-interactive` in the same work here bc we will swap it out in the extension and not be using it? Also, I think you need to update the pkg version number?

packages/tokens/src/colors.ts
```diff
@@ -55,11 +57,13 @@ export const colorThemes = {
   base: {
     'ink.text-primary': '#12100F',
     'ink.text-subdued': '#948677',
+    'ink.text-non-interactive': '##D8CEC4',
```
kyranjamie, line 60: typo

kyranjamie (approved):

PR #92, opened 2024-04-15
ci: upgrade eslint v9

kyranjamie:
In response to some lint errors in extension relating to `eslint`, and the drive towards esm, I tried to upgrade the monorepo to use their [new flat format](https://eslint.org/docs/latest/use/configure/migration-guide) with v9.

It turns out the [Expo native lint plugin doesn't work with v9 yet though](https://github.com/expo/expo/issues/28144), so this can sit here until ready.


kyranjamie:
Thanks @tigranpetrossian, that'd be awesome. I jumped the gun doing this because v9 had only just come out. 9.x versions have all kind of migration tools etc, so should be easier now

tigranpetrossian:
Closing in favor of https://github.com/leather-io/mono/pull/765

PR #99, opened 2024-04-22
chore: bootstrap releases for path: .

edgarkhanzadian:
Configuring release-please for monorepo packages

We would also need to setup github credentials for it: https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#github-credentials


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,46 @@
...
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v4
+        with:
+          # this assumes that you have created a personal access token (PAT)
+          token: ${{ secrets.GITHUB_RELEASE_PLEASE_TOKEN }}
```
pete-watters, line 19: Did you add this token?

kyranjamie: Use `LEATHER_BOT`

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,46 @@
...
+        uses: pnpm/action-setup@v3
+        with:
+          version: '^8.15'
+        if: ${{ steps.release.outputs.release_created }}
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org/'
+          cache: 'pnpm'
+        if: ${{ steps.release.outputs.release_created }}
```
kyranjamie, line 40: `if` statements in [github actions don't need wrapping in `${{` syntax btw](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif:~:text=matrix%20is%20applied.-,When%20you%20use%20expressions%20in%20an%20if%20conditional%2C%20you%20can%2C%20optionally%2C%20omit%20the%20%24%7B%7B%20%7D%7D,-expression%20syntax%20because)

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install pnpm
+        uses: pnpm/action-setup@v3
+        with:
+          version: '^8.15'
+        if: ${{ steps.release.outputs.release_created }}
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org/'
```
kyranjamie, line 38: Isn't this default? Would remove if it is

.release-please-manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  ".": "1.0.0",
```
kyranjamie, line 2: What does this do? Update root version?

release-please-config.json
```diff
@@ -0,0 +1,96 @@
...
+    },
+    "./packages/utils": {
+      "changelog-path": "CHANGELOG.md",
+      "bump-minor-pre-major": false,
+      "bump-patch-for-minor-pre-major": false,
+      "draft": false,
+      "prerelease": false
+    }
+  },
+  "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json"
```
kyranjamie, line 95: IMO schema belongs at top 👼🏼

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,46 @@
...
+      # The logic below handles the npm publication:
+      - uses: actions/checkout@v4
+        # these if statements ensure that a publication only occurs when
+        # a new release is created:
+        if: ${{ steps.release.outputs.release_created }}
+      - name: Install pnpm
+        uses: pnpm/action-setup@v3
+        with:
+          version: '^8.15'
+        if: ${{ steps.release.outputs.release_created }}
```
kyranjamie, line 34: I reckon good Github action hygiene is to separate the actual publish into a separate job. Then, we get more detailed action reports in PRs. 

One job can pass outputs to another in the same way you can from a step. Examples in extension of this.

kyranjamie (commented):
_Nice_. Bullish on this tool. The auto-release PR is part of changesets I missed, and this is perfect middle ground I think.

How does it handle dependency chains when releasing?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,48 @@
+on:
+  push:
+    branches:
+      - dev
+
+permissions:
+  contents: write
+  pull-requests: write
+
+name: release-please
```
kyranjamie, line 10: Inline with naming convention on extension, proper cased name here

```suggestion
name: Release packages
```

pete-watters:
> Tested the release please github action + deploy. Good news: It works! Bad news, some packages got updated and published to npm 😅 if that's fine, we can just merge this PR. If not, we can try deleting the packages from npm first and then moving back to the previous versions.
> 
> cc @kyranjamie @pete-watters

I think thats actually more great news than good or bad. It has just updated all the package versions and added the changelog. Now we have moved from patch versions to the first minor version.

Let's give Kyran final say on merging as he has some outstanding comments

kyranjamie (approved):
All good with the publishes. Let's merge 👍🏼

PR #100, opened 2024-04-22
feat: add turborepo + watch dev mode

edgarkhanzadian:
- Add turborepo to the repository
- Add watch dev mode to the monorepo. To run it:
- -  `pnpm install`
- - `pnpm dev`


packages/constants/index.ts
```diff
@@ -14,6 +14,14 @@ export const PERSISTENCE_CACHE_TIME = 1000 * 60 * 60 * 12; // 12 hours
 export const BTC_DECIMALS = 8;
 export const STX_DECIMALS = 6;
 
+// Units of `Money` should be declared in their smallest unit. Similar to
```
pete-watters, line 17: Is this necessary in this PR?

pete-watters (approved):
I tested this out and it looks good. I ran it in `dev` mode and made some changes and saw the live reload. 

I guess we have decided to go with `turborepo` instead of `Nx` then? I thought Kyran was working on setting up `Nx`? 

Once we have decided and started using something can we please initiate a release from `dev` to `main`? `main` is very outdated in this repo now

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 dist
 build
 tsconfig.tsbuildinfo
+.turbo/
```
kyranjamie, line 5: missing new line nit

packages/constants/index.ts
kyranjamie: It'd be nicer for this to live with money, but think it's okay to move here, as these are constant values.

kyranjamie (approved):

kyranjamie:
> I guess we have decided to go with turborepo instead of Nx then? I thought Kyran was working on setting up Nx?

I was, but hadn't gotten very far. Tbh I was reluctant turbo tbh, but since @edgarkhanzadian is eager to use it and set it up, it's worth giving it a shot.

pete-watters:
> @pete-watters I was researching both turborepo and nx, and i found out that turborepo is much smaller tool than nx, which would make it easier to understand and learn. For pnpm monorepos, it was just a drop-in replacement with a small json config for it. I discussed this with Kyran yesterday thus decided to move forward with this, but we also should be able to easily change this nx if we want to, it is quite easy to delete the turborepo from the mono.

OK sounds great 👍 I was just double checking. I'm happy to go with whatever is easier now. Thanks 🙇

PR #101, opened 2024-04-23
fix: update disabled submit color, show checkmark on successfull submit

edgarkhanzadian:
- Update disabled submit text color
- Show checkmark on successful submit
- Change button color with animation from loading to success state.

### DIFF
--- a/apps/mobile/src/components/button.tsx
+++ b/apps/mobile/src/components/button.tsx
@@ -1,5 +1,6 @@
-import { ComponentPropsWithoutRef, ReactNode } from 'react';
+import { ComponentPropsWithoutRef, ReactNode, forwardRef } from 'react';
 import { TouchableOpacity as RNTouchableOpacity } from 'react-native';
+import Animated from 'react-native-reanimated';
 
 import { Text, Theme, TouchableOpacity } from '@leather-wallet/ui/native';
 import {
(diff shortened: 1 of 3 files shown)


apps/mobile/src/components/welcome-screen/index.tsx
```diff
@@ -146,6 +147,16 @@ export function WelcomeScreen() {
...
 
+  // Show checkmark for 800ms before changing to the next screen state.
+  useEffect(() => {
+    if (isSubmitted) {
+      setShowSuccess(true);
+      setTimeout(() => {
+        setShowSuccess(false);
+      }, 800);
+    }
+  }, [isSubmitted]);
```
kyranjamie, line 158:
```suggestion
  useEffect(() => {
    if (isSubmitted) {
      setShowSuccess(true);
      const timeout = setTimeout(() => {
        setShowSuccess(false);
      }, 800);
      return () => clearTimeout(timeout);
    }
  }, [isSubmitted]);
```

kyranjamie: probs not necessary as it's a single page, but good practice to avoid the react runtime error

apps/mobile/src/components/button.tsx
```diff
@@ -46,75 +47,83 @@ function whenButtonState<T>(buttonState: ButtonState, match: Record<ButtonState,
...
-  const borderWidth = whenButtonState<number | undefined>(buttonState, {
-    default: undefined,
-    disabled: undefined,
-    success: undefined,
-    outline: 1,
-  });
+    const textVariant: VariantProps<Theme, 'textVariants'>['variant'] = 'label02';
 
-  const textVariant: VariantProps<Theme, 'textVariants'>['variant'] = 'label02';
+    const hasAGap = !!Icon && !!title;
```
fbwoolf, line 104: I think this could just be `hasGap`?

fbwoolf (approved):

PR #106, opened 2024-04-24
Feat: mobile store

edgarkhanzadian:
- Add protected store with zustand and Expo SecureStore (with FaceID authentication)
- Add persisted store with zustand and Expo SecureStore (without requiring authentication)
- Add temporary store with zustand for client side non persistent data, currently a placeholder.

Decided to separate those in 3 separate stores as these mostly don't depend on each other and it's easier to apply middleware on separate stores, rather than on slices.

NOTE: Had to patch the zustand v4 package with `pnpm patch` because metro doesn't properly support ESM yet. We are going to remove the patch as soon as metro supports ESM properly or zustand changes the package.json to support React Native properly. (which probably should happen in zustand v5)


apps/mobile/src/state/protected.ts
```diff
@@ -0,0 +1,102 @@
...
+  setHasHydrated(state: boolean): void;
+}
+
+export type ProtectedStore = ProtectedState & ProtectedAction;
+
+const INITIAL_STATE = {
+  mnemonics: {
+    ids: [],
+    entities: {},
+  } as Mnemonics,
```
kyranjamie, line 35: satisfies?

apps/mobile/src/state/temporary.ts
```diff
@@ -0,0 +1,9 @@
+import { StateCreator } from 'zustand';
+
+interface TemporaryState {}
+
+interface TemporaryAction {}
+
+export type TemporaryStore = TemporaryState & TemporaryAction;
+// Placeholder for client-side non-persistent data.
+export const createTemporaryStore: StateCreator<TemporaryStore> = () => ({});
```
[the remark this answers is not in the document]
fbwoolf: Agree here, don't love `temporary` for naming.

fbwoolf (approved):
Looks like a good start @edgarkhanzadian after addressing Kyran's comments. 👍

kyranjamie:
> Decided to separate those in 3 separate stores as these mostly don't depend on each other and it's easier to apply middleware on separate stores, rather than on slices.

Would love to see in this PR how we can use devtools. Doesn't making 3 separate ones make this harder?

apps/mobile/src/state/README.md
```diff
@@ -0,0 +1,9 @@
+# App state (client-side state)
+
+We keep the app state using `zustand`.
+
+We have 3 types of store:
```
pete-watters, line 5: Upon reflection, I think it could be a good idea to always protect all data in the store and require face ID to open the app, even if only to give users a more secure feeling. We could then also ask for it again when performing key interactions. 

Maybe that would simplify the implementation and we could just have a single protected store for all data? At least for persisted and protected. I'm not sure if temporary data even needs to be in a store, could that be inside application state instead?

I see you mention it's easier to have different middleware for different stores than slices so maybe that's not possible?

Do we have a fallback option where faceID is not supported? e.g. fingerprint / pin?

apps/mobile/src/state/persisted.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+interface PersistedAction {
+  setTheme(theme: string): void;
+  setHasHydrated(state: boolean): void;
+}
+
+export type PersistedStore = PersistedState & PersistedAction;
+
+export const createPersistedStore: StateCreator<
```
pete-watters, line 19: Excuse my ignorance but where do we persist this data? To `AsyncStorage`?

I think it may not be a bad idea to just encrypt everything to protected storage

pete-watters:
Great work on this! I added some thoughts but I trust your judgement

pete-watters (approved):

kyranjamie (changes requested):
This is a great start @edgarkhanzadian, requesting changes only because I believe we should continue discussions before merging. Changing persisted state later is a huge headache, and requires dealing with data migrations etc.

The current solution mixes together the concepts of mnemonics and xpubs which I think would be a mistake. I've thought a lot on this making state decisions in extension, and there's a lot to be learnt from Ledger IMO.

Conceptually, we're dealing with keys and accounts. A software mnemonic, and a Ledger device, are effectively the same data exposed differently. In software we have the whole key and can make near infinite accounts. Ledger devices hide the key, and expose account-level xpubs.

I think it'd be a smart to handle software accounts with the limitations of Ledger devices. This doesn't have any practical implications as far as what we can do with software wallets, but it allows us to manage the state in a uniform way.

_Wallets_ represent the top level keys or devices. _Accounts_ are maintained separately, and reference their respective wallet [by public key fingerprint](https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/doc/bitcoin.md#:~:text=05-,GET_MASTER_FINGERPRINT,-Return%20the%20fingerprint).

Using a derivation path based naming, like Ledger policies, allows us to keep a flat data structure. Otherwise we'd need to nest or divide data by, accounts, networks, keys etc.

This might look something like:
```json
// Suggestion 0
{
  "wallets": {
    "ids": ["publicKeyFingerprintA", "publicKeyFingerprintB"],
    "entities": {
      "publicKeyFingerprintA": {
        "type": "software"
      },
      "publicKeyFingerprintB": {
        "type": "ledger"      
      },
    }
  },
  "accounts": {
    "ids": [
      "[publicKeyFingerprintA]/0'/0", 
      "[publicKeyFingerprintA]/0'/1",
      "[publicKeyFingerprintB]/0'/0"
    ],
    "entities": {
      "[publicKeyFingerprintA]/0'/0": {},
      "[publicKeyFingerprintA]/0'/1": {},
      "[publicKeyFingerprintB]/0'/0": {},
    }
  }
}
```

Curious to hear everyone's thoughts on this.

pete-watters:
> This is a great start @edgarkhanzadian, requesting changes only because I believe we should continue discussions before merging. Changing persisted state later is a huge headache, and requires dealing with data migrations etc.
> 
> The current solution mixes together the concepts of mnemonics and xpubs which I think would be a mistake. I've thought a lot on this making state decisions in extension, and there's a lot to be learnt from Ledger IMO.
> 
> Conceptually, we're dealing with keys and accounts. A software mnemonic, and a Ledger device, are effectively the same data exposed differently. In software we have the whole key and can make near infinite accounts. Ledger devices hide the key, and expose account-level xpubs.
> 
> I think it'd be a smart to handle software accounts with the limitations of Ledger devices. This doesn't have any practical implications as far as what we can do with software wallets, but it allows us to manage the state in a uniform way.
> 
> _Wallets_ represent the top level keys or devices. _Accounts_ are maintained separately, and reference their respective wallet [by public key fingerprint](https://github.com/LedgerHQ/app-bitcoin-new/blob/develop/doc/bitcoin.md#:~:text=05-,GET_MASTER_FINGERPRINT,-Return%20the%20fingerprint).
> 
> Using a derivation path based naming, like Ledger policies, allows us to keep a flat data structure. Otherwise we'd need to nest or divide data by, accounts, networks, keys etc.
> 
> This might look something like:
> 
> ```json
> // Suggestion 0
> {
>   "wallets": {
>     "ids": ["publicKeyFingerprintA", "publicKeyFingerprintB"],
>     "entities": {
>       "publicKeyFingerprintA": {
>         "type": "software"
>       },
>       "publicKeyFingerprintB": {
>         "type": "ledger"      
>       },
>     }
>   },
>   "accounts": {
>     "ids": [
>       "[publicKeyFingerprintA]/0'/0", 
>       "[publicKeyFingerprintA]/0'/1",
>       "[publicKeyFingerprintB]/0'/0"
>     ],
>     "entities": {
>       "[publicKeyFingerprintA]/0'/0": {},
>       "[publicKeyFingerprintA]/0'/1": {},
>       "[publicKeyFingerprintB]/0'/0": {},
>     }
>   }
> }
> ```
> 
> Curious to hear everyone's thoughts on this.

I think it's good to leverage your experience with the extension and also if we handle things according to the limitations of ledger that will result in a more uniform approach with little implementation differences. Wallets will just have a different type

apps/mobile/src/state/entityAdapter/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    return removeOneById(state, id);
+  };
+
+  const removeOneById = (state: State, id: EntityId): State => {
+    return produce(state, draftState => {
+      delete draftState.entities[id];
+      draftState.ids.filter(v => v !== id);
+    });
+  };
```
kyranjamie, line 103: Elsewhere we're using function definitions

apps/mobile/src/queries/README.md
```diff
@@ -0,0 +1,55 @@
...
+2. Treat the query key like a dependency array. queryFn should receive same arguments as a queryKey.
+3. Use selectors to transform the data before usage. Example:
+
+```jsx
+export const useTodosQuery = select =>
+  useQuery({
+    queryKey: ['todos'],
+    queryFn: fetchTodos,
+    select,
+  });
```
kyranjamie, line 17:
```suggestion
export function useTodosQuery (select) {
  return useQuery({
    queryKey: ['todos'],
    queryFn: fetchTodos,
    select,
  });
}
```
consistency nit

apps/mobile/src/state/persisted.ts
```diff
@@ -0,0 +1,77 @@
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import merge from 'lodash.merge';
+import { StateCreator } from 'zustand';
+import { createJSONStorage, persist } from 'zustand/middleware';
+
+import { EntityAdapter, EntitySetter, createEntity } from './entityAdapter';
+import { PERSISTED_KEY, filterObjectKeys } from './utils';
+
+interface Accounts {
```
kyranjamie, line 9: This represents a single account, so better singular `Account`?

apps/mobile/src/state/protected.ts
```diff
@@ -0,0 +1,87 @@
...
+
+interface ProtectedAction {
+  setHasHydrated(state: boolean): void;
+}
+
+export type ProtectedStore = ProtectedState & ProtectedAction;
+
+const NOT_PERSIST_KEYS = ['_hasHydrated'];
+
+export const createProtectedStore: StateCreator<
```
kyranjamie, line 36: On naming, maybe secure store is a better name, more aligned with the storage medium?

kyranjamie (approved):
Great job @edgarkhanzadian 

Some nit picky comments, otherwise looks good. I reckon there are more improvements we can probably make to the organisation/state tree, but think it's better to work with it for a while, and reevaluate before launching.

One concern I'd like to figure out before then, for example, is how we mange migrations. ReduxPersist is outdated/unmaintained, but the built in migration utilities are super helpful.

apps/mobile/src/state/entityAdapter/index.ts
```diff
@@ -0,0 +1,126 @@
+import { Draft, produce } from 'immer';
```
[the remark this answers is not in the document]
kyranjamie: Can you look into why the linting tool didn't catch this? Pretty sure it should have 🤔

PR #107, opened 2024-04-25
fix: add missing 'hex' field in signPsbt response

victorkirov:
The hex field was missing from the response and requires a hack to get the typings correct.

Also, I'm pretty sure the txid field is only returned if broadcast is set to true, so made it nullable.


kyranjamie (approved):
Thanks @victorkirov !

fbwoolf:
I am working on psbt type changes on another branch, so can add these changes there?

fbwoolf:
I cherry-picked your commit and am adding it here: https://github.com/leather-wallet/mono/pull/108

PR #108, opened 2024-04-29
fix: reorg pkgs and refactor asset data models

fbwoolf:
This PR has become a bit larger effort in that it has refactored our package organization and started to look at how we can refactor our asset data models.

Suggestion here is to simplify our `CryptoAsset...` data models to follow what appears to be a fairly standard pattern in working with token protocol apis. As with Stacks, there is typically an endpoint for `balances` and one for `tokenInfo` or `tickerInfo`. Proposing here a `...Balance` and `...Info` model that remain separate as exported from our data models.

I ref'd the api response models for the assets we support in making these changes, I think we should keep the `snaked_cased` models separate (located with the queries) and build new models to use with only the `camelCased` props we need (located in the models pkg).


fbwoolf:
I might have messed up the types pkg here, so let me know via CR.

packages/types/src/network.ts
```diff
@@ -0,0 +1,162 @@
+import { Blockchains } from './blockchain';
+import {
+  BITCOIN_API_BASE_URL_MAINNET,
+  BITCOIN_API_BASE_URL_SIGNET,
+  BITCOIN_API_BASE_URL_TESTNET,
+  HIRO_API_BASE_URL_MAINNET,
+  HIRO_API_BASE_URL_TESTNET,
+} from './constants';
```
kyranjamie, line 8: I'm not sure types should import from constants.

Can't we put only the type values in here, and then import them in constants with the strings? So we export the networks from there?

kyranjamie:
@edgarkhanzadian changelog additions shouldn't happen during PRs I think. These should be added with the release PR right?

packages/types/src/network.ts
fbwoolf: Trying to understand this comment??? 🤔

fbwoolf:
> It seems to be an issue with LiteralUnion that models package imports from types package. As that is a very basic util that never changes, i would suggest just copying that interface to models package and delete the @leather-wallet/types dependency from @leather-wallet/models package.json. That should fix the issue of circular dependencies.

Not totally understanding this feedback, there isn't a circular dep error now?

kyranjamie:
This is super important to get right, and we should be aligned (and very strict) on these core package dependencies.

I'm currently seeing it this way, where types if the core most package with no external dependencies.

image

@fbwoolf @edgarkhanzadian thoughts on this? How might this alternative work?

fbwoolf:
> @fbwoolf @edgarkhanzadian thoughts on this? How might this alternative work?

Agree with this, but still not clear on your api url constants comment? We need those in types pkg for the networks, can you clarify what you meant there?

packages/types/src/network.ts
kyranjamie: See latest diagram, but this puts constant values in our types package, which I think we should avoid

kyranjamie:
To be honest, I think I'm not a fan of our `constants` package at all. It's too vague, just a place to go put anything without much thought because it's a `const` value? Why not have a `@leather-wallet/functions` package while we're at it /sarcasm.

I think having both `types` and `models` might be a bit of a confusing paradigm, so open to rethinking that. What if `/models` is our core most package, that anything can depend on, and types becomes `@leather-wallet/rpc` package.

image


> @leather-wallet/types is a package that is used by itself by external developers, it shouldn't contain anything that is related to our code.

but our APIs return data to external devs data that we combine in our code, so there has to be some relationship somewhere

fbwoolf:
Renaming to `rpc` sounds good to me, I think I would then call the models pkg `@leather-wallet/types` bc our data models are used as types with TypeScript, right? Also, won't apps want our network type exposed if used in the rpc pkg?

kyranjamie:
> So you suggest merging constants and models package together? 

No, would still keep models a very stripped down, types/interfaces/type-creator function purpose only. e.g. I think `createMoney` could belong in `models`, but nothing else really.

I dislike constants but maybe best to keep it for now. But it's really just network properties and a few other helpers that I reckon could live elsewhere.

fbwoolf:
Thx for feedback, will try to make these changes here so we can evaluate them.

.dependency-cruiser.cjs
```diff
@@ -1,17 +1,17 @@
...
-      comment: `@leather-wallet/types should not have any dependencies. It should be a standalone package.`,
-      from: {
-        path: '^packages/types',
-      },
-      to: {
-        path: '^(?!packages/types)',
-      },
-    },
+    // {
+    //   name: 'no-deps-for-models',
```
fbwoolf, line 5: Not sure we need this now?

packages/bitcoin/src/bitcoin-signer.ts
```diff
@@ -1,5 +1,5 @@
 import type { SignatureHash } from '@btckit/types';
```
fbwoolf, line 1: It seems like we can uninstall this and replace?

packages/models/src/network.model.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export const HIRO_API_BASE_URL_MAINNET = 'https://api.hiro.so';
+export const HIRO_API_BASE_URL_TESTNET = 'https://api.testnet.hiro.so';
+export const HIRO_INSCRIPTIONS_API_URL = 'https://api.hiro.so/ordinals/v1/inscriptions';
+
+export const BITCOIN_API_BASE_URL_MAINNET = 'https://blockstream.info/api';
+export const BITCOIN_API_BASE_URL_TESTNET = 'https://blockstream.info/testnet/api';
+export const BITCOIN_API_BASE_URL_SIGNET = 'https://mempool.space/signet/api';
+
+// Copied from @stacks/transactions to avoid dependencies
```
fbwoolf, line 11: Can remove if we are ok with dependencies in this pkg?

.dependency-cruiser.cjs
kyranjamie: I think models should be our core most package that has no deps

fbwoolf: This will be tough only bc we use types from Stacks in our models, guess the best thing is to then just copy them directly into our code?

fbwoolf: The other issue here is our use of `bignumber.js` in this package, I'm not sure how to handle that if any ideas?

kyranjamie: > I think it should have external dependencies

@edgarkhanzadian Which external deps might it need?

fbwoolf: @kyranjamie can you review the latest commit here at some point? I did only leave `bignumber.js` as a dependency and using only as a type. Proposing asset data model changes in last commit too, so hoping to get your thoughts.

fbwoolf: I updated the PR description bc this PR has expanded in scope.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export interface BtcCryptoAssetInfo extends CryptoAssetInfo {
+  name: 'stacks';
+  symbol: 'BTC';
+}
+
+export interface StxCryptoAssetInfo extends CryptoAssetInfo {
+  name: 'bitcoin';
+  symbol: 'STX';
+}
```
kyranjamie, line 16: wrong way around 🙃

packages/models/src/crypto-assets/crypto-asset.model.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface RuneCryptoAsset {
+  readonly blockchain: 'bitcoin';
+  readonly type: 'runes';
+  readonly asset: CryptoAssetBalance & RuneCryptoAssetInfo;
+}
+
+export interface InscriptionCryptoAsset {
+  readonly blockchain: 'bitcoin';
+  readonly type: 'inscriptions';
```
kyranjamie, line 62: As this represents a single item, shouldn't it be singular?

packages/models/src/crypto-assets/crypto-asset-balance.model.ts
```diff
@@ -0,0 +1,49 @@
...
+  readonly marketData: MarketData;
+}
+
+export interface BtcCryptoAssetBalance extends CryptoAssetBalance {
+  readonly balance: BtcBalance;
+}
+
+export interface StxCryptoAssetBalance extends CryptoAssetBalance {
+  readonly balance: StxBalance;
+}
```
kyranjamie, line 44: This creates a kind of nesting that seems odd to me. 

Separately, as before, in `CryptoAssetBalance` we're pairing models for a user's balance, and market data together. What's the rationale for creating this paring?

packages/models/src/crypto-assets/crypto-asset.model.ts
```diff
@@ -0,0 +1,99 @@
...
+export interface Brc20CryptoAsset {
+  readonly blockchain: 'bitcoin';
+  readonly type: 'brc-20';
+  readonly asset: CryptoAssetBalance & Brc20CryptoAssetInfo;
+}
+
+export interface RuneCryptoAsset {
+  readonly blockchain: 'bitcoin';
+  readonly type: 'runes';
+  readonly asset: CryptoAssetBalance & RuneCryptoAssetInfo;
```
kyranjamie, line 57: Similar to balance and market data, this pairs a single user's balance details together with the larger concept of the asset itself.

I find it helpful to think about building a database if Leather were a centralised app. We'd have one table for all the tokens supported. Another for user balances. The user balance table might reference the token for which it is a balance of. But you wouldn't put all token info and user balance data in the same table.

Putting these together, it's saying "in all possible cases, we'll pair a user's balance with the asset info". And that isn't necessarily the case. 

Consider in a signed out state we show all the tokens we support in the UI. We want to iterate though a list of known supported tokens, ideally as these types **but** oh this type expects us to have user balance, but there is no user. So, someone goes and creates dummy data like `const balance = createMoney(0, 'BTC') etc

kyranjamie (changes requested):
Afaict, these models couple together concerns that ought to be separate.

packages/models/src/crypto-assets/crypto-asset.model.ts
fbwoolf: The protocol name is typically plural, so wasn't sure here.

packages/models/src/crypto-assets/crypto-asset-balance.model.ts
fbwoolf: I keep getting confused on your feedback, I swear you asked me to make sure the price data was available and accessible thru the data model when we query it? So, we always `select` the query and make the model with the price data in a hook?

packages/models/src/crypto-assets/crypto-asset.model.ts
fbwoolf: I don't think what you are describing here is accurate? 🤔 Yes, that is true if you use this model where they are coupled together, but they will also be available separately with their individual queries? I was thinking this would be used just to make using them together easier, but we can def just keep them separate. Is that your pref?

fbwoolf: I understand this now after a dm exchange, if coupling happens it should be done at the presentation layer not in our models.

packages/models/src/crypto-assets/crypto-asset-balance.model.ts
fbwoolf: Will refactor and separate.

fbwoolf:
@kyranjamie maybe check the changes here when you can, I removed the coupled `CryptoAssets` and just left the two data models separate for `Balance` and `Info`. I added an `assetId` to both so that we could create some kind of relationship there, but it's hard to use one naming like `symbol` or `name` bc each protocol uses them differently, thoughts on keeping this? I also added an `address` to the balance model bc noticing we add that sometimes later ...`addressIndex` might be helpful too?

packages/models/src/crypto-assets/crypto-asset-balance.model.ts
```diff
@@ -0,0 +1,41 @@
...
+ * protectedBalance - balance of UTXOs with collectibles (if bitcoin)
+ * totalBalance - balance as confirmed on chain
+ * uneconomicalBalance - balance across UTXOs with need for larger fee than principal by UTXO given standard rate (if bitcoin)
+ * unlockedBalance - totalBalance minus lockedBalance
+ */
+
+export interface CryptoAssetBalance {
+  readonly address: string;
+  readonly addressIndex: number;
+  readonly assetId: string; // Name or symbol
```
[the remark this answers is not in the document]
fbwoolf: Yep, will do, comment was left by mistake bc I'm not sure the `assetId` is necessary? I was just thinking it would be a good identifier to have with the balance to understand what asset info it relates to.

packages/models/src/crypto-assets/crypto-asset-balance.model.ts
```diff
@@ -0,0 +1,41 @@
...
+ * pendingBalance - totalBalance plus inboundBalance minus outboundBalance
+ * protectedBalance - balance of UTXOs with collectibles (if bitcoin)
+ * totalBalance - balance as confirmed on chain
+ * uneconomicalBalance - balance across UTXOs with need for larger fee than principal by UTXO given standard rate (if bitcoin)
+ * unlockedBalance - totalBalance minus lockedBalance
+ */
+
+export interface CryptoAssetBalance {
+  readonly address: string;
+  readonly addressIndex: number;
```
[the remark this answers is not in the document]
fbwoolf: This appears to be added only for inscriptions rn, but I'm not nec seeing that it is used/needed later in the code ...so I might remove these for now ...think I will remove the assetId too bc I'm not sure if it is really nec for sure.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export interface StxCryptoAssetInfo extends CryptoAssetInfo {
+  readonly name: 'stacks';
+  readonly symbol: 'STX';
+}
+
+/**
+ * Bitcoin
+ */
```
kyranjamie, line 20: This comment now shows only for `Brc20CryptoAssetInfo`. 

If it's a "section" comment, better to use a `//` comment

kyranjamie (approved):
LGTM, nice work @fbwoolf

PR #112, opened 2024-05-01
feat: Prepare monorepo packages for extension development

edgarkhanzadian:
Spent a good chuck of time trying to link @leather-wallet/query package to extension repo for development.

Notes from the work:

Linking with just pnpm link doesn't work. That way, extension throws this runtime error:

```
Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

Well, after reviewing [react docs on this error](https://react.dev/warnings/invalid-hook-call-warning#duplicate-react), seems like @leather-wallet/query and extension use different types of React. One solution is to link @leather-wallet/query to the react version used in extension. For that:
```sh
pnpm link /Users/<username>/<path-to-extension>/extension/node_modules/.pnpm/react@18.2.0/node_modules/react
```
aghh. Doesn't work again. After some time reviewing this, seems like @leather-wallet/query now uses the correct, extension's, version of react. However, @tanstack/react-query uses monorepo's version of react. So to fix it, we now run (same command) in `packages/query/node_modules/@tanstack/react-query`
```
pnpm link /Users/<username>/<path-to-extension>/extension/node_modules/.pnpm/react@18.2.0/node_modules/react
```

Welp, now it works. We really need to find an easier way to connect @leather-wallet/query to extension for development.

Anyway, working steps for connecting @leather-wallet/query to extension.

1. `pnpm i` in extension
2. `pnpm i` in mono
3. find a relative path to extension’s react package, for me it is: `/Users/<username>/<path-to-extension>/extension/node_modules/.pnpm/react@18.2.0/node_modules/react`
4. cd `packages/query`
5. `pnpm link —global` to setup @leather-wallet/query as a linkable package.
6. `pnpm link /Users/<username>/<path-to-extension>/extension/node_modules/.pnpm/react@18.2.0/node_modules/react`
7. Go to tanstack package, `cd node_modules/@tanstack/react-query`
8. `pnpm link /Users/<username>/<path-to-extension>/extension/node_modules/.pnpm/react@18.2.0/node_modules/react`
9. Now go to extension repo
10. `pnpm link —global @leather-wallet/query`
11. Et voilà, with _*just*_ 10 easy steps we were able to set it up

### DIFF
--- a/packages/query/package.json
+++ b/packages/query/package.json
@@ -40,7 +40,6 @@
     "@tanstack/react-query-persist-client": "4.35.7",
     "axios": "1.6.7",
     "bignumber.js": "9.1.2",
-    "react": "18.2.0",
     "yup": "1.3.3"
   },
   "devDependencies": {
@@ -57,11 +56,20 @@
     "typescript": "5.3.3",
     "vitest": "0.34.6"
   },
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.js"
+    }
+  },
   "keywords": [
     "leather",
     "leather wallet",
     "query"
   ],
+  "peerDependencies": {
+    "react": "*"
+  },
   "prettier": "@leather-wallet/prettier-config",
   "publishConfig": {
     "access": "public"
--- a/packages/bitcoin/package.json
+++ b/packages/bitcoin/package.json
@@ -51,6 +51,12 @@
     "typescript": "5.3.3",
     "vitest": "0.34.6"
   },
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.js"
+    }
+  },
   "keywords": [
     "bitcoin",
     "leather",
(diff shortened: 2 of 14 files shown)


packages/query/package.json
```diff
@@ -40,7 +40,6 @@
     "@tanstack/react-query-persist-client": "4.35.7",
     "axios": "1.6.7",
     "bignumber.js": "9.1.2",
-    "react": "18.2.0",
```
pete-watters, line 43: Glad you got it working 👍 Maybe we can standardise the react version across everywhere we need it? 

I'm not sure why the extension would have a different version than other packages.

We could enforce react version and node version just to make life easier?

fbwoolf: Was think we could just update react-query to current version, but doesn't seem like the issue.

fbwoolf:
When doing dev work with the Connect mono repo, I typ used `yalc`. This issue addresses using pnpm in a similar way, but not sure if helpful, suggestion is to use `add file` not `link`?

https://github.com/pnpm/pnpm/issues/5589

fbwoolf (approved):

PR #121, opened 2024-05-09
fix: update inscription model and update inscription queries

edgarkhanzadian:
Update inscription model and some inscription queries for extension's PR: https://github.com/leather-wallet/extension/pull/5355

## Summary by CodeRabbit

- **New Features**
  - Introduced new functionalities for handling Bitcoin inscriptions, including creating and fetching inscriptions by various parameters.
  - Deprecated certain functions in favor of more efficient alternatives for Bitcoin transaction queries.

- **Enhancements**
  - Enhanced the processing of Bitcoin inscriptions to improve the handling of transaction-related data.

- **Bug Fixes**
  - Updated variable names and import statements to ensure consistency and correct data handling across modules.

- **Refactor**
  - Refactored existing functions to utilize newly introduced methods, improving code efficiency and readability.

- **Documentation**
  - Updated documentation to reflect new functionalities and deprecated features.

### DIFF
--- a/packages/models/src/crypto-assets/bitcoin/inscription.model.ts
+++ b/packages/models/src/crypto-assets/bitcoin/inscription.model.ts
@@ -1,3 +1,6 @@
+import { HIRO_INSCRIPTIONS_API_URL } from '../../network.model';
+import { InscriptionCryptoAssetInfo } from '../crypto-asset-info.model';
+
 /**
  * Inscriptions contain arbitrary data. When retrieving an inscription, it should be
  * classified into one of the types below, indicating that the app can handle it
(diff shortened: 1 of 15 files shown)


pete-watters (approved):

packages/models/src/crypto-assets/bitcoin/inscription.model.ts
```diff
@@ -53,3 +56,103 @@ export function whenInscriptionMimeType<T>(
...
+  tx_id: string;
+  offset: string;
+  address: string;
+  genesis_block_hash: string;
+  genesis_timestamp: number;
+  genesis_block_height: number;
+  value: string;
+}
+
+export function createInscription(inscription: RawInscription): Inscription {
```
kyranjamie, line 87: I see you use it here, but what if we want to create a `Inscription` type from another source? 

Two scenarios:
- What if we need to create an inscription from information passed in via an RPC call. Then we're having to create it by passing in parameters specific to the API (e.g. `tx_id` vs `txid`) etc
- Or, we use an additional API that has a `transactionId` prop name instead. Then we have to do some funky prop renaming: `createInscription({ tx_id: tx.transactionId, ...props })` when we really wanna stick to our `txid`.

Conceptually, I'm not sure @leather-wallet/models should "know" anything about API responses, and putting the response here means it does.

What do you think of doing the mapping of API response props → Inscription props all in the queries package? Where `createInscription` only really initialises the type?

Happy to jump on call and discuss live if I haven't explained this well

packages/query/src/bitcoin/ordinals/inscriptions.hooks.ts
```diff
@@ -40,3 +41,52 @@ export function useNumberOfInscriptionsOnUtxo({
     [inscriptions]
   );
 }
+
+export function useInscriptions({
+  taprootKeychain,
+  nativeSegwitAddress,
+}: {
+  nativeSegwitAddress: string;
+  taprootKeychain: HDKey | undefined;
```
kyranjamie, line 50: Any other ideas how we could do this? 

Here, we're passing the keychain to `@leather-wallet/queries` to derive addresses. Which, seems like a responsibility that shouldn't belong to this package.

It's tricky, because we can't just pass an array of string, as we need to recurse over. 

What if we pass a callback that generates a new address? `getNextTaprootAddress(index: number)`? For consistency we could do the same with nativeSegwit addresses too.

packages/models/src/crypto-assets/bitcoin/inscription.model.ts
```diff
@@ -53,3 +56,103 @@ export function whenInscriptionMimeType<T>(
...
+  output: string;
+  content_type: string;
+  tx_id: string;
+  offset: string;
+  address: string;
+  genesis_block_hash: string;
+  genesis_timestamp: number;
+  genesis_block_height: number;
+  value: string;
+}
```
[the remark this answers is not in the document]
kyranjamie: > we can just match it to this RawInscription interface before supplying it to createInscription function.

Understood, but aren't we then perpetuating mapping to mix-match of properties? That isn't the one in our main model?

**Current case**

If our model uses `genesisBlockHeight`, API 1 uses `genesis_block_height` and API 2 uses `genesis_height`

When integrating API2, we call `const i = createInscription({ genesis_block_height: resp.genesis_height })` all so we can consume `i.genesisBlockHeight`? It's messy, no?

**Alternative**

Versus mapping straight to the prop we use underneath.

```ts
const api1 = createInscription({ genesisBlockHeight: resp.genesis_block_height });
const api2 = createInscription({ genesisBlockHeight: resp.genesis_height });
```

In alternative case: models knows nothing about any of the API response details, and we do as little prop matching  as possible.

packages/models/src/crypto-assets/bitcoin/inscription.model.ts
kyranjamie: I get that it's more convenient now, but if we can anticipate that it'll happen, let's avoid it, given it's just a matter of prop naming?

packages/models/src/crypto-assets/bitcoin/inscription.model.ts
```diff
@@ -53,3 +56,103 @@ export function whenInscriptionMimeType<T>(
...
+  output: string;
+  content_type: string;
+  txid: string;
+  offset: string;
+  address: string;
+  genesisBlockHash: string;
+  genesisTimestamp: number;
+  genesisBlockHeight: number;
+  value: string;
+}
```
kyranjamie, line 85: Is this needed anymore?

kyranjamie (approved):
Great work Edgar

My last thought is whether an inscription, and its metadata, should be a single model or not. Are there some use-cases where we might want a model that only represents the bare minimum needed identify an inscription (and then not want all the metadata from the API). What're your thoughts on this?

Either way this is good to merge as is

packages/models/src/crypto-assets/bitcoin/inscription.model.ts
```diff
@@ -53,3 +56,103 @@ export function whenInscriptionMimeType<T>(
...
+
+interface RawInscription {
+  id: string;
+  number: number;
+  output: string;
+  content_type: string;
```
fbwoolf, line 77: @edgarkhanzadian think this was missed to camelCase?

packages/models/src/crypto-assets/bitcoin/inscription.model.ts
fbwoolf: Nvm here, I had missed the PR in the extension where more of this discussion happened.

PR #139, opened 2024-05-20
fix: bundle packages so that those work in node envs too

edgarkhanzadian:
Using @leather-wallet/query package in the extension is causing extension's tests to fail. Now the reason that happens is because currently monorepo packages are setup to compile packages for moduleResolution: Bundler. Using that moduleResolution doesn't include file extensions in the compiled files' imports, and that causes stuff like vitest (that can use only node environment) to fail. Now, I had pretty much 2 options to fix this issue.

1. Append .js file extension at the end of every import in our source code, which is ugly and annoying, though is an official [suggestion by Typescript](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-libraries)
2. Bundle the package into a single .mjs file.

After researching multiple projects it seemed like the 2nd option is the most popular one. In a package like `zod` they use [`rollup`](https://github.com/colinhacks/zod/blob/main/package.json#L94) for that. In `formik` - [`tsdx`](https://github.com/jaredpalmer/formik/blob/main/packages/formik/package.json#L45). After reviewing my options it seemed like rollup is the best way to go here.

So in this PR you can find my attempt at implementing rollup bundler for our packages. Tests in extension package are no longer failing using these rollup bundled packages.

## Summary by CodeRabbit

- **New Features**
  - Updated build scripts to utilize Rollup for more efficient bundling.
  - Implemented concurrent build and watch scripts for enhanced development workflow.

- **Improvements**
  - Enhanced export paths to include ES module imports and TypeScript declaration files for improved compatibility and type safety.

- **Dependencies**
  - Added `concurrently` for executing multiple commands concurrently.
  - Integrated shared Rollup configuration package `@leather-wallet/rollup-config`.

- **Documentation**
  - Included README for shared Rollup configuration settings.

### DIFF
--- a/packages/bitcoin/package.json
+++ b/packages/bitcoin/package.json
@@ -11,8 +11,8 @@
     "lint:fix": "eslint . --fix --ignore-path ../../.eslintignore",
     "test:coverage": "vitest run --coverage",
     "test:unit": "vitest run",
-    "ts:build": "tsc --build",
-    "ts:build:watch": "tsc --build --watch --preserveWatchOutput",
+    "ts:build": "tsup index.ts --format esm && tsc --emitDeclarationOnly",
+    "ts:build:watch": "concurrently  \"tsc --build --watch --preserveWatchOutput --emitDeclarationOnly\"  \"tsup index.ts --format esm --watch \"  ",
     "typecheck": "tsc --noEmit -p ./tsconfig.json"
   },
   "homepage": "https://github.com/leather-wallet/mono/tree/dev/packages/bitcoin",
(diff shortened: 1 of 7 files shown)


kyranjamie:
What are the advantages of using rollup vs tsup, that we already use in `@leather-wallet/rpc`? 

Tsup is a super minimal wrapper over Rollup. If we could use that instead, that'd be better than introducing a new tool. If there are reasons to use Rollup over Tsup, let's discuss.

Also, RE this build isn't, I _think_ this can be solved extension side with the tsconfig changes [here (that I never managed to get merged)](https://github.com/leather-wallet/extension/pull/5227/files)?

kyranjamie (approved):

PR #147, opened 2024-05-22
fix: add contract asset name to model

alter-eggo:
We need to use specific contract asset name for post conditions  

## Summary by CodeRabbit

- **New Features**
  - Updated crypto asset information to include a new `contractAssetName` property.
  - Renamed the `name` property to `tokenName` for better clarity.


kyranjamie (approved):
Can you share an example of what each of these values might be?

alter-eggo:
Screenshot 2024-05-22 at 15 52 50
Screenshot 2024-05-22 at 15 51 39

PR #161, opened 2024-05-29
feat: migrate UI components to `ui/packages`

pete-watters:
This PR builds on https://github.com/leather-wallet/mono/pull/151 and:
- sets up `panda-preset` in `ui/packages` + removal of panda config
- migration of `Button`, `Link`, `Avatar` components
- removal of legacy `Text` + `Caption` components

This will prove the concept of moving a UI component into a package and importing into the `extension` / wherever. I tested it and I got it working using `link:/` in the `extension`. 

## Summary by CodeRabbit

- **New Features**
  - Introduced an `Avatar` component with customizable styles and fallback options.
  - Added a `BtcAvatarIcon` component for Bitcoin-themed avatars.
  - Introduced a styled `Button` component with various props for customization.
  - Added a `Link` component with customizable properties.
  - Added new icons: `ArrowDownIcon`, `ChevronDownIcon`, and `PlaceholderIcon`.

- **Stories**
  - Added stories for `Avatar`, `Button`, `Link`, and `ExampleIcon` components to showcase different configurations and appearances.

- **Configuration**
  - Updated path mappings and included new dependencies for better component styling and functionality.


packages/ui/src/components/button/button.web.tsx
```diff
@@ -0,0 +1,23 @@
+import { styled } from 'leather-styles/jsx';
```
pete-watters, line 1: `Button` and `Link` are built using `recipes` which are now kept in `@packages/panda-preset`. 

I think it could make sense to move related `recipes` into `@packages/ui` instead. 

That would mean we would be unable to rebuild from the `recipe` using only the `panda-preset` but it could be easier when building new components to keep the recipes in the ui package

pete-watters:
@coderabbitai review

packages/ui/src/components/button/button.web.tsx
pete-watters: Added a note about this to the discussion https://github.com/orgs/leather-wallet/discussions/66#discussioncomment-9595101

PR #163, opened 2024-05-30
@packages/ui migration

pete-watters:
This PR migrates code from `leather-wallet/extension/ui` -> `@leather-wallet/ui`

During this work I was unsure where to place:
- test selectors 
- some common hooks we use `useOnMount`

In order to keep progressing I put the test selectors here in `tmp/selectors` and kept a local copy of `useOnMount` for now. 

I tried to setup nicer path alias's to avoid relative imports but I couldn't get this working when installing as a package in the extension. I found [this discussion](https://github.com/vercel/turbo/discussions/620) but there seems not to be a clean and clear way to do it.


pete-watters:
I'm getting a lot of CI errors here but all seems OK locally. Investigating 👁️

packages/ui/src/components/button/button.native.tsx
```diff
@@ -1,8 +1,9 @@
 import { TouchableOpacity, ViewStyle } from 'react-native';
 
 import { ResponsiveValue } from '@shopify/restyle';
-import { Theme } from 'native';
 
+// FIXME issues#64: This is a temporary fix to commit as I broke the direct import of the theme
+import { Theme } from '../../theme-native';
```
pete-watters, line 6: @edgarkhanzadian I'm not sure how I broke this but on refactor to a `function` to appease eslint, this import started failing so I changed it to a relative path

packages/ui/src/tmp/README.md
```diff
@@ -0,0 +1,5 @@
+# Temporary location
```
pete-watters, line 1: I added this README as I'm not sure where to put test selectors we need in the extension and the `ui/packages` . More info [here](https://github.com/orgs/leather-wallet/discussions/66#discussioncomment-9595101)

packages/ui/src/tmp/tests/selectors/home.selectors.ts
```diff
@@ -0,0 +1,15 @@
+export enum HomePageSelectors {
+  AssetList = 'asset-list',
+  HomePageContainer = 'home-page-container',
```
fbwoolf, line 3: Why do we need to move the test selectors to the `ui` package? Can't the actual selectors just be relocated off the ui component directly? Will those be consistent b/w web and mobile? I don't think all of these selectors are used on ui components?

packages/ui/src/components/button/button.native.tsx
```diff
@@ -1,8 +1,9 @@
...
 import { Box } from '../box/box.native';
 import { Text } from '../text/text.native';
```
fbwoolf, line 8: When did we decide on `.native` as the convention? Is that a RN standard I've forgotten?

pete-watters: We want to co-locate component code so Edgar had split the files into `.web` for web and `.native` for RN. 

The plan is to co-locate so we can share as much as we can but I'm not sure how much that will be

packages/ui/src/tmp/tests/selectors/home.selectors.ts
pete-watters: I just dumped the list here as some of them are used in UI components. I can curate it to only have what we need specifically in this package but I didn't want to spend too long on it if I'm not sure what to do. 

We need them both in the extension in the test code and here in the UI components. I added it to the [discussion](https://github.com/orgs/leather-wallet/discussions/66#discussioncomment-9595101) but I didn't want it to block this work.

We could:
- put them in a shared place like `utils` / a new test package
- pass them into the UI components 
- refactor our other code to use selectors on code that is in the extension and not these components

packages/ui/src/components/button/button.native.tsx
fbwoolf: I know abt the colocation, was asking abt the naming convention of the files ...wasn't sure why `.native` was being used vs `.mobile`?

pete-watters:
This pairs up with this draft PR which consumes it in the extension https://github.com/leather-wallet/extension/pull/5485

packages/ui/src/components/containers/dialog/dialog.web.tsx
```diff
@@ -0,0 +1,96 @@
...
+              boxShadow:
+                'hsl(206 22% 7% / 35%) 0 10px 38px -10px, hsl(206 22% 7% / 20%) 0 10px 20px -15px',
+              position: 'fixed',
+              top: '50%',
+              left: '50%',
+              transform: 'translate(-50%, -50%)',
+              width: { base: '100vw', md: '90vw' },
+              height: { base: '100%', md: 'auto' },
+              maxWidth: { base: '100vw', md: 'pageWidth' },
+              maxHeight: { base: '100vh', md: '90vh' },
```
alter-eggo, line 71: I believe we discussed it in slack... 
isn't it better to add animations in keyframes rather than delete them at all? 😕

pete-watters: I deleted them as I think they were never used and just mistakenly taken from the radix demo. 

I can add them into keyframes

packages/ui/package.json
```diff
@@ -54,8 +54,8 @@
     "metro-config": "0.80.5",
     "metro-resolver": "0.80.5",
     "prism-react-renderer": "2.3.1",
-    "react": "18.3.1",
-    "react-dom": "18.3.1",
+    "react": "18.2.0",
```
pete-watters, line 57: There are some issues when installing in the extension now, I think related to a conflicting react version:
Screenshot 2024-06-04 at 09 51 49

I tried updating it here but it had a knock-on effect causing type issues in the native components.

I'll work on that in another branch and leave this open as a reference

pete-watters:
I'll move this back to `DRAFT` so it can be still reviewed if desired however more work will be needed integrating it to the extension. 

It worked easily for a simple `Button` but for some of the more complex components I am hitting problems I will need to solve / exclude those from the initial update. 

See [here](https://github.com/leather-wallet/mono/pull/163/commits/67ff21ed3ddb3aad60fa69600c0164864cbee209#r1625628777)

kyranjamie (commented):
Great work @pete-watters 

Some thoughts
- Base level hooks like `onMount` and others that only handle browser/js/events/css etc I think are okay to live in UI, similar to how chakra has some. Media query hooks too
- I think some of the components in `ui/components` didn't belong there
  - All the mnemonic secret key grid components I think are extension specific and should remain there. Even if not, I reckon the ui pkg should be as domain-agnostic as possible, and not know about things like mnemonics.
  - I wonder about the containers, too. Are we going to have a 1:1 equivalent containers in mobile? Imagine that'll be quite different, if so, that should also remain in extension.
- There should be no selectors IMO, we should only ever pass them into the ui components. If we need to target specific elements, perhaps we should add data-* attrs to them
- Would omit all virtuoso components too

pete-watters:
Closing this off. I'll open a fresh PR with a leaner set of changes. 

Adding any of the more complicated components we build on top of radix was non-trivial so I will exclude them from the initial package as more investigation and work is needed there. 

When trying to import they crash the extension with variations of:
```
Cannot read properties of null (reading useMemo / useContext / useState
```

PR #169, opened 2024-05-31
feat: add stacks queries

fbwoolf:
This PR migrates all the Stacks queries from the extension. I have a WIP branch in the extension that has removed/installed all of them back into that code ...so I know all these are working. I will put that up too for review.

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive querying functionalities for `alex-sdk`, `market-data`, `stacks`, and `leather-query-provider`.
  - Added hooks and utility functions for managing Stacks transactions, balances, fees, and Blockchain Name System (BNS) interactions.
  - Integrated new functions for querying and interacting with blockchain-related data using the `alex` SDK.

- **Bug Fixes**
  - Improved fee estimation and transaction handling for Stacks blockchain transactions.

- **Chores**
  - Updated dependencies and package names to align with the latest standards and improvements.

- **Refactor**
  - Enhanced internal structures and interfaces for better code maintainability and performance.


kyranjamie:
`@leather-wallet/query` should be a leaf package, that no other package depends on. Can you add a dependency cruiser rule to enforce this?

packages/query/src/stacks/fetch-wrapper.ts
```diff
@@ -0,0 +1,48 @@
...
+  const initHeaders = init.headers || {};
+  // eslint-disable-next-line no-restricted-globals
+  const resp = await fetch(input, {
+    ...init,
+    credentials: 'omit',
+    headers: { ...initHeaders, ...leatherHeaders },
+  });
+  if (isErrorCode(resp.status)) trackApiError(resp.url, resp.status);
+  return resp;
+}
```
kyranjamie, line 36: What do you suggest we do with this file? Would be nice not to commit it with loads of commented out code. 

Can you first describe the problem? Then solutions we might go for?

packages/query/src/stacks/info/info.hooks.ts
```diff
@@ -0,0 +1,5 @@
+import { useGetStackNetworkBlockTimeQuery } from './block-time.query';
+
+export function useStacksBlockTime() {
+  return useGetStackNetworkBlockTimeQuery();
+}
```
kyranjamie, line 5: Pretty pointless I think? can we remove entirely?

packages/query/src/stacks/mempool/mempool.query.ts
```diff
@@ -0,0 +1,29 @@
+import { MempoolTransactionListResponse } from '@stacks/stacks-blockchain-api-types';
+import { useQuery } from '@tanstack/react-query';
+
+import { useHiroApiRateLimiter } from '../../hiro-rate-limiter';
+import { useStacksClient } from '../stacks-client';
+
+export interface UseAccountMempoolQueryArgs {
+  address: string;
+  onSuccess?(data: MempoolTransactionListResponse): void;
```
kyranjamie, line 9: We'll need to not rely on this in v5

kyranjamie (commented):
A little bit concerned we're packaging loads of logic into the queries package that goes beyond just fetching the data.

 I don't want this PR to hang around too long, so we can be lenient on what makes it in this time around, but would be good to plan out changes we need to make after merging

packages/query/src/.DS_Store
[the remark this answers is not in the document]
fbwoolf: Not sure where these came from, will delete them.

fbwoolf:
> A little bit concerned we're packaging loads of logic into the queries package that goes beyond just fetching the data.
> 
> I don't want this PR to hang around too long, so we can be lenient on what makes it in this time around, but would be good to plan out changes we need to make after merging
>
Agree, we may have to port code back but I think that is ok so we can refactor all the queries together here and establish the right pattern to use (bitcoin and stacks)?

packages/query/src/stacks/fetch-wrapper.ts
fbwoolf: @edgarkhanzadian am I right that adding the queryCache `onError` analytics would replace the need for using the `trackApiError` here?

packages/query/src/stacks/mempool/mempool.query.ts
fbwoolf: @kyranjamie nice article on why those callbacks were removed: https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose

packages/query/src/stacks/fetch-wrapper.ts
fbwoolf: @kyranjamie it seems like our use of `wrappedFetch` is outdated? If we are only using it to support the `FetchApi` for the generated Stacks api, should we just remove our dependency on that library and use axios directly for Stacks queries? We appear to be already intercepting the axios request to add the proper Hiro headers.

kyranjamie: Pretty sure we pass this into the hiro client library, that's why it's needed, in order to add the query parms

fbwoolf:
@kyranjamie check the changes made here removing the hiro client lib ...in `stacks-client.ts`. I confirmed the changes are working in the extension when installed.

packages/query/src/stacks/stacks-client.ts
```diff
@@ -0,0 +1,305 @@
...
+    },
+    async getAccountNonces(address: string, signal?: AbortSignal) {
+      const resp = await rateLimiter.add(
+        () =>
+          axios.get<AddressNonces>(`${basePath}/extended/v1/address/${address}/nonces`, { signal }),
+        { throwOnTimeout: true }
+      );
+      return resp.data;
+    },
+    // TODO: Need to replace, this endpoint has been deprecated
```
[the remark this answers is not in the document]
fbwoolf: We do need to address this deprecated endpoint.

packages/query/src/stacks/balance/account-balance.hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+  return {
+    availableBalance: subtractMoney(totalBalance, outboundBalance),
+    availableUnlockedBalance: subtractMoney(unlockedBalance, outboundBalance),
+    inboundBalance,
+    lockedBalance: createMoney(stxMoney.locked.amount, 'STX'),
+    outboundBalance,
+    pendingBalance: subtractMoney(sumMoney([totalBalance, inboundBalance]), outboundBalance),
+    totalBalance,
+    unlockedBalance,
+  };
```
kyranjamie, line 40: Here we're performing some business logic, but I kinda don't mind cause it's just simple addition subtraction. Whereas anything more complicated, I'd question where it should live

packages/query/src/stacks/bns/bns.hooks.ts
```diff
@@ -0,0 +1,7 @@
+import { useGetBnsNamesOwnedByAddressQuery } from './bns.query';
+
+export function useCurrentAccountNames(address: string) {
+  return useGetBnsNamesOwnedByAddressQuery(address, {
+    select: resp => resp.names ?? [],
+  });
+}
```
kyranjamie, line 7: Not needed here, but wonder if we should just remove entirely, as what's the real value? it saves us having to access a nested prop from useGetBnsNamesOwnedByAddressQuery?

packages/query/src/stacks/stacks-client.ts
```diff
@@ -0,0 +1,305 @@
...
+import { useLeatherNetwork } from '../leather-query-provider';
+import type {
+  AddressBalanceResponse,
+  ContractInterfaceResponseWithFunctions,
+  NonFungibleTokenHoldingListResult,
+  StacksTxFeeEstimation,
+} from './hiro-api-types';
+import type { Stx20BalanceResponse } from './stx20-api-types';
+
+function HiroAccountsApi(basePath: string) {
```
kyranjamie, line 31: Let's not use this casing. It's not a class anymore, so functions begin with lowercase.

Suggest action based naming: `createHiroAccountApi`

packages/query/src/stacks/stacks-client.ts
```diff
@@ -0,0 +1,305 @@
...
+  hiroFeesApi: ReturnType<typeof HiroFeesApi>;
+  hiroInfoApi: ReturnType<typeof HiroInfoApi>;
+  hiroNamesApi: ReturnType<typeof HiroNamesApi>;
+  hiroNetworkApi: ReturnType<typeof HiroNetworkApi>;
+  hiroNonFungibleTokensApi: ReturnType<typeof HiroNonFungibleTokensApi>;
+  hiroSmartContractsApi: ReturnType<typeof HiroSmartContractsApi>;
+  hiroTokensApi: ReturnType<typeof HiroTokensApi>;
+  hiroTransactionsApi: ReturnType<typeof HiroTransactionsApi>;
+  stx20Api: ReturnType<typeof Stx20Api>;
+}
```
kyranjamie, line 284: I feel like the grouping of these comes from the old way of doing it with the generated client. Why not just have `client.getBnsNames` vs having to access via a nested object? DX is better to have a long list because I can search/browse all available methods.

packages/query/src/stacks/token-metadata/non-fungible-tokens/non-fungible-token-holdings.query.ts
```diff
@@ -0,0 +1,32 @@
...
+const queryOptions = { cacheTime: staleTime, staleTime, refetchhOnFocus: false } as const;
+
+function fetchNonFungibleTokenHoldings(client: StacksClient) {
+  return async (address: string) => {
+    if (!address) return;
+    return client.hiroNonFungibleTokensApi.getNftHoldings(address);
+  };
+}
+
+function makeNonFungibleTokenHoldingsQuery(address: string, network: string, client: StacksClient) {
```
kyranjamie, line 18: Should be exported.

packages/query/src/stacks/stacks-client.ts
fbwoolf: Yeah, agree, I'll look at refactoring the BitcoinClient too. I was trying to follow that effort and missed some more changes we can make. Same for comment below.

packages/query/src/stacks/token-metadata/non-fungible-tokens/non-fungible-token-holdings.query.ts
fbwoolf: I am actually going to merge this one so I can remove the queries from the extension. I'm getting quite a few conflicts there from other pkg changes. I'll start refactoring the queries in a follow up PR.

PR #171, opened 2024-06-03
fix(rpc): do not augment global type implicitly

kyranjamie:
Reading up on Typescript bundles, we shouldn't implicitly augment the global object. It might not be desired, can cause performance issues, and is prohibited by some tools.

Here, I've removed the global augmentation, and added README and TSDoc comments to explain how to use the provider object.

image

## Summary by CodeRabbit

- **Documentation**
  - Added Typescript typings for the `LeatherProvider` global object in the `@leather-wallet/rpc` package.
  - Updated documentation to guide developers on using `LeatherProvider` in a Typescript project.
  - Documented the `LeatherProvider` interface and its integration with the global `window` object.

### DIFF
--- a/packages/rpc/README.md
+++ b/packages/rpc/README.md
@@ -1,3 +1,18 @@
 # @leather-wallet/rpc
 
 This package provides Typescript typings Leather developers can use when interacting with the `LeatherProvider` global object.
+
+## Consuming as global object
+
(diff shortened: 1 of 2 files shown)


kyranjamie:
Yes it will be, though imagine we have very few users given we renamed it.

PR #175, opened 2024-06-04
Populate @leather-wallet/ui

pete-watters:
This PR introduces a set of simpler packages to `@leather-wallet/ui` to start with E2E integration to the extension. 

I have excluded some components mentioned on the list [here](https://github.com/leather-wallet/mono/pull/163#pullrequestreview-2095782535) . 

You can see the updated components in storybook in the mono-repo here:
https://github.com/leather-wallet/mono/assets/2938440/68f4d795-b2f4-476e-89fa-3b66fcbfe489


pete-watters:
Please see [comment here](https://github.com/leather-wallet/issues/issues/64#issuecomment-2149739337) 

**TLDR:**
- I am introducing some simpler components at first to prove the concept
- Integrating more complicated ones (e.g. `Input` and installing to the extension was causing it to crash)
- I thought that the icons and some other components were working right away but I was fooled by the cache and need to do more work investigating the packaging

fbwoolf (approved):
🚀

PR #178, opened 2024-06-05
feat: fix release process

kyranjamie:
Previous release build failed because pnpm version hadn't been updated in release job. Here, we reuse the provision action with the later version of pnpm.


pete-watters (approved):
👍 FYI I updated it in the extension also as I hit issues earlier installing the UI 

https://github.com/leather-wallet/extension/pull/5487

PR #186, opened 2024-06-06
chore: use tsup to prep UI package

pete-watters:
## Summary by Human

This PR:
- updates `panda` in the monorepo
- uses `tsup` to package `leather-wallet/uis`

## Summary by CodeRabbit

- **New Features**
  - Improved build process for web assets using `tsup` for faster and more efficient builds.

- **Chores**
  - Updated `@pandacss/dev` dependency to version 0.40.1 for enhanced development features and stability.
  - Added `tsup` dependency to streamline the build process.

## Summary by CodeRabbit

- **New Features**
  - Introduced new build configuration using `tsup` for web assets in the UI package.

- **Chores**
  - Updated `@pandacss/dev` dependency version from "0.26.2" to "0.40.1".
  - Added `tsup` as a dependency in the UI package.
  - Updated build scripts and export paths in the UI package's `package.json`.
  - Adjusted TypeScript configuration options in `tsconfig.web.json` for optimized builds.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -56,7 +56,7 @@
     "@babel/runtime": "7.23.2",
     "@leather-wallet/eslint-config": "workspace:*",
     "@leather-wallet/panda-preset": "workspace:*",
-    "@pandacss/dev": "0.26.2",
+    "@pandacss/dev": "0.40.1",
     "@react-native-async-storage/async-storage": "1.23.1",
     "@storybook/addon-actions": "7.6.17",
     "@storybook/addon-docs": "7.6.15",
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -11,7 +11,7 @@
     "build:all": "pnpm assets:web && pnpm assets:native && tsc --build ./tsconfig.json",
     "build:native": "pnpm assets:native && tsc -p ./tsconfig.native.json",
     "build:watch": "concurrently  \"pnpm panda:watch\"  \"pnpm build:native --watch --preserveWatchOutput\"  \"pnpm build:web --watch --preserveWatchOutput\"  \"pnpm build:all --watch --preserveWatchOutput\" ",
-    "build:web": "pnpm assets:web && tsc -p ./tsconfig.web.json",
+    "build:web": "pnpm assets:web && tsup --config tsup.config.web.ts",
     "format": "prettier . --write \"src/**/*.{ts,tsx}\" --ignore-path ../../.prettierignore",
     "format:check": "prettier . --check \"src/**/*.{ts,tsx}\" --ignore-path ../../.prettierignore",
     "lint": "eslint . --ignore-path ../../.eslintignore",
@@ -25,7 +25,11 @@
     "typecheck": "tsc --noEmit --project ./tsconfig.json"
   },
   "exports": {
-    ".": "./dist-web/web.js",
+    ".": {
+      "types": "./dist-web/web.d.ts",
+      "import": "./dist-web/web.js",
+      "require": "./dist-web/web.js"
+    },
     "./native": "./dist-native/native.js"
   },
   "dependencies": {
@@ -49,7 +53,8 @@
     "metro-resolver": "0.80.5",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "react-native": "0.74.1"
+    "react-native": "0.74.1",
+    "tsup": "8.1.0"
   },
   "devDependencies": {
     "@babel/core": "7.24.6",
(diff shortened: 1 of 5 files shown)


packages/ui/package.json
```diff
@@ -25,7 +25,11 @@
     "typecheck": "tsc --noEmit --project ./tsconfig.json"
   },
   "exports": {
-    ".": "./dist-web/web.js",
+    "./web": {
+      "types": "./web.ts",
```
pete-watters, line 29: I'm not sure if this is right for `types` but it works. Maybe I need to package and move them to `dist` also.

Right now we export and consume some types for extending with:
- ButtonProps
- FlagProps
- TagProps
- TagVariants
- HrProps

This is a temporary step until I move the remaining components so maybe we can ignore for now. 

The types in `index.d.ts` have no exports so if I instead include that it fails to install

packages/ui/package.json
```diff
@@ -25,7 +25,11 @@
     "typecheck": "tsc --noEmit --project ./tsconfig.json"
   },
   "exports": {
-    ".": "./dist-web/web.js",
+    "./web": {
```
kyranjamie, line 28: Let's try and keep the naked path, and only have `/native` for native, as react is `react` and not `react/web`

packages/ui/tsup.config.web.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  entry: ['web.ts'],
+  format: ['esm'],
+  tsconfig: 'tsconfig.web.json',
+  outDir: 'dist-web/',
```
kyranjamie, line 7: Needs `dts: true` no?

packages/ui/package.json
pete-watters: I changed it, thanks 👍

packages/ui/tsup.config.web.ts
pete-watters: I added this and some other options with inspiration from https://github.com/fungible-systems/micro-stacks/blob/main/packages/core/tsup.config.ts. 

Thanks!

packages/ui/package.json
```diff
@@ -25,7 +25,11 @@
     "typecheck": "tsc --noEmit --project ./tsconfig.json"
   },
   "exports": {
-    ".": "./dist-web/web.js",
+    ".": {
```
pete-watters, line 28: I am not sure if we need to keep the `build:all` step above. 

We bundle into 3 folders:
- `dist-all`
- `dist-web`
- `dist-native`

Maybe we can remove the output to `dist-all` ? I'm not sure if we need it. 

We will do our imports for web from `leather-wallet/ui` and native from `leather-wallet-ui/native`.

kyranjamie: > Maybe we can remove the output to dist-all ? I'm not sure if we need it.

👍🏼 

> We will do our imports for web from leather-wallet/ui and native from leather-wallet-ui/native.

Why not `@leather-wallet/ui` and `@leather-wallet/ui/native`?

packages/ui/package.json
```diff
@@ -11,7 +11,7 @@
     "build:all": "pnpm assets:web && pnpm assets:native && tsc --build ./tsconfig.json",
     "build:native": "pnpm assets:native && tsc -p ./tsconfig.native.json",
     "build:watch": "concurrently  \"pnpm panda:watch\"  \"pnpm build:native --watch --preserveWatchOutput\"  \"pnpm build:web --watch --preserveWatchOutput\"  \"pnpm build:all --watch --preserveWatchOutput\" ",
-    "build:web": "pnpm assets:web && tsc -p ./tsconfig.web.json",
+    "build:web": "pnpm assets:web && tsup --config tsup.config.web.ts",
```
kyranjamie, line 14: Do we still need the `assets` task if we're using tsup?

packages/ui/package.json
pete-watters: I just noticed a typo in what I had said, apologies. 

It's exactly what you said: 
- Web files - `@leather-wallet/ui` 
- Native files - `@leather-wallet/ui/native`

kyranjamie: wondered after if it was a typo

packages/ui/package.json
pete-watters: We currently do but I will investigate configuring `tsup` to copy the assets over. 

I didn't setup `tsup` for the native components yet, only for web

pete-watters: On initial investigation, we could add a `postbuild` script to copy the assets once `tsup` build finishes but its almost the same as just having the `assets` task. 

Right now `tsup` doesn't support moving the assets itself from what I can tell.

pete-watters: I'll merge for now so I can continue testing, we can improve on later

PR #187, opened 2024-06-06
feat: add stacks tx and status types

alter-eggo:
this pr adds stacks tx and tx status types

## Summary by CodeRabbit

- **New Features**
  - Enhanced transaction handling with new type declarations for `StacksTx` and `StacksTxStatus`.
  - Improved integration with blockchain API by adding `@stacks/stacks-blockchain-api-types` as a dependency.

### DIFF
--- a/packages/models/package.json
+++ b/packages/models/package.json
@@ -25,13 +25,13 @@
   },
   "bugs": "https://github.com/leather-wallet/mono/issues",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


kyranjamie:
also wonder, shouldn't these live in the stacks package?

kyranjamie:
> get dep lint error here, though @stacks/stacks-blockchain-api-types is actually types library 

this is fixed with `import type`

PR #199, opened 2024-06-09
feat: add lingui

edgarkhanzadian:
- Use pnpm lingui:extract to extract all translation strings from code
- Use pnpm lingui:compile to compile js files for translation
- Added check-translation-strings to test if the translation strings were updated in the PR
- Run pnpm lingui:watch to extract and compile strings while in development

## Summary by CodeRabbit

- **New Features**
  - Introduced internationalization support using Lingui in the mobile app.
  - Added French translations for the welcome screen component.
  - Added a pseudo-locale for testing translations.

- **Bug Fixes**
  - Updated `.eslintignore` and `.prettierignore` to exclude `messages.js`.

- **Chores**
  - Added a new job in GitHub Actions to check for uncommitted translation strings.
  - Updated Babel configuration to include the `macros` plugin.
  - Added Lingui-related scripts and dependencies to `package.json`.


apps/mobile/src/locales/en/messages.po
```diff
@@ -0,0 +1,46 @@
...
+msgid "Stay in the loop and be the first one to hear about new developments"
+msgstr "Stay in the loop and be the first one to hear about new developments"
+
+#: src/components/welcome-screen/index.tsx:296
+msgid "Submit"
+msgstr "Submit"
+
+#: src/components/welcome-screen/index.tsx:226
+msgid "Thanks for subscribing! We'll notify you when Leather's mobile app launches. In the meantime, please <0>check out our browser extension.</0>"
+msgstr "Thanks for subscribing! We'll notify you when Leather's mobile app launches. In the meantime, please <0>check out our browser extension.</0>"
```
kyranjamie, line 46: There's no need to commit this if we're generating the file right? 

Would be much nicer to incorporate generation into build flow, as it works with Panda, rather than committing generated files. This can cause big repo-bloat and conflict issues.

fbwoolf (approved):

apps/mobile/src/locales/en/messages.po
kyranjamie: What happens if we human author updates some text, and another then re-generates the file?

pete-watters (approved):
This LGTM Edgar, nice work! 

Right now I am more in favour of a wrapping `Trans` component but I wonder how high of a level that can be and if will have have lots of them in the code.

I like the `msgid` style in the `.po` files

kyranjamie (approved):

PR #213, opened 2024-06-12
fix: hiro api testnet url path

fbwoolf:
@alter-eggo I think this needs to be the `old` path? It seems out of sync with the extension?


fbwoolf:
I'm not sure you can post new transactions to the old endpoint? 🤔 I have no balance with the new one, but no ability to broadcast txs using the `old` path.

fbwoolf:
Issue: https://github.com/leather-wallet/extension/issues/5529

PR #225, opened 2024-06-19
refactor: upgrade react query to v5

fbwoolf:
Part 1 of refactoring queries, this upgrades us to React Query v5. I wanted to get one PR done just upgrading the lib and making sure things are working.

I have this installed in the extension and things seem to be working. The only real trouble I had was with our use/implementation of `useInfiniteQuery` and changes to the `pageParam` requiring a new `initialPageParam`. There example only shows a `number` type for this, but we use an object.

Also, I ran into this, so I had to upgrade typescript too as the fix:
https://github.com/TanStack/query/discussions/6670

## Summary by CodeRabbit

- **New Features**
  - Updated TypeScript to version `5.4.5` across all relevant packages for better performance and new features.
  - Upgraded `@tanstack/react-query` to `5.45.1` for improved query handling.

- **Refactor**
  - Renamed `isLoading` to `isPending` in the `WelcomeScreen` component for enhanced state clarity.
  - Changed `cacheTime` to `gcTime` in multiple query configurations to refine cache management.
  - Simplified query handling and improved code structure in several query functions.

- **Bug Fixes**
  - Corrected button state logic in `WelcomeScreen` to reflect the loading state accurately.
  
- **Chores**
  - Removed unnecessary comments and imports to clean up the codebase.


apps/mobile/package.json
```diff
@@ -97,7 +97,7 @@
     "react-test-renderer": "18.2.0",
     "ts-jest": "29.1.4",
     "tslib": "2.6.2",
-    "typescript": "5.3.3"
+    "typescript": "5.4.5"
```
pete-watters, line 100: Good job updating these. Kyran mentioned to me previously about setting up `syncpack` to keep these in check. I opened an issue for it [here](https://github.com/leather-wallet/mono/issues/226) 

That could help us keep versions of certain packages the same throughout the monorepo

packages/query/src/bitcoin/ordinals/brc20/brc20-tokens.query.ts
```diff
@@ -39,8 +39,8 @@ export function useGetBrc20TokensQuery({
   const query = useInfiniteQuery({
     queryKey: [QueryPrefixes.GetBrc20Tokens, currentNsBitcoinAddress, network.id],
     async queryFn({ pageParam }) {
-      const fromIndex: number = pageParam?.fromIndex ?? 0;
-      let addressesWithoutTokens = pageParam?.addressesWithoutTokens ?? 0;
+      const fromIndex = pageParam.fromIndex;
+      let addressesWithoutTokens = pageParam.addressesWithoutTokens;
```
pete-watters, line 43: I know this isn't your code but maybe we could refactor this to not use `let`. Not in this PR but just flagging that using `let` can be confusing and harder to follow what's happening.  

So that we set:
```
const addressesWithoutTokens = pageParam.addressesWithoutTokens;

const brc20addressesWithoutTokens = brc20Tokens.filter(tokens => tokens.length === 0).length

...

return {
...
 addressesWithoutTokens: addressesWithoutTokens += brc20addressesWithoutTokens
}
```

packages/query/src/bitcoin/ordinals/inscriptions.query.ts
```diff
@@ -65,30 +55,17 @@ async function fetchInscriptions({
...
-  {
-    taprootKeychain,
-    nativeSegwitAddress,
-  }: {
-    nativeSegwitAddress: string;
-    taprootKeychain: HDKey | undefined;
-  },
-  options?: UseInfiniteQueryOptions<GetInscriptionsInfiniteQuery, unknown, T>
-) {
+export function useGetInscriptionsInfiniteQuery({
```
pete-watters, line 62: Do you know if we can do the refactor mentioned above? It could be nice to do it now we are changing this but maybe it's complicated

packages/query/src/bitcoin/ordinals/inscriptions.query.ts
```diff
@@ -193,12 +171,19 @@ export function useGetInscriptionsInfiniteQuery<T extends unknown = GetInscripti
...
+      total: 0,
+    },
+    getNextPageParam(lastPage) {
       const { offset, total, stopNextFetch, fromIndex, addressesWithoutOrdinalsNum, addressesMap } =
-        prevInscriptionsQuery;
+        lastPage;
       if (stopNextFetch) return undefined;
 
-      // calculate offset for next fetch
+      // Calculate offset for next fetch
```
pete-watters, line 186: For this, it could be nice to extract into a function - `calculateOffset`. 

Not something necessary to bundle in right now but calling it as I see it and adding as a suggestion for how trying to avoid using `let` can force a more understandable style

```
function calculatedOffset(offset, inscriptionsLazyLoadLimit, total) {
  const initialOffset = offset + inscriptionsLazyLoadLimit; 
 const offsetAboveLimit = offset + inscriptionsLazyLoadLimit > total;

   if (offsetAboveLimit) {
       return offset + (total - offset);
      }
 return initialOffset;

}

```

packages/query/src/bitcoin/runes/runes.hooks.ts
```diff
@@ -23,12 +23,10 @@ export function useRuneTokens(addresses: string[]) {
   const results = useGetRunesTickerInfoQuery(runesBalances);
 
   return useMemo(() => {
-    // We can potentially use the 'combine' option in react-query v5 to replace this?
```
pete-watters, line 26: 💪

pete-watters (approved):
Great work @fbwoolf 👏 

I added some coding style suggestions but they are for pre-existing stuff not relevant to this work. Just mentioning as I was reading the code anyway

apps/mobile/package.json
kyranjamie: Edgar already set up syncpack, and we have a CI job for it? 🤔

kyranjamie (approved):
LGTM @fbwoolf

apps/mobile/package.json
pete-watters: OK. I guess I checked for it in the `extension` by mistake and don't understand how it works. I thought it should help us avoid having to update all the dependancies in each package.json?

packages/query/src/bitcoin/ordinals/inscriptions.query.ts
fbwoolf: I'm actually not real sure who left that comment to refactor as it was relocated code, so the git lens is off. I would assume it was @alter-eggo bc our use of `useInfiniteQuery` was mostly implemented by him? @alter-eggo can I create an issue for you to relook at these query hooks once I'm done with the Stacks query refactors? These hooks are a bit complicated to understand and think could be refactored? There are several comments left here by @pete-watters with suggestions.

pete-watters: Let's merge this anyway and we can consider improvements in the future

alter-eggo: yeah, it's quite complex logic here as we need to fetch inscriptions from native segwit address, loop through taproot addresses, and also support dynamic loading once reach bottom of home page (if user has many inscriptions)

agree, it would be great to create an issue for it

packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.query.ts
```diff
@@ -83,7 +83,7 @@ export function useGetFungibleTokensMetadataQuery(keys: string[]) {
         enabled: !!principal,
         queryKey: ['get-ft-metadata', principal, network.chain.stacks.url],
         queryFn: fetchFungibleTokenMetadata(client)(principal),
-        select: (resp: FtAssetResponse) => {
+        select: (resp: any) => {
```
alter-eggo, line 86: hm, why any here 🤔

alter-eggo (approved):

packages/query/src/bitcoin/ordinals/inscriptions.query.ts
fbwoolf: Made issue here: https://github.com/leather-io/issues/issues/109

PR #228, opened 2024-06-19
fix: test export for png, ref leather-wallet/issues#64

pete-watters:
This PR changes the entry point for the UI package to load from `dist-web` / `dist-native` instead of `dist-all`. 

I need to do some more work investigating the best way of doing this however I want to change this now so I can finish migrating more components. 

If I don't do this, I get an installation error in the `extension` as it cannot find the `.png` source files for `Stx20AvatarIcon` and `StampsAvatarIcon`   

## Summary by CodeRabbit

- **Chores**
	- Updated internal paths for type definitions, imports, and dependencies to improve module organization.
	- Changed import statements to use absolute paths for better consistency in the codebase.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -6,11 +6,10 @@
   "scripts": {
     "assets:native": "mkdir -p dist-native/src && cp -r src/assets-native ./dist-native/src/",
     "assets:web": "mkdir -p dist-web/src && cp -r src/assets-web ./dist-web/src",
-    "build": "pnpm panda && pnpm build:native && pnpm build:web && pnpm build:all",
+    "build": "pnpm panda && pnpm build:native && pnpm build:web",
     "build-storybook:web": "storybook build -c ./src/.storybook-web",
-    "build:all": "pnpm assets:web && pnpm assets:native && tsc --build ./tsconfig.json",
     "build:native": "pnpm assets:native && tsc -p ./tsconfig.native.json",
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/avatar/stx20-avatar-icon.web.tsx
```diff
@@ -1,4 +1,5 @@
-import Stx20AvatarIconSrc from '../../assets-web/avatars/stx20-avatar-icon.png';
+import Stx20AvatarIconSrc from 'src/assets-web/avatars/stx20-avatar-icon.png';
```
pete-watters, line 1: Trying this as I can't see it loading when testing locally 
Screenshot 2024-06-19 at 14 10 16

packages/ui/package.json
```diff
@@ -26,11 +26,11 @@
   },
   "exports": {
     ".": {
-      "types": "./dist-all/web.d.ts",
-      "import": "./dist-all/web.js",
-      "require": "./dist-all/web.js"
+      "types": "./dist-web/web.d.ts",
```
pete-watters, line 29: Trying this as `assets` aren't bring properly copied to `dist-all` so I want to check if this will work for the `.png`. Then I can try and fix the copy to `dist-all`

kyranjamie (approved):
I wonder if there's a larger issue here that image assets shouldn't be being delivered via a package?

pete-watters:
> I wonder if there's a larger issue here that image assets shouldn't be being delivered via a package?

Yes I think we shouldn't do that. My plan here is to merge this as otherwise I get a failure when installing the package. 

Then I can import all the Avatars, except for the two `.png` ones and we can investigate what to do with those. 

I opened an issue for the static assets task - https://github.com/leather-wallet/issues/issues/103

PR #243, opened 2024-06-26
feat: implement bis api for inscription by address query

alter-eggo:
Inscription by address query  is used for filtering native segwit utxos.
This pr changes the the source of data to bis api, other queries will be refactored later

### DIFF
--- a/packages/query/src/bitcoin/ordinals/inscriptions.query.ts
+++ b/packages/query/src/bitcoin/ordinals/inscriptions.query.ts
@@ -5,7 +5,10 @@ import { useInfiniteQuery } from '@tanstack/react-query';
 import axios from 'axios';
 
 import { getTaprootAddress } from '@leather.io/bitcoin';
-import { HIRO_INSCRIPTIONS_API_URL } from '@leather.io/models';
+import {
+  HIRO_INSCRIPTIONS_API_URL,
+  WalletDefaultNetworkConfigurationIds,
+} from '@leather.io/models';
 import { createNumArrayOfRange } from '@leather.io/utils';
 import { ensureArray } from '@leather.io/utils';
 
@@ -13,6 +16,7 @@ import type { InscriptionResponseHiro } from '../../../types/inscription';
 import { useLeatherNetwork } from '../../leather-query-provider';
 import { QueryPrefixes } from '../../query-prefixes';
 import { useHiroApiRateLimiter } from '../../rate-limiter/hiro-rate-limiter';
+import { useBitcoinClient } from '../clients/bitcoin-client';
 
 const stopSearchAfterNumberAddressesWithoutOrdinals = 5;
 const addressesSimultaneousFetchLimit = 5;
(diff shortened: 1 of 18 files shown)


packages/models/src/crypto-assets/bitcoin/inscription.model.ts
```diff
@@ -156,3 +157,82 @@ export function createInscription(inscription: RawInscription): Inscription {
...
+    offset,
+    address: inscription.owner_wallet_address,
+    preview,
+    title,
+    genesisBlockHeight: inscription.genesis_height,
+
+    // TODO: currently no data from BestinSlot API
+    genesisBlockHash: '0',
+    genesisTimestamp: 0,
+    value: '0',
```
alter-eggo, line 185: wonder if we need this info? 🤔 it doesn't come from bis api, should we ask to fix it?

packages/models/src/inscriptions/inscriptions.model.ts
```diff
@@ -0,0 +1,21 @@
+export interface BestinslotInscriptionResponse {
```
fbwoolf, line 1: I think we were making an effort to keep api models with the queries and only use the models pkg for the models we map/create/use in the app ...like a DTO pattern?

fbwoolf (approved):

packages/models/src/inscriptions/inscriptions.model.ts
kyranjamie: +1, this is specific to the API so belongs with the fetcher/query. 

`@leather.io/models` shouldn't "know" about BIS.

packages/query/src/bitcoin/bitcoin-client.ts
```diff
@@ -156,6 +137,23 @@ interface RunesOutputsByAddressResponse {
   data: RunesOutputsByAddress[];
 }
 
+interface BestinSlotInscriptionByAddressArgs {
```
kyranjamie, line 140:
```suggestion
interface BestInSlotInscriptionByAddressArgs {
```

kyranjamie (approved):
LGTM. I think it'd be better to use zod + type inference to get the type though, this way we can validate the schema and ensure it's as we expect

alter-eggo:
@kyranjamie added zod validation, will double check it, once bis team send type definitions

packages/query/src/bitcoin/ordinals/inscriptions.query.ts
```diff
@@ -208,33 +212,38 @@ export function useGetInscriptionsInfiniteQuery({
...
+      return lastPage.offset + bestinslotInscriptionsRequestNum;
     },
-    refetchOnMount: false,
-    refetchOnReconnect: false,
-    refetchOnWindowFocus: false,
-    staleTime: 3 * 60 * 1000,
+    // refetchOnMount: false,
+    // refetchOnReconnect: false,
+    // refetchOnWindowFocus: false,
+    // staleTime: 3 * 60 * 1000,
```
fbwoolf, line 246: Did you mean to leave these commented out?

alter-eggo: yes, I've reverted it in main pr

PR #245, opened 2024-06-27
fix: update BitcoinNetworkModes, ref leather-wallet/extension#4874

pete-watters:
Having a go at solving https://github.com/leather-io/extension/issues/4874

### DIFF
--- a/packages/bitcoin/src/bitcoin.utils.ts
+++ b/packages/bitcoin/src/bitcoin.utils.ts
@@ -36,11 +36,11 @@ export function initBitcoinAccount(derivationPath: string, policy: string): Bitc
  * has a number of networks, its often only necessary to consider the higher
  * level concepts of mainnet and testnet
  */
-export const bitcoinNetworkToCoreNetworkMap: Record<BitcoinNetworkModes, NetworkModes> = {
+export const bitcoinNetworkToCoreNetworkMap: Record<BitcoinNetworkModes, BitcoinNetworkModes> = {
   mainnet: 'mainnet',
   testnet: 'testnet',
-  regtest: 'testnet',
-  signet: 'testnet',
+  regtest: 'regtest',
+  signet: 'signet',
 };
 export function bitcoinNetworkModeToCoreNetworkMode(mode: BitcoinNetworkModes) {
   return bitcoinNetworkToCoreNetworkMap[mode];
@@ -52,9 +52,11 @@ export function bitcoinNetworkModeToCoreNetworkMode(mode: BitcoinNetworkModes) {
  * @example
  * `m/86'/1'/0'/0/0`
  */
-export const coinTypeMap: Record<NetworkModes, 0 | 1> = {
+export const coinTypeMap: Record<BitcoinNetworkModes, 0 | 1 | 2 | 3> = {
   mainnet: 0,
   testnet: 1,
+  regtest: 2,
+  signet: 3,
 };
 
 export function getBitcoinCoinTypeIndexByNetwork(network: BitcoinNetworkModes) {


packages/bitcoin/src/bitcoin.utils.ts
```diff
@@ -36,11 +36,11 @@ export function initBitcoinAccount(derivationPath: string, policy: string): Bitc
  * has a number of networks, its often only necessary to consider the higher
  * level concepts of mainnet and testnet
  */
-export const bitcoinNetworkToCoreNetworkMap: Record<BitcoinNetworkModes, NetworkModes> = {
+export const bitcoinNetworkToCoreNetworkMap: Record<BitcoinNetworkModes, BitcoinNetworkModes> = {
```
pete-watters, line 39: @kyranjamie , I'm having a go at solving https://github.com/leather-io/extension/issues/4874 but I'm not sure if this is correct. 

Is `bitcoinNetworkToCoreNetworkMap` and `bitcoinNetworkModeToCoreNetworkMode` intended to:

1. block us from using `regtest` + `signet` in some cases
2. not setup correctly with all possible networks

In the extension I can see cases where we call `bitcoinNetworkModeToCoreNetworkMode` and in the bug report you say to write a test to make sure the BTC send flow should work when choosing the `regtest` network. 

Maybe my change here is incorrect but instead I need to change the `extension` to stop using `bitcoinNetworkModeToCoreNetworkMode` for:
- `useBtcSendForm`
- `useSendInscriptionForm`

pete-watters: @fbwoolf , I should have tagged you yesterday. Maybe you can help me understand the history here and how I should fix this?

packages/bitcoin/src/bitcoin.utils.ts
```diff
@@ -52,9 +52,11 @@ export function bitcoinNetworkModeToCoreNetworkMode(mode: BitcoinNetworkModes) {
  * @example
  * `m/86'/1'/0'/0/0`
  */
-export const coinTypeMap: Record<NetworkModes, 0 | 1> = {
+export const coinTypeMap: Record<BitcoinNetworkModes, 0 | 1 | 2 | 3> = {
   mainnet: 0,
   testnet: 1,
+  regtest: 2,
+  signet: 3,
```
kyranjamie, line 59: Breaking change. There is no coin type 2 and 3.

kyranjamie (changes requested):
I don't think this is the correct solution. This map is now pointless, it just turns a string into the same string (and will def break wallet functionality).

There are distinct but similar concepts touched in this map. The exact representation of each network, and the concept of there being a mainnet and a testnet.

One mainnet. Many testnets. 

Some logic needs to know the specific network, other the broader category. Consider coinType of derivation paths: Mainnet `m/86'/0'/` and Testnet `m/86'/1'/`. The latter is share across testnets, so it's helpful to have a fn to turn the specific network into it's category of either mainnet or testnet.

packages/bitcoin/src/bitcoin.utils.ts
kyranjamie: The problem of this issue doesn't lie in this map/function, which serves the purpose of turning a specific network into the less granular, broader concept of mainnet/testnet.

What is likely happening is that we're writing logic considering only mainnet/testnet, when we need logic specific to regtest. If that's the case, the logic needs to change to never call this function in the first place, not change this fn.

pete-watters: OK thanks 👍

PR #246, opened 2024-06-27
fix: inscriptions query by address

fbwoolf:
Need this change to fix:
https://github.com/leather-io/issues/issues/114

cc @alter-eggo to be aware of `useInfiniteQuery` bug:
https://stackoverflow.com/questions/72753877/useinfinitequerys-enabled-option-in-react-query-doesnt-work

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability by preventing actions when the address parameter is missing or invalid, ensuring smoother user experience.

### DIFF
--- a/packages/query/src/bitcoin/ordinals/inscriptions.query.ts
+++ b/packages/query/src/bitcoin/ordinals/inscriptions.query.ts
@@ -227,6 +227,7 @@ export function useInscriptionsByAddressQuery(address: string) {
     },
     initialPageParam: 0,
     getNextPageParam(lastPage) {
+      if (!address) return undefined;
       if (lastPage.offset >= lastPage.total) return undefined;
       return lastPage.offset + 60;
     },
@@ -238,8 +239,9 @@ export function useInscriptionsByAddressQuery(address: string) {
 
   // Auto-trigger next request
   useEffect(() => {
+    if (!address) return;
     void query.fetchNextPage();
-  }, [query, query.data]);
+  }, [address, query, query.data]);
 
   return query;
 }


fbwoolf:
@alter-eggo I need this to fix a P1 bug in the extension, sorry I know it will impact your open PR for BIS ...minor change though I hope.

alter-eggo:
@fbwoolf no worries

pete-watters:
Nice working finding this issue @fbwoolf . Checking the code and I see we also use `useInfiniteQuery`  in:
- `useGetInscriptionsInfiniteQuery`
- `useGetBrc20TokensQuery`

Maybe we also need to add this defensive code there? If it's something we will need often we could investigate a custom hook? 

I can open a task for it if you think we need to

fbwoolf:
> Nice working finding this issue @fbwoolf . Checking the code and I see we also use `useInfiniteQuery` in:
> 
> * `useGetInscriptionsInfiniteQuery`
> * `useGetBrc20TokensQuery`
> 
> Maybe we also need to add this defensive code there? If it's something we will need often we could investigate a custom hook?
> 
> I can open a task for it if you think we need to
>
I saw that too, good point. I was hesitant to do more in these queries rn w/ @alter-eggo working on them and planning to refactor. We can open an issue or let him tackle in his refactor issue which already exists?

pete-watters:
> * `useGetInscriptionsInfiniteQuery`
> * `useGetBrc20TokensQuery`

I just made a new issue for it anyway in case it gets lost: https://github.com/leather-io/issues/issues/124 

@alter-eggo if you want to fix it in your current work that would be great. Otherwise we can do it separately

alter-eggo:
I'd work separately

PR #250, opened 2024-06-28
feat: implement animated action bar

edgarkhanzadian:
Implement animated action bar:
- When scrolling down, hide action bar after passing 50px
- When scrolling up, immediately show the action bar
- When reaching the bottom of the scroll list, show the action bar

https://github.com/leather-io/mono/assets/22010816/ac9defc0-77d7-4a95-b88f-31c719dbbc06

## Summary by CodeRabbit

- **New Features**
	- Introduced animated `ActionBar` with show/hide capabilities and smooth transitions.

- **Enhancements**
	- Improved scroll handling in `SendScreen` for better action bar visibility control.
	- Updated text content and localization strings for key actions like "Receive," "Send," "Submit," and "Swap."

- **Refactor**
	- Refactored and modularized `ActionBar` component to use React context and forward references for better state management and scalability.

### DIFF
--- a/apps/mobile/src/app/(tabs)/send.tsx
+++ b/apps/mobile/src/app/(tabs)/send.tsx
@@ -1,5 +1,5 @@
-import { useEffect, useState } from 'react';
-import { FlatList, StyleSheet, View } from 'react-native';
+import { useContext, useEffect, useState } from 'react';
+import { FlatList, NativeScrollEvent, NativeSyntheticEvent, StyleSheet, View } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
 import { ACTION_BAR_TOTAL_HEIGHT } from '@/components/action-bar';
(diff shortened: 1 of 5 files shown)


kyranjamie (commented):
Logic lgtm

> When reaching the bottom of the scroll list, show the action bar

Is this expected behaviour you think? I'd imagine we only show again when scrolling back up, regardless of being at bottom?

apps/mobile/src/app/(tabs)/send.tsx
```diff
@@ -53,24 +55,109 @@ const DATA = [
   },
 ];
 
+const SCROLL_UNTIL_CLOSE = 50;
```
fbwoolf, line 58: Typ we use camelCase for local const in extension if following same in mobile?

fbwoolf: Maybe these will be shared at some point?

fbwoolf (approved):

PR #253, opened 2024-07-02
fix: outdated get address fn

fbwoolf:
This one took me a bit of time to track down. At first we believed it was the update to the signer lib that caused this bug, but as I went through the process to revert/downgrade the signer libs nothing fixed it. I finally tracked it to this function. It was outdated in the mono repo and then installed into the extension with outdated code.

## Summary by CodeRabbit

- **Refactor**
	- Improved the `getAddressFromOutScript` function for better readability and maintainability, now handling multiple script types in a more structured manner.

- **Tests**
	- Updated test suite to use new functions for deriving BIP39 mnemonic and root BIP32 keychain, ensuring compatibility with the latest cryptographic standards.

### DIFF
--- a/packages/bitcoin/src/bitcoin.utils.ts
+++ b/packages/bitcoin/src/bitcoin.utils.ts
@@ -89,32 +89,31 @@ export function decodeBitcoinTx(tx: string): ReturnType<typeof btc.RawTx.decode>
 export function getAddressFromOutScript(script: Uint8Array, bitcoinNetwork: BtcSignerNetwork) {
   const outputScript = btc.OutScript.decode(script);
 
-  if (outputScript.type === 'pk' || outputScript.type === 'tr') {
-    return btc.Address(bitcoinNetwork).encode({
-      type: outputScript.type,
-      pubkey: outputScript.pubkey,
-    });
+  switch (outputScript.type) {
+    case 'pkh':
+    case 'sh':
+    case 'wpkh':
+    case 'wsh':
+      return btc.Address(bitcoinNetwork).encode({
+        type: outputScript.type,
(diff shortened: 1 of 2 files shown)


packages/bitcoin/src/bitcoin.utils.ts
```diff
@@ -89,32 +89,33 @@ export function decodeBitcoinTx(tx: string): ReturnType<typeof btc.RawTx.decode>
...
+    case 'tr':
+      return btc.Address(bitcoinNetwork).encode({
+        type: outputScript.type,
+        pubkey: outputScript.pubkey,
+      });
+    case 'ms':
+      return btc.p2ms(outputScript.m, outputScript.pubkeys).address ?? '';
+    case 'pk':
+      return btc.p2pk(outputScript.pubkey, bitcoinNetwork).address ?? '';
+    case 'tr_ms':
```
pete-watters, line 110: I think we can get rid of `logger.error(`Cannot currently handle script type ${outputScript.type}`)` and catch `tr_ns` and `tr_ms` with the `default` case

pete-watters: I just made that change, hopefully that's OK with you @fbwoolf ?

pete-watters (approved):
Nice work tracking this down 👏

fbwoolf:
@pete-watters has this been released and installed in the extension?

pete-watters:
> @pete-watters has this been released and installed in the extension?

Not yet, I have a PR pending here: https://github.com/leather-io/extension/pull/5594 

I was planning to get it into the next release once approved. Is that OK?

PR #254, opened 2024-07-02
feat: add chevron right icon

fbwoolf:
Adds the chevron right icon for: https://github.com/leather-io/extension/pull/5568#pullrequestreview-2136627292

cc @fabric-8 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `ChevronRightIcon` component offering a right-pointing chevron icon, available in small or regular sizes.
  - Exported the `ChevronRightIcon` component for easier use in other parts of the application.

### DIFF
--- a/packages/ui/src/icons/index.ts
+++ b/packages/ui/src/icons/index.ts
@@ -7,6 +7,7 @@ export * from './checkmark-circle-icon';
 export * from './checkmark-icon';
 export * from './chevron-down-icon';
 export * from './chevron-up-icon';
+export * from './chevron-right-icon.web';
 export * from './chevrons-right-icon';
 export * from './circle-icon';
 export * from './close-icon';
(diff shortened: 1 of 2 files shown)


packages/ui/src/icons/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './checkmark-circle-icon';
 export * from './checkmark-icon';
 export * from './chevron-down-icon';
 export * from './chevron-up-icon';
+export * from './chevron-right-icon.web';
```
pete-watters, line 10: I just added this export so that it will get picked up properly in `web.ts` and will be available in the extension. 

In `web.ts` we load all the icons from index with:
```
export * from './src/icons';
```

FYI I didn't name all the icons as `.web` yet as I was thinking we could possibly share them with mobile. It doesn't make much difference calling them `.web` or not, just explaining the rationale for having a few `.web` and most not

pete-watters (approved):

PR #255, opened 2024-07-02
fix: change spam replacement text, add test cases, ref leather-wallet…

pete-watters:
This PR:
- changes our spam token replacement from `Unknown` to `Suspicious`
- adds explicit test cases to highlight that `.fund` and `.com` will be caught

Any token name or symbol with a `.` is already caught


markmhendrickson (approved):

pete-watters:
I will hold off merging until we are 100% it solves the issue

pete-watters:
Merging this as this function is working but another fix is needed in the extension

PR #259, opened 2024-07-03
refactor: inscriptions infinite query to use bis api

alter-eggo:
This pr refactors  inscriptions infinite query to use BiS api

### DIFF
--- a/packages/query/src/bitcoin/clients/best-in-slot.ts
+++ b/packages/query/src/bitcoin/clients/best-in-slot.ts
@@ -2,8 +2,6 @@ import axios from 'axios';
 import { ZodError, z } from 'zod';
 
 import {
-  BESTINSLOT_API_BASE_URL_MAINNET,
-  BESTINSLOT_API_BASE_URL_TESTNET,
   type BitcoinNetworkModes,
   MarketData,
   Money,
@@ -134,8 +132,7 @@ interface RunesOutputsByAddressResponse {
   data: RunesOutputsByAddress[];
 }
 
-interface BestInSlotInscriptionByAddressArgs {
-  address: string;
+interface BestInSlotInscriptionByAddressDefaultArgs {
   network?: WalletDefaultNetworkConfigurationIds;
   sortBy?: 'inscr_num';
   order?: 'asc' | 'desc';
@@ -144,6 +141,14 @@ interface BestInSlotInscriptionByAddressArgs {
   signal?: AbortSignal;
   exclude_brc20?: boolean;
   cursed_only?: boolean;
+}
+
+interface BestInSlotInscriptionByAddressArgs extends BestInSlotInscriptionByAddressDefaultArgs {
+  address: string;
+}
+
+interface BestInSlotInscriptionByAddressesArgs extends BestInSlotInscriptionByAddressDefaultArgs {
+  addresses: string[];
 }
 
 interface BestinSlotInscriptionByAddressResponse {
@@ -185,22 +190,12 @@ const inscriptionsByAddressSchema = z.object({
   data: z.array(bestInSlotInscriptionSchema),
 });
 
-export function BestinSlotApi() {
-  const url = BESTINSLOT_API_BASE_URL_MAINNET;
-  const testnetUrl = BESTINSLOT_API_BASE_URL_TESTNET;
-
-  const defaultOptions = {
-    headers: {
-      'x-api-key': `${process.env.BESTINSLOT_API_KEY}`,
-    },
-  };
-
+export function BestinSlotApi(basePath: string) {
   /**
    * @see https://docs.bestinslot.xyz/reference/api-reference/ordinals-and-brc-20-and-runes-and-bitmap-v3-api-mainnet+testnet+signet/wallets#get-wallet-inscriptions
    */
   async function getInscriptionsByAddress({
     address,
-    network = WalletDefaultNetworkConfigurationIds.mainnet,
     sortBy = 'inscr_num',
     order = 'asc',
     offset = 0,
@@ -209,8 +204,6 @@ export function BestinSlotApi() {
     exclude_brc20 = false,
     signal,
   }: BestInSlotInscriptionByAddressArgs) {
-    const baseUrl = network === 'mainnet' ? url : testnetUrl;
-
     const queryParams = new URLSearchParams({
       address,
       sort_by: sortBy,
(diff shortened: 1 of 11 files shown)


packages/query/src/bitcoin/ordinals/inscriptions-infinite.query.ts
```diff
@@ -0,0 +1,161 @@
...
+import { QueryPrefixes } from '../../query-prefixes';
+import { useBestInSlotApiRateLimiter } from '../../rate-limiter/best-in-slot-limiter';
+import { useBitcoinClient } from '../clients/bitcoin-client';
+
+const stopSearchAfterNumberAddressesWithoutOrdinals = 5;
+const addressesSimultaneousFetchLimit = 5;
+
+const inscriptionsLazyLoadLimit = 20;
+
+export function useBestInSlotGetInscriptionsInfiniteQuery({
```
alter-eggo, line 19: I've added sep request, as "hiro" one is used in some other hooks, like e.g. `useNumberOfInscriptionsOnUtxo`
I'll refactor it separately, bc `useNumberOfInscriptionsOnUtxo` doesn't work properly and actually shouldn't use infinite query

packages/query/src/bitcoin/clients/best-in-slot.ts
```diff
@@ -222,7 +223,44 @@ export function BestinSlotApi() {
...
+    network = WalletDefaultNetworkConfigurationIds.mainnet,
+    sortBy = 'inscr_num',
+    order = 'asc',
+    offset = 0,
+    // 2000 is the maximum count
+    count = 100,
+    exclude_brc20 = false,
+    signal,
+  }: BestInSlotInscriptionByAddressesArgs) {
+    const baseUrl = network === 'mainnet' ? url : testnetUrl;
```
kyranjamie, line 250: whenNetwork?

packages/query/src/bitcoin/ordinals/inscriptions-infinite.query.ts
```diff
@@ -0,0 +1,161 @@
...
+        {}
+      );
+    },
+    [taprootKeychain, network.chain.bitcoin.bitcoinNetwork]
+  );
+
+  const query = useInfiniteQuery({
+    queryKey: [QueryPrefixes.GetInscriptions, 'bis', nativeSegwitAddress, network.id],
+    async queryFn({ pageParam, signal }) {
+      const responsesArr: any[] = [];
```
kyranjamie, line 50: Can we avoid use of any?

packages/query/src/bitcoin/clients/best-in-slot.ts
```diff
@@ -222,7 +223,44 @@ export function BestinSlotApi() {
...
-      { ...defaultOptions, signal }
+      { signal }
+    );
+
+    try {
+      return inscriptionsByAddressSchema.parse(resp.data);
+    } catch (e) {
+      // TODO: should be analytics
+      // eslint-disable-next-line no-console
+      if (e instanceof ZodError) console.log('schema_fail', e);
```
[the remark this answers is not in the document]
pete-watters: I think analytics isn't setup in the mono-repo at all yet

packages/query/src/bitcoin/ordinals/inscriptions-infinite.query.ts
alter-eggo: good catch, missed that
I believe we need at least warning for any?

packages/query/src/bitcoin/clients/best-in-slot.ts
alter-eggo: there is type error bc of we pass to whenNetwork `NetworkModes`, but here network type is `BitcoinNetworkModes`

kyranjamie: `bitcoinNetworkModeToCoreNetworkMode` for cases like this

packages/query/src/bitcoin/ordinals/inscriptions-infinite.query.ts
```diff
@@ -0,0 +1,161 @@
...
+        fromIndex + addressesSimultaneousFetchLimit
+      );
+
+      let addressesData = getTaprootAddressData(
+        fromIndex,
+        fromIndex + addressesSimultaneousFetchLimit
+      );
+
+      let offset = pageParam.offset;
+      // Loop through addresses until we reach the limit, or until we find an address with many inscriptions
```
pete-watters, line 64: Maybe this could be extracted into a well named standalone function so we wouldn't need the comment?

packages/query/src/rate-limiter/best-in-slot-limiter.ts
```diff
@@ -0,0 +1,36 @@
...
+  timeout: 60000,
+});
+
+export function useBestInSlotApiRateLimiter(): PQueue {
+  const currentNetwork = useCurrentNetworkState();
+
+  switch (currentNetwork.mode) {
+    case 'mainnet':
+      return bestInSlotMainnetApiLimiter;
+    case 'testnet':
```
pete-watters, line 28: Maybe we should add a `default` here too for `bestInSlotTestnetApiLimiter` in case its a custom network?

pete-watters (approved):
This LGTM, great work 👍 I asked a couple of Q's but no blockers so resolve the comments if you need to

packages/query/src/bitcoin/ordinals/inscription.utils.ts
```diff
@@ -31,7 +26,7 @@ export function createBestinSlotInscription(
   inscription: BestinslotInscriptionResponse
 ): Inscription {
   const id = inscription.inscription_id;
-  const contentSrc = `${HIRO_INSCRIPTIONS_API_URL}/${id}/content`;
+  const contentSrc = inscription.content_url;
```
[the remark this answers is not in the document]
alter-eggo: hm, not sure that we need it. we expect inscription to be there, otherwise there is problem in the code above

packages/query/src/rate-limiter/best-in-slot-limiter.ts
alter-eggo: network mode could be only `mainnet` or `testnet`. we have helper fn for such cases `bitcoinNetworkModeToCoreNetworkMode`

packages/query/src/bitcoin/ordinals/inscriptions-infinite.query.ts
alter-eggo: this whole logic could be refactored/rethink, I'd prefer to do it overall later rather than partially now

PR #272, opened 2024-07-15
refactor: export stacks query fns - part 1

fbwoolf:
Hoping to get feedback just to make sure this is how people are thinking these queries should be refactored. I used the naming convention `createDefaultFungibleTokenMetadataQueryOptions` rather than referring to it as just a function to export the queryFn, it is really inclusive of all default query options, correct? This is working, so I can move forward to refactor them all if there are no comments.

I know @alter-eggo questioned needing to keep the separate fetch function, not sure if nec or if it should be exported? I can remove as I go if preferred?

Interesting thread here, but it is using a useQuery wrapper:
https://x.com/TkDodo/status/1491451513264574501

codesandbox example from bottom of thread:
https://codesandbox.io/p/sandbox/bold-platform-zgvsl?file=%2Fsrc%2FApp.tsx%3A22%2C3-23%2C75

## Summary by CodeRabbit

- **Refactor**
  - Improved the organization and maintainability of fetching fungible token metadata.
  - Centralized query options creation for fungible token metadata fetching functions.

### DIFF
--- a/packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.query.ts
+++ b/packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.query.ts
@@ -22,20 +22,36 @@ const queryOptions = {
   staleTime,
 };
 
-function fetchFungibleTokenMetadata(client: StacksClient, signal: AbortSignal) {
-  return (principal: string) =>
-    client.getFtMetadata(principal, signal) as unknown as FtAssetResponse;
+interface CreateFungibleTokenMetadataQueryOptionsArgs {
+  address: string;
+  client: StacksClient;
+  network: string;
+}
+export function createFungibleTokenMetadataQueryOptions({
+  address,
+  client,
+  network,
+}: CreateFungibleTokenMetadataQueryOptionsArgs) {
+  return {
+    enabled: !!address,
+    queryKey: ['get-ft-metadata', address, network],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      client.getFtMetadata(address, signal) as unknown as FtAssetResponse,
+    ...queryOptions,
+  } as const;
 }
 
-export function useGetFungibleTokenMetadataQuery(principal: string) {
+export function useGetFungibleTokenMetadataQuery(address: string) {
   const client = useStacksClient();
   const network = useCurrentNetworkState();
 
-  return useQuery({
-    queryKey: ['get-ft-metadata', principal, network.chain.stacks.url],
-    queryFn: ({ signal }) => fetchFungibleTokenMetadata(client, signal)(principal),
-    ...queryOptions,
-  });
+  return useQuery(
+    createFungibleTokenMetadataQueryOptions({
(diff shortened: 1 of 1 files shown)


packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.query.ts
```diff
@@ -22,20 +22,41 @@ const queryOptions = {
...
+  client,
+  network,
+}: CreateFungibleTokenMetadataQueryOptionsArgs) {
+  return {
+    enabled: !!address,
+    queryKey: ['get-ft-metadata', address, network],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      fetchFungibleTokenMetadata(client, signal)(address),
+    ...queryOptions,
+  };
```
kyranjamie, line 46: Should this be `as const` for stricter typings?

packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.query.ts
```diff
@@ -22,20 +22,41 @@ const queryOptions = {
   staleTime,
 };
 
-function fetchFungibleTokenMetadata(client: StacksClient, signal: AbortSignal) {
-  return (principal: string) =>
-    client.getFtMetadata(principal, signal) as unknown as FtAssetResponse;
+export function fetchFungibleTokenMetadata(client: StacksClient, signal: AbortSignal) {
```
kyranjamie, line 25: With this pattern, should `signal` be optional, otherwise if we do call outside RQ, and we don't need cancellation, we'll have to pass an empty `new AbortController()`

kyranjamie (approved):
> I know @alter-eggo questioned needing to keep the separate fetch function, not sure if nec or if it should be exported? I can remove as I go if preferred?

It seems conceivable we'd sometimes want to fire requests outside the context of react query (we do in some cases already), so think it's a good idea to keep separate

Naming looks good, perhaps "default" isn't necessary if we want to keep a little more succinct 

> Interesting thread here, but it is using a useQuery wrapper: https://x.com/TkDodo/status/1491451513264574501

This looks like a variation of `UseAppQueryConfig`. IMO the complexity added to get it working with Typescript, is more of a pain than simply repeating the `useQuery({ ...createDefaultQueryObjectForSomeRequest(), ...andConfiguringFromThereIfNeeded })` where it works perfectly with implicit typings

packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.query.ts
fbwoolf: I don't see that we make the signal optional anywhere currently, is that something I should address at the client level if I remove the extra fetch function? Or, do you all want to keep that additional fetch function with each query?

fbwoolf:
Going to merge this with these changes as Part 1 so reviews don't get confusing. I can address more comments in another PR if people have them.

PR #283, opened 2024-07-17
refactor: stacks queries part 2

fbwoolf:
This PR refactors all the Stacks queries to follow the pattern established  here: https://github.com/leather-io/mono/pull/272

I tried to coordinate a lot of naming, etc with the `...Prefixes` for queryKeys. I'll have to address/coord with the Bitcoin queries next. I kept the naming conventions using, `Get...Query` for the hooks NOT extending the default query options, which are exported from the `.query.ts` file. For those extending the options, with things like `select`, I kept those in `.hooks.ts` ...but open to feedback here on establishing a convention with this work that everyone wants to follow.

`AppUseQueryConfig` has been removed from all Stacks queries, but has been left with a `deprecated` tag in the code until I refactor the Bitcoin queries, at that point I can remove it completely.


kyranjamie (approved):
Great work Fara 👏🏼

PR #310, opened 2024-07-24
fix: handle long token names gracefully, ref leather-wallet/extension…

pete-watters:
This PR improves the display of `ItemLayout` at the base + `sm` breakpoint to stop long token names breaking the UI. I implemented it based on this [bug](https://github.com/leather-io/extension/issues/5673) for Runes. 

You can see here how the change helps the UI for Runes tokens with long names:
https://github.com/user-attachments/assets/5e4173d8-c2d6-43be-92c1-aa10da1e6dcd

Changing `ItemLayout` here also affects some other places where we will use ellipsis in the action popup. I could implement a Runes specific fix but I think it's better to restrict this on small views in a more global manner.

**SWAP** - `Ethereum WBTC via All...`
Screenshot 2024-07-24 at 10 41 51

**Activity List** - `age000-governance`
Screenshot 2024-07-24 at 10 39 25

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `ItemLayout` component to improve handling of the `titleLeft` prop with new styling properties, ensuring better display control and preventing text overflow.
  
- **Bug Fixes**
	- Implemented a maximum width constraint and text overflow handling for titles, allowing text to truncate with an ellipsis when exceeding the specified width.

### DIFF
--- a/packages/ui/src/components/item-layout/item-layout.web.tsx
+++ b/packages/ui/src/components/item-layout/item-layout.web.tsx
@@ -44,7 +44,14 @@ export function ItemLayout({
             {isValidElement(titleLeft) ? (
               titleLeft
             ) : (
-              <styled.span textStyle="label.02">{titleLeft}</styled.span>
+              <styled.span
+                textStyle="label.02"
+                maxWidth={{ base: '175px', md: 'unset' }}
+                overflow="hidden"
+                textOverflow="ellipsis"
+              >
+                {titleLeft}
+              </styled.span>
             )}
             {isSelected && (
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):

pete-watters:
> truncateString

I think we should deprecate `truncateString`, we only use it once for `AccountNameTitle` in `AccountNameSuspense`. 

If we use CSS it's easier to have it adjust based on breakpoint. In this change we only add ellipsis if its on the small breakpoint

PR #320, opened 2024-07-29
feat: link to `Add Wallet` if no wallets created

pete-watters:
This PR makes the Action Bar dynamic so that if no wallets have been created it shows `Add Wallet` and links to `APP_ROUTES.WalletCreateNewWallet`

### DIFF
--- a/apps/mobile/src/app/wallet/(tabs)/_layout.tsx
+++ b/apps/mobile/src/app/wallet/(tabs)/_layout.tsx
@@ -1,84 +1,19 @@
-import { RefObject, createContext, useRef } from 'react';
+import { useRef } from 'react';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
 import EmojiSmile from '@/assets/emoji-smile.svg';
-import Inbox from '@/assets/inbox.svg';
 import Menu from '@/assets/menu.svg';
-import PaperPlane from '@/assets/paper-plane.svg';
-import Swap from '@/assets/swap.svg';
-import { ActionBar, ActionBarMethods } from '@/components/action-bar';
+import { ActionBarMethods } from '@/components/action-bar';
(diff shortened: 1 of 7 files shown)


kyranjamie (changes requested):
Why does this return a anon function, rather than a component that accepts props? 🤔

Underneath the JSX, a component is just a function that accepts arguments, as props, so afaik it isn't necessary to wrap in a fn like this.

apps/mobile/src/app/wallet/(tabs)/_layout.tsx
```diff
@@ -187,7 +218,7 @@ export default function TabLayout() {
           }}
         />
       </Tabs>
-      {createFilledActionBar(ref, router)()}
+      {CreateFilledActionBar(ref, router)()}
```
kyranjamie, line 221: I see this pattern was here from before, but very odd imo and a react anti-pattern, unless there's some RN quirk I'm not aware of

kyranjamie: image

pete-watters:
> Why does this return a anon function, rather than a component that accepts props? 🤔
> 
> Underneath the JSX, a component is just a function that accepts arguments, as props, so afaik it isn't necessary to wrap in a fn like this.

I went for getting it working quickly using what was there already but I can refactor it 👍

kyranjamie (approved):
Way cleaner, thanks @pete-watters

pete-watters:
> Way cleaner, thanks @pete-watters

No problem. I'll make more effort to refactor as I go

pete-watters:
> Looks awesome @pete-watters !! We will also need this action bar on home screen. Also, can you, please, attach a video of UI changes? It's a bit easier to review that way as a reviewer shouldn't change branches and build app on that PR that way.

Ok, I will make sure to add a video in future and add it to the Home Screen

PR #321, opened 2024-07-29
feat: add inscription output value

alter-eggo:
This pr adds inscription output value, genesis ts and block hash to create bis inscription fn


packages/query/src/bitcoin/clients/zod-schemas.ts
```diff
@@ -76,6 +76,9 @@ export const bestInSlotInscriptionSchema = z.object({
   render_url: z.string().nullable().optional(),
   bitmap_number: z.number().nullable().optional(),
   delegate: bestInSlotInscriptionDelegateSchema.nullable().optional(),
+  output_value: z.number(),
+  genesis_ts: z.string(),
+  genesis_block_hash: z.string(),
```
kyranjamie, line 81: They're definitely always there?

kyranjamie (approved):

PR #323, opened 2024-07-30
feat: add network config schema

alter-eggo:
This pr adds network config schema


.dependency-cruiser.cjs
```diff
@@ -11,6 +11,7 @@ module.exports = {
       to: {
         path: '^(?!packages/models)',
         dependencyTypesNot: ['type-only'],
+        pathNot: ['^packages/models/[^/]+/[^/]+', 'node_modules/.pnpm/zod@.*'],
```
pete-watters, line 14: Do you need to do this to use `zod` in `models` ?

pete-watters (approved):

PR #324, opened 2024-07-30
refactor: account state

kyranjamie:
This PR adjusts some of the logic [per discussion on managing accounts](https://trustmachines.slack.com/archives/C05LRS7G44R/p1722327643740619).

I've added a `deleted` status to accounts on which to filter. Intentionally not using a boolean to allow for potential future other states.

Removes `ofKeychains` in account slice as it isn't needed, I'm pretty sure. We can infer the keychains to select based on `fingerprint + accountIndex`.

Improved typing of `handleEntityActionWith` function.


apps/mobile/src/components/wallet-manager/account-list.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+        <Text style={{ fontSize: 16, marginBottom: 8 }}>Account {account.accountIndex}</Text>
+        <Pressable
+          style={{ marginBottom: 12, marginLeft: 8 }}
+          onPress={() => onRemoveAccount(account.accountIndex)}
+        >
+          <Text>❌</Text>
+        </Pressable>
+      </View>
+      {renderKeychains(fingerprint, account.accountIndex)}
```
kyranjamie, line 29: Some blog posts will say: _😡 why u use render props, hooks replace need entirely in every possible use case._ IMO this is still a clean pattern and creates a nice separation between high level state-consuming components from dumb components.

apps/mobile/src/components/wallet-manager/account-list.tsx
```diff
@@ -0,0 +1,32 @@
...
+  onRemoveAccount: (accountIndex: number) => void;
+  renderKeychains(fingerprint: string, accountIndex: number): ReactNode;
+}
+export function AccountsLayout({
+  fingerprint,
+  accounts,
+  onRemoveAccount,
+  renderKeychains,
+}: AccountsLayoutProps) {
+  return accounts.map(account => (
```
pete-watters, line 18: How about de-structuring `account` to keep things DRY? 

```
return accounts.map({accountIndex, id} => (
```

apps/mobile/src/components/wallet-manager/account-list.tsx
pete-watters: I agree and although hooks have their place they can be over-used

apps/mobile/src/components/wallet-manager/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { WalletListLayout } from './wallets';
+
+interface BitcoinKeychainProps {
+  fingerprint: string;
+  accountIndex: number;
+}
+function BitcoinKeychains({ fingerprint, accountIndex }: BitcoinKeychainProps) {
+  const bitcoinKeychains = useBitcoinKeychains();
+  const keychains = bitcoinKeychains.fromAccountIndex(fingerprint, accountIndex);
+  return keychains.map(a => (
```
pete-watters, line 18: Whats an `a`? Clearly account I know but I would again destructure early here - `map({address, descriptor}`

apps/mobile/src/state/utils.ts
```diff
@@ -28,18 +28,20 @@ export const useAppDispatch: () => AppDispatch = useDispatch;
...
+}
+
+type AdapterMethod<T> = (state: EntityState<T, string>, args: any) => void;
 
-export function handleEntityActionWith<T, P>(
-  adapterMethod: AdapterMethod<T>,
-  selectPayload: (payload: P) => unknown
+export function handleEntityActionWith<State, Payload, R extends AdapterMethod<State>>(
+  adapterMethod: R,
+  // Payload selector fn expected to return the value passsed to second parmeter
```
pete-watters, line 39: You have a typo here in `parmeter`

apps/mobile/src/utils/types.ts
```diff
@@ -0,0 +1,5 @@
+import { ReactNode } from 'react';
```
pete-watters, line 1: Should we add a `types` package to share this with the extension? It's not a big deal for one type

pete-watters (approved):
Nice work 🚀 I added a couple of suggestions but am happy to merge. 

Maybe we could add some unit tests for state manipuation in `wallets.slice` and `accounts.slice`?

apps/mobile/src/components/wallet-manager/index.tsx
kyranjamie: Was previously named account, renamed

apps/mobile/src/state/utils.ts
kyranjamie: ta

apps/mobile/src/utils/types.ts
kyranjamie: Was thinking this, but doesn't really belong in models pkg as it's React specific. If anything, maybe a helper type from ui pkg makes sense?

apps/mobile/src/components/wallet-manager/account-list.tsx
kyranjamie: Yep, good point @alter-eggo, map moved

kyranjamie:
I tried using Claude to do some refactoring here and worked pretty well. For readability, it's really important to separate layout, styles, and animation components from those that read and manipulate state, exposing events with the `*.layout.ts` component pattern of similar.

This prompt pretty much does all the work:

```
Take the following component, and break it into two: a layout component, named the same but ending in "Layout` with all
the views and content of the component, and the other, that contains all the usage of React hooks, and reading of state. 
If methods are called from within the Layout component, create callback props with event-based names, such as
`onUserClicksButton`. Prefer interfaces over type declarations.
```

apps/mobile/src/components/wallet-manager/wallets.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+export function WalletListLayout({
+  wallets,
+  onRemoveWallet,
+  onCreateNewAccount,
+  renderAccount,
+}: WalletListLayoutProps) {
+  return (
+    <View style={{ padding: 20 }}>
+      {wallets.map(wallet => (
```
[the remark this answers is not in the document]
kyranjamie: Yes, good point, this is only a demo UI though, so we'll remove this soon. Was mostly tidying up the demo components to make sure reading from the store would work well in the actual UI.

PR #326, opened 2024-07-31
feat: implement security screen + toast + simple headers + localization

edgarkhanzadian:
- Security screen
- Toaster UI
- Update localization strings. Implement TransText component that is a Trans macro wrapped in Text component
- Use simple (no blur) headers on some screens

https://github.com/user-attachments/assets/602afebf-1733-43ac-9494-14848e80e044


apps/mobile/src/assets/cross-large.svg
```diff
@@ -1,3 +1,4 @@
 <svg width="25" height="24" viewBox="0 0 25 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M5.5 5L19.5 19M19.5 5L5.5 19" stroke="#12100F" stroke-width="2" stroke-linecap="round"/>
+  <path d="M5.5 5L19.5 19M19.5 5L5.5 19" stroke="currentColor" stroke-width="2"
+    stroke-linecap="round" />
```
kyranjamie, line 3: Editor pretty printer?

apps/mobile/src/components/blurred-header.tsx
```diff
@@ -21,6 +21,61 @@ type NativeStackHeaderProps = Parameters<
 
 export const HEADER_HEIGHT = 64;
 
+export function createSimpleHeader(props: {
+  left?: ReactNode;
+  center?: ReactNode;
+  right?: ReactNode;
+  insets: EdgeInsets;
+}) {
```
kyranjamie, line 29: Why not just a component accepting these props? Not clear why this needs to be a HOC

apps/mobile/src/components/secure-your-wallet/skip-secure-wallet-modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { Box, Theme, TouchableOpacity } from '@leather.io/ui/native';
+
+import { CLOSED_ANIMATED_POSITION, Modal } from '../bottom-sheet-modal';
+import { Button } from '../button';
+import { TransText } from '../trans-text';
+
+interface SkipSecureWalletModalProps {
+  skipSecureWalletModalRef: RefObject<BottomSheetModal>;
+  onSkip(): unknown;
```
kyranjamie, line 16:
```suggestion
  onSkip(): void;
```

more common typing here if returning nothing?

apps/mobile/src/components/blurred-header.tsx
```diff
@@ -21,6 +21,61 @@ type NativeStackHeaderProps = Parameters<
...
+          <Box height="100%" flex={1} alignItems="flex-end" justifyContent="center">
+            {props?.right}
+          </Box>
+        </Box>
+      </Box>
+    );
+  };
+}
+
+export function HeaderAvoidingContainer(props: PropsWithoutRef<Parameters<typeof Box>>['0']) {
```
pete-watters, line 66: Maybe this could be named a bit better?

apps/mobile/src/app/wallet/secure-your-wallet.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export default function () {
+  const { bottom } = useSafeAreaInsets();
+  const theme = useTheme<Theme>();
+  const skipSecureWalletModalRef = useRef<BottomSheetModal>(null);
+  const router = useRouter();
+  const toastContext = useToastContext();
+
+  function createWallet() {
+    toastContext.displayToast({ type: ToastType.success, title: 'Wallet added successfully' });
```
kyranjamie, line 27:
```suggestion
    toastContext.displayToast({ type: 'success', title: 'Wallet added successfully' });
```

Using a string literal > enum means better ts autocompletion & no import statement needed. Overall better DX imo

apps/mobile/src/components/toast/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const top = useSharedValue<number>(TOP_POSITION.closed);
+
+  useImperativeHandle<ToastMethods, ToastMethods>(toastRef, () => ({
+    display(data) {
+      setToastData(data);
+      top.value = withTiming(TOP_POSITION.opened);
+    },
+  }));
+
+  function getIcon(type: ToastType) {
```
pete-watters, line 31: I think we could move `getIcon` + `getBackground` out of the main `Toast` scope into their own standalone helpers

pete-watters (approved):
Great work! 👏 

Can you clean up the commit history a bit to squash the commits and also use a rebase from `dev` rather than a merge?

apps/mobile/src/app/wallet/_layout.tsx
```diff
@@ -26,17 +26,17 @@ export default function StackLayout() {
     center: <AccountHeader />,
     right: <MenuHeader />,
   });
-  const NavigationHeaderHome = createBlurredHeader({
+  const NavigationHeaderHome = createSimpleHeader({
     insets,
     left: <LeatherLogoHeader />,
     right: <OptionsHeader onPress={() => router.navigate(APP_ROUTES.WalletDeveloperConsole)} />,
   });
```
kyranjamie, line 33:
```
function NavigationHeaderHome () {
  return <SimpleHeader {...elementProps} {...otherProps}  />
}
``` 
?

kyranjamie (commented):
What patterns should we use going forward re `TransText` `t` etc?

apps/mobile/src/components/blurred-header.tsx
kyranjamie: isn't that just the difference between `<SomeComponent />` and `() => <SomeComponent />`?

apps/mobile/src/components/secure-your-wallet/skip-secure-wallet-modal.tsx
kyranjamie: Sure, not particularly fussed either way. Claude has this to say:

_In Typescript, what's the difference between `callback(): void` and `callback(): unknown`_
```
Usage Scenarios
void:

Use void when you do not expect a callback to provide any return value and you are only interested in the side effects of the function.
Suitable for event handlers, logging functions, or any situation where the return value is not needed.
unknown:

Use unknown when you expect the callback to return some kind of value, but you don't know the specific type at the time of defining the callback.
Suitable for situations where functions can return various types of results, and you need to handle each case safely with type checks.
```

apps/mobile/src/components/blurred-header.tsx
pete-watters: Ah OK don't worry then as I just didn't know about `KeyboardAvoidingView`

PR #331, opened 2024-08-02
Stacks package and key generation

kyranjamie:
Initial work adding stacks key gen and fleshing out `@leather.io/stacks` package.

I went back and forward thinking how to save Stacks keys, and figured reusing format of BIP380 makes most sense. This spec covers regular pub keys, not just xpubs as with bitcoin accounts.

image

https://github.com/user-attachments/assets/291206d5-ed2d-45cf-a49d-6375b90b8cfa

Todo: 
- [x] Refactor to make code unit testable


fbwoolf (approved):
🚀

PR #354, opened 2024-08-09
fix: display active developer console functionality first

edgarkhanzadian:
Just for convenience and faster development, we can change it later if this becomes a customer facing screen
console


kyranjamie (approved):
Nice, scrolling down the list was becoming a real chore haha

PR #356, opened 2024-08-09
feat: add lingui eslint and localize strings

edgarkhanzadian:
lingui eslint plugin: https://github.com/lingui/eslint-plugin

Generally, let's try to always use `t` unless we need to encompass multiple components in Trans component, e.g.:
```
            <Trans>
              <Animated.View style={title1Style}>
                <Text variant="heading02" color="ink.text-primary">
                  Bitcoin for
                </Text>
              </Animated.View>
              <Animated.View style={title2Style}>
                <Text variant="heading02" color="ink.text-primary">
                  the rest of us
                </Text>
              </Animated.View>
            </Trans>
```


kyranjamie (approved):
Nice 👍🏼 

Also noticed we're missing the react eslint plugin to that warns about missing effect deps

fbwoolf (approved):

PR #358, opened 2024-08-13
fix: remove dropdown group if no children

alter-eggo:
Fix per @markmhendrickson [comment](https://github.com/leather-io/extension/pull/5726#issuecomment-2275067678)
There is no sense in rendering dropdown group wrapper if there are no children.

### DIFF
--- a/packages/ui/src/components/dropdown-menu/dropdown-menu.web.tsx
+++ b/packages/ui/src/components/dropdown-menu/dropdown-menu.web.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, forwardRef } from 'react';
+import { Children, ReactNode, forwardRef } from 'react';
 
 import * as RadixDropdownMenu from '@radix-ui/react-dropdown-menu';
 import { css } from 'leather-styles/css';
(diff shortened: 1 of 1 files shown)


kyranjamie (changes requested):
I don't think this should be implemented here at the UI library level, in the same way that Radix doesn't include this logic. These should be dumb components unaware of their contents.

The question is: why does a component with no children render `DropdownMenu.Group` in the first place?

kyranjamie:
But even with these changes, we'd still render the bottom divider, which doesn't fix the problem.

The responsibility of this component is currently to only add a style wrapper around Radix. Here that rule is broken by adding business logic. We can't be sure we always want this behaviour. [React docs also suggest caution](https://react.dev/reference/react/Children) with `Children`. If this behaviour was so obvious, it'd be in Radix.

IMO It's a very dangerous game trying to solve problems you encounter in **one instance** of a UI component's use, by making changes to the UI component itself.

If you really want to use this approach, create a wrapper component of DropdownMenu.Group that has this conditional rendering `Children` logic.

PR #361, opened 2024-08-13
feat: load balances and additional data separately

alter-eggo:
This pr separates fetching balances from additional data (pending txs, inscriptions, etc)

### DIFF
--- a/packages/query/src/bitcoin/address/utxos-by-address.hooks.ts
+++ b/packages/query/src/bitcoin/address/utxos-by-address.hooks.ts
@@ -65,19 +65,32 @@ interface UseFilterUtxosByAddressArgs {
 
 type filterUtxoFunctionType = (utxos: UtxoResponseItem[]) => UtxoResponseItem[];
 
+function useUtxosByAddressQuery(address: string) {
+  const client = useBitcoinClient();
+
+  return useQuery({
+    ...createGetUtxosByAddressQueryOptions({ address, client }),
+  });
+}
+
 export function useNativeSegwitUtxosByAddress({
   address,
   filterInscriptionUtxos,
   filterPendingTxsUtxos,
   filterRunesUtxos,
 }: UseFilterUtxosByAddressArgs) {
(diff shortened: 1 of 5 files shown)


packages/query/src/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -71,13 +71,20 @@ export function useNativeSegwitUtxosByAddress({
   filterPendingTxsUtxos,
   filterRunesUtxos,
 }: UseFilterUtxosByAddressArgs) {
+  const client = useBitcoinClient();
+
+  const initialUtxosQuery = useQuery({
+    ...createGetUtxosByAddressQueryOptions({ address, client }),
+  });
```
kyranjamie, line 78: Think we should make a local wrapper component than using directly in the component?

kyranjamie (approved):

packages/query/src/stacks/balance/account-balance.hooks.ts
```diff
@@ -55,13 +53,45 @@ export function useStxCryptoAssetBalance(address: string) {
...
+    isLoadingAdditionalData: query.isLoading,
+  };
 }
 
 export function useStxAvailableUnlockedBalance(address: string) {
-  return useStxCryptoAssetBalance(address).data?.unlockedBalance ?? createMoney(0, 'STX');
+  return (
+    useStxCryptoAssetBalance(address).processedBalanceQuery.data?.unlockedBalance ??
+    createMoney(0, 'STX')
+  );
```
kyranjamie, line 94:
```suggestion
  const stxBalance = useStxCryptoAssetBalance(address);
  return (
    stxBalance.processedBalanceQuery.data?.unlockedBalance ??
    createMoney(0, 'STX')
  );
```

also `processedBalanceQuery` is odd name to export from a hook I think?

kyranjamie (approved):

packages/query/src/stacks/balance/account-balance.hooks.ts
kyranjamie: What is the relevance that it's been filtered or processed?

PR #366, opened 2024-08-14
fix(mobile): status bar colour

kyranjamie:
Before

image

After

https://github.com/user-attachments/assets/de81565c-2223-4cfb-848c-8e9bb95f2f4a

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -1,4 +1,4 @@
-import { StatusBar, useColorScheme } from 'react-native';
+import { StatusBar } from 'react-native';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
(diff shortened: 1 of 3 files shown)


apps/mobile/src/hooks/use-theme.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function useTheme() {
+  const settings = useSettings();
+  const systemTheme = useColorScheme();
+  const themeMode = (settings.theme === 'system' ? systemTheme : settings.theme) ?? 'light';
+  return {
+    theme: themeMode,
+    whenTheme: whenTheme(themeMode),
+  };
+}
```
kyranjamie, line 21: wonder if should just put this in use settings?

kyranjamie: done

PR #367, opened 2024-08-14
feat: notify on existing wallet

edgarkhanzadian:
https://github.com/user-attachments/assets/c6a4ea6f-6f01-4a6d-9751-685cd07d1196


kyranjamie (approved):
LGTM

Only thought is if it'd make sense for the error type / id, and function the check it's this specific error, to be colocated in same file?

```
function throwKeyExistsError() {}
function isKeyExistsError() {}
```

and for these to imported to respected components?

PR #373, opened 2024-08-16
feat: change stacks client requests priorities

alter-eggo:
This pr changes stacks requests queue priorities, to load balance specific data first

### DIFF
--- a/packages/query/src/stacks/stacks-client.ts
+++ b/packages/query/src/stacks/stacks-client.ts
@@ -39,6 +39,26 @@ export interface CallReadOnlyFunctionArgs {
   signal?: AbortSignal;
 }
 
+const priorityLevels = {
+  getNetworkStatus: 6,
+  getAccountTransactionsWithTransfers: 5,
+  getNamesOwnedByAddress: 5,
+  getAccountBalance: 4,
+  getAccountNonces: 4,
+  getNameInfo: 4,
+  getNftHoldings: 4,
+  getAddressMempoolTransactions: 4,
+  getRawTransactionById: 4,
+  getTransactionById: 4,
+  getContractInterface: 4,
+  getNetworkBlockTimes: 3,
+  postFeeTransaction: 2,
+  getFtMetadata: 2,
+  getNftMetadata: 2,
+  callReadOnlyFunction: 2,
+  getStx20Balances: 1,
+};
+
 export function stacksClient(basePath: string) {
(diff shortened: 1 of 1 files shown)


kyranjamie (commented):
These are feeling a bit magic number-y.

Do you have any ideas how we might improve this? What if we do want a priority call on `getAccountBalance` in another occasion?

packages/query/src/stacks/stacks-client.ts
```diff
@@ -49,7 +49,7 @@ export function stacksClient(basePath: string) {
           axios.get<AddressBalanceResponse>(`${basePath}/extended/v1/address/${address}/balances`, {
             signal,
           }),
-        { priority: 1, signal, throwOnTimeout: true }
+        { priority: 4, signal, throwOnTimeout: true }
```
kyranjamie, line 52: Why is get account balance less important that network status?

packages/query/src/stacks/stacks-client.ts
```diff
@@ -39,6 +39,26 @@ export interface CallReadOnlyFunctionArgs {
   signal?: AbortSignal;
 }
 
+const PRIORITY_LEVELS = {
```
kyranjamie, line 42:
```suggestion
const priorityLevels = {
```

no need for upper casing here

kyranjamie (approved):
This is more readable, though think I'd have gone for 

```
const priority = {
  high: 5,
  medium: 3
  ...etc
}
```

packages/query/src/stacks/stacks-client.ts
kyranjamie: What if we need different priorities in web and mobile?

kyranjamie: It could yeah. Imagine this:

```ts
// mobile
const query = useSomeQuery(createQueryConfig({ ...args, priority: 'high' }))
// web
const query = useSomeQuery(createQueryConfig({ ...args, priority: 'low' }))
```

kyranjamie: We can choose a scale of whatever granularity we need, so we can add very-high, or use numbers. Important thing is each platform decides.

> also managing request priorities when they are in request itself is quite painful

Agree. I'm not really sure we want to go too far down the path of juggling request priorities. What is most important isn't fixed per app, it varies depending on route/what the users doing etc, so while setting a priority can be helpful, maybe we shouldn't rely on it too much 🤹🏼

PR #376, opened 2024-08-18
feat: mobile ui svg icons

fbwoolf:
This PR relocates all svgs from the mobile app into the UI library. They are now exported from the UI library as React components with a default size of 24x24 and a 'small' variant of 16x16. All other SvgProps can be passed to change color, size, etc. The theme is used in the UI library to set the 'currentColor' as 'ink.action-primary-default', so the icons work with dark mode.

The native UI library build is now using tsup (aka 'sup'). 😆 Once this is merged, I can refactor the existing 'web' icons to use the svgs, but I wanted to separate the work into two PRs.

Preview of use...
Screenshot 2024-08-15 at 4 21 42 PM

signal-2024-08-18-192320_002

signal-2024-08-18-192320_003


packages/ui/app.json
```diff
@@ -31,6 +31,17 @@
...
-    "plugins": ["expo-font"]
+    "plugins": [
+      "expo-font",
+      {
+        "fonts": [
+          "../assets-native/fonts/SpaceMono-Regular.ttf",
+          "../assets-native/fonts/ABCDiatype-Regular.otf",
+          "../assets-native/fonts/ABCDiatype-Light.otf",
+          "../assets-native/fonts/ABCDiatype-Medium.otf",
+          "../assets-native/fonts/marche-super-pro.otf"
```
kyranjamie, line 42: Can we rename to be all the same casing?

packages/ui/package.json
```diff
@@ -4,12 +4,12 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
-    "assets:native": "mkdir -p dist-native/src && cp -r src/assets-native ./dist-native/src/",
-    "assets:web": "mkdir -p dist-web/src && cp -r src/assets-web ./dist-web/src",
+    "assets:native": "mkdir -p dist-native/src && cp -r src/assets-native ./dist-native/src && cp -r src/assets ./dist-native/src",
+    "assets:web": "mkdir -p dist-web/src && cp -r src/assets-web ./dist-web/src && cp -r src/assets ./dist-web/src",
```
kyranjamie, line 8: Not a fan of having these copy commands in the package.json. tsup doesn't have a "beforeBuild" like it does `onSuccess` but [per comment here](https://github.com/egoist/tsup/issues/1094#issuecomment-1987318528), anything in the build script runs before the config.

So, could we move these directory/asset copying actions there, using a node script to do the same?

kyranjamie (approved):
Fantastic work Fara, this is a really tidy set up. Definitely worth the time you spent investigating this to get to this result.

pete-watters (approved):
Great work @fbwoolf 👏

packages/ui/src/icons/arrow-left-icon.native.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <Icon {...props}>
+        <ArrowLeftSmall />
+      </Icon>
+    );
+  return (
+    <Icon {...props}>
+      <ArrowLeft />
+    </Icon>
+  );
+}
```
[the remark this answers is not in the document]
kyranjamie: @edgarkhanzadian there are separate SVG files to handle small pixel rendering differences at smaller sizes

packages/ui/package.json
fbwoolf: I meant to come back to this bc came across a suggestion to use this plugin to copy assets in the tsup config:
https://github.com/LinbuduLab/esbuild-plugins/tree/main/packages/esbuild-plugin-copy

packages/ui/app.json
fbwoolf: Let me check this.

packages/ui/src/icons/arrow-left-icon.native.tsx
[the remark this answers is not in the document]
fbwoolf: @edgarkhanzadian the designers asked us not to scale svgs b/w 16x16 and 24x24, so we have two separate exports for the 'default' and 'small' variants. If another size is needed, we can then scale from the default.

fbwoolf:
@kyranjamie I'm getting a 'not-to-dev' dependency error here with tsup. Docs say it should be installed as a dev dependency, but we had it as a dependency. I moved it and added `tsup` to the `pathNot`, let me know if that isn't the correct way to handle that error. I also gitignored the generated `.tsup` folder.

kyranjamie:
Config files like tsup shouldn't be subject to rules of depcruiser, so good to ignore (even globally)

packages/ui/app.json
fbwoolf: The fonts aren't broken that I can see, do we need them in both places?

fbwoolf: Nvm, think I am wrong here.

PR #377, opened 2024-08-19
fix: prevent unnecessary nft requerying

alter-eggo:
This pr prevents re-querying of nft metadata using only useQuery, without keeping ids in store like here https://github.com/leather-io/extension/pull/5747

### DIFF
--- a/packages/query/src/stacks/token-metadata/non-fungible-tokens/non-fungible-token-metadata.query.ts
+++ b/packages/query/src/stacks/token-metadata/non-fungible-tokens/non-fungible-token-metadata.query.ts
@@ -2,7 +2,7 @@ import { hexToCV } from '@stacks/transactions';
 import { QueryFunctionContext, type UseQueryResult, useQueries } from '@tanstack/react-query';
 import { AxiosError } from 'axios';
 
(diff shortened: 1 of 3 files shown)


packages/query/src/stacks/token-metadata/non-fungible-tokens/non-fungible-token-metadata.query.ts
```diff
@@ -34,10 +35,22 @@ export function createGetNonFungibleTokenMetadataQueryOptions({
...
+
   return {
     enabled: !!tokenId,
-    queryKey: [StacksQueryPrefixes.GetNftMetadata, address, tokenId],
-    queryFn: ({ signal }: QueryFunctionContext) => client.getNftMetadata(address, tokenId, signal),
+    queryKey,
+    queryFn: async ({ signal }: QueryFunctionContext) => {
+      try {
+        const response = await client.getNftMetadata(address, tokenId, signal);
+        return response;
```
kyranjamie, line 46:
```suggestion
        return await client.getNftMetadata(address, tokenId, signal);
```

kyranjamie (approved):
This is a more simple solution than what I'd proposed with a new slice 👍🏼

PR #379, opened 2024-08-19
fix: dialog header height + refactor dialog footer

pete-watters:
This PR:
- refactors `Dialog` to locate the `Footer` component local to it and avoid using it in the extension
- Adds a min height to dialog headers

### DIFF
--- a/packages/ui/src/components/dialog/dialog-footer.web.tsx
+++ b/packages/ui/src/components/dialog/dialog-footer.web.tsx
@@ -1,0 +1,26 @@
+import { Flex, FlexProps } from 'leather-styles/jsx';
+
+interface DialogFooterProps extends FlexProps {
+  children: React.ReactNode;
+}
+
+export function DialogFooter({ children, ...props }: DialogFooterProps) {
+  return (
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/dialog/dialog-footer.web.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export function DialogFooter({ children, ...props }: DialogFooterProps) {
+  return (
+    <Flex
+      gap="space.05"
+      p="space.05"
+      bottom={0}
+      width="100vw"
+      maxWidth="100%"
```
alter-eggo, line 14: just `width: 100%` is not the same? 🤔

alter-eggo (approved):

packages/ui/src/components/dialog/dialog-footer.web.tsx
pete-watters: Thanks @alter-eggo ✋ 

You're right, using `width: 100%` works here. I tested it and updated it

PR #381, opened 2024-08-20
fix: ui storybook

edgarkhanzadian:
I was hitting this issue running storybook native: 

Simulator Screenshot - iPhone 15 Pro - 2024-08-20 at 10 46 44

https://trustmachines.slack.com/archives/C05LAP952E7/p1724147525864839 

@edgarkhanzadian kindly helped me spot the problem and this PR resolves it

PR #383, opened 2024-08-20
chore: add native flag component, ref ref leather-io/issues#262

pete-watters:


packages/ui/src/components/flag/flag.native.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type { Meta, StoryObj } from '@storybook/react';
+
+import { Box } from '../box/box.native';
+import { Flag } from './flag.native';
+
+const meta: Meta<typeof Flag> = {
+  title: 'Layout/Flag',
+  component: Flag,
+  tags: ['autodocs'],
+  argTypes: {
```
pete-watters, line 12: Ideally these controls would work in the simulator but we will need to investigate fixing that as [we disabled add ons](https://github.com/leather-io/mono/pull/381) for now so storybook would load

packages/ui/src/components/flag/flag.shared.tsx
```diff
@@ -0,0 +1 @@
+export type FlagAlignment = 'top' | 'middle' | 'bottom';
```
pete-watters, line 1: I tried adding these in a `types.ts` file but struggled with the CI failing with `flag.types.ts` is not listed within the file list of the project. 

I tried to configure `tsconfig` but couldn't get it working. 

@fbwoolf I see this is what you did to share types in icons. Did you hit similar issues? 

I thought that just adding `types.ts` to `tsconfig.json` should work

fbwoolf: Basically, with ts and builds, any files with extension `.native` or `.web` are included in those separate builds AND `.shared` are included in both.

packages/ui/src/components/flag/flag.native.tsx
```diff
@@ -0,0 +1,52 @@
...
+          flexDirection: reverse ? 'row-reverse' : 'row',
+          alignItems: getFlagAlignment(align),
+        },
+        viewBoxStyles,
+      ]}
+    >
+      <View style={styles.container}>{img}</View>
+      <View style={styles.container}>{children}</View>
+    </View>
+  );
```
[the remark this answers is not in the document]
pete-watters: Thanks @edgarkhanzadian , I'll change this. I wasn't sure what to use and said I may as well ask via a PR

packages/ui/src/components/flag/flag.shared.tsx
fbwoolf: I think what you have is fine tbh, but could just be `flag.shared.ts`.

packages/ui/src/components/flag/flag.native.tsx
```diff
@@ -0,0 +1,52 @@
...
+      style={[
+        styles.container,
+        {
+          flexDirection: reverse ? 'row-reverse' : 'row',
+          alignItems: getFlagAlignment(align),
+        },
+        viewBoxStyles,
+      ]}
+    >
+      <View style={styles.container}>{img}</View>
```
fbwoolf, line 42: I think I'm not understanding `styles.container` here and how it is handling what the web version of `Flag` does? This code: https://github.com/leather-io/mono/blob/68bb5075d03fcb3e8b530df2ad457679952eb5ff/packages/ui/src/components/flag/flag.web.tsx#L50

pete-watters: I didn't add any spacing yet as I wasn't sure. I can add it, no problem. 

Are we working off a visual design here? I can just add the same spacing as web `space.03`

packages/ui/src/components/flag/flag.shared.tsx
pete-watters: Thanks @fbwoolf . I did try adding it to the includes yesterday but I couldn't get it working. 

I'll investigate more as I add more components to see if I can move it but at least now its doing what we want - sharing types

packages/ui/src/components/flag/flag.native.tsx
pete-watters: I've implemented this properly now to apply `margin`  appropriately on `reverse`

pete-watters:
@edgarkhanzadian : Thanks for the approval 🙇 

I have refactored it to use `Box` as you suggested and to use inline props. Do you mind taking another look to double check?

packages/ui/src/components/flag/flag.native.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+};
+
+export function Flag({
+  align = 'middle',
+  img,
+  children,
+  reverse = false,
+  spacing = '3',
+  boxProps = {},
```
pete-watters, line 35: @edgarkhanzadian : I have a Q about how I want this component to be extensible. 

On web we have `Flag` which extends `FlexProps`. In this PR I try to achieve the same thing by allowing `style={boxStyles}` so that we could pass extra props in if we need to: 

```
<Flag marginLeft={X
```

Is this the right pattern to follow?

packages/ui/src/components/flag/flag.native.tsx
fbwoolf: Yeah, from what I understand, with these two components (Flag and ItemLayout) we are just trying to match the web version as they are layout components which should be consistent b/w web and mobile.

cc @mica000 if I'm not correct here?

fbwoolf (approved):

packages/ui/src/components/flag/flag.native.tsx
pete-watters: Thanks @fbwoolf . I'll just build them with the same setup and we can then update them when consuming them if needed

PR #384, opened 2024-08-21
feat: prevent ft tokens unnecessary requerying

alter-eggo:
This pr prevents ft tokens unnecessary requerying. same logic as for nfts [here](https://github.com/leather-io/mono/pull/377)

### DIFF
--- a/packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.hooks.ts
+++ b/packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.hooks.ts
@@ -27,7 +27,7 @@ export function useStacksFungibleTokensBalance(
           client,
           network: network.chain.stacks.url,
         }),
-        select: (resp: FtAssetResponse) => {
+        select: (resp: FtAssetResponse | null) => {
           if (!(resp && isFtAsset(resp))) return;
           const { contractAssetName } = getStacksContractIdStringParts(key);
           const name = resp.name || contractAssetName;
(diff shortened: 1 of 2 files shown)


packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.hooks.ts
```diff
@@ -27,7 +27,7 @@ export function useStacksFungibleTokensBalance(
           client,
           network: network.chain.stacks.url,
         }),
-        select: (resp: FtAssetResponse) => {
+        select: (resp: unknown) => {
```
pete-watters, line 30: Does this need to be `unknown`?

pete-watters (approved):

packages/query/src/stacks/token-metadata/fungible-tokens/fungible-token-metadata.hooks.ts
fbwoolf: That is bc you removed `client.getFtMetadata(address, signal) as unknown as FtAssetResponse` on L33 in the queryFn.

alter-eggo: in catch block of query fn we return null if there is api 404 err, so this won't help here

pete-watters: If there are errors it would be good to try fix them if we can to avoid losing the type. Or else provide `FtAssetResponse | unknown`

alter-eggo: I was getting false errors in vscode... it worked ok once restarted ts server...
fixed the types to `FtAssetResponse | null`

alter-eggo: thanks guys 👍

PR #387, opened 2024-08-21
feat: wallet settings

edgarkhanzadian:
- Add wallets settings screen
- Add Wallet config screen
- Add Account config screen
- Add account rename functionality
- Add account hide functionality
- Add wallet rename functionality
- Add wallet delete functionality
- Add wallet view secret key functionality
- Add view hidden accounts screen

https://github.com/user-attachments/assets/8ce83449-a6ef-40c3-85d1-329706290e6f


apps/mobile/src/app/wallet/wallet-settings/configure/[wallet]/view-secret-key.tsx
```diff
@@ -0,0 +1,105 @@
...
+} from '@leather.io/ui/native';
+
+function ViewSecretKey({ fingerprint }: { fingerprint: string }) {
+  const { bottom } = useSafeAreaInsets();
+  const theme = useTheme<Theme>();
+  const { navigateAndCreateWallet } = useCreateWallet();
+  const { displayToast } = useToastContext();
+  const [mnemonic, setMnemonic] = useState<string | null>(null);
+  useEffect(() => {
+    mnemonicStore(fingerprint)
```
Comment (pete-watters) [inline apps/mobile/src/app/wallet/wallet-settings/configure/[wallet]/view-secret-key.tsx:31]: Maybe we can start adding `await` for some of these promises? I see this warning a few times

apps/mobile/src/components/avatar-icon.tsx
```diff
@@ -0,0 +1,88 @@
+import AlienIcon from '@/assets/avatar-icons/alien.svg';
```
pete-watters, line 1: I think that these should go in `packages/ui`? 

We can make a new task to move them

apps/mobile/src/components/cell.tsx
```diff
@@ -0,0 +1,75 @@
+import { ResponsiveValue, useTheme } from '@shopify/restyle';
```
pete-watters, line 1: This is the `Cell` from [here](https://github.com/leather-io/issues/issues/229) right? 

Maybe we can just add it to the UI lib from the start?

apps/mobile/src/components/wallet-settings/account-name-modal.tsx
```diff
@@ -0,0 +1,30 @@
+import { RefObject } from 'react';
```
pete-watters, line 1: Design want to call all `modal`s as 'Sheet` so we should change the name of this. See https://github.com/leather-io/issues/issues/268 

I guess this also includes these files

apps/mobile/src/state/global-action.ts
```diff
@@ -2,7 +2,7 @@ import { createAction } from '@reduxjs/toolkit';
 
 import { BitcoinKeychainStore } from './keychains/bitcoin/bitcoin-keychains.slice';
 import { StacksKeychainStore } from './keychains/stacks/stacks-keychains.slice';
-import { WalletStore } from './wallets/wallets.slice';
+import { LooseWalletStore } from './wallets/wallets.slice';
```
pete-watters, line 5: What does `Loose` mean here? Just double checking for my own learning

apps/mobile/src/components/cell.tsx
pete-watters: I did some initial work on this in https://github.com/leather-io/mono/pull/398 

I can finish moving it once this is merged

apps/mobile/src/app/wallet/wallet-settings/configure/[wallet]/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+          }}
+        >
+          <Box px="5" gap="6">
+            <Text variant="heading05">{wallet.name}</Text>
+            <Cell
+              title={t`View Secret Key`}
+              Icon={Eye1ClosedIcon}
+              onPress={() => {
+                router.navigate({
+                  pathname: APP_ROUTES.WalletWalletsSettingsConfigureViewSecretKey,
```
Comment (pete-watters) [inline apps/mobile/src/app/wallet/wallet-settings/configure/[wallet]/index.tsx:86]: I branched off here to get started on `Sheet` and it seems like `APP_ROUTES.WalletWalletsSettingsConfigureViewSecretKey` doesn't exist?

pete-watters (changes requested):
I approved but I think theres a missing route when I dig in more - https://github.com/leather-io/mono/pull/387/files#r1732508519 

`APP_ROUTES.WalletWalletsSettingsConfigureViewSecretKey`

apps/mobile/src/components/avatar-icon.tsx
fbwoolf: I created an issue to make the AvatarIcon in the UI library for mobile. @pete-watters is correct, this is already used in the extension and should be part of the UI lib.

https://github.com/leather-io/issues/issues/290

apps/mobile/src/components/cell.tsx
fbwoolf: The `Cell` should be comparable to the `Pressable` component in the extension, it should often be just a wrapper arnd `ItemLayout`?

apps/mobile/src/components/avatar-icon.tsx
fbwoolf: We can just pull these out with that issue so it doesn't delay getting this merged.

fbwoolf (approved):
Some of these files are getting quite large and hard to read, should we start looking at the file structure of the extension and using `.utils` files for helper functions, etc?

apps/mobile/src/components/avatar-icon.tsx
pete-watters: Thanks! Yes lets not block merging this, I just wanted to mention it

apps/mobile/src/components/cell.tsx
pete-watters: OK thanks @fbwoolf . Maybe I will should rename `Pressable` as `Cell`?

apps/mobile/src/state/global-action.ts
pete-watters: Haha ok thanks! I was just reading it as `Lose` and then thought `Loose` could be some computer science term I didn't know 😄

apps/mobile/src/app/wallet/wallet-settings/configure/[wallet]/index.tsx
Comment (pete-watters) [reply apps/mobile/src/app/wallet/wallet-settings/configure/[wallet]/index.tsx:86]: OK, maybe it's my mistake when re-basing and moving between branches 👍

pete-watters (approved):
I can see route is here https://github.com/leather-io/mono/pull/387/files#diff-28f23f082360d9fb23113eb05a44316dbe98628b6828bdc0f036ce603e5ae725R24 so my mistake I guess 😊

apps/mobile/src/components/cell.tsx
fbwoolf: I am pretty open to naming and have a hard time tracking where the names come from, but seems to make sense to have them coordinated if the same in web/mobile. `Pressable` prob isn't a great name for web, and RN already has a `Pressable` component.

pete-watters: I thought `Cell` was an Apple term from https://developer.apple.com/design/human-interface-guidelines/ but it's not. 

Let's just call it `Cell` so that we can group the components together. I honestly don't mind what we call them but `Cell` could be an upgrade on `Pressable` if it avoids clashing with RN `Pressable` also

fbwoolf:
@edgarkhanzadian now working with this code, wouldn't it make more sense to start nesting the routes based on the flows. For example, `/wallet/settings/wallets-and-accounts` rather than `/wallet/settings/wallet-settings`. Or, is that not how things are typ done in mobile? I was trying to add off settings `/wallet/settings/display`, it seems strange to add `/wallet/settings/display-settings` as a new folder?

fbwoolf:
Like this...

Screenshot 2024-08-28 at 3 03 34 PM

apps/mobile/src/state/accounts/accounts.slice.ts
```diff
@@ -1,26 +1,52 @@
...
+    if (!draftAccount.name) {
+      draftAccount.name = t`Account ${accountIdx}`;
+    }
+    if (!draftAccount.status) {
+      draftAccount.status = 'active';
+    }
+    return draftAccount;
+  });
+  return updatedAccount as AccountStore;
+}
```
kyranjamie, line 48: Is the immer code here needed? Isn't this handled internally by RTK entity methods?

Also see the `accountIdx` is for order, but would suggest another name, orderBy or something, just because "account index" is already a used concept for it's path account index

PR #388, opened 2024-08-21
feat: web svg icons

fbwoolf:
This PR refactors the icons used in the extension to be generated from the shared .svg file with mobile and it adjusts the tsup build to handle it. It also coordinates naming with Figma so that all icon names are descriptive rather than adjusted for use case. All .svg files have been named according to the exports direct from Figma: `small` is `...-16-16.svg` and `default` is `...24-24.svg`. I have created icon files for what are in use, but if one is missing please use the .svgs and follow the pattern to create the icon for web or mobile accordingly.

Tests from extension following the same tests in mobile...
Screenshot 2024-08-19 at 4 47 58 PM

Screenshot 2024-08-19 at 4 48 10 PM

The extension has been tested and no icons seem broken.

EDIT: I have a working branch in the extension, so I'll put up an ext PR too so this work can be installed.


packages/tokens/src/tokens.ts
```diff
@@ -42,8 +42,6 @@ export const tokens = {
     popupHeight: { value: '756px' },
     popupHeaderHeight: { value: '68px' },
     headerContainerHeight: { value: '40px' },
-    logoHeight: { value: '32px' },
```
pete-watters, line 45: Good work getting rid of these 💪

packages/ui/src/icons/console-icon.native.tsx
```diff
@@ -1,6 +1,7 @@
 import Console from '../assets/icons/console.svg';
 import { Icon, IconProps } from './icon/icon.native';
 
+// Is this icon outdated?
```
pete-watters, line 4: Lets double check this with design

packages/ui/src/icons/emoji-smile-icon.native.tsx
```diff
@@ -1,6 +1,7 @@
 import EmojiSmile from '../assets/icons/emoji-smile.svg';
 import { Icon, IconProps } from './icon/icon.native';
 
+// Is this an icon or an AvatarIcon?
```
pete-watters, line 4: Lets check this also

packages/ui/src/icons/x-logo-icon.native.tsx
```diff
@@ -1,8 +1,8 @@
 import XLogo from '../assets/icons/x-logo.svg';
 import { Icon, IconProps } from './icon/icon.native';
 
-export function XLogoIcon({ variant, ...props }: IconProps) {
-  // TODO: Need svg for small variant
+// Should this be an AvatarIcon?
```
pete-watters, line 4: Lets double check this

packages/ui/tsup.config.web.ts
```diff
@@ -1,7 +1,26 @@
+import { copy } from 'esbuild-plugin-copy';
+import svgrPlugin from 'esbuild-plugin-svgr';
 import { defineConfig } from 'tsup';
 
 export default defineConfig({
   entry: ['web.ts'],
+  esbuildPlugins: [
```
pete-watters, line 7: Nice work 😎

packages/ui/package.json
```diff
@@ -24,7 +23,7 @@
     "typecheck": "tsc --noEmit --project ./tsconfig.json"
   },
   "exports": {
-    ".": "./dist-all/web.js",
+    ".": "./dist-web/web.js",
```
pete-watters, line 26: Thanks for changing this to make it clearer. I think you probably know but just in case not, this will require an update in the extension's `panda.config.ts` to change the path. 

I also specify it in the `packages/ui/README.md` so it would be good to update that also.

pete-watters (approved):
Great work on this @fbwoolf , it's looking really good. 

I mentioned a couple of comments about checking `Icons` I saw so it would be good to do that and then remove the comments. 

It would be good [to update the `README` also with the new path](https://github.com/leather-io/mono/pull/388/files#r1726643428) to `dist-web/`

packages/ui/package.json
fbwoolf: Yep, I have those changes on a branch in the extension, thx.

packages/ui/src/icons/console-icon.native.tsx
fbwoolf: Thx, I'm communicating with design team on these and might need to circle back to them separately, but I'll try to address them here before merging.

packages/ui/package.json
fbwoolf: Made the change to the readme, thought I had done that already. I pushed up the change here.

PR #389, opened 2024-08-22
chore: item layout mobile

pete-watters:
This adds a native version of `ItemLayout`. Using the web styles and its not looking quite right:

https://github.com/user-attachments/assets/ad0b40dd-8657-4548-bfbf-74841ae043fd

On native, it seems like `alignItems` doesn't support `start` / `end` so I have instead used `flex-start` and `flex-end`, maybe thats the issue? 

When I check the **web.storybook** `ItemLayout` is not loading at all there so I can investigate fixing that and compare

Screenshot 2024-08-23 at 14 14 28

### DIFF
--- a/packages/ui/src/components/box/hstack.native.tsx
+++ b/packages/ui/src/components/box/hstack.native.tsx
@@ -1,0 +1,11 @@
+import { createBox } from '@shopify/restyle';
+
+import { type Theme } from '../../theme-native';
+
+export const HStack = createBox<Theme>();
+
+HStack.defaultProps = {
+  flex: 1,
+  flexDirection: 'row',
+  alignItems: 'center',
+};
--- a/packages/ui/src/components/item-layout/item-layout.native.tsx
+++ b/packages/ui/src/components/item-layout/item-layout.native.tsx
@@ -1,0 +1,70 @@
+import { ReactNode, isValidElement } from 'react';
+
+import { Flag } from '../../components/flag/flag.native';
+import { Text } from '../../components/text/text.native';
+import { ChevronUpIcon } from '../../icons/chevron-up-icon.native';
+import { Box } from '../box/box.native';
+import { HStack } from '../box/hstack.native';
+import { Stack } from '../box/stack.native';
+
+export interface ItemLayoutProps {
+  captionLeft: ReactNode;
+  captionRight?: ReactNode;
+  flagImg: ReactNode;
+  isDisabled?: boolean;
+  isSelected?: boolean;
+  showChevron?: boolean;
+  titleLeft: ReactNode;
+  titleRight: ReactNode;
+}
+
+export function ItemLayout({
+  captionLeft,
+  captionRight,
+  flagImg,
+  showChevron,
+  titleLeft,
+  titleRight,
+}: ItemLayoutProps) {
+  return (
+    <Flag img={flagImg} spacing="3">
+      <Box flex={1} alignItems="center" justifyContent="space-between">
+        <Stack
+          alignItems="flex-start"
+          justifyContent="space-between"
+          flexGrow={2}
+          overflow="hidden"
+        >
+          <HStack>{isValidElement(titleLeft) ? titleLeft : <Text>{titleLeft}</Text>}</HStack>
+          {isValidElement(captionLeft) ? (
+            captionLeft
+          ) : (
+            <Text variant="caption01" color="ink.text-subdued">
+              {captionLeft}
+            </Text>
+          )}
+        </Stack>
+
+        <HStack gap="3">
+          <Stack alignItems="flex-end" gap="2" height={42}>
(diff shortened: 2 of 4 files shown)


packages/ui/src/components/item-layout/item-layout.native.tsx
```diff
@@ -0,0 +1,103 @@
...
+  captionRight,
+  flagImg,
+  showChevron,
+  titleLeft,
+  titleRight,
+}: ItemLayoutProps) {
+  return (
+    <Flag img={flagImg} spacing="3">
+      <Box style={styles.container}>
+        <Box
```
pete-watters, line 31: @edgarkhanzadian : I have some Q's here. 

- Do you think its worth making wrappers for native - `Stack` + `HStack`? Then the code would match up better?
- Is `Box` the right element to use here? On web we are aligning `start` and `end` and not `flex-start` and `flex-end`

packages/ui/src/components/item-layout/item-layout.native.tsx
```diff
@@ -0,0 +1,103 @@
...
+            {
+              alignItems: 'flex-start', // this start on web
+              justifyContent: 'space-between',
+              flexGrow: 2,
+              overflow: 'hidden',
+            },
+          ]}
+        >
+          <Box style={styles.hstack}>
+            {/* how to handle text overflow? or do i need to */}
```
pete-watters, line 43: @edgarkhanzadian do we need to do some handling for text overflow?

packages/ui/src/components/item-layout/item-layout.native.tsx
```diff
@@ -0,0 +1,103 @@
...
+    flex: 1,
+    flexDirection: 'column',
+    alignItems: 'center',
+  },
+  hstack: {
+    flex: 1,
+    flexDirection: 'row',
+    alignItems: 'center',
+  },
+  pressableCaptionStyles: {
```
pete-watters, line 100: Adding this on web as we have:
```
export const pressableCaptionStyles = css({
  _groupDisabled: { color: 'ink.text-non-interactive' },
  color: 'ink.text-subdued',
});

export const pressableChevronStyles = css({
  _groupDisabled: { color: 'ink.text-non-interactive' },
  color: 'ink.action-primary-default',
});
```

I think maybe we can just add `color` only instead as `_groupDisabled` doesn't seem to apply here

fbwoolf:
Looks good @pete-watters 👍

pete-watters:
> Looks great! 🚀

Thanks 🙇

packages/ui/src/components/item-layout/item-layout.native.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <Box
+          style={[
+            styles.stack,
+            {
+              alignItems: 'flex-start', // this start on web
+              justifyContent: 'space-between',
+              flexGrow: 2,
+              overflow: 'hidden',
+            },
+          ]}
```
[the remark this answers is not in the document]
pete-watters: Thanks @edgarkhanzadian . I'll move the styles inline to the components and create some wrappers for `Stack` + `HStack`.

I won't use `style` anymore either, it was just easier to demonstrate `Stack` + `HStack` using it here.

packages/ui/src/components/item-layout/item-layout.native.tsx
fbwoolf: _groupDisabled is from panda: https://panda-css.com/docs/concepts/conditional-styles#group-selectors

pete-watters: Thanks @fbwoolf . I'll exclude it for now and we can assess how it looks without, is that OK?

packages/ui/src/components/box/hstack.native.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { type Theme } from '../../theme-native';
+
+export const HStack = createBox<Theme>();
+
+HStack.defaultProps = {
+  flex: 1,
+  flexDirection: 'row',
+  alignItems: 'center',
+};
+
```
pete-watters, line 12: I left this commented code as I'm not sure what the best way of doing this is. On web we often extend for example:
```
interface PeteBox extends BoxProps {
...
```

This way of re-creating custom boxes seems to work but I am curious as to whether it's best to do that or a variation of the commented code below

packages/ui/src/components/item-layout/item-layout.native.tsx
fbwoolf: O yeah, totally fine, was just letting @edgarkhanzadian know where it came from. It was used to style the children based on the parent being disabled. Not sure that is even relevant in mobile?

pete-watters:
I'll merge this as it is adding the component correctly although the alignment of items [(see demo)](https://github.com/leather-io/mono/pull/389#issue-2481152147) could need some work which we can tweak when consuming it.

PR #391, opened 2024-08-23
fix: web ui build missing src folder

fbwoolf:
Hoping this fixes the web ui build, going to use the same plugin as I did for the native build which seems to work, maybe the clean plugin ended up removing the src files? I'm not sure why it was working for me locally.

cc @pete-watters

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -86,7 +86,6 @@
     "babel-preset-expo": "11.0.6",
     "concurrently": "8.2.2",
     "css-loader": "6.10.0",
-    "esbuild-plugin-clean": "1.0.1",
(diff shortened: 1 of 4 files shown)


fbwoolf:
@pete-watters let me know if this doesn't look good to you and you'd prefer me to put back the `dist-all` setup you had? I was just trying to simplify things a bit by having only two build folders for `web` and `mobile`. I also tested this before w/out using the `clean` plugin to remove the `assets-native` folder, and that seemed to work to if we want to avoid using the copy plugin for web.

fbwoolf:
I am going to merge this to see if builds correctly when released. The build is already broken in the last release and I'd like to see if I can fix things today.

PR #393, opened 2024-08-26
fix: revert UI web svgs

fbwoolf:
@pete-watters @alter-eggo putting this up so you can revert these commits if my changes are blocking people from working locally on the extension. I realized the extension PR here, https://github.com/leather-io/extension/pull/5807, needs to be merged quickly after I make these changes bc the webpack svg loader has to change. I have a working branch to address this: https://github.com/leather-io/extension/pull/5807#issuecomment-2307959644, but I didn't get it finished over the wkend. I don't want to block anyone in your AM. Feel free to leave this if not needed.


pete-watters:
Thanks @fbwoolf 👍 no stress over this as it's tricky to get it working correctly. 

We can just use an older version of `ui` library  in the extension if we need to

fbwoolf:
Going to close this bc I have a working branch to get this all fixed today.

pete-watters:
If you have time can you double check that the web storybook is working too? I think some stories may need to update to use the new icons like `Dialog` for example:

Screenshot 2024-08-26 at 14 40 09

You can run ` pnpm storybook:web` in `packages/ui/` to start it. I can take a look into it soon if you don't have time, no problem. 

In future I think we could export `.stories` from the library too. Then we can use them to compose stories in the extension and have our published storybook showing them also (right now it only shows the few components we didn't move)

PR #395, opened 2024-08-26
fix: add config hook to disable brc-20 sends, ref leather-io/issues#5773

pete-watters:
This PR adds a config hook so that we will be able to remotely enable / disable BRC-20 sends

https://github.com/leather-io/extension/issues/5773

### DIFF
--- a/packages/query/src/common/remote-config/remote-config.query.ts
+++ b/packages/query/src/common/remote-config/remote-config.query.ts
@@ -120,3 +120,8 @@ export function useConfigSwapsEnabled() {
   const config = useRemoteConfig();
   return get(config, 'swapsEnabled', false);
 }
+
+export function useBrc20SendEnabled() {
+  const config = useRemoteConfig();
+  return get(config, 'brc20SendEnabled', false);
+}


pete-watters:
I'm not sure if it makes sense to add this as we don't do this to stop send of other assets e.g. SIP-10, SRC-20 etc. 

See [here](https://github.com/leather-io/extension/pull/5804#issuecomment-2309714899)

fbwoolf:
Do we need to be able to toggle this on/off quickly w/out a release? That is the benefit of using the config.

pete-watters:
> Do we need to be able to toggle this on/off quickly w/out a release? That is the benefit of using the config.

I don't think so as there are some in BRC-20 send anyway with the form crashing sometimes so we want to just make it ready only. 

@alter-eggo suggested using a config setting [here](https://github.com/leather-io/extension/pull/5804#issuecomment-2304659028) but when I looked into it, we already have a pattern of using `isReadOnly` in the `AssetList` to disable **Send**. 

I can right more code so that we can toggle it with a setting in the future but we seem to only do that to totally disable / enable things like - `runesEnabled` , `swapEnabled` etc.

fbwoolf:
> I can right more code so that we can toggle it with a setting in the future but we seem to only do that to totally disable / enable things like - `runesEnabled` , `swapEnabled` etc.

Yeah, I would only add it to the config if it's something we need to control easily w/out doing a release.

pete-watters:
> > I can right more code so that we can toggle it with a setting in the future but we seem to only do that to totally disable / enable things like - `runesEnabled` , `swapEnabled` etc.
> 
> Yeah, I would only add it to the config if it's something we need to control easily w/out doing a release.

I don't think we need to do that so I will close this PR and I have removed the config settings from https://github.com/leather-io/extension/pull/5804/files

PR #396, opened 2024-08-26
fix: solve type errors in ui lib

pete-watters:
I was trying to help solve the type issues which I didn't however I think these updates should be made to help banish `dist-all`

### DIFF
--- a/packages/ui/tsup.config.web.ts
+++ b/packages/ui/tsup.config.web.ts
@@ -43,6 +43,7 @@ export default defineConfig({
   target: 'esnext',
   loader: {
     '.js': 'jsx',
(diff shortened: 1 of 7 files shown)


packages/ui/tsup.config.web.ts
```diff
@@ -43,6 +43,7 @@ export default defineConfig({
   target: 'esnext',
   loader: {
     '.js': 'jsx',
+    '.jsx': 'jsx',
```
fbwoolf, line 46: This is default from what I saw in docs?

fbwoolf:
I'll include anything I missed from here in my branch.

PR #397, opened 2024-08-27
fix: web ui build

fbwoolf:
This PR should fix the web build issues currently blocking: https://github.com/leather-io/extension/pull/5807

I followed the suggestion here to use the tsup `onSuccess` callback to generate the declaration maps, rather than using build scripts: 
https://tsup.egoist.dev/#generate-typescript-declaration-maps--d-ts-map
From issue: https://github.com/egoist/tsup/issues/564

@pete-watters let me know if this looks like a good change, or I can revert back to using the `dist-all` approach.


turbo.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://turbo.build/schema.json",
   "tasks": {
     "build": {
-      "outputs": ["dist/**"],
+      "outputs": ["dist/**", "dist-*/**"],
```
fbwoolf, line 5: Isn't this needed here? 🤔

pete-watters: 🤷

pete-watters:
I tested this in the extension and it's working well, great stuff 👍  Well done on banishing `dist-all` as the setup is more intuitive now too 👏 

FYI I also tested it in on the `dev` branch of the `extension` when I was still a bit sleepy and can confirm it's broken on `dev` but then working on your branch `refactor/icons-using-svgs`

pete-watters (approved):

fbwoolf:
> FYI I also tested it in on the `dev` branch of the `extension` when I was still a bit sleepy and can confirm it's broken on `dev` but then working on your branch `refactor/icons-using-svgs`

It is expected to be broken on dev, so that is ok. Hoping to get the ext PR merged quickly after with the new svg loader to make it work. Will get this release and hope it continues to work once installed there. 🤞

PR #398, opened 2024-08-27
feat: add native cell component to UI library

pete-watters:
This PR moves the new `Cell` component to the UI library


fbwoolf:
@pete-watters was thinking on this last night, is `Cell` actually just a RN `TouchOpacity` wrapping `ItemLayout`, maybe `Cell` isn't even nec now that we have `ItemLayout`?

fbwoolf:
Was thinking bc we created `Pressable` in the extension to handle different states, but there already is a RN `Pressable` (or `TouchableOpacity`) that handles all the states, we just needed the layout components for mobile, right?

pete-watters:
> Was thinking bc we created `Pressable` in the extension to handle different states, but there already is a RN `Pressable` (or `TouchableOpacity`) that handles all the states, we just needed the layout components for mobile, right?

I think that could be right 👍  I don't mind having a wrapper or else just using `<TouchableOpacity`. It might be easier to mentally connect with the design system if we use a `<Cell` wrapper.

pete-watters:
> @pete-watters was thinking on this last night, is `Cell` actually just a RN `TouchableOpacity` wrapping `ItemLayout`, maybe `Cell` isn't even nec now that we have `ItemLayout`?

I think it's a similar case in `Sheet` which is essentially just a wrapper for `ScrollView` - https://github.com/leather-io/mono/pull/400 

Maybe it's easier for semantics just to keep the wrappers for consistent naming? (`Sheet` on web + native, `Cell` also)

pete-watters:
@fbwoolf - opening this for review now as I don't want it to block your next work. 

If you think it looks OK and you need it feel free to merge.

apps/mobile/src/components/button.tsx
```diff
@@ -66,7 +66,7 @@ export function getTextColor(buttonState: ButtonState) {
   );
 }
 
-// TODO: Move to UI library
+// TODO 289: Move to UI library
```
pete-watters, line 69: I added this tag as we aren't actually using the `button` from the ui package yet https://github.com/leather-io/issues/issues/289

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,13 +1,6 @@
...
 
-import {
-  Box,
-  ChevronRightIcon,
-  IconProps,
-  Text,
-  Theme,
-  TouchableOpacity,
-} from '@leather.io/ui/native';
+import { Box, ChevronRightIcon, IconProps, Text, Theme, TouchableOpacity } from '../../../native';
```
pete-watters, line 3: This is really just moving the existing `Cell` we had out. 

I can do more work to refactor it and get it to receive `ItemLayout` instead

fbwoolf: No rush here, I can work with the existing `Cell`, if some of this code is just a dup of `ItemLayout` then let's get that change in here. The stuff I'm doing is just setting up the UI, so not difficult to swap out. I think it's better to have these components setup correctly.

pete-watters: I'll merge this to move `Cell` out of the app code but will then refactor it more if it's the same as item layout and add this variant - https://github.com/leather-io/mono/pull/398#pullrequestreview-2268577677

PR #400, opened 2024-08-27
feat: add native sheet component

pete-watters:
This PR adds a `Sheet` component for native. It seems to me like `Sheet` is a wrapper for `ScrollView`. 

Looking at the code and I can set it up to have variants but it could be more composable to just accept `props` for style as sometimes it has no `style` or `contentContainerStyle` for example

### DIFF
--- a/packages/ui/native.ts
+++ b/packages/ui/native.ts
@@ -11,3 +11,6 @@ export { TouchableOpacity } from './src/components/button/touchable-opacity.nati
 export type { Theme } from './src/theme-native';
 export { Icon } from './src/icons/icon/icon.native';
 export * from './src/icons/index.native';
+export { Flag } from './src/components/flag/flag.native';
+export { ItemLayout } from './src/components/item-layout/item-layout.native';
+export { Sheet } from './src/components/sheet/sheet.native';
--- a/packages/ui/src/components/sheet/sheet.native.tsx
+++ b/packages/ui/src/components/sheet/sheet.native.tsx
@@ -1,0 +1,34 @@
+import { NativeScrollEvent, NativeSyntheticEvent } from 'react-native';
+import { ScrollView } from 'react-native-gesture-handler';
+
+interface SheetProps {
+  children: React.ReactNode;
+  contentContainerStyle?: Record<string, unknown>;
+  contentInset?: Record<string, unknown>;
+  horizontal?: boolean;
+  onScroll?: (event: NativeSyntheticEvent<NativeScrollEvent>) => void;
+  style?: Record<string, unknown>;
+}
+
+export function Sheet({
+  children,
+  horizontal = false,
+  style,
+  contentContainerStyle,
+  contentInset,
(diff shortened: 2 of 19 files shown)


packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface CellProps {
+  title: string;
+  subtitle?: string;
+  onPress?(): unknown;
+  Icon: React.FC<IconProps>;
+  variant?: CellVariant;
+}
+
+export function Cell({ title, subtitle, onPress, Icon, variant = 'active' }: CellProps) {
```
fbwoolf, line 48: Should the `Cell` code be here?

packages/ui/native.ts
```diff
@@ -11,3 +11,6 @@ export { TouchableOpacity } from './src/components/button/touchable-opacity.nati
 export type { Theme } from './src/theme-native';
 export { Icon } from './src/icons/icon/icon.native';
 export * from './src/icons/index.native';
+export { Flag } from './src/components/flag/flag.native';
+export { ItemLayout } from './src/components/item-layout/item-layout.native';
```
pete-watters, line 15: I had forgotten to include `ItemLayout` + `Flag` here so they weren't available in the public package

packages/ui/src/components/sheet/sheet.native.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface SheetProps {
+  children: React.ReactNode;
+  contentContainerStyle?: Record<string, unknown>;
+  contentInset?: Record<string, unknown>;
+  horizontal?: boolean;
+  onScroll?: (event: NativeSyntheticEvent<NativeScrollEvent>) => void;
+  style?: Record<string, unknown>;
+}
+
```
pete-watters, line 12: `Sheet` is basically just a wrapper of `ScrollView`. I'm not sure if we even need to do that. 

Sometimes we pass more props to `ScrollView`, for example I had to add `contentInset` + `onScroll`. 

It helps us to link it to `Sheet` in design but I'm not sure if it's better than just consuming `ScrollView`. 

It might make more sense having it here when the footers are implemented https://github.com/leather-io/issues/issues/231

pete-watters:
@edgarkhanzadian - is this the right component for `Sheet`? 

Having done it and checking the code I found this other component - `BottomSheetModal` from `@gorhom/bottom-sheet`. 

I created this based on looking at the demo and where we open a drawer with a scroll view

packages/ui/src/components/sheet/sheet.native.tsx
fbwoolf: @edgarkhanzadian can you provide feedback here on what you've experienced working with RN, is it helpful to wrap `ScrollView` like this so it is coordinated with designs and the extension? It does make sense to me to do, but I'm also not sure if it is overkill?

PR #402, opened 2024-08-28
refactor: all inscriptions requests to use bis api

alter-eggo:
This pr refactors all inscriptions related requests to use BiS api instead of Hiro


packages/query/src/bitcoin/clients/best-in-slot.ts
```diff
@@ -10,12 +10,15 @@ import {
 
 import {
   bestInSlotInscriptionSchema,
+  bestinslotInscriptionBatchInfoSchema,
```
pete-watters, line 13: This should be in camelCase  -> `bestInSlot`

packages/query/src/bitcoin/balance/btc-taproot-balance.hooks.ts
```diff
@@ -26,15 +26,15 @@ export function useCurrentTaprootAccountUninscribedUtxos({
...
+  const query = useBestInSlotGetInscriptionsInfiniteQuery({
     taprootKeychain,
     nativeSegwitAddress,
   });
 
   return useMemo(() => {
     const inscriptionsResponse = query.data?.pages?.flatMap(page => page.inscriptions) ?? [];
 
-    const inscriptions = inscriptionsResponse.map(createHiroInscription);
+    const inscriptions = inscriptionsResponse.map(createBestinSlotInscription);
```
pete-watters, line 37:
```suggestion
    const inscriptions = inscriptionsResponse.map(createBestInSlotInscription);
```

packages/query/src/bitcoin/ordinals/inscription.hooks.ts
```diff
@@ -1,11 +1,13 @@
...
-import { createHiroInscription } from './inscription.utils';
+import { createBestinSlotInscription } from './inscription.utils';
 
 export function useInscription(id: string) {
+  const client = useBitcoinClient();
   return useQuery({
-    ...createGetInscriptionQueryOptions(id),
-    select: resp => createHiroInscription(resp),
+    ...createGetInscriptionQueryOptions(id, client.BestinSlotApi),
+    select: resp => createBestinSlotInscription(resp.data),
```
pete-watters, line 11: Maybe this can have issues if `resp` has no `data` ?

packages/query/src/bitcoin/ordinals/inscriptions-by-param.hooks.ts
```diff
@@ -4,36 +4,52 @@ import { useQuery } from '@tanstack/react-query';
...
   return useQuery({
-    ...createGetInscriptionsByParamQueryOptions({ isPending, param }),
+    ...createGetInscriptionsByParamQueryOptions({
+      isPending,
+      param,
+      bestInSlotApi: client.BestinSlotApi,
+    }),
     select(resp) {
-      const inscriptionResponse = resp.results[0];
+      const inscriptionResponse = resp.data[0].result?.[0];
```
pete-watters, line 34: This will fail if no `data` / `data[0]`

pete-watters (commented):
Nice work 👍 

Some suggestions:
- use consistent naming for `bestInSlot`
- add defensive code in case `resp.data` undefined (maybe we don't need to)
- tag the issue in your commit message so it's easier when doing releases - `-m refactor: all inscriptions requests to use bis api, ref leather-io/issues#267`

pete-watters (approved):

PR #407, opened 2024-08-30
feat: account selector

edgarkhanzadian:
Implement account selector with animations

https://github.com/user-attachments/assets/d9637c43-fcb4-4878-ba18-7720ebd812d2


apps/mobile/src/components/account-selector-sheet.tsx
```diff
@@ -0,0 +1,213 @@
...
+  return (
+    <>
+      <Animated.View style={placeholderUpAnimatedStyles} />
+      <GestureDetector gesture={gestures}>
+        <Animated.View
+          onLayout={e => {
+            const { height, width } = e.nativeEvent.layout;
+            thisItemDimensions.value = { height, width };
+          }}
+          // @ts-expect-error some ref type issues, it should be fine
```
pete-watters, line 125: 😰

pete-watters (approved):
🚀

apps/mobile/src/components/account-selector-sheet.tsx
fbwoolf: Can this be explained? I'm not sure we typically do this? 🤔

apps/mobile/src/components/account-selector-sheet.tsx
```diff
@@ -0,0 +1,213 @@
...
+import { useRouter } from 'expo-router';
+
+import { Box, IconProps } from '@leather.io/ui/native';
+
+import { getAvatarIcon } from './avatar-icon';
+import { Modal } from './bottom-sheet-modal';
+import { AccountCard } from './wallet-settings/account-card';
+
+const APPROXIMATE_DEFAULT_ITEM_SIZE = { height: 150, width: 300 };
+function DraggableAccount({
```
fbwoolf, line 25: Would it make sense to start pulling some of this code apart and using the `features` pattern? Some of these files are quite large and hard to follow. Would `account-selector-sheet` be it's own feature with local component files like `draggable-account`? In the extension, we typ try to only have one component per file if possible for readability.

fbwoolf: Also, will the `draggable` code be reused? 🤔 Maybe it needs to be more generic?

pete-watters: Good points 👍 I've been a bit trigger happy approving mobile stuff to increase velocity but as more of us are working on it now we should take extra care to increase the quality

fbwoolf (approved):

apps/mobile/src/features/account-selector-sheet.tsx
```diff
@@ -0,0 +1,88 @@
...
+            return (
+              <Draggable
+                idx={idx}
+                direction={direction}
+                scrollViewRef={scrollViewRef}
+                placeholderIdx={placeholderIdx}
+                accountsLength={accounts.length}
+                key={account.id}
+                accountId={account.id}
+                Component={() => (
```
fbwoolf, line 76: Curious on this bc I keep seeing it, we never have captialized props in the extension (even if we are passing ReactNodes), is that typ in RN code? Not opposed to it, but I think in the icon work I refactored some to lowercase `icon` out of habit.

apps/mobile/src/components/account-selector-sheet.tsx
fbwoolf: Lgtm, thanks for doing! 👍

apps/mobile/src/features/account-selector-sheet.tsx
pete-watters: This looks weird to me also, but perhaps is it just when the prop is `Component`? 

There is a convention I'm familiar with where components are capitalised to indicate that e.g. `SomeComponent.tsx`

apps/mobile/src/features/account-selector-sheet.tsx
```diff
@@ -0,0 +1,88 @@
...
+        idx1 === null ||
+        idx2 === null ||
+        !checkIdxWithinBounds(idx1) ||
+        !checkIdxWithinBounds(idx2)
+      )
+        return;
+      const temp = accountIds[idx1];
+      /** @ts-expect-error we know that both indexes have both value in accounts array */
+      accountIds[idx1] = accountIds[idx2];
+      /** @ts-expect-error we know that both indexes have both value in accounts array */
```
kyranjamie, line 44: Any way we can avoid doing this with some runtime checks? Would say ts-expect-error isn't something we want to rely on too much

apps/mobile/src/features/account-selector-sheet.tsx
kyranjamie: Why pass `() => ReactNode` instead of `ReactNode`?

apps/mobile/src/components/draggable.tsx
```diff
@@ -0,0 +1,127 @@
...
+  placeholderIdx,
+  scrollViewRef,
+  direction,
+  onAccountPress,
+  swapAccountIndexes,
+  Component,
+}: {
+  idx: number;
+  accountId: string;
+  accountsLength: number;
```
kyranjamie, line 26: This component being called `Draggable` makes me think it's a generic component that can be used elsewhere. But it can't, it's specific to the account re-ordering use-case.

Suggest either refactoring to be reusable, or renaming and relocating to the feature folder

fbwoolf: I was trying to get him to refactor it so it was reusable, but I overlooked there were still account specific prop names. My thought was we will have other uses for the draggable code and it should be generic?

apps/mobile/src/components/wallet-settings/account-card.tsx
```diff
@@ -7,10 +9,14 @@ export function AccountCard({
...
   name,
   onPress,
+  onLongPress,
+  onLayout,
 }: {
   Icon: React.FC<IconProps>;
   name: string;
   onPress?(): unknown;
+  onLongPress?(): unknown;
+  onLayout?(e: LayoutChangeEvent): void;
```
kyranjamie, line 19: Not sure why mixing void and unknown. If no return value expected, `void` should be used imo, as done in docs https://www.typescriptlang.org/docs/handbook/2/functions.html

PR #411, opened 2024-08-30
feat: add native Skeleton Loader to UI library

adrianocola:
This PR adds a native Skeleton Loader to the UI library.

* replicates the animation of the web version;
* if a new Skeleton component is added to the page after the initial render, its animation will automatically synchronize with existing Skeletons

To achieve this, the libraries `react-native-reanimated` and `expo-linear-gradient` have been included in the UI package.

**Team feedback and any requests for changes are greatly appreciated!**

Web, iOS, and Android storybooks:

https://github.com/user-attachments/assets/001836d1-b0b5-438a-bff6-0b865897de2a

Story showing new skeletons being added/removed:

https://github.com/user-attachments/assets/f2236d77-317c-4ca1-9aaa-f95fd530e5ad

PR #417, opened 2024-09-02
feat: add native Skeleton Loader to UI library

pete-watters:
This PR is a clone of [this one](https://github.com/leather-io/mono/pull/411) from an external contributor. 

Thanks @adrianocola 👏

### DIFF
--- a/packages/ui/src/components/skeleton-loader/skeleton-loader.native.tsx
+++ b/packages/ui/src/components/skeleton-loader/skeleton-loader.native.tsx
@@ -1,0 +1,94 @@
+import { useLayoutEffect } from 'react';
+import { StyleSheet } from 'react-native';
+import Animated, {
+  Easing,
+  interpolate,
+  useAnimatedStyle,
+  useSharedValue,
+  withRepeat,
+  withSequence,
+  withTiming,
+} from 'react-native-reanimated';
+
+import { useTheme } from '@shopify/restyle';
+import { LinearGradient } from 'expo-linear-gradient';
+
+import { Box } from '../box/box.native';
+
+const AnimatedLinearGradient = Animated.createAnimatedComponent(LinearGradient);
+
+export const ANIMATION_DURATION = 1_000;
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/skeleton-loader/skeleton-loader.native.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { LinearGradient } from 'expo-linear-gradient';
+
+import { Box } from '../box/box.native';
+
+const AnimatedLinearGradient = Animated.createAnimatedComponent(LinearGradient);
+
+export const ANIMATION_DURATION = 1_000;
+const ANIMATION_EASING = Easing.ease;
+const LEFT_POS_START = -100;
+const LEFT_POS_END = 100;
```
fbwoolf, line 23: Is there a reason these are all caps? Typically in the extension we only use this when constants are shared globally. Should these be?

fbwoolf (approved):

packages/ui/src/components/skeleton-loader/skeleton-loader.native.tsx
pete-watters: **NO REASON** 😄 

I can update, thanks

kyranjamie: We have a handful of conventions mentioned in Notion https://www.notion.so/trustmachines/Naming-constants-1a66b44cd8234f11aadaadff5e45a3f1?pvs=4

Would love to grow this and add more

adrianocola: @kyranjamie, can these conventions and standards be made public? I need to request access to view them.

pete-watters: @adrianocola notion is not public but here's what it says: 
Screenshot 2024-09-05 at 06 08 45

I will rename them here so we can merge more efficiently

PR #422, opened 2024-09-03
fix: update mnemonic key view

edgarkhanzadian:
https://github.com/user-attachments/assets/ca8af526-9df8-4456-88fa-9742bf3dc439

### DIFF
--- a/apps/mobile/src/app/wallet/create-new-wallet.tsx
+++ b/apps/mobile/src/app/wallet/create-new-wallet.tsx
@@ -3,20 +3,17 @@ import { ScrollView } from 'react-native-gesture-handler';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
 import { MnemonicDisplay } from '@/components/create-new-wallet/mnemonic-display';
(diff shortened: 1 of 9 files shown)


apps/mobile/src/state/storage-persistors.ts
```diff
@@ -137,18 +137,23 @@ export function useMnemonic({
...
     } else {
       throw new Error('No mnemonic found for this fingerprint');
     }
+
+    if (response.passphrase) {
+      setPassphrase(response.passphrase);
+    }
   }
   useEffect(() => {
     if (shouldLoadOnInit) {
```
kyranjamie, line 154: Does `useMnemonic` mean we always keep it in memory?

fbwoolf (approved):

apps/mobile/src/state/storage-persistors.ts
pete-watters: I just made [this issue ](https://github.com/leather-io/issues/issues/315) for tracking this so we can merge this PR

PR #427, opened 2024-09-04
feat: radio button and theme modal

fbwoolf:
This PR add a `RadioButton` to the UI library and wires up the settings theme modal. I removed the theme toggle in the dev console, but let me know if people prefer we keep it?

signal-2024-09-04-130502_004 signal-2024-09-04-130502_002

*Are we wanting to default to `light` or `system`? Currently, we are defaulting to `light`.


apps/mobile/src/app/wallet/settings/display/index.tsx
```diff
@@ -40,7 +43,7 @@ export default function SettingsDisplayScreen() {
       >
         <Cell
           title={t`Theme`}
-          subtitle={t`System`}
+          subtitle={capitalize(settings.themeStore)}
```
fbwoolf, line 46: I added the `capitalize` util to our utils pkg bc there are times we don't always want to do a text transform to capitalize the string passed. Here, the subtitle is often all cap tickers.

apps/mobile/src/components/headers/modal-header.tsx
```diff
@@ -9,6 +9,8 @@ import {
   TouchableOpacity,
 } from '@leather.io/ui/native';
 
+import { MODAL_HEADER_HEIGHT } from './constants';
```
fbwoolf, line 12: Should I start moving these to global constants?

fbwoolf: I moved this `ModalHeader` into `headers` bc I actually created one not realizing a component was in another location, `modal-headers`.

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -87,11 +98,10 @@ export function Cell({ title, subtitle, onPress, Icon, ...props }: CellProps) {
           )}
         </Box>
       </Box>
-      {props.variant !== 'radio' && props.variant !== 'switch' && (
+      {_variant === 'active' && (
         <ChevronRightIcon color={theme.colors['ink.text-primary']} variant="small" />
       )}
-      {/* TODO: Temp placeholder - implement custom radio button */}
-      {props.variant === 'radio' && <></>}
+      {props.variant === 'radio' && <RadioButton isSelected={props.isSelected} />}
```
fbwoolf, line 104: Adding to the `Cell` variants here, but wondering if we should just have a better composed `Cell` to work with? The `active` version is still very much a duplicate of `ItemLayout`, but `ItemLayout` is more flexible. I added a conditional here to cover when there is no `Icon`.

fbwoolf: Alternatively, maybe we just want to kill `ItemLayout` and `Flag` for mobile, maybe `Cell` is all we need? I know I keep asking this... 🤣 don't laugh @pete-watters.

kyranjamie: Find it hard to believe that we won't use Flag again, it's such a common layout pattern? Cell can use Flag, no?

apps/mobile/src/features/settings/theme-modal.tsx
```diff
@@ -1,21 +1,67 @@
...
 export function ThemeModal({ modalRef }: ThemeModalProps) {
   const animatedIndex = useSharedValue<number>(CLOSED_ANIMATED_SHARED_VALUE);
+  const { bottom } = useSafeAreaInsets();
+  const theme = useTheme<Theme>();
+  const settings = useSettings();
+  const { displayToast } = useToastContext();
+
+  function onPressUpdateTheme(theme: ThemeStore) {
+    settings.changeTheme(theme);
+    displayToast({ title: t`Theme updated`, type: 'success' });
```
kyranjamie, line 27: Do we need toast here? Will be pretty obvious that the theme has changed

kyranjamie (commented):
system as default imo

apps/mobile/src/features/settings/theme-modal.tsx
fbwoolf: @mica000 it is shown in the design, what do you think?

packages/ui/src/components/cell/cell.native.tsx
fbwoolf: Yeah, hoping we can refactor Cell to use it, but I don't want to step on Pete's toes bc he was working on it. @pete-watters you want me to do?

packages/utils/src/index.ts
```diff
@@ -197,3 +197,7 @@ export function mapObject<T extends object, U>(
 export function assertIsTruthy<T>(val: T): asserts val is NonNullable<T> {
   if (!val) throw new Error(`expected: true, actual: ${val}`);
 }
+
+export function capitalize(val: string) {
```
pete-watters, line 201: Curious if we need this. Can't we just use the different variants of `text-transform`? 

```
text-transform: uppercase;
text-transform: capitalize;
```

I don't think we should do string manipulation for display purposes

apps/mobile/src/app/wallet/settings/display/index.tsx
pete-watters: I'm not sure we need to do this - https://github.com/leather-io/mono/pull/427/files#r1744823802

packages/ui/src/components/cell/cell.native.tsx
pete-watters: I can add `Flag` into `Cell`, no problem. I did it already so I can do it again. See this discussion @kyranjamie : https://github.com/leather-io/mono/pull/416#discussion_r1740683610 

We can use `Flag` in `Cell` but that's the only place to use it. We can also manipulate `ItemLayout` to fit into the usage of `Cell`. But then you may ask, what's the point as we use `Cell` everywhere and the only place we use `Flag` and `ItemLayout` will be in `Cell`. 

Of course they could be of use in the future but at this point in time we are just moving code around and not adding specific value. I think we can work smarter and build features during which we figure out what components we need rather than building loads of components that we might need or not

apps/mobile/src/app/wallet/settings/networks/index.tsx
```diff
@@ -29,24 +28,24 @@ export default function SettingsNetworksScreen() {
         }}
       >
         <Cell
+          isSelected
```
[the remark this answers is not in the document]
pete-watters: I tested this and I can't change the values? 🤔 

😂

apps/mobile/src/components/headers/modal-header.tsx
pete-watters: I can refactor this. `Modal` is to become `Sheet` and I can locate the header alongside that in `ui/packages/`. 

For the extension, we use some tokens for these types of heights. So we could move them to the tokens? 

Having a localised constants file is OK unless we need to use them in loads of places

apps/mobile/src/features/settings/theme-modal.tsx
```diff
@@ -1,21 +1,67 @@
...
+          gap: theme.spacing[5],
+        }}
+      >
+        <ModalHeader
+          Icon={SunIcon}
+          modalVariant="normal"
+          onPressSupport={() => {}}
+          title={t`Theme`}
+        />
+        <Cell
```
pete-watters, line 46: Maybe we could `map` / `switch` these options?

pete-watters (approved):
Nice work @fbwoolf 👍 

I added some feedback but no blockers. I'm curious if we can forever banish `capitalize(` or if I am missing the point

apps/mobile/src/features/settings/theme-modal.tsx
```diff
@@ -1,21 +1,67 @@
...
 
 interface ThemeModalProps {
   modalRef: RefObject<BottomSheetModal>;
 }
 export function ThemeModal({ modalRef }: ThemeModalProps) {
   const animatedIndex = useSharedValue<number>(CLOSED_ANIMATED_SHARED_VALUE);
+  const { bottom } = useSafeAreaInsets();
+  const theme = useTheme<Theme>();
+  const settings = useSettings();
+  const { displayToast } = useToastContext();
```
kyranjamie, line 23: This component could be better broken apart imo

Ideal outer component API might be something like

```
export function ThemeModal({ modalRef }: ThemeModalProps) {
  return (
    <ThemeModalLayout>
      <ThemeSwitcher activeTheme=}{ onUpdateTheme={() => {}}  />
    </ThemeModalLayout>
  );
}
```

apps/mobile/src/app/wallet/settings/display/index.tsx
kyranjamie: I don't think we can say we'll never need it, on grounds we should prefer css capitalisation. Working with external libraries for example where you can't access css. 

Either way, this is now no longer localised? How do we add a conditional, dynamic localised string?

pete-watters: Thanks @fbwoolf 👍 no problem using it, I was just wondering why

apps/mobile/src/features/settings/theme-modal.tsx
mica000: Good point, probably not but it's mostly added to keep consistently with all the other areas of the settings. Wouldn’t be weird to have all of others showing a toaster and only this one not?

mica000: Although, we might want to keep as there is also a case where users wont be able to see the change which is when you switch to the same theme the system is already using.

packages/utils/src/index.ts
fbwoolf: Alternatively, we could remove and always pass a `Text` `ReactNode` with the text transformed instead of passing the string?

apps/mobile/src/app/wallet/settings/display/index.tsx
fbwoolf: @kyranjamie it appears you have to use the `useLingui` hook for translating variables...

```ts
const { i18n } = useLingui();
i18n._(capitalize(themeStore))
```

fbwoolf:
@kyranjamie @pete-watters can you take a look at my second commit here? I spent quite a bit of time refactoring `Cell` today. You will see the old one renamed `CellToReplace` temporarily, I'll swap those out once we all agree my changes are ok. Look at the new `Cell` and `CellWithAvatar` (open to feedback on naming). `Cell` does not use the `Flag` so covers the last one here, while `CellWithAvatar` uses `Flag` to handle the others. I added an `actionIcon` so we can pass in the `Switch` or `RadioButton` as needed (or chevron). This removes needing the variant. Look at `ThemeSwitcher` for sample implementation.

Screenshot 2024-09-05 at 8 20 37 PM

packages/ui/src/components/item-layout/item-layout.native.stories.tsx
```diff
@@ -23,11 +22,8 @@ export default meta;
 
 export const ItemLayoutStory = {
   args: {
+    actionIcon: <ChevronRightIcon color="ink.action-primary-default" variant="small" />,
     captionLeft: 'captionLeft',
-    captionRight: 'captionRight',
```
pete-watters, line 27: Any reason to remove these? I added them to give the full picture of what `ItemLayout` can show

packages/ui/src/components/flag/flag.native.tsx
```diff
@@ -41,7 +40,7 @@ export function Flag({
       alignItems={getFlagAlignment(align)}
       style={[boxProps]}
     >
-      <Box flex={1} marginLeft={reverse ? spacing : '0'} marginRight={!reverse ? spacing : '0'}>
+      <Box marginLeft={reverse ? spacing : '0'} marginRight={!reverse ? spacing : '0'}>
```
pete-watters, line 43: I think this `flex={1}` is necessary for it to render properly. You need to give things at least a `flex={1`. 

You can see in storybook without it:
Screenshot 2024-09-06 at 05 51 34

This is the storybook with `flex={1`

packages/utils/src/index.ts
pete-watters: I think that could be good if it's always a localised string. I don't think it's a big deal though, up to you

pete-watters:
> @kyranjamie @pete-watters can you take a look at my second commit here? I spent quite a bit of time refactoring `Cell` today. You will see the old one renamed `CellToReplace` temporarily, I'll swap those out once we all agree my changes are ok. Look at the new `Cell` and `CellWithAvatar` (open to feedback on naming). `Cell` does not use the `Flag` so covers the last one here, while `CellWithAvatar` uses `Flag` to handle the others. I added an `actionIcon` so we can pass in the `Switch` or `RadioButton` as needed (or chevron). This removes needing the variant. Look at `ThemeSwitcher` for sample implementation.
> 
> Screenshot 2024-09-05 at 8 20 37 PM

This looks great to me, thanks for taking it on @fbwoolf 👍 It's much cleaner using `ItemLayout` and worth taking the time now to then update all the `Cell`s. 

I just have one Q about removing the `flex={1` as thats something I think all views / box's need. We could even just add it to our `box.native` wrapper though to avoid having to always add it. 

For context @kyranjamie , this is the discussion we originally had about it https://github.com/leather-io/mono/pull/416#discussion_r1740683610

packages/ui/src/components/radio-button/radio-button.native.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box } from '../box/box.native';
+
+interface RadioButtonProps {
+  isSelected: boolean;
+}
+export function RadioButton({ isSelected }: RadioButtonProps) {
```
kyranjamie, line 6: This looks good, but would've thought the UI lib has an actual button with radio like props and events, e.g. `onChange`

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,113 +1,31 @@
...
+import { TouchableOpacity } from '../button/touchable-opacity.native';
+import { ItemLayout } from '../item-layout/item-layout.native';
+
+interface CellWithoutAvatarProps {
+  actionIcon?: React.ReactNode;
+  captionLeft?: string;
+  captionRight?: string;
+  onPress?(): void;
+  titleLeft: string;
+  titleRight?: string;
```
kyranjamie, line 10: Thanks for refactoring @fbwoolf 

This makes me question whether we need "Cell" at all. It's just a wrapper around existing components, that feeds the props down. 

I don't think it adds any value versus composing this in every place we need it. 

The concept of _Cell_ could be a design-only concept that we simply implement on a case-by-case basis. Sometimes it's pressable, sometimes it isnt'. Sometimes has ItemLayout, sometimes has Flag only. There are too many variants to assume ahead of time, and I see no problem in using 2–3 extra lines of JSX for a Cell definition in mobile, if it provides us this flexibility and avoids us making assumptions.

@fabric-8 @mica000

kyranjamie:
Some of these conflicts are because of me, but hopefully just import path nes

packages/ui/src/components/item-layout/item-layout.native.stories.tsx
fbwoolf: Not sure why I removed, maybe was during some testing so I can add back. Tbh, this is what I encourntered with `ItemLayout` in storybook, so for me it just wasn't working as expected when I used in `ThemeSwitcher`...

image

packages/ui/src/components/flag/flag.native.tsx
fbwoolf: This is what I encountered trying to use `ItemLayout` with `Flag` in code, did you actually use it anywhere?

signal-2024-09-06-084504_002

Our web version of Flag doesn't behave that way? If I don't remove it, what else do you suggest for the Avatar here?

packages/ui/src/components/radio-button/radio-button.native.tsx
fbwoolf: The `RadioButton` isn't actually a button bc the entire row triggers the behavior, so maybe that is weird. If called a `...Button` maybe it should also have those properties so can be used on own?

packages/ui/src/components/cell/cell.native.tsx
fbwoolf: Yeah, it is a bit weird ...we also create our own `TouchableOpacity` so quite a few components. `ItemLayout` for mobile is interesting bc in our web version we always assume it needs to be a `Flag`, but that is not the case in mobile. So, maybe in mobile we just remove the `Flag` from `ItemLayout`, which is what I've done here, and compose them together where needed?

I'll remove my changes to `Cell`, and my use of it here. If it ends up working out better, I'll remove `Cell` in a different PR.

packages/ui/src/components/flag/flag.native.tsx
fbwoolf: Flag... https://csswizardry.com/2013/05/the-flag-object/

packages/ui/src/components/radio-button/radio-button.native.tsx
fbwoolf: I added the `TouchableOpacity` here and passing the props so it can be used like a button on own. When triggered by an onPress on a row, I just used `disabled` prop or the two buttons conflicted.

packages/ui/src/components/flag/flag.native.tsx
pete-watters: I was building `Flag` based on web and know that link. 

I had it working but the only place I was using it was in `ItemLayout` so far. This is the original PR I did:
https://github.com/leather-io/mono/pull/383 

I just know that not having this `flex={1}` has broken the display in the story. 

Are you using it anywhere? I was planning to use it in [the headers ](https://github.com/leather-io/issues/issues/230) so I can fix this then and let you merge this

fbwoolf: I am using it, yes, on the settings `Display` screen. It had to be changed to work. The web version, if you are following it, doesn't have the flex 1 on it. That is why I posted the link and mentioned it. I don't think you are building a Flag correctly?

My change (and web version w/out flex 1):
IMG_1689

Your version with flex 1:
IMG_1690

If I put a fixed width/height on the Avatar:
IMG_1691

What am I missing? Maybe the story you composed is the 'broken' part? Why are you feeling like the image part of the Flag should flex?

fbwoolf: @kyranjamie you built the web Flag, correct? Thoughts here?

kyranjamie: The idea of Flag is to simply position text next to an image. The implementation can vary, flex, float, grid etc, so long as the it doesn't make assumptions of the content inside.

As far as I understand flex: 1, this sizes an element relative to its siblings, which is not an assumption Flag should make.

fbwoolf: Yeah, I did this in native story that is similar to implementation in web story...

Screenshot 2024-09-06 at 12 54 16 PM

fbwoolf: That is with removing flex 1, so not really broken. @pete-watters let me know if you understand a prob with this?

pete-watters: My understanding is that in React Native, you have to declare `flex={1` [for each view for it to appear](https://reactnative.dev/docs/flexbox), maybe that's incorrect though and children can force the parent to view if they have dimensions supplied.  

I can see it is working in your updated story, passing a box with dimensions 50x50. Before, it wasn't showing the `Img` section - (the green part of the story) without using Flex = 1. It could have been that my demo `Box` for `img` was lacking this dimension. 

The Flag I implemented is basically the same as this but with a `flex=1` on the first element so I do know how to build it. It just looked broken when I checked storybook so flagged it. It looks fine now 👍

PR #428, opened 2024-09-04
feat: settings network

fbwoolf:
This PR is branched off my theme settings branch. It wires up being able to change networks.

signal-2024-09-04-163006_003 signal-2024-09-04-163006_002

### DIFF
--- a/apps/mobile/src/app/wallet/settings/networks/index.tsx
+++ b/apps/mobile/src/app/wallet/settings/networks/index.tsx
@@ -1,82 +1,27 @@
-import { ScrollView } from 'react-native-gesture-handler';
-import { useSafeAreaInsets } from 'react-native-safe-area-context';
-
+import { useToastContext } from '@/components/toast/toast-context';
+import { useSettings } from '@/store/settings/settings.write';
 import { t } from '@lingui/macro';
-import { useTheme } from '@shopify/restyle';
 
-import {
-  Avatar,
-  Box,
-  Flag,
-  GlobeIcon,
-  ItemLayout,
-  PlaygroundFormsIcon,
-  RadioButton,
-  TestTubeIcon,
-  Theme,
-  TouchableOpacity,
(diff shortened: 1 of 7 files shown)


apps/mobile/src/app/wallet/settings/networks/index.tsx
```diff
@@ -28,23 +38,26 @@ export default function SettingsNetworksScreen() {
         }}
       >
         <Cell
-          isSelected
+          isSelected={settings.network.id === 'mainnet'}
+          onPress={() => onPressChangeNetwork('mainnet')}
           title={t`Mainnet`}
-          subtitle={t`Enabled`}
+          subtitle={settings.network.id === 'mainnet' ? t`Enabled` : t`Disabled`}
```
pete-watters, line 44: It seems like we could clean this up a bit but I think it's fine as is. 

Maybe destructure and renaming `settings.network.id` earlier? 

```
const { network: { id: networkId } } = settings
```

Or we could have a `switch` to pass props to `<Cell` based on `settings.network.id`

pete-watters (approved):
Nice work 👍 I added a suggestion but fine as is

apps/mobile/src/app/wallet/settings/networks/index.tsx
kyranjamie: Just a style preference, but isn't accessing a few properties deep of an object cleaner than a nested destructure?

On grounds that `{ network: { id: networkId } }` is more syntax heavy than `settings.network.id`?

kyranjamie (approved):

apps/mobile/src/app/wallet/settings/networks/index.tsx
pete-watters: I think it's cleaner to do one destructure to get what you need than multiple chains. Just my preference though

PR #430, opened 2024-09-05
feat: shared stacks signing logic

kyranjamie:
This PR adds code for the Stacks signer. This gets initialised in part of the keychain selector.

Next, I aim to figure out the best way to consume these signers in the app. I am **very reluctant** to continue using the pattern from the extension where we do something like

```ts
const signer = useCurrentAccountStacksSigner(); // `signer | undefined`
```

Instead I want a pattern where we pass signers around dynamically, which might look something like:

```tsx
function Assets () {
  return signers.map(signer => {
    switch (signer.type) {
      case 'stacks': return <StacksSigner signer={signer} />
      case 'bitcoin': return <BitcoinSigner signer={signer} />
    }
  });
}
```

This way, we can consume the signer well in the knowledge it exists

or 

```tsx
<SignerLoader requiredKeys={[keyOrigin1, keyOrigin2]} fallback={<AddMoreKeysOrWarningComponent />}>
  {signers => <SendForm signers={signers} />}
</SignerLoader>


kyranjamie:
Merging this, and will push a follow up with proposal to above concerns

PR #434, opened 2024-09-09
feat: refactor ModalHeader to SheetHeader, ref leather-io/issues#230

pete-watters:
This PR refactors `ModalHeader` -> `SheetHeader` and simplifies it by making it more composable and removing variant lookup functions


apps/mobile/src/components/wallet-settings/remove-wallet-modal.tsx
```diff
@@ -1,33 +1,53 @@
...
+        <Box gap="3">
+          <Button onPress={onSubmit} buttonState="default" title={t`Proceed`} />
+          <Button
+            onPress={() => sheetRef.current?.dismiss()}
+            buttonState="ghost"
+            title={t`Cancel`}
+          />
+        </Box>
+      </Box>
+    </Sheet>
```
kyranjamie, line 51: Previously this component was a really clean wrapper around a layout-like component. Any chance we can continue to separate logic and styles per [layout convention](https://www.notion.so/trustmachines/Use-layout-components-b94ad409bbf5417395a2bc9d22574da1?pvs=4)

packages/ui/src/components/sheet/components/sheet-header.native.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export function SheetHeader({
+  title,
+  icon,
+  onPressSupport,
+}: {
+  title: string;
+  icon: React.ReactNode;
+  onPressSupport?(): unknown;
+}) {
```
kyranjamie, line 13: Can we create a separate interface for the props, [re convention here](https://www.notion.so/trustmachines/Naming-Component-Props-13e6ba7dd5ab4404a1cda0de6175cda6?pvs=4)

apps/mobile/src/components/wallet-settings/remove-wallet-modal.tsx
pete-watters: I can refactor it again and try clean it up. 

I changed it as the original was accepting `variant="critical"` that fed into lookup functions to determine icon colour. 

I can refactor but this is actually a layout component with no real logic , just avoiding a wrapper layer to drill props down to.

pete-watters: I reverted this but with a cleaner version of `WarningSheet` without using lookup functions for colours. 

I'm not sure about this as we have several layers of wrapper components in different folders. Heres an example:

- `wallet-settings/configure/[wallet]/index.tsx` imports `<RemoveWalletSheet` from `src/components/wallet-settings/remove-wallet-sheet`
- `<RemoveWalletSheet` is a wrapper component that loads another wrapper - `WarningSheet` from `src/components/sheets`
- `WarningSheet` is in itself a layout component configuring `Sheet` from the UI library. 

What this achieves is passing a different icon / icon colour and some text. 

I think it would be cleaner to extend the code of `Sheet` and then load that directly without all of the layers and drilling.

apps/mobile/src/components/secure-your-wallet/skip-secure-wallet-sheet.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { t } from '@lingui/macro';
+
+import { SheetRef } from '@leather.io/ui/native';
+
+import { WarningSheet } from '../sheets/warning-sheet.layout';
+
+interface SkipSecureWalletSheetProps {
+  sheetRef: RefObject<SheetRef>;
+  onSubmit(): unknown;
```
kyranjamie, line 11:
```suggestion
  onSubmit(): void;
```
onSubmit doesn't expect a return value

kyranjamie (approved):

PR #421, opened 2024-09-03
feat: add avatar selector

edgarkhanzadian:
https://github.com/user-attachments/assets/ad8b261e-6ca6-480d-ae23-46261ab9c56c

PR #435, opened 2024-09-09
fix: fix issue with translations leather-io/issues#136

pete-watters:
I made a mistake of some kind when rebasing https://github.com/leather-io/mono/pull/421/ and it introduced some translation errors which I have now fixed. 

It added a lot of `<<<<<<< HEAD` comments into the file


kyranjamie (approved):
yeah I've seen this happen as well

PR #436, opened 2024-09-09
refactor(mobile): split store files

kyranjamie:
Finishes moving around store files per naming convention


apps/mobile/src/features/account-selector-sheet.tsx
```diff
@@ -60,7 +61,7 @@ export function AccountSelectorSheet({ sheetRef }: { sheetRef: RefObject<SheetRe
   );
 
   return (
-    <Sheet isScrollView shouldHaveContainer={false} ref={sheetRef} themeVariant={themeVariant}>
+    <Sheet isScrollView ref={sheetRef} themeVariant={themeVariant}>
```
pete-watters, line 64: Did you mean to remove `shouldHaveContainer`? This seems to be the only place that uses it to avoid wrapping sheet children with `BottomSheetComponent`.

pete-watters (approved):
This LGTM I am just unsure about [removing shouldHaveContainer={false}](https://github.com/leather-io/mono/pull/436/files#r1750317884)

If this isn't needed here we can in fact also update the `Sheet` component to deprecate it

apps/mobile/src/features/account-selector-sheet.tsx
kyranjamie: Currently, the My Accounts button is broken

image

Removing this seems to fix it

kyranjamie:
It's still used for now in one place, let's ask Edgar about it when he's back

pete-watters:
> It's still used for now in one place, let's ask Edgar about it when he's back

Just double checking this and I can't see it used anywhere now? We can check with Edgar when he's back but removing this can simplify `Sheet` and avoid this logic:
```
  {shouldHaveContainer && (
          <BottomSheetComponent
            style={{
              backgroundColor: theme.colors['ink.background-primary'],
              paddingBottom: bottom,
              borderTopLeftRadius: theme.borderRadii.lg,
              borderTopRightRadius: theme.borderRadii.lg,
              overflow: 'hidden',
            }}
          >
            {children}
          </BottomSheetComponent>
        )}
        {!shouldHaveContainer && children}
```

PR #439, opened 2024-09-11
feat: accounts widget ref leather-io/issues#136

pete-watters:
This PR adds the basic accounts widget: 

https://github.com/user-attachments/assets/a624ac2c-b451-47c1-9f25-8a4c4c85085a

There is some more work to do:
- fetching real account balances
- creating a new page for `AccountOverview`

I can continue on this branch or add in a follow up, I just wanted to open for earlier review


apps/mobile/src/components/widgets/widget.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
pete-watters, line 1: I'm not sure if I should move `widgets` to the UI library? 

I can as I work on building them out but we don't really share them on web yet. I could to enforce a cleaner pattern of UI in the UI lib

apps/mobile/src/components/widgets/account/utils/format-balance.ts
```diff
@@ -0,0 +1,11 @@
+// FIXME: This is a temporary solution to format the balance. We need to use the same format as the extension.
```
pete-watters, line 1: I got AI to fire this out for me. I will migrate the `extension` format balance in https://github.com/leather-io/issues/issues/138

apps/mobile/src/components/widgets/account/account-widget.tsx
```diff
@@ -0,0 +1,105 @@
...
+      balance: index + 10,
+    };
+  });
+
+  const mockedWalletBalance = mockedAccounts.reduce((sum, account) => sum + account.balance, 0);
+
+  if (selectedAccount) {
+    return (
+      <>
+        {/* FIXME: This is throwaway code to add a fake header
```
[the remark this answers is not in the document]
fabric-8: fine to use a regular page transition (from right-left) for now!

apps/mobile/src/components/widgets/widget.tsx
fbwoolf: What is a `widget` exactly? It looks like just a `Box` component? Anything special about it?

pete-watters: Nope, nothing special just a `Box` but I thought if we are building a few of them better to have a wrapper than repeat the box and padding.

I can also get rid of it?

apps/mobile/src/components/widgets/account/account-widget.tsx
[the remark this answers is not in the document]
pete-watters: It just needs a bit more work than that Fab, as I will need to move to a different page entirely I think that includes the `Tokens` widget.

So we will have the `Home` page screen then move to a new page for `AccountOverview` and update the header.

apps/mobile/src/components/widgets/widget.tsx
pete-watters: I guess it's similar to `Cell` in that it helps match design to dev. 

`Box`s are essentially `div`s anyway 😄

apps/mobile/src/components/home/home.tsx
```diff
@@ -1,53 +1,20 @@
...
-            <Earn />
-          </Box>
-        </ScrollView>
-      </Box>
-      <ActionBarContainer ref={actionBarRef} />
-    </ActionBarContext.Provider>
+    <HomeLayout>
+      <AccountWidget accounts={accounts.list} wallets={wallets.list} />
+      <Earn />
+    </HomeLayout>
```
kyranjamie, line 18: 👏🏼 thanks Pete, exactly how a root feature component should look IMHO

apps/mobile/src/components/widgets/account/account-header.tsx
```diff
@@ -0,0 +1,34 @@
...
+        onPress={() => sheetRef.current?.present()}
+        flexDirection="row"
+        gap="1"
+        alignItems="center"
+      >
+        <AccountHeaderText />
+        <ChevronRightIcon variant="small" />
+      </TouchableOpacity>
+
+      <AccountSelectorSheet sheetRef={sheetRef} />
```
kyranjamie, line 31: Is this a smart component? If so, should it be passed in a children instead?

apps/mobile/src/components/widgets/account/account-widget.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const insets = useSafeAreaInsets();
+
+  const router = useRouter();
+
+  const hasWallets = wallets.length > 0;
+  const hasAccounts = accounts.length > 0;
+
+  // FIXME: this is fake data for now
+  // need to implement fetch of account?.type + balances
+  const mockedAccounts = accounts.map((account, index) => {
```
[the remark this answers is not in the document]
kyranjamie: watch this space, hope to have something asap to remove all mocked code

apps/mobile/src/components/widgets/account/ledger-badge.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const theme = useTheme<Theme>();
+  return (
+    <Box
+      position="absolute"
+      p="1"
+      borderColor="ink.border-transparent"
+      borderRadius="xs"
+      borderWidth={1}
+      right={theme.spacing[5]}
+      top={theme.spacing[5]}
```
kyranjamie, line 16: can't this just be {5}

kyranjamie (approved):
Great start Pete, some thoughtful component separation here. 

Not 100% on widget yet and how exactly this component should look, but can start with this

apps/mobile/src/components/widgets/widget.tsx
fbwoolf: I did the same with `Avatar` bc the code was being repeated, hoping we can expand on what I relocated to the UI lib too.

fbwoolf (approved):
🚀

apps/mobile/src/components/widgets/account/cards/account-card.layout.tsx
```diff
@@ -0,0 +1,52 @@
+import { WalletStore } from '@/store/wallets/wallets.write';
+import { useTheme } from '@shopify/restyle';
+
+import { Box, IconProps, Text, TouchableOpacity } from '@leather.io/ui/native';
+
+import { LedgerBadge } from '../ledger-badge';
+
+export interface AccountCardLayoutProps {
+  Icon: React.FC<IconProps>;
```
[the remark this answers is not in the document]
pete-watters: I was changing them to be lowercase but then as I worked I saw that we need to pass props to them. 

We use a function `getAvatarIcon('string')` to determine the icon, then we pass it in a slot as a prop and when we render it then we often need to pass color / dimensions:

```
// wallet-card
 {accountsList.map(account => {
            return (
              <AccountCard
                Icon={getAvatarIcon(account.icon)}
```

```
// account-card
export function AccountCard({ Icon}){
        <Icon color={theme.colors['ink.background-primary']}  width={32} height={32} />
...
```

I agree it looks weird but then it seems to work right now. I'm agnostic as to removing / using, it seems useful for this icon case but then it's a weird pattern we don't really use elsewhere

apps/mobile/src/components/widgets/account/account-header.tsx
pete-watters: I'm thinking that perhaps I will integrate this into `Widget` as most of them seem to have headers. 

I'll think about how to improve it and refactor to pass it as children to improve when working on https://github.com/leather-io/issues/issues/138

pete-watters:
Thanks @fbwoolf + @kyranjamie . I'm going to merge this to keep the momentum going but will take on board the suggestions and add into the next batch of work

apps/mobile/src/components/widgets/widget.tsx
pete-watters: Sounds good. We can move fast but take notes as we go and implement more re-usable patterns. Even if we have simple components not doing much I think it's nice to read code like this:
```
 <Widget>
   {something.map(thing => 
       <Cell name={thing.name}/>
 }
</Widget>

```

apps/mobile/src/components/widgets/account/ledger-badge.tsx
pete-watters: I will experiment with this, I copied this from elsewhere. 

If we set it to `top={5}` I think that's 5 pixel units but I see other areas where we set `px='5'` and that uses our tokens so I will try and refactor away `theme.spacing[5]` if possible. 

I opened a task for it so we don't forget https://github.com/leather-io/issues/issues/321

apps/mobile/src/components/widgets/widget.tsx
fbwoolf: I suppose with this approach, the refactor I did to keep `Cell` would make sense even though it just wrapped `TouchableOpacity` to create consistent formatting?

PR #440, opened 2024-09-11
feat(bitcoin): signer logic

kyranjamie:
This PR implements the signing logic discussed in https://github.com/leather-io/issues/issues/243

**Performance**
By default, the wallet will do a lot of re-deriving of the same data. xpubs → `HDKey`s, `HDKey`s → payment objects etc. Selectors have their own memoization to reduce this load, but when a new wallet is created, the store changes, and so _all accounts_ are re-derived. This noticeably slows the wallet. I've built this such that the expensive derivation functions are memoized. In the demo, you can see how many wallet operations are taken place, and we're only deriving the newly created data.

<details>

<summary>Memoization demo</summary>

https://github.com/user-attachments/assets/291f0add-6f2c-4979-b3b6-5982f76cfa04

</details>

**Differences with extension**
As described in https://github.com/leather-io/issues/issues/243 signing metadata is annotated _when a transaction is authored_. Much of the logic in `@leather.io/bitcoin` is serialising/deserialising data from transactions. This might look kinda verbose, but it's much simpler than creating a whole system of passing around metadata with a PSBT. [See comment →](https://github.com/leather-io/mono/pull/440#discussion_r1758668522).

**Naming**
In the extension we use the nomenclature of _Accounts_ and _Signers_, where _Account_ refers to a given account-level keychain, and _Signer_ is the payment object, with address, and methods to sign inputs. 

[Due to differences in the way we store keys, this doesn't translate well to mobile.](https://github.com/leather-io/issues/issues/243#issuecomment-2340138405).

In this PR I'm using the terms _Account_ and _Payer_. Where the _Payer_ contains all "payment" information, such as address, witness script etc, key origin, but doesn't contain signing methods. Open to feedback on this naming, but fairly confident this naming works. Reluctant to call `BitcoinAddress` or similar, and this describes what it is, rather than the purpose of the object (and also contains more useful props for creating txs).

**Misc. refactoring**
I've created some types `WalletId` and `AccountId` as we very frequently use these to refer to a given wallet or account. I don't think we need to call the enum `APP_ROUTES` just because its globally used. Enums are usually PascalCased.

**Todo**
- [x] Remove comments
- [x] Derive keys and sign tx from mnemonic in final signing step
- [x] Figure out typing for Bitcoin accounts
- [x] Remove dummy keys function


apps/mobile/src/app/wallet/developer-console/bitcoin-scratch-pad.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Button
+        buttonState="default"
+        title={t`Sign`}
+        mt="5"
+        onPress={() => {
+          const tx = new btc.Transaction();
+          tx.addInput({
+            txid: '19c52a02cf783c73028b32aabad3b97afec7958052a45bd1504b84900e419bc6',
+            index: 0,
+            bip32Derivation: [payerToBip32Derivation(nativeSegwitPayer)],
```
kyranjamie, line 59: See here how, when creating a transaction, we decorate the transaction with `bip32Derivation` or `tapBip32Derivation`. This is required for Ledger transactions, and is also useful for our software implementation.

If we omit this, by the time the transaction reaches PsbtSigner, we won't know which key corresponds to each input. In our extension, we use an InputSigningConfig that includes metadata indicating which key owns each input. This works, but isn't ideal as we 1) have to create our own kinda schema to live alongside a PSBT, 2) duplicate the information anyway to support Ledger.

Much of the new logic this PR is to make it trivial to declare, and later read, the key derivation info of each key, such that by the time we get to signing a tx, we have full instructions on how to sign a input.

apps/mobile/src/app/wallet/(tabs)/_layout.tsx
```diff
@@ -5,7 +5,7 @@ import { ActionBarMethods } from '@/components/action-bar';
 import { ActionBarContainer, ActionBarContext } from '@/components/action-bar/container';
 import { BlurredHeader } from '@/components/headers/containers/blurred-header';
 import { TabBar } from '@/components/tab-bar';
-import { APP_ROUTES } from '@/routes';
+import { AppRoutes } from '@/routes';
```
pete-watters, line 8: Maybe we should change this to be `RouteUrls` to align with the extension ?

apps/mobile/src/components/add-wallet/add-wallet-sheet.tsx
```diff
@@ -19,12 +19,12 @@ export function AddWalletSheet({ addWalletSheetRef }: AddWalletSheetBaseProps) {
   const router = useRouter();
   const [optionData, setOptionData] = useState<OptionData | null>(null);
   const createWallet = useCallback(() => {
-    router.navigate(APP_ROUTES.WalletCreateNewWallet);
+    router.navigate(AppRoutes.WalletCreateNewWallet);
```
pete-watters, line 22: Maybe you can fix the `hooks` lint error now too? Just adding `addWalletSheetRef` to the dependancies?

apps/mobile/src/components/wallet-manager/index.tsx
```diff
@@ -1,45 +1,39 @@
 import { Text } from 'react-native';
 
+import { AccountId, WalletId } from '@/models/domain.model';
 import { useAccountsByFingerprint } from '@/store/accounts/accounts.read';
 import { useKeyStore } from '@/store/key-store';
-import { useBitcoinKeychains } from '@/store/keychains/bitcoin/bitcoin-keychains.read';
-import { useStacksKeychains } from '@/store/keychains/stacks/stacks-keychains.read';
+import { useBitcoinAccounts } from '@/store/keychains/bitcoin/bitcoin-keychains.read';
```
pete-watters, line 6: I like this renaming to `Accounts`, `Signers` + `Payer` 👍

packages/bitcoin/src/bitcoin.utils.ts
```diff
@@ -153,17 +157,30 @@ export function whenPaymentType(mode: PaymentTypes | BtcSignerLibPaymentTypeIden
...
  * Below we see path has 86 in it, per convention, this refers to taproot payments
  * @example
  * `m/86'/1'/0'/0/0`
  */
 export function inferPaymentTypeFromPath(path: string): PaymentTypes {
-  if (path.startsWith('m/84')) return 'p2wpkh';
-  if (path.startsWith('m/86')) return 'p2tr';
-  if (path.startsWith('m/44')) return 'p2pkh';
-  throw new Error(`Unable to infer payment type from path=${path}`);
+  const purpose = extractPurposeFromPath(path);
```
pete-watters, line 173: Nice refactor 👍

packages/bitcoin/src/bitcoin-signer.ts
```diff
@@ -1,13 +1,40 @@
-import { HDKey } from '@scure/bip32';
+import { HARDENED_OFFSET, HDKey } from '@scure/bip32';
```
pete-watters, line 1: Will we split this up into a sub package / smaller files?

pete-watters:
This looks good on initial glance. Nice work with the sematic re-naming

packages/bitcoin/src/bitcoin-signer.ts
kyranjamie: We can create sub-esm packages like `@leather.io/bitcoin/signers`, but would only bother with that when bloat becomes an issue

fbwoolf (approved):
Looks good, thanks for explaining the changes you made from the extension implementation. 👍

pete-watters (approved):

PR #447, opened 2024-09-13
feat: settings bitcoin unit

fbwoolf:
This PR wires up bitcoin unit in the display settings flow.


packages/models/src/bitcoin-unit.model.ts
```diff
@@ -0,0 +1,46 @@
...
+export const bitcoinUnitsKeyedByName: Record<WalletBitcoinUnit, BitcoinUnitInfo> = {
+  [WalletBitcoinUnit.bitcoin]: {
+    name: WalletBitcoinUnit.bitcoin,
+    symbol: 'BTC',
+    decimal: BigNumber(1),
+  },
+  [WalletBitcoinUnit.millibit]: {
+    name: WalletBitcoinUnit.millibit,
+    symbol: 'mBTC',
+    decimal: BigNumber(0.001),
```
fbwoolf, line 29: @kyranjamie using here, this seem ok?

packages/models/src/bitcoin-unit.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  },
+  [WalletBitcoinUnit.satoshi]: {
+    name: WalletBitcoinUnit.satoshi,
+    symbol: 'sat',
+    decimal: BigNumber(0.000001),
+  },
+  [WalletBitcoinUnit.millisatoshi]: {
+    name: WalletBitcoinUnit.millisatoshi,
+    symbol: 'msat',
+    decimal: BigNumber(0.00000000001),
```
kyranjamie, line 44: doesn't BN need a `new` initialiser?

packages/models/src/bitcoin-unit.model.ts
kyranjamie: wouldn't this be something for constant pkg?

packages/models/src/bitcoin-unit.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  [WalletBitcoinUnit.bit]: {
+    name: WalletBitcoinUnit.bit,
+    symbol: 'μBTC',
+    decimal: BigNumber(0.000001),
+  },
+  [WalletBitcoinUnit.satoshi]: {
+    name: WalletBitcoinUnit.satoshi,
+    symbol: 'sat',
+    decimal: BigNumber(0.000001),
+  },
```
kyranjamie, line 40: satoshi is the smallest unit, so afaik this is wrong, and there is no millisatoshi?

packages/models/src/bitcoin-unit.model.ts
```diff
@@ -0,0 +1,46 @@
...
+  [WalletBitcoinUnit.bit]: {
+    name: WalletBitcoinUnit.bit,
+    symbol: 'μBTC',
+    decimal: BigNumber(0.000001),
+  },
+  [WalletBitcoinUnit.satoshi]: {
+    name: WalletBitcoinUnit.satoshi,
+    symbol: 'sat',
+    decimal: BigNumber(0.000001),
```
kyranjamie, line 39:
```suggestion
    decimal: BigNumber(0.00000001),
```

packages/models/src/bitcoin-unit.model.ts
fbwoolf: I can move there ...I forgot the `new` syntax btw, haha.

packages/models/src/bitcoin-unit.model.ts
fbwoolf: Copied from: https://en.bitcoin.it/wiki/Units

packages/models/src/bitcoin-unit.model.ts
fbwoolf: Yep, see I missed that now.

packages/models/src/bitcoin-unit.model.ts
fbwoolf: Meant to put this here: https://en.bitcoin.it/wiki/Units

fbwoolf: Screenshot 2024-09-13 at 10 24 04 AM

packages/models/src/bitcoin-unit.model.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type BitcoinUnitSymbol = 'BTC' | 'sat';
+
+export interface BitcoinUnitInfo {
+  name: WalletBitcoinUnit;
+  symbol: BitcoinUnitSymbol;
+  decimal: string;
+}
+
+export const bitcoinUnitsKeyedByName: Record<WalletBitcoinUnit, BitcoinUnitInfo> = {
```
fbwoolf, line 15: We use a similar pattern for `networks`, should both be moved to `constants`?

apps/mobile/src/features/settings/bitcoin-unit-sheet/bitcoin-unit-cell.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { BitcoinUnitInfo, WalletBitcoinUnit } from '@leather.io/models';
+import { ItemLayout, RadioButton, TouchableOpacity } from '@leather.io/ui/native';
+
+interface BitcoinUnitCellProps {
+  activeBitcoinUnit: WalletBitcoinUnit;
+  onUpdateBitcoinUnit(unit: WalletBitcoinUnit): void;
+  unit: BitcoinUnitInfo;
+}
+export function BitcoinUnitCell({
```
fbwoolf, line 11: @pete-watters curious your thoughts here bc just using the `Cell` ref'd in designs as the component/file name here rather than using a shared `Cell` from the UI library?

packages/models/src/bitcoin-unit.model.ts
```diff
@@ -0,0 +1,26 @@
...
+  [WalletBitcoinUnit.bitcoin]: {
+    name: WalletBitcoinUnit.bitcoin,
+    symbol: 'BTC',
+    decimal: '1',
+  },
+  [WalletBitcoinUnit.satoshi]: {
+    name: WalletBitcoinUnit.satoshi,
+    symbol: 'sat',
+    decimal: '0.00000001',
```
fbwoolf, line 24: Went with a string here, we will have to use `initBigNumber` helper when used, thoughts?

fbwoolf:
@markmhendrickson need your feedback here on Monday bc the team talked abt reducing offered bitcoin units to just `BTC` and `sats` to start. We weren't sure why the others were being prioritized?

pete-watters:
I tested this but when I try and reach `Display` - `Settings` it's crashing with: 
```
 ERROR  TypeError: Cannot read property 'symbol' of undefined

This error is located at:
    in SettingsDisplayScreen
```


https://github.com/user-attachments/assets/40099822-56b3-4fcb-9137-1bc32a112d35

Maybe I'm missing a step where I need to regenerate lingui files?

apps/mobile/src/features/settings/bitcoin-unit-sheet/bitcoin-unit-cell.tsx
pete-watters: I think now we have a lot of separate specific `Cell`s almost doing the same thing but we can compose them on a specific case basis to meet our needs which is good.

I think it could be better to improve the shared `Cell` in the library to:
- accept a `flagImg` and if received wrap `ItemLayout` in a flag
- get rid of the `variant` in the `Cell` and accept `actionItem` instead of `actionIcon` that can dictate what it does - e.g. `Switch` etc. 

We also seem to often have a `Sheet` mapping `Cell`s so maybe we could even abstract this as a variant of `Sheet` that has data and then maps it to `Cell`s? 

I think what you have done here is perfect for now as we build out features as it will ensure they work independently, then in future we can assess overlap between `Cells` and do a refactor if we want, having already delivered a POC

packages/models/src/bitcoin-unit.model.ts
pete-watters: We have `networkModes` in `network.models`, maybe we can use those? 

I'm not sure I understand totally. I am happy to keep these types local to the model and load from there if needed rather than move to constants

packages/models/src/bitcoin-unit.model.ts
pete-watters: I think thats OK

pete-watters (approved):
Great work here @fbwoolf . I had [trouble getting to the settings screen](https://github.com/leather-io/mono/pull/447#issuecomment-2352051898) with the APP crashing but I guess it's some step I'm missing

packages/models/src/bitcoin-unit.model.ts
kyranjamie: The idea of models is that it represents the structure of the data we use, not the data itself. Have also thought that network should be moved elsewhere, probably constants

apps/mobile/src/features/settings/bitcoin-unit-sheet/bitcoin-unit-cell.tsx
kyranjamie: We removed `Cell` entirely from the library right? 

Just a few comments here, if we use a Flag inside a component to do some kind of layout, that's an implementation detail of the component that doesn't need to be exposed. So `img` not `flagImg` as a prop.

pete-watters: We [still have](https://github.com/leather-io/mono/blob/dev/packages/ui/src/components/cell/cell.native.tsx) `Cell` and are using it in a few places. It's similar to these new `Cell`s but it's still using `Box`s rather than `ItemLayout` / `Flag` and supporting other variants.

I think we can take the improvements made here and centralise it to that `Cell`

kyranjamie (approved):

markmhendrickson:
> need your feedback here on Monday bc the team talked abt reducing offered bitcoin units to just BTC and sats to start, we weren't sure why the others were being prioritized?

The idea here is that users may want a wider range of options, given just how much BTC they're dealing with in the app. bitcoin and sats are indeed the most common units these days; however, the other ones may help users with more mid-sized or smaller holdings.

My presumption is that the work involved would be relatively the same regardless (between two or five options) in that we need to support an array of options, and whichever is chosen, we'll apply that to a formatting function throughout the app? However, I'm happy to release with just two to start if there is indeed a significant work difference.

Reference: https://bitcoin.design/guide/designing-products/units-and-symbols/

apps/mobile/src/features/settings/bitcoin-unit-sheet/bitcoin-unit-cell.tsx
fbwoolf: `Cell` only still exists bc the issue to remove it entirely hasn't been done yet: https://github.com/leather-io/issues/issues/314

My proposal here is to still remove it, and instead just use the naming design team has settled on in our mobile code directly. When I map thru the list, typically we would name the component `...Item` that is rendered. Here, I was proposing we use, `...Cell`. Not critical to do, was just an idea to relate it to designs rather than having a UI lib component named `Cell`.

fbwoolf:
> I tested this but when I try and reach `Display` - `Settings` it's crashing with:
> 
> ```
>  ERROR  TypeError: Cannot read property 'symbol' of undefined
> 

This was happening to me until I purged the persisted redux store. This is likely exposing a prob with our store setup, bc we shouldn't haven't to do that, right?

pete-watters:
> > I tested this but when I try and reach `Display` - `Settings` it's crashing with:
> > ```
> >  ERROR  TypeError: Cannot read property 'symbol' of undefined
> > ```
> 
> This was happening to me until I purged the persisted redux store. This is likely exposing a prob with our store setup, bc we shouldn't haven't to do that, right?

OK let me check it out. Yes I don't think we should need to purge the store but maybe it's something we have to deal with

PR #448, opened 2024-09-13
feat: tokens widget

pete-watters:
This PR adds a static UI for the tokens widget and related UI components / migrations to the monorepo.

I improvised how to handle locked STX balances so we have support for that from the start 

Simulator Screenshot - iPhone 15 Pro - 2024-09-17 at 06 41 51

**Storybook**
- chip
- bullet separator

https://github.com/user-attachments/assets/76d77086-ecb4-40f5-a70f-7fa415184132

**NOTE** This is not 100% as per design which includes a network label but I decided to improvise and make it more like the extension for now so it can be connected fully. 

The design has: 
- a nested icon for the blockchain
- the blockchain label underneath the coin name
- the ticker on the right 

Screenshot 2024-09-17 at 09 09 38


packages/ui/tsconfig.native.json
```diff
@@ -6,7 +6,7 @@
     "tsBuildInfoFile": "tsconfig.native.tsbuildinfo",
     "jsx": "react-native",
     "jsxImportSource": "react",
-    "lib": ["es2017", "DOM"],
+    "lib": ["ESNext", "DOM"],
```
pete-watters, line 9: Adding this to add support for `flatMap` in `BulletSeparator`

packages/utils/src/abbreviate-number/abbreviate-number.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { abbreviateNumber } from './abbreviate-number';
```
pete-watters, line 1: Migrating `abbreviateNumber` from the `extension` and getting claude to generate these tests.

I'll need to refactor + remove from the `extension`

packages/utils/src/money/format-balance/format-balance.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { formatBalance } from './format-balance';
```
pete-watters, line 1: Migrating `formatBalance` from the extension and getting claude to generate these tests. We do some specific formatting here, it could be good to refactor these tests with more detail on the business logic. 

I'll need to refactor + remove from the extension

packages/utils/src/money/format-balance/format-balance.ts
```diff
@@ -0,0 +1,22 @@
+import { abbreviateNumber } from '../../abbreviate-number/abbreviate-number';
+import { removeCommas } from './remove-commas';
+
+/**
```
pete-watters, line 4: When I dug into this it seems like we could perhaps refactor this a bit based on the data we get from the API.

packages/utils/src/money/format-balance/remove-commas.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { removeCommas } from './remove-commas';
```
pete-watters, line 1: More `claude` tests based on the current function behaviour

apps/mobile/src/components/widgets/tokens/tokens.mocks.tsx
```diff
@@ -0,0 +1,121 @@
+import { t } from '@lingui/macro';
```
pete-watters, line 1: I'm using this mock file to think through what we actually need to do here. 

In the extension, we have separate files with `BtcCryptoAssetItem`, `StacksCryptoAssetItem` and the code can be a bit hard to follow. 

That could be something we need to do based on having to have a `Loader` per token. 

I will try to create a serialiser that gathers the data then creates a `tokens` array of objects with all the data needed and map over that.

apps/mobile/src/components/widgets/components/balance/fiat-balance.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useMemo } from 'react';
```
pete-watters, line 1: I will try refactor `FiatBalance` and `TokenBalance` to merge them

packages/ui/src/components/avatar/btc-avatar-icon.native.tsx
```diff
@@ -0,0 +1,13 @@
+import BtcIcon from '../../assets/icons/bitcoin.svg';
+import { Icon } from '../../icons/icon/icon.native';
+import { Avatar } from './avatar.native';
+
+export function BtcAvatarIcon() {
```
pete-watters, line 5: @fbwoolf Do you think this is enough to satisfy https://github.com/leather-io/issues/issues/290? 

If we accept a color to `<Avatar` then we can pass `Icon` and an `svg` as children like this

apps/mobile/src/components/widgets/components/balance/format-locked-balance.ts
```diff
@@ -0,0 +1,11 @@
+interface Balance {
+  amount: string;
+  symbol: string;
+  decimals: number;
+}
+export function formatLockedBalance(balance: Balance) {
+  if (!balance || !balance.amount || !balance.decimals) return null;
+
+  const amount = parseFloat(balance.amount) / Math.pow(10, balance.decimals);
```
kyranjamie, line 9: Use BigNumber for math to avoid floating point imprecision

apps/mobile/src/components/widgets/tokens/tokens.mocks.tsx
```diff
@@ -0,0 +1,119 @@
...
+// provided by useTotalBalance hook in extension/src/app/common/hooks/balance/use-total-balance.tsx
+
+export const mockTotalBalance = {
+  totalUsdBalance: '$126.74',
+  totalBtcBalance: '0.00215005',
+  totalStxBalance: '0.0024',
+};
+
+export interface Token {
+  availableBalanceString: string;
```
kyranjamie, line 14: This should be `Money` and later is JSX we call `formatMoney`

packages/ui/src/components/chip/chip.native.tsx
```diff
@@ -0,0 +1,28 @@
...
+  borderRadius: 'xs',
+  paddingLeft: '1',
+  paddingRight: '1',
+  paddingTop: '0', // says 2px in figma but no token for that
+  paddingBottom: '0', // says 2px in figma but no token for that
+  flexDirection: 'column',
+  justifyContent: 'center',
+  alignItems: 'center',
+  gap: '3', // 12px - says 10px in figma but no token for that
+};
```
kyranjamie, line 28: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis

Believe we shouldn't be using `defaultProps`. But we can just add the props to the compoent, and allow them to be overridden instead?

packages/utils/src/abbreviate-number/abbreviate-number.ts
```diff
@@ -0,0 +1,8 @@
+export const abbreviateNumber = (n: number) => {
+  if (n < 1e3) return n.toString();
+  if (n >= 1e3 && n < 1e6) return +(n / 1e3).toFixed(2) + 'K';
+  if (n >= 1e6 && n < 1e9) return +(n / 1e6).toFixed(2) + 'M';
+  if (n >= 1e9 && n < 1e12) return +(n / 1e9).toFixed(2) + 'B';
+  if (n >= 1e12) return +(n / 1e12).toFixed(2) + 'T';
+  return n.toString();
+};
```
kyranjamie, line 8:
```suggestion
export function abbreviateNumber(n: number) {
  if (n < 1e3) return n.toString();
  if (n >= 1e3 && n < 1e6) return +(n / 1e3).toFixed(2) + 'K';
  if (n >= 1e6 && n < 1e9) return +(n / 1e6).toFixed(2) + 'M';
  if (n >= 1e9 && n < 1e12) return +(n / 1e9).toFixed(2) + 'B';
  if (n >= 1e12) return +(n / 1e12).toFixed(2) + 'T';
  return n.toString();
}
```

🙃

kyranjamie (commented):
Great start @pete-watters looks like we should avoid defaultProps though

packages/ui/src/components/chip/chip.native.tsx
pete-watters: OK let me check that. I am doing it this way as it's easer to extend `Chip` to accept more props if you use `createBox<Theme, ChipProps>`.

I did a similar thing in `hstack.native.tsx`. I'm not sure how to do our usual `interface Chip extends BoxProps {`  with the native `Box` component but I can check it out

packages/utils/src/abbreviate-number/abbreviate-number.ts
pete-watters: I'll change, thanks! It's just lifted from the extension. Good old Thomas 

Screenshot 2024-09-17 at 11 16 25

apps/mobile/src/components/home/home.tsx
```diff
@@ -14,7 +15,7 @@ export function Home() {
   return (
     <HomeLayout>
       <AccountWidget accounts={accounts.list} wallets={wallets.list} />
-      <Earn />
+      <TokenWidget tokens={getMockTokens()} totalBalance={mockTotalBalance.totalUsdBalance} />
```
pete-watters, line 18: @kyranjamie - I'm realising I have an erratic naming convention here. 

Should I go with plural or singular?

- `AccountWidget` vs `AccountsWidget`
    - `widgets/account/` vs  ``widgets/account/`
- `TokenWidget` vs `TokensWidget`
    - `widgets/token/token-widget` vs  ``widgets/tokens/tokens-widget` 
    
    etc.

kyranjamie: I think I'd error on side of plural, or `AccountListWidget` if this component reflects data of many

apps/mobile/src/components/widgets/tokens/token-widget.layout.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface TokenWidgetProps {
+  balance?: React.ReactNode;
+  children: React.ReactNode;
+  header?: React.ReactNode;
+  sheetRef?: React.RefObject<SheetRef>;
+}
+
+export function TokenWidgetLayout({ children, header }: TokenWidgetProps) {
+  const theme = useTheme<Theme>();
```
kyranjamie, line 18: Would be great to wrap this somewhere so we can just import the hook without passing the generic

apps/mobile/src/components/widgets/tokens/token-widget.layout.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { ScrollView } from 'react-native-gesture-handler';
+
+import { useTheme } from '@shopify/restyle';
+
+import { Box, SheetRef, Theme } from '@leather.io/ui/native';
+
+import { Widget } from '../widget';
+
+interface TokenWidgetProps {
```
fbwoolf, line 10: `TokenWidgetLayoutProps`

packages/ui/src/components/avatar/btc-avatar-icon.native.tsx
fbwoolf: Seems good, yep. 👍 Thx for expanding on it.

apps/mobile/src/components/widgets/tokens/token-widget.layout.tsx
pete-watters: I opened this issue to do that https://github.com/leather-io/issues/issues/323

packages/ui/src/components/chip/chip.native.tsx
pete-watters: @kyranjamie I just made a [different task](https://github.com/leather-io/issues/issues/324) for refactoring this as I don't want to get deep into it right now. 

I'll change it shortly once I come up with a better way

fbwoolf (approved):
Nice work @pete-watters. 👍

PR #449, opened 2024-09-16
fix: use margin on scrollview to allow elements reach edge, closes le…

pete-watters:
This PR updates the Account widget to use a margin on `ScrollView` instead of padding so that elements can scroll to the edge of the screen:

https://github.com/user-attachments/assets/33831d2c-b004-4372-8fca-0e83fbc5c301

It doesn't seem to respect the margin on the right, but that could be just as the scroll view keeps bouncing back

### DIFF
--- a/apps/mobile/src/components/widgets/account/account-widget.layout.tsx
+++ b/apps/mobile/src/components/widgets/account/account-widget.layout.tsx
@@ -18,14 +18,17 @@ export function AccountWidgetLayout({ balance, children, header }: AccountWidget
   const theme = useTheme<Theme>();
   return (
     <Widget>
-      <Box>
+      <Box marginHorizontal="5">
         {header}
         {balance}
       </Box>
       <ScrollView
         showsHorizontalScrollIndicator={false}
         horizontal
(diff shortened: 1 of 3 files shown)


apps/mobile/src/components/widgets/account/account-widget.layout.tsx
```diff
@@ -18,14 +18,17 @@ export function AccountWidgetLayout({ balance, children, header }: AccountWidget
...
         {header}
         {balance}
       </Box>
       <ScrollView
         showsHorizontalScrollIndicator={false}
         horizontal
-        contentContainerStyle={{ gap: theme.spacing['3'] }}
+        contentContainerStyle={{
+          gap: theme.spacing['3'],
+          marginHorizontal: theme.spacing['5'],
```
pete-watters, line 30: I'm not sure if I should use `marginLeft` only or `marginHorizontal`

kyranjamie (approved):
Thanks @pete-watters

pete-watters:
> Thanks @pete-watters

No problem. Thanks for pointing it out. Some new things to learn and watch out for on mobile but still the same old padding style problems as in 2004

PR #453, opened 2024-09-16
feat: settings conversion unit

fbwoolf:
This PR adds all the new supported currencies to our `FiatCurrency` (changed in this work to singular) type to be used with our market data code. Bc of the type name change, I'll have to put up a PR in the extension to track the change, but I think it is the direction we want to go with singular naming. I also changed a few others, like `Currencies` > `Currency` and `CryptoCurrencies` > `CryptoCurrency`.

At some point our code in the query pkg to get market data will have to be refactored a bit to handle a passed fiat currency, right now it defaults to USD.


packages/models/src/currencies.model.ts
```diff
@@ -1,7 +1,7 @@
 import { LiteralUnion } from './types.utils';
 
-export type CryptoCurrencies = LiteralUnion<'BTC' | 'STX', string>;
+export type CryptoCurrency = LiteralUnion<'BTC' | 'STX', string>;
```
pete-watters, line 3: Maybe we can call this `Token` instead of `CryptoCurrency` if we are changing it anyway?

pete-watters (approved):
Great work, as always 👏

kyranjamie (approved):
Looks good Fara.

It's ugly as it breaks two lines, but also unnecessary I think to say _Chinese Yuan Renminbi_. 

In the same respect we don't say British Pound Stirling here, and that the Y in CNY means Yuan, vote for saying _Chinese Yuan_ only.

packages/models/src/currencies.model.ts
fbwoolf: I'm not sure `Token` is the right choice here considering an NFT is a `token`? The naming here is meant to identify those that are native gas tokens, or `cryptocurrencies`:
https://www.gemini.com/cryptopedia/cryptocurrencies-vs-tokens-difference

pete-watters: OK 👍 I just see us labelling them as `tokens` in the mobile designs now, we have `tokens` + `collectibles`

PR #456, opened 2024-09-17
feat: settings account identifier

fbwoolf:
This PR wires up the settings `account identifier`...

*Renamed `Blockchains` type to `Blockchain`
*Wondering if `Native SegWit` should have `address` after it? 🤔


packages/models/src/types.ts
```diff
@@ -1,15 +1,10 @@
...
-  | 'stx'
-  | 'brc-20'
-  | 'inscription'
-  | 'rune'
-  | 'sip-9'
-  | 'sip-10'
-  | 'src-20'
-  | 'stamp'
-  | 'stx-20';
+export type AccountIdentifier = 'ns' | 'tr' | 'bns' | 'stacks';
```
kyranjamie, line 5: We already have `AccountId` referring to an object that specifies identifies an account.

As this is a purely presentational concern (that we'll deal with much less frequently), maybe a longer more descriptive name would help avoid any confusion. Something like `AccountDisplayIdentifier` or `UserAccountIdPreference`. Ugly name here is okay.

kyranjamie: *in my un merged PR

fbwoolf: Ok, yep, was concerned about that bc at first I actually had `accountId` which def seemed problematic. I'll change.

kyranjamie: `AccountDisplayPreference` might work?

fbwoolf: Changed it to `AccountDisplayPreference` except for I left the component name as `AccountIdentifierSheet` and `AccountIdentifierCell` so it is still coordinated with designs.

kyranjamie (approved):

PR #458, opened 2024-09-18
feat: collectibles widget

pete-watters:
This PR adds the collectibles widget using mock data 

https://github.com/user-attachments/assets/c2c52277-ebf4-4101-af2a-e0a2c6f10d29

### DIFF
--- a/apps/mobile/src/components/widgets/collectibles/collectibles-header.tsx
+++ b/apps/mobile/src/components/widgets/collectibles/collectibles-header.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+
+import { t } from '@lingui/macro';
+
+import { Box, ChevronRightIcon, Chip, SheetRef, Text } from '@leather.io/ui/native';
+
+import { FiatBalance } from '../components/balance/fiat-balance';
+
+interface CollectiblesHeaderProps {
+  collectibleCount: number;
+  sheetRef: React.RefObject<SheetRef>;
+  totalBalance: string;
+}
+
+function CollectiblesHeaderText() {
+  return <Text variant="heading05">{t`My collectibles`}</Text>;
+}
+
+export function CollectiblesHeader({ collectibleCount, totalBalance }: CollectiblesHeaderProps) {
+  const hasCollectibles = collectibleCount > 0;
+  if (!hasCollectibles) return <CollectiblesHeaderText />;
+  return (
+    <Box flexDirection="row" gap="1" alignItems="center" marginHorizontal="5">
+      <CollectiblesHeaderText />
+      <Chip label={collectibleCount} />
+      <ChevronRightIcon variant="small" />
+      <Box flex={1} justifyContent="flex-end" alignItems="flex-end">
+        <FiatBalance balance={totalBalance} color="ink.text-subdued" />
+      </Box>
+    </Box>
+  );
+}
--- a/apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
+++ b/apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
@@ -1,0 +1,14 @@
+import { HasChildren } from '@/utils/types';
+import { BaseTheme, BoxProps } from '@shopify/restyle';
+
+import { Box, Theme } from '@leather.io/ui/native';
+
+type CollectiblesCardLayoutProps<Theme extends BaseTheme> = BoxProps<Theme> & HasChildren;
+
+export function CollectiblesCardLayout({ children, ...props }: CollectiblesCardLayoutProps<Theme>) {
+  return (
+    <Box width={200} height={200} borderRadius="lg" overflow="hidden" {...props}>
+      {children}
+    </Box>
+  );
+}
--- a/apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
+++ b/apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
@@ -1,0 +1,45 @@
+// FIXME: Lifted from extension
+// // TODO: migrate `collectible-type` components to monorepo
+// FIXME when using real data sanitization is needed
+// import { sanitize } from 'dompurify';
+import { useEffect, useState } from 'react';
+
+import { Text } from '@leather.io/ui/native';
+
+import { CollectiblesCardLayout } from './collectible-card-layout';
+
+interface CollectibleTextProps {
+  src: string;
+}
+
+export function CollectibleText({ src }: CollectibleTextProps) {
+  const [content, setContent] = useState<string | null>(null);
+
+  useEffect(() => {
+    const fetchContent = async () => {
+      try {
+        const response = await fetch(src);
+        if (!response.ok) {
+          throw new Error('Network response was not ok');
+        }
+        const data = await response.json();
+        setContent(JSON.stringify(data, null, 2));
+      } catch (err) {
+        // eslint-disable-next-line no-console, lingui/no-unlocalized-strings
+        console.log('Error fetching content:', err);
+      }
+    };
+
+    void fetchContent();
+  }, [src]);
+
+  return (
+    <CollectiblesCardLayout bg="ink.text-primary" p="4">
+      <Text color="ink.background-secondary" textAlign="left">
+        {/* FIXME when using real data sanitization is needed  */}
+        {/* {sanitize(children)} */}
+        {content}
+      </Text>
+    </CollectiblesCardLayout>
+  );
+}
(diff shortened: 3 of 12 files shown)


apps/mobile/src/locales/en/messages.po
```diff
@@ -172,19 +172,19 @@ msgstr "Avatar"
...
+msgstr "BIP39 passphrase"
 
 #: src/components/widgets/tokens/tokens.mocks.tsx:50
 msgid "Bitcoin"
 msgstr "Bitcoin"
 
 #: src/components/widgets/tokens/tokens.mocks.tsx:27
 msgid "Bitcoin blockchain"
-msgstr "Bitcoin blockchain>>>>>>> bbb518c (feat: add tokens widget, ref leather-io/issues#221)"
+msgstr "Bitcoin blockchain"
```
pete-watters, line 183: I keep getting these weird rebase issues in the locales file. It's annoying and takes up time fixing.

I wonder how to avoid?

apps/mobile/src/components/widgets/collectibles/collectibles-header.tsx
```diff
@@ -0,0 +1,32 @@
...
+  collectibleCount: number;
+  sheetRef: React.RefObject<SheetRef>;
+  totalBalance: string;
+}
+
+function CollectiblesHeaderText() {
+  return <Text variant="heading05">{t`My collectibles`}</Text>;
+}
+
+export function CollectiblesHeader({ collectibleCount, totalBalance }: CollectiblesHeaderProps) {
```
pete-watters, line 19: All the widgets have this same header so I will refactor this to keep things DRY in the next PR

apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
```diff
@@ -0,0 +1,13 @@
+import { Box, BoxProps } from '@leather.io/ui/native';
+
+interface CollectiblesCardLayoutProps extends BoxProps {
```
pete-watters, line 3: @kyranjamie : I'm no sure how to extend the shopify `Box`. 
```
export type { BoxProps } from '@shopify/restyle';
```

Doing this works (applying the props) but gives a type error `Generic type 
 requires between 1 and 2 type arguments.`


That's why I have been using `defaultProps` so I can more easily extend `Box`. 

See [issue here](https://github.com/leather-io/issues/issues/324). 

I will try and figure out how to do it without `defaultProps`. 

Maybe you have an idea on how to fix it?

pete-watters:
Opening this for review but it has a known [type error](https://github.com/leather-io/mono/pull/458#discussion_r1765263378) I need to solve

apps/mobile/src/locales/en/messages.po
fbwoolf: It is annoying, I just delete the files and run `pnpm lingui` in mobile root to regenerate the files.

apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
fbwoolf: Maybe look in our UI lib `touchable-opacity.native.tsx`, I think @edgarkhanzadian did something with shopify props there that might be helpful?

apps/mobile/src/locales/en/messages.po
pete-watters: I haven't been deleting them, I'll do that in future and look into adding a shortcut to do that. Thanks!

apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
pete-watters: Thanks, I'll check it out. 50% of my work is `I love coding` and 50% is `I really hate typescript`

kyranjamie: [The deprecation docs](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops) have examples of this

image

Default props are a runtime concern, so not sure what the types have to do with this?

```
function CollectiblesCardLayout({ 
  children, 
  somePropThatNeedsADefault = false, 
  ...props 
}: CollectiblesCardLayoutProps) {ts
```

pete-watters: I was setting default props as a way to extend the shopify theme components. I will try what Fara suggested in `touchable-opacity.native.tsx` as it seems like that's what I need

pete-watters: That worked, thanks @fbwoolf 👏

pete-watters:
> Opening this for review but it has a known [type error](https://github.com/leather-io/mono/pull/458#discussion_r1765263378) I need to solve

Fixed

apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
```diff
@@ -0,0 +1,45 @@
+// FIXME: Lifted from extension
```
pete-watters, line 1: Should I move this code into the UI library alongside the extension code for displaying collectibles? 

I didn't setup `sanitize` from `'dompurify'` here yet as I was thinking I could install that once in the UI library instead and then load these components from there

apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
```diff
@@ -0,0 +1,14 @@
+import { HasChildren } from '@/utils/types';
```
pete-watters, line 1: Do we need a shared `types` package? 

We have this type here and also in the `extension`

apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
fbwoolf: I think it makes to colocate code if the components are the same in both.

apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
fbwoolf: MIght make sense if we have quite a few that are the same, which I bet is the case ...especially once we start to refactor the extension after mobile.

apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
pete-watters: Thanks. They are almost the same apart from some nuances ( e.g. `<Image` vs `<image`) 

Do you know where I could find ordinal collectibles of type `image` / `audio` ? No worries if not, I can ask support or try and source some.

When I test with the dev wallet all I could find was `html` / `text` which I have handled here. My code for `CollectibleImage` should handle image ordinals but I haven't tested it.

pete-watters:
> I wonder what's the reason for it not going all the way to the right:
> 
>  collectibles.mov 
> And i think we should also add "5" padding to the inside of the scrollview so that it doesn't stick to the end of the phone. You can do it with `contentContainerStyles`

Thanks, let me check. I face the same issue with the scrollview on the `Accounts` widget. I tried adding a right margin but it didn't work. I will play with it and see if I can figure it out

I had padding but no margin and it wasn't going the full width of the screen: https://github.com/leather-io/issues/issues/322 

The design has padding of `5` 
Screenshot 2024-09-19 at 15 00 47

apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
fbwoolf: Good question, I may have just hardcoded it to test ...think we've changed apis since then. Can you hardcode or buy from bestinslot.xyz directly?

pete-watters:
I'm going to close this for now and spend more time investigating https://github.com/leather-io/mono/pull/458#issuecomment-2361030784 and also moving the extension `Collectible` components to the monorepo

apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
pete-watters: Sure, I will figure it out and test it properly. Thanks!

apps/mobile/src/components/widgets/collectibles/components/collectible-card-layout.tsx
pete-watters: I did this in a new PR https://github.com/leather-io/mono/pull/481 as it wouldn't let me reopen this

apps/mobile/src/components/widgets/collectibles/components/collectible-text.tsx
pete-watters: I found samples of most inscription types apart from audio + video and added mocks. 

I can inscribe audio + video on ordinals bot if we think that's useful. 

For video the network fee is high so it's ~ $392 for what I tested. Audio can be cheaper ~ $122.02

Issue #34, opened 2023-12-05
Add code coverage tool


kyranjamie:
In order to keep code coverage high, and remind us to write tests, I suggest we add a code coverage tool to the monorepo.

I haven't done much research, but [Codecov](https://about.codecov.io/) is a tool acquired by Sentry that is probably a good bet. Here's a guide I found to set it up https://www.curtiscode.dev/post/tools/codecov-monorepo/.

#14 not necessary if we take on this issue

PR #459, opened 2024-09-18
ci: add codecov integration to Github test action

camerow:
## Summary in brief

Integrate with codecov to get pretty code coverage metrics in our mono repo. This is just the initial integration commit to test the functionality. I am following their quick start docs and not adding any frills or extras, just to test functionality. 

### Overview of changes

1. Add `codecov` integration to repository, add `test-coverage` Github action
2. Use `v8` reporter for code coverage, remove references to `istanbul` coverage reporter
3. Break coverage reporting into components so we can see coverages changes per `package` via `codecov.yml`
4. Use `pnpm` to recursively test all of the `packages` directory. '
5. Currently this is being merged as `informational` - as a team we should discuss and commit to thresholds

**Note**: `mobile` was failing because of `@/` import path aliases. I believe this must be fixed in the `vitest.config.ts` file but not quite sure how to do so yet. For now I simply didn't include a `pnpm test:coverage` command in there. 

### Future Enhancements

[Specify levels of coverage per package type](https://docs.codecov.com/docs/github-5b-setting-coverage-standards-with-flags). For example `apps/*` should probably have a different level of acceptable coverage than most things in `packages/*`.


fbwoolf (approved):
Typically we do squash into one commit before merging.

.github/workflows/code-checks.yml
```diff
@@ -85,5 +85,24 @@ jobs:
 
       - run: pnpm build
 
-      - name: Run unit tests
+      - name: Run unit tests with code coverage
```
kyranjamie, line 88: I wonder if we still need `test-unit` as here I think we're running the tests twice?

kyranjamie (commented):
+1 to @fbwoolf's comment. As this is a small code change, it's more helpful when looking back at git history to see a single commit, and also renders nicer change logs.

Also, we can reference the issue in that commit e.g. `ci: add code coverage, closes #34`

pete-watters:
> +1 to @fbwoolf's comment. As this is a small code change, it's more helpful when looking back at git history to see a single commit, and also renders nicer change logs.
> 
> Also, we can reference the issue in that commit e.g. `ci: add code coverage, closes #34`

Nice work Will! To add to what Kyran said, when working an an issue from a different repo, you need to specify the repo to get the issue to link correctly. 

For example this is an issue from `leather-io/issues` but the repo is `leather-io/mono` so you need to use:
```
git commit -a -m "fix: some message, closes leather-io/issues#324"
```

`ci:` is also perfect to use in this case but FYI when working on packages that you want to be added to the release please auto release you need to specify `fix:` / `feat:` for it to deem it worthy of creating a release

.github/workflows/code-checks.yml
camerow: It's a good question, likely we don't, though I split them for the sake of the initial testing (at the very least) for a bit of sanity. I should probably remove the `test-coverage` command  and just stick with `unit` for the ci

PR #460, opened 2024-09-19
fix: deprecate usage of `defaultProps` and extend BoxProps

pete-watters:
This PR addresses an issue with how I was using `defaultProps` in some of our UI components

### DIFF
--- a/packages/ui/src/components/avatar/avatar.native.tsx
+++ b/packages/ui/src/components/avatar/avatar.native.tsx
@@ -1,11 +1,16 @@
-import { createBox } from '@shopify/restyle';
+import { type BoxProps } from '@shopify/restyle';
 
 import { type Theme } from '../../theme-native';
+import { Box } from '../box/box.native';
 
-export const Avatar = createBox<Theme>();
+interface AvatarProps extends BoxProps<Theme> {
+  children: React.ReactNode;
+}
 
-Avatar.defaultProps = {
-  bg: 'ink.background-secondary',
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/avatar/avatar.native.tsx
```diff
@@ -1,11 +1,16 @@
-import { createBox } from '@shopify/restyle';
+import { type BoxProps } from '@shopify/restyle';
 
 import { type Theme } from '../../theme-native';
+import { Box } from '../box/box.native';
 
-export const Avatar = createBox<Theme>();
+interface AvatarProps extends BoxProps<Theme> {
```
pete-watters, line 6: I could use `HasChildren` here but I don't want to create it a third time in `packages/ui`. 

Should we have a `types` package? Or where should we keep shared `types`?

kyranjamie (approved):

packages/ui/src/components/avatar/avatar.native.tsx
kyranjamie: I just added it to `ui` in my approver work. Figure it's okay for UI package to have some helpers like `useOnMount` and type ones like HasChildren, [in the same vein Chakra has hook helpers](https://v2.chakra-ui.com/docs/hooks/use-boolean)

pete-watters: OK, thanks!

PR #462, opened 2024-09-19
feat: settings security

fbwoolf:
This PR wires up the security settings flows, and:

- Separates settings read/write api to be exposed thru `useSettings` in `settings.ts`
- Creates a UI library component for `Switch` (note, I didn't replace the one in Cell bc it will be removed)

  
 

cc @mica000 We will need final text for `Description` here on both sheets. Also, is `App authorization` the right naming here? I'm not sure I even understood what it meant and the toast says `metrics`?


pete-watters (approved):
Nice work 🚀

apps/mobile/src/store/settings/settings.ts
```diff
@@ -0,0 +1,128 @@
...
+  userChangedNetwork,
+  userChangedTheme,
+  userChangedWalletSecurityLevel,
+  userToggledAnalyticsEnabled,
+  userToggledAppAuthEnabled,
+} from './settings.write';
+
+export const defaultNetworks = ['mainnet', 'testnet', 'signet'] as const;
+
+export const defaultThemes = ['light', 'dark', 'system'] as const;
```
kyranjamie, line 38: Been thinking on this naming, because `system` isn't really a theme, it's a setting that points to a theme.

Maybe we have `ThemeSetting`/`ThemePreference` and `Theme`?

apps/mobile/src/store/settings/settings.ts
```diff
@@ -0,0 +1,128 @@
...
+  appAuthEnabled: false,
+  bitcoinUnit: 'bitcoin',
+  conversionUnit: 'USD',
+  createdOn: new Date().toISOString(),
+  network: WalletDefaultNetworkConfigurationIds.mainnet,
+  theme: 'system',
+  walletSecurityLevel: 'undefined',
+};
+
+export function useSettings() {
```
kyranjamie, line 68: Thanks for the refactor. Think this is a really nice pattern where we expose all store read/writes via a uniform hook, abstracting any store-specific implementation details.

Will follow up with a depcruiser rule to prevent components reading `.{read,write}.ts` files

apps/mobile/src/store/settings/settings.ts
```diff
@@ -0,0 +1,128 @@
...
+    theme,
+    themeStore,
+    walletSecurityLevel,
+    whenTheme: whenTheme(theme),
+    changeAccountDisplayPreference(type: AccountDisplayPreference) {
+      dispatch(userChangedAccountDisplayPreference(type));
+    },
+    toggleAnalyticsEnabled() {
+      dispatch(userToggledAnalyticsEnabled());
+    },
```
kyranjamie, line 97: Not big on this name, kinda sounds like we're enabling it, but also toggling it which doesn't make sense. We call the store item `analyicsEnabled` but really what's happening here is the user is changing the setting. So this is almost exposing a store detail—only the name, but still something the rest of the app doesn't care about. `toggleAnalyticsPreference`?

packages/ui/src/components/switch/switch.native.tsx
```diff
@@ -0,0 +1,28 @@
+import { Switch as RNSwitch, SwitchProps as RNSwitchProps } from 'react-native';
+
+import { useTheme } from '@shopify/restyle';
+import { Theme } from 'from ../../../native';
+
+interface SwitchProps extends RNSwitchProps {
+  onToggleValue?(): void;
+  value: boolean;
+}
+export function Switch({ onToggleValue, value, ...props }: SwitchProps) {
```
kyranjamie, line 10: Nice, think Pete made an issue for this we can close

apps/mobile/src/store/settings/settings.ts
```diff
@@ -0,0 +1,128 @@
...
+  createdOn: string;
+  network: DefaultNetworkConfigurations;
+  theme: ThemeStore;
+  walletSecurityLevel: WalletSecurityLevel;
+}
+
+export const initialState: SettingsState = {
+  accountDisplayPreference: 'ns',
+  analyticsEnabled: false,
+  appAuthEnabled: false,
```
kyranjamie, line 59: Can't this be inferred from `walletSecurityLevel`?

kyranjamie (commented):
Great work @fbwoolf 

Lots of comments, pretty much all naming related

apps/mobile/src/app/wallet/settings/security/security-call.tsx
```diff
@@ -0,0 +1,29 @@
+import { ReactNode } from 'react';
```
pete-watters, line 1: Just realising now that there is a typo in the filename - `security-call`

apps/mobile/src/store/settings/settings.ts
fbwoolf: Agree, been struggling a bit with some of the naming so thx for the feedback on all comments. I'll refactor them here.

apps/mobile/src/store/settings/settings.ts
fbwoolf: Agree, was struggling with it and tried a few things. Will change.

apps/mobile/src/store/settings/settings.ts
fbwoolf: I need clarification on what this setting is for the from design team. I am unclear what it is, the toast says `Metrics updated`. 🤔 cc @mica000 Can someone clarify for me what `walletSecurityLevel` is for?

kyranjamie: cc/ @edgarkhanzadian

fbwoolf: It does appear this is used with biometrics true/false, so I'll remove `appAuthEnabled` bc it seems the same.

mica000: Just registering that the copy should be App authentication updated.

fbwoolf:
@kyranjamie thoughts renaming here, I went in a direction where I coordinated all settings to be `...Preference` so when encountered we know it is associated to a preferred setting chosen by the user.

fbwoolf:
@markmhendrickson can you confirm, is the app authentication turning on/off using biometrics? I am still a little unclear if I am interpreting this correctly?

fbwoolf:
> Indeed, app authentication == whether the user needs to use device-level authentication (e.g. biometrics like Face ID or PIN as backup) to perform certain actions. If they do, they'll see the unlock view each time: [leather-io/issues#160](https://github.com/leather-io/issues/issues/160)
> 
> [These](https://www.notion.so/trustmachines/App-authentication-triggers-cb17c03938f544f993cb1d300b5bb000?pvs=4#d076c7b4017c44358405cb008dddd846) are the intended triggers / actions that require unlock if enabled.

Got it, thanks for clarifying. 👍

fbwoolf:
@edgarkhanzadian where is the code that triggers the face id to popup? I see `biometrics` being passed around, but I'm not understanding what I can use to trigger it if the user toggles app auth on/off?

apps/mobile/src/features/settings/app-authentication-sheet.tsx
```diff
@@ -1,19 +1,41 @@
...
 }
 export function AppAuthenticationSheet({ sheetRef }: AppAuthenticationSheetProps) {
-  const { theme: themeVariant } = useSettings();
+  const settings = useSettings();
+  const { displayToast } = useToastContext();
+
+  useEffect(() => {
+    if (settings.securityLevelPreference === 'undefined')
+      settings.changeSecurityLevelPreference('insecure');
+  });
```
kyranjamie, line 21: Mmm, this seems weird? Why wouldn't we just set the preference as `insecure` in the store if it it changed automatically when landing on this page?

fbwoolf: Yeah, I'll remove. Prob not needed but was asking for clarification on this setting from Edgar. I'm not sure why undefined is even needed in the type?

kyranjamie (approved):

PR #471, opened 2024-09-24
feat: eas update

edgarkhanzadian:


.github/actions/eas-deploy/action.yml
```diff
@@ -47,65 +52,74 @@ runs:
...
+    #     script: |
+    #       const { BUILD_LINK } = process.env
+    #       if(BUILD_LINK) {
+    #         github.rest.issues.createComment({
+    #           issue_number: context.issue.number,
+    #           owner: context.repo.owner,
+    #           repo: context.repo.repo,
+    #           body: "Expo simulator build link: " + BUILD_LINK
+    #         })
+    #       }
```
kyranjamie, line 115: Rather than a new comment, can we update the PR body like we do with extension builds?

https://github.com/leather-io/extension/blob/32b064f7e4f0fef0bbc588ae09280fdeffb42ab6/.github/workflows/build-extension.yml#L20-L29

PR #472, opened 2024-09-24
Create a simple event tracking package

camerow:
Update Oct 18, 2024

I totally refactored this and opted to keep the existing event names and naming conventions, as well as parameters. This will allow for a seamless first pass at adding this to `extension`, essentially will be a 1:1 replacement with some nice type enforcement. 

Instead of adding methods to a class per event I also moved to the recommendation of using a type and enforcing what can be passed based on the type. 

Also, tests!

### Overview

- ~Created two entry points for analytics package: one for web using analytics browser client from segment, and one for native using segment client from react native.~
- Injected into a generic analytics client with configurable options.
- Planning to make generic functions private and add specific methods for tracking events.
- ~Proposing to use uppercase or human-readable event names for better readability.~
- Updating functions to be more generic and customizable via options.

### Assumptions

- Making generic functions private will not impact current usage in apps.
- ~Uppercase event names will improve readability and consumption.~
- Adding customizable options for functions will enhance flexibility.

### Future steps

I decided to just create the package and not include it anywhere for now. This is just to scope out the work such that we can identify more clearly when we added this to the respective applications, update stakeholders on name changes, etc.

### DIFF
--- a/docs/extension-development.md
+++ b/docs/extension-development.md
@@ -2,17 +2,17 @@
 
 The extension will heavily rely on monorepo packages. Soon, we will need to apply changes to these packages as part of the work on the extension repository, thus necessitating a way to test changes made to the monorepo within the extension.
 
-### Hard Linking
+## Hard Linking
 
-#### Initialization
+### Initialization
 
 One method to achieve this is through hard linking. Here is a step-by-step guide to help you set it up correctly:
 
 1. Clone the repositories:
 
-   ```
-   $ git clone https://github.com/leather-io/mono
-   $ git clone https://github.com/leather-io/extension
+   ```git
+   git clone https://github.com/leather-io/mono
+   git clone https://github.com/leather-io/extension
    ```
 
    For simplicity, do this in the same folder so that `mono` and `extension` are sibling directories.
@@ -57,7 +57,7 @@ That's it; the setup should now be fully functional.
 
 Now, when we update a package in the `mono` repo, it doesn't automatically move the changes to the `extension`'s `node_modules`. To do this, run the following command in the extension each time you update a package in the monorepo:
 
-```
+```sh
 pnpm i
 ```
 
--- a/packages/analytics/README.md
+++ b/packages/analytics/README.md
@@ -1,0 +1,55 @@
+# @leather/analytics
+
+This package provides a client for sending analytics events to various analytics services. Currently, it supports Segment but could be extended to support other services in the future.
+
+## Installation
+
+```bash
+pnpm install @leather/analytics
+```
+
+## Usage
+
+Before making any analytics calls, you must configure the analytics client with your write key.
+
+```ts
+import { configureAnalyticsClient } from '@leather.io/analytics';
+
+configureAnalyticsClient({
+  writeKey: 'YOUR_WRITE_KEY',
+  defaultProperties: {
+    platform: 'web',
+  },
+});
+```
+
+Now you can make analytics calls.
+
+```ts
+import { analyticsClient } from '@leather/analytics';
+
+analyticsClient.track('My Event', {
+  property: 'value',
+});
+```
+
+You can also inject your own analytics client if you are using a custom analytics service or want to use a stubbed client for testing.
+
+````ts
+import { configureAnalyticsClient } from '@leather/analytics';
+
+configureAnalyticsClient({
+  client: myAnalyticsClient,
+});
+
+## Development
+
+```bash
+pnpm build
+````
+
+or
+
+```bash
+pnpm build:watch
+```
--- a/packages/analytics/src/client.ts
+++ b/packages/analytics/src/client.ts
@@ -1,0 +1,38 @@
+import {
+  AnalyticsClientConfig,
+  EventProperties,
+  ExternalAnalyticsClientInterface,
+  Json,
+} from './types';
+
+export type AnalyticsClientOptions = Pick<
+  AnalyticsClientConfig,
+  'defaultProperties' | 'defaultTraits'
+>;
+
+export function AnalyticsClient<T extends ExternalAnalyticsClientInterface>(
+  client: T,
+  options: AnalyticsClientOptions
+) {
+  return {
+    async track<K extends keyof EventProperties>(event: K, properties: EventProperties[K]) {
+      return client.track(event, { ...properties, ...options.defaultProperties });
+    },
+
+    async untypedTrack(event: string, properties?: Json) {
+      return client.track(event as any, { ...properties, ...options.defaultProperties });
+    },
+
+    async screen(name: string, properties?: Json) {
+      return client.screen(name, { ...properties, ...options.defaultProperties });
+    },
+
+    async group(groupId: string, traits?: Json) {
+      return client.group(groupId, { ...traits, ...options.defaultTraits });
+    },
+
+    async identify(userId: string, traits?: Json) {
+      return client.identify(userId, { ...traits, ...options.defaultTraits });
+    },
+  };
+}
--- a/packages/analytics/src/client.ts
+++ b/packages/analytics/src/client.ts
@@ -20,6 +20,9 @@ export function AnalyticsClient<T extends ExternalAnalyticsClientInterface>(
     },
 
     async untypedTrack(event: string, properties?: Json) {
+      if (event.match(/^[a-zA-Z0-9\s][a-zA-Z0-9\s]*$/)) {
+        throw new Error('Event must be snake_case');
+      }
       return client.track(event as any, { ...properties, ...options.defaultProperties });
     },
 
--- a/packages/analytics/src/types.ts
+++ b/packages/analytics/src/types.ts
@@ -12,7 +12,7 @@ export interface ExternalAnalyticsClientInterface {
 export interface AnalyticsClientConfig {
   writeKey: string;
   client: ExternalAnalyticsClientInterface;
-  defaultProperties: DefaultProperties;
+  defaultProperties?: DefaultProperties;
   defaultTraits?: Record<string, Json>;
 }
 
--- a/packages/analytics/src/types.ts
+++ b/packages/analytics/src/types.ts
@@ -1,0 +1,70 @@
+export interface DefaultProperties {
+  platform: 'web' | 'extension' | 'mobile';
+}
+
+export interface ExternalAnalyticsClientInterface {
+  track: (event: string, properties?: Json) => Promise<void>;
+  screen: (name: string, properties?: Json) => Promise<void>;
+  group: (groupId: string, traits?: Json) => Promise<void>;
+  identify: (userId: string, traits?: Json) => Promise<void>;
+}
+
+export interface AnalyticsClientConfig {
+  writeKey: string;
+  client: ExternalAnalyticsClientInterface;
+  defaultProperties: DefaultProperties;
+  defaultTraits?: Record<string, Json>;
+}
+
+export type Json =
+  | Json[]
+  | {
+      [index: number]: Json;
+    }
+  | {
+      [key: string]: Json;
+    };
+
+export type OptionalIfUndefined<T> = T extends undefined ? undefined : T;
+
+export interface EventProperties {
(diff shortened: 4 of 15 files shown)


fbwoolf:
Is there an issue to link here?

camerow:
> Is there an issue to link here?

There was but I'm struggling to find it. I think I left it in "draft" and possibly lost it.

camerow:
How does the team feel about reusing issues for multiple PRs? I am re-creating this issue, and wondering do we normally create an issue per task or can one issue be represented in multiple PRs?

fbwoolf:
> ~How does the team feel about reusing issues for multiple PRs? I am re-creating this issue, and wondering do we normally create an issue per task or can one issue be represented in multiple PRs?~
> 
> EDIT: Feel like I answered my own question looking at how these things are linked. One issue per PR seems to be the only real logical choice.

You can link multiple issues to one PR if needed ...sometimes a PR will close multiple open issues.

fbwoolf:
@camerow are you using GH notifications? They can be super helpful. I keep a tab pinned with my notifications always open.

packages/analytics/src/client/client.ts
```diff
@@ -0,0 +1,121 @@
...
+  public async trackNativeSegwitTxHexToLedgerTx(properties: { success: boolean }) {
+    return this.track('Native Segwit Transaction Hex to Ledger', properties);
+  }
+  // TODO: replace psbt_sign_request_p2tr_missing_taproot_internal_key with this method call.
+  public async trackPsbtSignRequest(properties: { status: 'missing_taproot_internal_key' }) {
+    return this.track('PSBT Sign Request', properties);
+  }
+  // TODO: replace ledger_nativesegwit_add_nonwitnessutxo with this method call.
+  // TODO: replace ledger_nativesegwit_skip_add_nonwitnessutxo with this method call.
+  public async trackLedgerNativesegwitAddNonwitnessUtxo(properties: {
```
fbwoolf, line 112:
```suggestion
  public async trackLedgerNativeSegwitAddNonWitnessUtxo(properties: {
```

packages/analytics/src/client/client.ts
```diff
@@ -0,0 +1,121 @@
...
+  private errorToJsonMap(error: Error) {
+    return {
+      errorMessage: error.message,
+      errorName: error.name,
+    };
+  }
+
+  // TODO: replace view_transaction_confirmation with this method call.
+  public async trackViewTransactionConfirmation(properties: { symbol: string }) {
+    return this.track('Transaction Confirmation Viewed', properties);
```
fbwoolf, line 42: Not sure about the naming pattern change here? @markmhendrickson is this change ok? I know we coordinated the `snake_case` for naming quite awhile ago.

packages/analytics/src/client/client.ts
```diff
@@ -0,0 +1,121 @@
+import { AnalyticsClientConfig, GenericAnalyticsClient, JsonMap } from '../utils/types.shared';
+
+export type AnalyticsClientOptions = Pick<
+  AnalyticsClientConfig,
+  'defaultProperties' | 'defaultTraits'
+>;
+
+export class AnalyticsClient<T extends GenericAnalyticsClient> {
```
fbwoolf, line 8: Why the choice here to use a class over a function? We just refactored our api clients away from using classes.

docs/extension-development.md
```diff
@@ -57,7 +57,7 @@ That's it; the setup should now be fully functional.
 
 Now, when we update a package in the `mono` repo, it doesn't automatically move the changes to the `extension`'s `node_modules`. To do this, run the following command in the extension each time you update a package in the monorepo:
```
pete-watters, line 58: As an FYI, we thought about moving the `extension` into the `mono` repo at some stage but it has a more complicated CI release setup. 

It could be good to eventually migrate it over there if we can maintain the release process

pete-watters (approved):
Nice work @camerow 👍  

This code LGTM so good to merge, especially as the package is not yet consumed. 

I like the readability of the uppercase events better but I'm not sure if we need `snake_case` for some reason

packages/analytics/README.md
```diff
@@ -0,0 +1,55 @@
...
+```bash
+pnpm install @leather/analytics
+```
+
+## Usage
+
+Before making any analytics calls, you must configure the analytics client with your write key.
+
+```ts
+import { configureAnalyticsClient } from '@leather/analytics';
```
kyranjamie, line 16:
```suggestion
import { configureAnalyticsClient } from '@leather.io/analytics';
```

packages/analytics/src/index.native.ts
```diff
@@ -0,0 +1,29 @@
...
+    createClient({
+      writeKey,
+    });
+
+  analyticsClient = new AnalyticsClient(nativeClient, { defaultProperties });
+
+  return analyticsClient;
+};
+
+export { analyticsClient, configureAnalyticsClient };
```
kyranjamie, line 29:
```suggestion
export function configureAnalyticsClient ({
  writeKey,
  defaultProperties,
  client,
}: AnalyticsClientConfig) {
  const nativeClient =
    client ??
    createClient({
      writeKey,
    });

  analyticsClient = new AnalyticsClient(nativeClient, { defaultProperties });

  return analyticsClient;
}
```

Convention used elsewhere is to use [function declarations](https://www.notion.so/trustmachines/Function-declarations-22f7055df3124d868a242adb04f0f60d?pvs=4)

packages/analytics/src/utils/types.shared.ts
```diff
@@ -0,0 +1,31 @@
...
+export type GenericAnalyticsClient = SegmentClient | AnalyticsBrowser | AnalyticsClientInterface;
+
+export type JsonValue = boolean | number | string | null | JsonList | JsonMap | undefined;
+
+export interface JsonMap {
+  [key: string]: JsonValue;
+  [index: number]: JsonValue;
+}
+
+export type JsonList = JsonValue[];
```
kyranjamie, line 31: Seems like a candidate for utils pkg

kyranjamie (commented):
Great start @camerow

packages/analytics/src/client/client.ts
markmhendrickson: I'm cool with updating the naming pattern. Let's just be sure to alias old names to the new ones if we update them, so we can continue reporting consistently.

kyranjamie: Mentioned on slack, but question why we'd make this change? What's so superior about this naming scheme that warrants the effort of doing the aliasing? [Per Mixpanel docs](https://docs.mixpanel.com/docs/data-structure/events-and-properties#name-events-and-properties-consistently:~:text=Generally%2C%20adopting%20snake_case%20for%20your%20event%20and%20property%20names%20tend%20to%20be%20more%20robust%2C%20especially%20if%20you%20plan%20to%20export%20your%20Mixpanel%20data%20to%20downstream%20processes%20such%20as%20data%20warehouses.)

> Generally, adopting snake_case for your event and property names tend to be more robust, especially if you plan to export your Mixpanel data to downstream processes such as data warehouses.

packages/analytics/src/client/client.ts
camerow: I would like to store a reference to `options` and creating a class seemed to be the best way. WHat is the disadvantage of a class over a function?

docs/extension-development.md
camerow: Yeah, @kyranjamie and I discussed this - one advantage he brought up having it external is having to consume our packages as truly external which potentially increases their quality, though I see plenty of advantages to bringing `extension` into `mono`

packages/analytics/src/client/client.ts
kyranjamie: Nothing inherently wrong using a `class` here, and think it's okay for a package like this, but I think Fara's mentioning because we typically favour a more functional approach. Classes often lead to inheritance over composition, mutative methods over pure functions etc. Use of classes in Javascript is a hotly debated topic.

fbwoolf: Yep, was just questioning based on Kyran's feedback here. Some suggest to avoid classes in JS if possible, so was just curious about your choice.

camerow: I have historically written in codebases that prefer one or the other. No strong preference but depending on the interface I like classes more. Happy to conform my code to our preferences.

camerow:
@edgarkhanzadian / @kyranjamie ok, totally revamped this and I think it's much better for it. 

Not going for replacing any events or implementing this in any of the apps with this PR, just looking to get the foundations merged in.

packages/analytics/src/types.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export type JsonValue = boolean | number | string | null | JsonList | JsonMap | undefined | unknown;
+
+export interface JsonMap {
+  [key: string]: JsonValue;
+  [index: number]: JsonValue;
+}
+
+export type JsonList = JsonValue[];
```
kyranjamie, line 25: If we need to type JSON, you can use a recursive type 

```
type Json = string | number | boolean | null | Json[] | { [key: string]: Json };
```

packages/analytics/src/client.ts
```diff
@@ -0,0 +1,36 @@
...
+  'defaultProperties' | 'defaultTraits'
+>;
+
+export class AnalyticsClient<T extends AnalyticsClientInterface> {
+  constructor(
+    private readonly client: T,
+    private readonly options: AnalyticsClientOptions
+  ) {
+    this.client = client;
+    this.options = options;
```
kyranjamie, line 14: private keyword of the class already defines the parameter to the class, so these lines are redundant afaik?
```suggestion
```

packages/analytics/src/client.ts
```diff
@@ -0,0 +1,36 @@
+import { AnalyticsClientConfig, AnalyticsClientInterface, EventProperties, JsonMap } from './types';
+
+export type AnalyticsClientOptions = Pick<
+  AnalyticsClientConfig,
+  'defaultProperties' | 'defaultTraits'
+>;
+
+export class AnalyticsClient<T extends AnalyticsClientInterface> {
```
kyranjamie, line 8: Nothing inherently wrong using a class here, but it would more inline with other packages we have to use a more functional style

```ts
export function createAnalyticsClient<T extends AnalyticsClientInterface>(
  client: T,
  options: AnalyticsClientOptions
) {
  return {
    async track<K extends keyof EventProperties>(event: K, properties: EventProperties[K]) {
      return client.track(event, { ...properties, ...options.defaultProperties });
    },

    async untypedTrack(event: string, properties?: JsonMap) {
      return client.track(event, { ...properties, ...options.defaultProperties });
    },

    async screen(name: string, properties?: JsonMap) {
      return client.screen(name, { ...properties, ...options.defaultProperties });
    },

    async group(groupId: string, traits?: JsonMap) {
      return client.group(groupId, { ...traits, ...options.defaultTraits });
    },

    async identify(userId: string, traits?: JsonMap) {
      return client.identify(userId, { ...traits, ...options.defaultTraits });
    }
  };
}
```

kyranjamie (approved):

packages/analytics/src/client.ts
camerow: Right @fbwoolf did mention that - I will update.

packages/analytics/src/client.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.options = options;
+  }
+
+  public async track<K extends keyof EventProperties>(event: K, properties: EventProperties[K]) {
+    return this.client.track(event, { ...properties, ...this.options.defaultProperties });
+  }
+
+  public async untypedTrack(event: string, properties?: JsonMap) {
+    return this.client.track(event, { ...properties, ...this.options.defaultProperties });
+  }
```
kyranjamie, line 23: I wonder if this function could enforce_the_naming_convention? Throwing an error if another format is added?

camerow: Good call, done and added tests for all the behaviors we expect with the client.

PR #473, opened 2024-09-25
fix(mobile): handle missing wallets case

kyranjamie:
Closes leather-io/issues#338

`ApproverSheet` is always rendered regardless of whether it's open (separately, think this is a performance issue to have loads of views rendered, but hidden even if not needed).

Here, I use `WalletLoader` to ensure at least one wallet exists. Open to a conversation about how we can best handle reading of wallet state. This has been a recurring problem in the extension. The Loader pattern is the best I've come up with so far, but hopefully there are more friendly alternatives (that aren't hooks)


apps/mobile/src/app/(home)/developer-console/index.tsx
```diff
@@ -119,12 +119,16 @@ export default function DeveloperConsoleScreen() {
         <PressableListItem title={t`Drawer`} />
         <PressableListItem title={t`Page`} />
       </ScrollView>
-      <ApproverSheet
-        fingerprint={wallets.list[0] ? wallets.list[0].fingerprint : ''}
-        accountIndex={0}
-        message={getAddressesMessage}
-        sendResult={() => setGetAddressesMessage(null)}
-      />
+      <WalletLoader fingerprint={wallets.list[0]?.fingerprint ?? ''}>
```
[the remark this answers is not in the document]
kyranjamie: I made a conditional, to avoid the empty  string

apps/mobile/src/app/(home)/developer-console/index.tsx
```diff
@@ -119,12 +119,18 @@ export default function DeveloperConsoleScreen() {
         <PressableListItem title={t`Drawer`} />
         <PressableListItem title={t`Page`} />
       </ScrollView>
-      <ApproverSheet
-        fingerprint={wallets.list[0] ? wallets.list[0].fingerprint : ''}
-        accountIndex={0}
-        message={getAddressesMessage}
-        sendResult={() => setGetAddressesMessage(null)}
-      />
+      {wallets.list[0]?.fingerprint && (
```
fbwoolf, line 122: Was curious about our wallet state yest working on the app security level, should we expose something from the wallet read state that is the `wallets.list[0]`, I've seen it repeated in multiple places.

kyranjamie: I don't think we should really ever be doing `wallets.list[0]`. Only reason we are here is for testing purposes.

fbwoolf (approved):
I know in the extension we use `gates` for routes, could we start one here for routes that need the `WalletLoader`?

apps/mobile/src/app/(home)/developer-console/index.tsx
fbwoolf: Got it, thanks for clarifying, I missed this was just in the dev console. 👍

PR #479, opened 2024-09-25
feat: mpc and hardware wallets UI

edgarkhanzadian:
- Update notify modal UI
- Create hardware wallet list
- Create MPC wallet list

https://github.com/user-attachments/assets/12accea3-b578-4ea9-a3dd-1b1c1a7ac6e2


apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
```diff
@@ -87,6 +63,42 @@ export function AddWalletSheetLayout({
     marginBottom: interpolate(animatedIndex.value, [-1, 0], [200, 0], Extrapolation.CLAMP),
   }));
 
+  function getUnavailableFeatures(theme: Theme) {
```
pete-watters, line 66: Maybe this exist in it's own file to format the data then pass in a simpler way? 

So this could accept `featureEntries` where `featureEntries = Object.entries(getUnavailableFeatures(theme))`

The `.layout` file should just be UI formatting

apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
```diff
@@ -128,12 +140,13 @@ export function AddWalletSheetLayout({
...
+                  function fallbackOnPress() {
+                    onOpenNotificationsSheet({ title: feature.title, id: featureKey });
+                  }
                   return (
                     <AddWalletListItem
                       key={featureKey}
-                      onPress={() =>
-                        onOpenNotificationsSheet({ title: feature.title, id: featureKey })
-                      }
+                      onPress={feature.onPress ?? fallbackOnPress}
```
pete-watters, line 149: Maybe we can keep the function anonymous here as otherwise the `.map` will create `fallbackOnPress` even if it has `feature.onPress`

packages/ui/src/icons/index.native.ts
```diff
@@ -61,4 +61,19 @@ export * from './support-icon.native';
 export * from './test-tube-icon.native';
 export * from './key-icon.native';
 export * from './terminal-icon.native';
+export * from './logo-mpc-bitgo.native';
```
pete-watters, line 64: Maybe we could add a `logo/` folder here to help group these together?

apps/mobile/src/app/(home)/mpc-wallets.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const notifyUserSheetRef = useRef<SheetRef>(null);
+  const [optionData, setOptionData] = useState<OptionData | null>(null);
+  function onOpenNotificationsSheet(option: OptionData) {
+    setOptionData(option);
+    notifyUserSheetRef.current?.present();
+  }
+  function onCloseNotificationsSheet() {
+    setOptionData(null);
+  }
+  function getUnavailableFeatures() {
```
pete-watters, line 39: Does this have to be a function? It seems like it just returns a config object we could move into another file

apps/mobile/src/app/(home)/mpc-wallets.tsx
```diff
@@ -0,0 +1,111 @@
...
+      portal: { title: t`Portal`, Icon: LogoMPCPortal },
+      privy: { title: t`Privy`, Icon: LogoMPCPrivy },
+      qredo: { title: t`Qredo`, Icon: LogoMPCQredo },
+    };
+
+    return UNAVAILABLE_FEATURES;
+  }
+  return (
+    <>
+      <Box
```
pete-watters, line 55: It would be good to make a `.layout` for this and have the `{Object.entries(getUnavailableFeatures()).map(featureEntry => {` part as a separate component passed as `{children}`

pete-watters (approved):
Nice work Edgar 💪 I added some suggested improvements but overall LGTM

apps/mobile/src/app/(home)/hardware-wallets.tsx
```diff
@@ -0,0 +1,106 @@
...
+    bitgo: { title: t`Bitkey`, Icon: LogoHardwareBitkey },
+    capsule: { title: t`Ledger`, Icon: LogoHardwareLedger },
+    copper: { title: t`OneKey`, Icon: LogoHardwareOnekey },
+    fireblocks: { title: t`Passport`, Icon: LogoHardwareFoundation },
+    foredefi: { title: t`Ryder`, Icon: LogoHardwareRyder },
+    portal: { title: t`Trezor`, Icon: LogoHardwareTrezor },
+  };
+}
+
+export default function HardwareWalletsScreen() {
```
kyranjamie, line 36: Can we make a layout component here that separates the logic from the styles 🙏🏼

apps/mobile/src/app/(home)/mpc-wallets.tsx
```diff
@@ -0,0 +1,110 @@
...
+    copper: { title: t`Copper`, Icon: LogoMPCCopper },
+    fireblocks: { title: t`Fireblocks`, Icon: LogoMPCFireblocks },
+    foredefi: { title: t`Foredefi`, Icon: LogoMPCFordefi },
+    portal: { title: t`Portal`, Icon: LogoMPCPortal },
+    privy: { title: t`Privy`, Icon: LogoMPCPrivy },
+    qredo: { title: t`Qredo`, Icon: LogoMPCQredo },
+  };
+}
+
+export default function MPCWalletsScreen() {
```
kyranjamie, line 40:
```suggestion
export default function MpcWalletsScreen() {
```

elsewhere we treat acronyms like words when usig PascalCase

apps/mobile/src/app/(home)/mpc-wallets.tsx
```diff
@@ -0,0 +1,110 @@
...
+              {Object.entries(getUnavailableFeatures()).map(featureEntry => {
+                const [featureKey, feature] = featureEntry;
+                const mpcWalletName = feature.title;
+                function onPress() {
+                  onOpenNotificationsSheet({
+                    title: t`Connect MPC wallet: ${mpcWalletName}`,
+                  });
+                }
+                return (
+                  <Cell
```
fbwoolf, line 90: I don't think we should add more uses of `Cell` as it is currently?

apps/mobile/src/app/(home)/hardware-wallets.tsx
fbwoolf: Agree here, these components are quite hard to read.

apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
```diff
@@ -124,22 +100,44 @@ export function AddWalletSheetLayout({
...
-                      onPress={() =>
-                        onOpenNotificationsSheet({ title: feature.title, id: featureKey })
-                      }
-                      title={feature.title}
-                      subtitle={feature.subtitle}
-                      icon={feature.icon}
-                    />
-                  );
-                })}
+            {!moreOptionsVisible ? null : (
```
fbwoolf, line 103:
```suggestion
            {moreOptionsVisible && (
```

apps/mobile/src/app/(home)/mpc-wallets.tsx
```diff
@@ -0,0 +1,110 @@
...
+                    py="4"
+                    key={featureKey}
+                    onPress={onPress}
+                    title={feature.title}
+                    Icon={feature.Icon}
```
fbwoolf, line 95: To follow our code conventions, we should start refactoring capitalized props in mobile? They should all be lowercase.

packages/ui/src/icons/logos/logo-mpc-bitgo.native.tsx
```diff
@@ -0,0 +1,10 @@
+import LogoMPCBitgoLogo from '../../assets/icons/logos/logo-mpc-bitgo-24-24.svg';
+import { Icon, IconProps } from '../icon/icon.native';
+
+export function LogoMPCBitgo({ variant, ...props }: IconProps) {
```
fbwoolf, line 4: Think all these should be `Mpc` in naming.

apps/mobile/src/app/(home)/mpc-wallets.tsx
pete-watters: I think that for `Icon`, the way things are setup, it needs to be passed as upper case as it's a component and sometimes we need to pass more props to it. 

We could call it `icon` and load it with `{icon}` but then if we need to set the color it needs to be capital to give it props `<Icon color={...`. Maybe there is a way to do this while keeping it lower case.

pete-watters: If we are refactoring away `Cell` maybe we don't need to do this anymore? 

In there we are using it: 
```
   {Icon && (
          <Box flexDirection="row" p="2" bg={getIconBackgroundColor(_variant)} borderRadius="round">
            <Icon color={theme.colors[getIconColor(_variant)]} />
          </Box>
        )}
```

Maybe we can instead just pass the icon with the colour needed?

fbwoolf: Yeah, I refactored in some places so we can pass as lowercase. We need to do it in all places when encountered. The icons can just be type `ReactNode` now.

PR #480, opened 2024-09-26
feat: settings notifications

fbwoolf:
This PR implements screens for settings notifications, adds tabs for 'push' and 'email' notifications, and uses an input to store a preferred email address to receive notifications.

I used zod to validate the email address. I'll need design feedback on what to show for an invalid email address, rn I'm showing a toast but we only have a `success` and `info` types currently. I think I need to add a `fail` type?


apps/mobile/src/app/(home)/settings/notifications/_layout.tsx
```diff
@@ -0,0 +1,66 @@
...
+          },
+          title: t`Email`,
+          isActive: pathname === AppRoutes.SettingsNotificationsEmail,
+        },
+      ]}
+    />
+  );
+}
+
+export default function SetttingsNotificationsLayout() {
```
pete-watters, line 36: You have a typo here or else you're excited about this feature `Setttings`

apps/mobile/src/app/(home)/settings/notifications/email.tsx
```diff
@@ -0,0 +1,58 @@
...
+} from '@leather.io/ui/native';
+
+export default function SettingsNotificationsEmailScreen() {
+  const { bottom } = useSafeAreaInsets();
+  const emailAddressSheetRef = useRef<SheetRef>(null);
+  const theme = useTheme<Theme>();
+
+  return (
+    <Box flex={1} backgroundColor="ink.background-primary">
+      <ScrollView
```
pete-watters, line 27: Not for this PR but we seem to repeat this `<ScrollView` with these props a lot (5 times). We could look into a layout for it?

apps/mobile/src/app/(home)/settings/notifications/_layout.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { BackButtonHeader } from '@/components/headers/back-button';
+import { SimpleHeader } from '@/components/headers/containers/simple-header';
+import { TitleHeader } from '@/components/headers/title';
+import { TabBar } from '@/components/tab-bar';
+import { AppRoutes } from '@/routes';
+import { t } from '@lingui/macro';
+import { Tabs, router, usePathname, useRouter } from 'expo-router';
+
+function HeaderBottomTabs() {
```
pete-watters, line 11: Maybe `HeaderBottomTabs` + `NavigationHeader` could be in their own `Header` component?

pete-watters (approved):
Great work 🚀  

I added suggestions but it's fine as is, bar typo

apps/mobile/src/store/settings/settings.ts
```diff
@@ -54,8 +57,8 @@ export const initialState: SettingsState = {
   accountDisplayPreference: 'native-segwit',
   analyticsPreference: 'consent-given',
   bitcoinUnitPreference: 'bitcoin',
+  emailAddressPreference: '',
   fiatCurrencyPreference: 'USD',
-  createdOn: new Date().toISOString(),
```
kyranjamie, line 58: Why is this removed? This is useful metadata to keep even if we don't display anywhere.

fbwoolf: I searched and it wasn't being used, so I thought it was just done temporarily for dev purposes. I can add it back. I think at one point it was being used as a placeholder where designers want us to show the app version number? We have the same prop on wallet state.

apps/mobile/src/app/(home)/settings/notifications/email.tsx
```diff
@@ -0,0 +1,58 @@
...
+              </Avatar>
+            }
+          >
+            <ItemLayout
+              actionIcon={<ChevronRightIcon variant="small" />}
+              captionLeft={t`Awaiting verification`}
+              titleLeft={t`Email address`}
+            />
+          </Flag>
+        </TouchableOpacity>
```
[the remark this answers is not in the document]
fbwoolf: The opacity does drop on my other uses of this, so let me check why different here.

apps/mobile/src/app/(home)/settings/notifications/_layout.tsx
fbwoolf: Our headers really need some work, so I'll see if there is an issue to coordinate all these a bit better. If not, I'll make one.

apps/mobile/src/app/(home)/settings/notifications/email.tsx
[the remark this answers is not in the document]
fbwoolf: I tested this and the opacity change DOES work for me, so not sure why wasn't for you.

apps/mobile/src/app/(home)/settings/notifications/email.tsx
fbwoolf: I forgot I created a `SettingsScreenLayout` to use here, so will do. Thx for reminder.

fbwoolf:
Updated the toast types here cc @fabric-8, also the copy on the success was to wordy so I simplified to fit on screen better...

apps/mobile/src/app/(home)/settings/notifications/_layout.tsx
pete-watters: Thanks @fbwoolf 👍 It seems like we have the headers we need but they could do with some refactoring

PR #481, opened 2024-09-26
feat: collectibles widget

pete-watters:
This PR:
- Adds collectibles widget to the native app using mock data 
- Migrates some similar UI components from the extension to the monorepo 

Once the UI package is released I can complete https://github.com/leather-io/extension/pull/5876

**Mobile demo**

https://github.com/user-attachments/assets/607f8ce1-bf87-40bf-884e-1c53ba948507

**Mobile storybook demo**

https://github.com/user-attachments/assets/5c4f585e-4893-4ddd-9729-5760234213a9

There are some remaining items to figure out:

- [ ] An alternative for `dompurify.sanitize` for text ordinals 
- [ ] Stopping scroll on `WebView` ordinals
- [ ] Fixing padding and margin on all widgets @edgarkhanzadian - I'll do this [in](https://github.com/leather-io/issues/issues/349)


packages/ui/src/hooks/use-pressable.web.tsx
```diff
@@ -0,0 +1,67 @@
+import { Square, SquareProps } from 'leather-styles/jsx';
```
pete-watters, line 1: I moved this as it was the only other thing using `use-events` so I can also move that package here. 

We use that in [`CollectibleItemLayout`](https://github.com/leather-io/mono/pull/481/files#diff-ef02bb76527afe8ac113dc4b66eea143904026a1bd6c05e7f439e646b9ef81f5) which I migrated here but could revert to the `extension` if desired. 

The hook `usePressable` is only used once in the extension `accounts.tsx` = `AddAccountAction`

packages/ui/src/components/collectibles/collectibles.shared.tsx
```diff
@@ -0,0 +1,319 @@
...
+  genesisBlockHeight: number;
+  genesisBlockHash: string;
+  genesisTimestamp: number;
+  value: string;
+  mimeType: string;
+  name: string;
+  src: string;
+}
+
+const mockOrdinals: Ordinal[] = [
```
pete-watters, line 21: I obtained these mocks by logging all the inscriptions we have on Leather Dev wallet and picking out different samples. 

The structure here is what we are sending to the components but it seems different than the [BIS docs](https://docs.bestinslot.xyz/reference/api-reference/ordinals-and-brc-20-and-runes-and-bitmap-v3-api-mainnet+testnet+signet/inscriptions) and also what we have in `leather.io/query` for `[mock-inscriptions](https://github.com/leather-io/mono/blob/dev/packages/query/test/mock-inscriptions.ts)`

Then today I found this other `utils` that is mapping BIS to this format - [`createBestInSlotInscription`](https://github.com/leather-io/mono/blob/dev/packages/query/src/bitcoin/ordinals/inscription.utils.ts) 

I think it's OK for now and we can improve this and share the mocks better if needed / ditch these altogether when connecting the API. 

I found mocks

packages/ui/src/components/collectibles/native/collectible-card.native.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  title: 'Collectibles/CollectibleCard',
+  component: CollectibleCard,
+  tags: ['autodocs'],
+  argTypes: {},
+  parameters: {},
+  decorators: [Story => <Story />],
+};
+
+export default meta;
+
```
pete-watters, line 15: I added some mocks of what we use to actually render inscriptions here.

- Video seems to be handled the same as `html` / `svg` in an iframe anyway. 
- Audio seems to just [show an icon](https://github.com/leather-io/extension/blob/dev/src/app/features/collectibles/components/_collectible-types/collectible-audio.tsx) now 

 I can inscribe audio + video on ordinals bot if we think that's useful. For video the network fee is high so it's ~ $392 for what I tested. Audio can be cheaper ~ $122.02.

packages/ui/src/components/collectibles/native/collectible-html.native.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { CollectibleCardLayout } from './collectible-card-layout.native';
+
+interface CollectibleHtmlProps {
+  src: string;
+}
+
+export function CollectibleHtml({ src }: CollectibleHtmlProps) {
+  return (
+    <CollectibleCardLayout>
+      <WebView source={{ uri: src }} width={200} height={200} style={{ overflow: 'hidden' }} />
```
pete-watters, line 12: I need to do a bit more work here as I still see some scroll bars

https://github.com/user-attachments/assets/eee72595-9e99-46d1-8359-94342ac8d0cb

apps/mobile/src/components/widgets/collectibles/collectibles-serializer.ts
```diff
@@ -0,0 +1,21 @@
+import { type Collectible, type CollectibleCardProps } from '@leather.io/ui/native';
```
pete-watters, line 1: I added this as a way to just get only the data we need to display from the mocks. 

Later I found `createBestInSlotInscription` which is actually giving us this data and more. 

I will investigate more when hooking up APIs but maybe it could be good to have a BE proxy for this type of stuff - data we need for App Home for example.

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <CollectibleCardLayout bg="ink.text-primary" p="4">
+      <Text color="ink.background-secondary" textAlign="left">
+        {/* FIXME 
+        - implement alternative for dompurify in native
+        - I tried using jsdom as a polyfill but then hit an error Can't resolve 'vm' in jsdom
+        - maybe we should write our own sanitizer?
+        */}
+        {/* {content ? sanitize(content) : ''} */}
```
pete-watters, line 37: I tried using `dompurify.sanitize` here but it seems not to work on native. 

I tried a polyfill of `jsdom` but then that failed and needed another `polyfill` so I need to investigate a better way. 

Maybe we could run `sanitize` in the query instead? So that we just provide the data ready to render

packages/ui/src/components/collectibles/native/collectible-card.native.tsx
```diff
@@ -0,0 +1,31 @@
...
+  src: string;
+  mimeType?: string | null;
+}
+
+export function CollectibleCard({ name, type, src, mimeType }: CollectibleCardProps) {
+  const isOrdinal = type === 'inscription';
+
+  if (isOrdinal) {
+    switch (mimeType) {
+      // TODO: add audio support
```
pete-watters, line 17: I didn't implement `mimeType === 'audio'` yet. It seems like on web we just show a Head phone icon unless I am missing something

fbwoolf: Yep, that is correct. We didn't want the user to interact with the controls in the iframe so we just render an icon with a link to the audio.

apps/mobile/src/components/widgets/collectibles/collectibles-widget.layout.tsx
```diff
@@ -0,0 +1,30 @@
+import { ScrollView } from 'react-native-gesture-handler';
+
+import { useTheme } from '@shopify/restyle';
+
+import { Box, SheetRef, Theme } from '@leather.io/ui/native';
+
+import { Widget } from '../widget';
+
+interface CollectiblesWidgetProps {
```
fbwoolf, line 9:
```suggestion
interface CollectiblesWidgetLayoutProps {
```

fbwoolf (approved):
Looks great, @pete-watters.

packages/ui/src/components/approver/animations/approver-animation.web.tsx
```diff
@@ -1,6 +1,6 @@
 import { motion, stagger, useAnimate } from 'framer-motion';
 import { css } from 'leather-styles/css';
-import { HasChildren } from 'src/utils/has-children';
+import { HasChildren } from 'src/utils/has-children.shared';
```
kyranjamie, line 3: Can you rebase on dev and make sure there are no `src/` imports? Think they should cause type errors on rebase

kyranjamie (approved):

packages/ui/src/components/collectibles/native/collectible-card.native.tsx
pete-watters: OK thanks. I updated to add basic support just showing an icon so something will appear

packages/ui/src/components/approver/animations/approver-animation.web.tsx
pete-watters: I rebased and cleaned this up. I think we should add path aliases to the `tsconfig` to clean up all of the relative path imports. Is there a reason not to do that? 

It would mean we need to have the `src/` path in `ui/tsconfig.json`:

```
"paths": {
      "leather-styles/*": ["leather-styles/*"],
      "@components/*": ["src/components/*"],
      "@icons/*": ["src/icons/*"],
      "@hooks/*": ["src/hooks/*"],
      "@utils/*": ["src/utils/*"],
```

packages/ui/src/components/collectibles/collectibles.shared.tsx
```diff
@@ -0,0 +1,319 @@
...
+  preview: string;
+  title: string;
+  genesisBlockHeight: number;
+  genesisBlockHash: string;
+  genesisTimestamp: number;
+  value: string;
+  mimeType: string;
+  name: string;
+  src: string;
+}
```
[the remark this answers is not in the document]
pete-watters: I'll move it when connecting to real data

PR #484, opened 2024-09-27
refactor/ui-package-bundling

kyranjamie:
This PR fixes several issues with the bundling of the UI package targeting web. Was close to calling Mission Abort on this one, but eventually trial-and-errored a way to fixing the extension styles 😅

- Native code no longer bundled into `web-dist` folder
- Prevents double-bundling with `tsup` and `tsc`
- Removes tsconfig settings causing default to absolute path imports
- Moves export file into `src` folder
- Adds export for UI lib styles


pete-watters:
Great improvements, thanks for doing this 👏

PR #493, opened 2024-09-30
feat: app auth

fbwoolf:
This PR implements the biometrics trigger for when a user toggles the app auth setting.

*Note, when testing, Face ID often doesn't work, you will instead be prompted to enter a passode:
https://github.com/expo/expo/issues/21694

### DIFF
--- a/apps/mobile/src/features/settings/account-identifier/account-identifier-cell.tsx
+++ b/apps/mobile/src/features/settings/account-identifier/account-identifier-cell.tsx
@@ -1,7 +1,9 @@
+import { Pressable } from 'react-native';
+
 import { t } from '@lingui/macro';
 
 import { AccountDisplayPreference, AccountDisplayPreferenceInfo } from '@leather.io/models';
-import { ItemLayout, RadioButton, TouchableOpacity } from '@leather.io/ui/native';
+import { ItemLayout, RadioButton } from '@leather.io/ui/native';
 
 interface AccountIdentifierCellProps {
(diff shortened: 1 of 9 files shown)


apps/mobile/src/features/settings/account-identifier/account-identifier-cell.tsx
```diff
@@ -19,7 +21,7 @@ export function AccountIdentifierCell({
   title,
 }: AccountIdentifierCellProps) {
   return (
-    <TouchableOpacity onPress={() => onUpdateAccountDisplayPreference(accountDisplayPref.type)}>
+    <Pressable onPress={() => onUpdateAccountDisplayPreference(accountDisplayPref.type)}>
```
pete-watters, line 24: Should I be refactoring my work to use `Pressable` instead of `TouchableOpacity` ?

pete-watters (approved):

apps/mobile/src/features/settings/account-identifier/account-identifier-cell.tsx
fbwoolf: cc @fabric-8 here bc not sure everywhere? In the settings design review we were asked to remove the opacity change on press of the 'cells'.

pete-watters: Let's sync on this @fabric-8 and let me know if I need to update the homepage. I can gather some places where we use opacity to check with you

PR #495, opened 2024-10-01
fix: refactor widget code, ref leather-io/issues#349

pete-watters:
This PR makes some basic design changes requested in https://github.com/leather-io/issues/issues/349 and performs a refactor of `widget` code to share code better and reduce un-necessary component wrapping 

Visual changes are:
- title font change
- removal of `Chip` count 

https://github.com/user-attachments/assets/cf9851c7-6258-42e6-809f-548acf14886f


apps/mobile/src/components/widgets/components/widget/widget-title.tsx
```diff
@@ -0,0 +1,25 @@
+import { useTheme } from '@shopify/restyle';
+
+import { Box, ChevronRightIcon, Text, Theme } from '@leather.io/ui/native';
+
+import { FiatBalance } from '../balance/fiat-balance';
+
+interface WidgetTitleProps {
+  title: string;
+  totalBalance?: string;
```
kyranjamie, line 9: Shouldn't we be passing `Money`?

kyranjamie (approved):

apps/mobile/src/components/widgets/components/widget/widget-title.tsx
pete-watters: We should, good spot. Right now it's a mocked string though. 

I'll add a note to update in my next work

PR #498, opened 2024-10-01
fix(ui): approver updates, animations

kyranjamie:
Some small changes to `<Approver />` that tighten up some of the animations, and makes the inner container full width.


pete-watters (approved):

kyranjamie:
> would be great to attach video as well

You can see it in the extension PR, these are only very subtle changes

PR #499, opened 2024-10-01
Fix/connect home to data

pete-watters:


kyranjamie:
I'd encourage looking at this with fresh eye, versus how we handle balances on the extension. There are a few difference with how we store accounts here, that hopefully makes it a little easier. Though, I do wonder just how much we can reuse (at least in its current state)

I'm thinking something like `useQueries` rather than `useQuery` that just accepts an array of all the accounts a user has would be ideal?

pete-watters:
> I'd encourage looking at this with fresh eye, versus how we handle balances on the extension. There are a few difference with how we store accounts here, that hopefully makes it a little easier. Though, I do wonder just how much we can reuse (at least in its current state)
> 
> I'm thinking something like `useQueries` rather than `useQuery` that just accepts an array of all the accounts a user has would be ideal?
> 
> Generally think it's worth setting the expectation that this is going to be quite hard to get right.

Thanks @kyranjamie 👍 yes I just wanted to flag that it's a non trivial task and could be tricky to get right. 

We could also potentially share the solution in the `extension` where that passes an array with just the one accounts we have there as I guess we will need to support multiple accounts there too. 

It seems like something worth discussing and planning as a team to make sure we get it right (I'll close this draft for now)

PR #501, opened 2024-10-01
feat: settings privacy mode

fbwoolf:
This PR adds privacy mode to settings so that a user can show or hide their account balances. In addition to toggling the setting itself, a user can also now press on any balance to show/hide all balances...

### DIFF
--- a/apps/mobile/src/app/(home)/settings/display/index.tsx
+++ b/apps/mobile/src/app/(home)/settings/display/index.tsx
@@ -11,9 +11,11 @@ import { useLingui } from '@lingui/react';
 import {
   BitcoinCircleIcon,
   DollarCircleIcon,
+  Eye1Icon,
   PackageSecurityIcon,
   SheetRef,
   SunInCloudIcon,
+  Switch,
 } from '@leather.io/ui/native';
 import { capitalize } from '@leather.io/utils';
 
@@ -28,10 +30,16 @@ export default function SettingsDisplayScreen() {
   const {
     accountDisplayPreference,
     bitcoinUnitPreference,
+    changePrivacyModePreference,
     fiatCurrencyPreference,
+    privacyModePreference,
     themePreference,
   } = useSettings();
   const { i18n } = useLingui();
+
+  function onUpdatePrivacyMode() {
+    changePrivacyModePreference(privacyModePreference === 'visible' ? 'hidden' : 'visible');
+  }
 
   return (
     <>
@@ -40,7 +48,7 @@ export default function SettingsDisplayScreen() {
           title={t`Theme`}
           caption={i18n._(capitalize(themePreference))}
           icon={<SunInCloudIcon />}
-          onCreateSheetRef={() => {
+          onPress={() => {
             themeSheetRef.current?.present();
           }}
         />
@@ -48,7 +56,7 @@ export default function SettingsDisplayScreen() {
           title={t`Bitcoin unit`}
           caption={i18n._(bitcoinUnitPreference.symbol)}
           icon={<BitcoinCircleIcon />}
-          onCreateSheetRef={() => {
+          onPress={() => {
             bitcoinUnitSheetRef.current?.present();
           }}
         />
@@ -56,7 +64,7 @@ export default function SettingsDisplayScreen() {
           title={t`Conversion unit`}
           caption={i18n._(fiatCurrencyPreference)}
           icon={<DollarCircleIcon />}
-          onCreateSheetRef={() => {
+          onPress={() => {
             conversionUnitSheetRef.current?.present();
           }}
         />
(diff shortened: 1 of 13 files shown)


fbwoolf:
Continuing work here to implement privacy mode, so I'll change to draft temporarily.

pete-watters:
Feel free to keep going on this and get it done. 

In case it helps, on the homepage, I have two balance components - `FiatBalance` + `TokenBalance`.

They are very similar so I was thinking to refactor and merge them. Then we could check the state in there and decide to show / hide

fbwoolf:
> They are very similar so I was thinking to refactor and merge them. Then we could check the state in there and decide to show / hide

I saw this after I implemented it on both, so let me know if I should refactor them into one? Maybe it is ok to keep them separate?

fbwoolf:
**Also, note there were a couple of build issues I encountered with icons, so I fixed them here.**

apps/mobile/src/app/(home)/settings/display/index.tsx
```diff
@@ -28,46 +30,59 @@ export default function SettingsDisplayScreen() {
...
+
   return (
     <>
       <SettingsScreenLayout>
         <DisplayCell
           title={t`Theme`}
           caption={i18n._(capitalize(themePreference))}
           icon={<SunInCloudIcon />}
-          onCreateSheetRef={() => {
+          onPress={() => {
```
camerow, line 51: I hope you don't mind me peeking in on some PRs - quick question about passing anon-functions - are we picking and choosing where to use `useCallback` based on performance concerns? These seem like good candidates to create via `useCallback` to ensure we're not needlessly re-rendering cells.

apps/mobile/src/components/private-text.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const defaultPrivateText = '***';
+
+interface PrivateTextProps {
+  color?: TextProps['color'];
+  privateText?: string;
+  text: string;
+  variant?: TextProps['variant'];
+}
+export function PrivateText({
```
camerow, line 16: What do you think about this component not knowing anything about value it is wrapping? eg 

```
<PrivateValue>
  <Text>some text</Text>
</PrivateValue>
```
so that this component just either returns it's children or the private text?

apps/mobile/src/components/private-text.tsx
```diff
@@ -0,0 +1,36 @@
...
+}: PrivateTextProps) {
+  const { changePrivacyModePreference, privacyModePreference } = useSettings();
+  const isPrivate = usePrivacyMode();
+
+  function onTogglePrivacyMode() {
+    changePrivacyModePreference(privacyModePreference === 'visible' ? 'hidden' : 'visible');
+  }
+
+  return (
+    <Pressable onPress={() => onTogglePrivacyMode()}>
```
camerow, line 30: *thought*: Another good candidate for `useCallback`?

apps/mobile/src/app/(home)/settings/display/index.tsx
fbwoolf: We have had debates about overusing `useMemo` and `useCallback`:
https://kentcdodds.com/blog/usememo-and-usecallback

Not sure it is necessary on the cases you flagged?

apps/mobile/src/components/private-text.tsx
fbwoolf: I thought about this, but the private text has to take on the styles of the 'children' ...so it didn't seem like the right approach. And, the private text itself isn't always the same. I'll look at it again to see if I missed something about doing the wrapper.

apps/mobile/src/components/private-text.tsx
fbwoolf: See comment above, I'm not sure `useCallback` is nec here?

apps/mobile/src/app/(home)/settings/display/index.tsx
camerow: You're very likely right - probably safe (and wise) to not overuse it and add when necessary.

apps/mobile/src/components/private-text.tsx
camerow: Nice looks good! I like this encapsulation as the consumer needs to know a lot less about the implementation, and it has a smaller API surface area. Thanks for updating!

packages/ui/src/icons/wallet-plus-icon.native.tsx
```diff
@@ -1,12 +1,12 @@
-import WalletSmall from '../assets/icons/wallet-16-16.svg';
+import WalletPlusSmall from '../assets/icons/wallet-plus-16-16.svg';
```
pete-watters, line 1: Thanks for fixing this too! 👏

pete-watters:
Nice work @fbwoolf 👏 thanks for cleaning up some of my work too. 

I will think about merging `TokenBalance` / `FiatBalance` but I think it's OK for now. 

I'll merge this so I can rebase on top

pete-watters (approved):

PR #504, opened 2024-10-03
fix(ui): add tooltip exit animation

kyranjamie:
- Animates the tooltip on exit, otherwise it feels kinda janky when it animates in then abruptly disappears 
- Renames `info` element in Approver to avoid being tooltip specific
- I've also added back Flag controls, so we can highlight how it can be used for different layouts
- Fixes Flag not responding to all controls 

https://github.com/user-attachments/assets/6c688d01-292a-40e9-8efe-88811ccd7749

### DIFF
--- a/packages/ui/src/components/approver/components/approver-header.web.tsx
+++ b/packages/ui/src/components/approver/components/approver-header.web.tsx
@@ -12,14 +12,10 @@ import { useApproverContext, useRegisterApproverChild } from '../approver-contex
 
 interface ApproverHeaderProps extends HasChildren {
   title: ReactNode;
-  iconTooltip?: ReactNode;
+  info?: ReactNode;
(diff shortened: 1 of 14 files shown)


packages/ui/src/components/approver/components/approver-header.web.tsx
```diff
@@ -31,7 +27,7 @@ export function ApproverHeader({
       pos="relative"
     >
       <ApproverHeaderAnimation>
-        <styled.h1 textStyle="heading.03" mr={iconTooltip ? 'space.06' : ''}>
+        <styled.h1 textStyle="heading.03" mr={info ? 'space.06' : ''}>
```
pete-watters, line 30: Maybe `mr` should be `undefined` if no `info`?

pete-watters (approved):

fbwoolf (approved):

PR #505, opened 2024-10-03
fix: use Action instead of ItemLayout

edgarkhanzadian:
Example use of Action:

```jsx
// for chevron
<Cell.Root
  title={t`title`}
  caption={t`caption`}
  onPress={onPress}
  Icon={<PlaceholderIcon />}
>
  <Cell.Chevron />
</Cell.Root>

// for Switch
<Cell.Root
  title={t`title`}
  caption={t`caption`}
  onPress={onPress}
  Icon={<PlaceholderIcon />}
>
  <Cell.Switch onPress={onPress} isSelected={/** some boolean value */ true} />
</Cell.Root>
```

The only annoying thing right now is that we need to manually pass same onPress to Switch too. We can't disable it because we lose switch animation that way.


packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
+import { RadioButtonProps } from '../radio-button/radio-button.native';
+import { SwitchProps } from '../switch/switch.native';
 
 type RegularCellVariant = 'active' | 'inactive' | 'critical';
-type SwitchCellVariant = 'switch';
-type CellVariant = RegularCellVariant | SwitchCellVariant;
 
 function getIconColor(
-  variant: CellVariant
+  variant: RegularCellVariant
```
fbwoolf, line 22: What does `Regular...` mean here? I think there are some naming inconsistencies that make the code hard to follow. Here, the three type options don't all seem to make sense together? `active` and `inactive` make sense, but what does `critical` mean then in relation to those two states? Can the cell be both `active` and `critical`?

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
 
-interface SwitchCell {
-  variant: SwitchCellVariant;
-  switchValue: boolean;
-  toggleSwitchValue(): void;
-}
+export function Cell({ caption, Icon, title, children, variant, ...props }: CellProps) {
+  const theme = useTheme<Theme>();
+
+  const itemLayout = <ItemLayout actionIcon={children} captionLeft={caption} titleLeft={title} />;
```
fbwoolf, line 44: So, this proposal would not cover cells that don't have an action and actually has right content in the `ItemLayout`?

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
+    </Pressable>
+  );
 }
 
-interface BaseCell {
-  title: string;
-  subtitle?: string;
-  onPress?(): unknown;
-  Icon?: React.FC<IconProps>;
+function Press() {
```
fbwoolf, line 67: Wonder if there is a better name here? All of them are using `Pressable` ...if specific to using the chevron icon, maybe just `Chevron`?

fbwoolf (commented):
Thanks for the proposed changes @edgarkhanzadian, I left a few questions.

kyranjamie (commented):
Great work Edgar, I like the ideas here. Would love to work on some naming choices, but the composition looks good. 

I think it's unnecessary to pass Switch and Radio through the Cell component. The grouping has no real meaning, as it'd work just as well passing them in directly. The "actions" in my mind, are the form fields, not the cell, so I imagine an API like

```tsx
<Cell.Root>
  <Cell.Content {...allItemLayoutProps} />
  <Cell.Action>
    <Radio />
  </Cell.Action>
</Cell.Root>
```

Imagine we add a checkbox. Why should I first need to add it to `Cell` in the style of `Cell.Checkbox`, if it works just as well to pass it in directly?

apps/mobile/src/app/(home)/hardware-wallets.tsx
```diff
@@ -51,13 +51,16 @@ export default function HardwareWalletListScreen() {
...
             return (
-              <Cell
+              <Action.Cell
                 py="4"
                 key={featureKey}
-                onPress={onPress}
                 title={feature.title}
                 Icon={feature.Icon}
-              />
+                variant="active"
```
fbwoolf, line 59: Thinking `active` should be the default rather than having to set it on each?

apps/mobile/src/app/(home)/settings/help/index.tsx
```diff
@@ -4,9 +4,14 @@ import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { t } from '@lingui/macro';
```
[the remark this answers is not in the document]
pete-watters: I agree and if saving the import path is desired you can still have a simpler `index.ts` that exports the component.

Development can be hard when you have lots of files open and all are called `index.tsx`

apps/mobile/src/app/(home)/settings/networks/index.tsx
```diff
@@ -7,22 +7,27 @@ import {
...
+  PlaceholderIcon,
+  PlaygroundFormsIcon,
+  TestTubeIcon,
+} from '@leather.io/ui/native';
 import { capitalize } from '@leather.io/utils';
 
 import SettingsScreenLayout from '../settings-screen.layout';
-import { NetworkCell } from './network-cell';
 
 function getNetworkIcon(network: DefaultNetworkConfigurations) {
```
pete-watters, line 21: Maybe `getNetworkIcon` could instead be a component that returns a JSX version of the icon to help avoid the uppercase `Icon=`

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
-      return 'ink.background-secondary';
-    case 'critical':
-      return 'red.background-primary';
-  }
+interface CellProps extends PressableProps<Theme> {
+  caption?: string;
+  Icon?: React.FC<IconProps>;
+  title: string;
+  children: NonNullable<ReactNode>;
+  variant: RegularCellVariant;
```
pete-watters, line 38: The `variant` seems to just feed into `getIconColor` if I'm not mistaken. 

Why can't we just pass the icons with the colour they need instead of the `variant` prop? 

Most `Cell`'s seem to set it to `active` anyway

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
+
+  const itemLayout = <ItemLayout actionIcon={children} captionLeft={caption} titleLeft={title} />;
 
-interface RegularCell {
-  variant?: RegularCellVariant;
+  const content = Icon ? (
+    <Flag
+      img={
+        <Avatar>
+          <Icon color={theme.colors[getIconColor(variant)]} />
```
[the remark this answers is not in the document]
pete-watters: See feedback [here](https://github.com/leather-io/mono/pull/505/files#r1787175530) I think it could be nice to have the icon just passed in with the correct colour set and avoid the `variant` prop altogether if needed.

 Most of them are `active` which should render the right colour by default?
 
 Then for `critical` we can pass the icon as `<TrashIcon color='red.action-primary-default' />`

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
-  onPress?(): unknown;
-  Icon?: React.FC<IconProps>;
+function Press() {
+  return <ChevronRightIcon variant="small" />;
 }
 
-type CellProps = BaseCell & (RegularCell | SwitchCell) & TouchableOpacityProps<Theme>;
+function Switch(props: SwitchProps) {
+  return <UISwitch {...props} />;
+}
```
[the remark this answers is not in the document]
pete-watters: I feel the same about `Radio` and `Press`. We pass a single prop to each of those which I can understand but I still think it's better to not bother wrapping them call them directly

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,84 @@
...
-          value={props.switchValue}
-        />
-      )}
-    </TouchableOpacity>
-  );
+function Radio(props: RadioButtonProps) {
+  return <RadioButton disabled {...props} />;
 }
+
+export const Action = {
```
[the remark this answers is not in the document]
pete-watters: I'd just call it `Cell` and not bother exporting the other parts as they are just wrapper layers passing on props

apps/mobile/src/app/(home)/settings/networks/index.tsx
```diff
@@ -39,14 +44,16 @@ export default function SettingsNetworksScreen() {
   return (
     <SettingsScreenLayout>
       {defaultNetworkPreferences.map(network => (
-        <NetworkCell
-          key={network}
-          caption={settings.networkPreference.id === network ? t`Enabled` : t`Disabled`}
-          icon={getNetworkIcon(network)}
-          isSelected={settings.networkPreference.id === network}
-          onChangeNetwork={() => onChangeNetwork(network)}
+        <Action.Cell
```
pete-watters, line 47: I think that these types of maps like `defaultNetworkPreferences.map` seem more efficient but actually make things harder to follow and add complexity. 

In this case if we want to minimise repetition we could have a `NetworkCell` that wraps `Action.Cell` and just call it more times: 

```
    <SettingsScreenLayout>
        <NetworkCell network='mainnet' icon={<GlobeIcon />} />
        <NetworkCell network='testnet' icon={<TestTubeIcon />} />
        <NetworkCell network='signet'  icon={<PlaygroundFormsIcon />}/>
   </SettingsScreenLayout>

... 
// change defaultNetworkPreferences to a type 
type Networks = 'mainnet' | 'testnet' | 'signet';

interface NetworkCell {
 network: Networks;
icon: React.ReactNode;
}

function NetworkCell({network, icon}: NetworkCell){
   return  (<Action.Cell> 
          icon={icon}
          title={i18n._(capitalize(network))}
          caption={settings.networkPreference.id === network ? t`Enabled` : t`Disabled`}
          onPress={() => onChangeNetwork(network)}>
           <Action.Radio isSelected={settings.networkPreference.id === network} />
       
   </Action.Cell>)
}
```

pete-watters (commented):
Great work here @edgarkhanzadian 👍 

I like having a shared `Cell` instead of repeating lots of similar variants of it. 

I added some of my opinions to open discussions as I think this is a great PR we can use to discuss patterns and feed it into our conventions

kyranjamie:
Icon colour setting ideas so far:
1. Passing component directly and adding prop 
```tsx
// <Component Icon={Icon} />
function Component({ Icon }) {
  return <Icon color="new.color" />
}
```
2. Passing JSX and adding prop with `cloneElement`
```tsx
// <Component icon={<Icon />} />
function Component(props) {
  const icon = cloneElement(props.icon, { color: 'new.color' });
}
```
3. Using context
```tsx
// <IconColorProvider color="new.color">
// <Component />
// </IconColorProvider>
function Icon(props) {
  const context = useIconColorContext();
  return <Icon color={props.color ?? context.color ?? defaultColor} />
}
```
4. currentColor. Not sure if this works in React Native but some sources indicate might
```tsx
// <View color="new.color">
// <Component />
// </View>

File: packages/ui/src/components/cell/cell.native.tsx (hunk 7 above)
Comment (fbwoolf) [reply packages/ui/src/components/cell/cell.native.tsx:38]: Agree here.

**Comment by fbwoolf** [comment]
> Icon colour setting ideas so far:
> 
I agree with @pete-watters in other feedback here, there just isn't a need to pass an icon w/out the props it needs. It is just how things were coded initially, but there is a better way to do it imo. The icons can just be passed with the color (or variant) it needs w/out needing to functions to get the color later ...or am I missing something?

**Comment by kyranjamie** [comment]
I'm not advocating for `cloneElements`. I've added another option per the alternates stated in the docs, using a render prop.

I think the feeling brewing from the discussions with Fara and Pete here and on Slack is that: there's surely a better way to do it than this? and I'm inclined to agree. 

The approach used here also reinitialises the component on every render, and is fairly unintuitive React, which encourage composition by passing JSX.

**Comment by fbwoolf** [comment]
I am currently looking at a similar situation we have built with our native Button. There, we named the prop `buttonState` instead of `variant` ...and we export `getButtonTextColor(buttonState)` to use and apply the color before passing the icon with the color applied to the `Button` as a ReactNode.

**Comment by fbwoolf** [comment]
> Soo made a few changes to the PR:
> 
> 1. Export UI as Cell not as Action
> 2. Icons are now passed as a ReactNode as `icon={<PlaceholderIcon />}`. Color is assigned through `cloneElement`, but it doesn't override a color that was passed during icon node creation.
> 3. No variants anymore, the color should be assigned when creating icon node.

Thx for working on these changes @edgarkhanzadian, fwiw it might make more sense to leave the `variant` prop and assign the color in the Cell component ...I just didn't understand what the variant prop was doing and tried to point out the defined types were unclear ...and inconsistent with other components. I didn't realize it was just being used for styling now. Whatever we decide, I think we should do the same with `buttonState` (prob change that to `variant`) ...with Cell, wouldn't using similar types be better: `'default' | 'disabled' | 'critical'`? I was just confused, and not really following the intention of your changes well, so if you feel it is better to leave/refine your previous direction, please feel free to. It seem like the `variant` prop would need to do some other style changes in Cell too? Wouldn't the chevron icon change color? Wouldn't the `ItemLayout` text colors change in some way if the Cell is 'disabled'? With our custom web `Pressable` the `disabled` prop handles a lot more styling changes than just the avatar icon.

Just a note too, if you leave these changes, the `variant` prop hasn't been removed entirely. It doesn't seem like it needs to be kept as a prop or type?

File: packages/ui/src/components/cell/cell.native.tsx (hunk 9 above)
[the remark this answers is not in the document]
Comment (kyranjamie) [reply packages/ui/src/components/cell/cell.native.tsx:60]: You may as well define it as, which spares the prop passing.

```suggestion
const Switch = UISwitch;
```

I think of `Cell` as this generic kind of layout row component, that may have some kind of action, optionally. So the decision to wrap these components seem unnecessary coupling between the core component and speicific use-cases

apps/mobile/src/app/(home)/settings/index.tsx
```diff
@@ -35,51 +34,74 @@ export default function SettingsScreen() {
...
             onPress={() => router.navigate(AppRoutes.SettingsWallet)}
-          />
+          >
+            <Cell.Chevron />
+          </Cell.Root>
         </Box>
         <Divider />
-        <Box gap="5" paddingHorizontal="5" paddingVertical="5">
-          <SettingsCell
+        <Box gap="2" paddingHorizontal="5" paddingVertical="5">
```
kyranjamie, line 47: Would `VStack` be more appropriate here?

apps/mobile/src/app/(home)/settings/help/index.tsx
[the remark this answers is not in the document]
kyranjamie: Open for debate here, proposed naming rules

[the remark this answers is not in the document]
kyranjamie: https://www.notion.so/trustmachines/Naming-File-naming-117d74b694c780a8abe6d6d29588aaa5?pvs=4

apps/mobile/src/app/(home)/settings/wallet/index.tsx
```diff
@@ -37,21 +37,25 @@ export default function SettingsScreen() {
         <Box>
           <Divider />
           <Box px="5" pt="5" style={{ paddingBottom: theme.spacing['5'] + bottom }} gap="6">
-            <Cell
+            <Cell.Root
               title={t`Hidden accounts`}
-              subtitle={t`${hiddenAccountsLength} hidden accounts`}
-              Icon={Eye1ClosedIcon}
+              caption={t`${hiddenAccountsLength} hidden accounts`}
```
kyranjamie, line 42: Not from this PR, but what if there is only 1 hidden account?

packages/ui/src/components/cell/cell.native.tsx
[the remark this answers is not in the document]
pete-watters: I don't agree with these name spaced repeated functions 🤷‍♂️ 

If we ever change the design for the switch / radio we will probably need to update them all anyway so it still doesn't make sense. 

Also I don't think we should code favouring things that might happen like this e.g. design changes. We should future proof making sure it's scalable etc.

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -1,108 +1,71 @@
...
-interface SwitchCell {
-  variant: SwitchCellVariant;
-  switchValue: boolean;
-  toggleSwitchValue(): void;
-}
+  const content = icon ? (
+    <Flag
+      img={
+        <Avatar>
+          {cloneElement<IconProps>(icon, {
```
pete-watters, line 35: I still don't get why we have to do this. We already have a `cloneElement` in `Icon` which does the same thing: 
```
export function Icon({ children, ...props }: IconProps) {
  const child = React.Children.only(children) as React.ReactElement;

  return (
    <>
      {React.cloneElement(child, {
        color: token('colors.ink.action-primary-default'),
        ...props,
      })}
    </>
  );
}

```

Is it because that's setting them to `ink.action-primary-default` and you want them to be `ink.background-secondary` without passing it all the time?

fbwoolf: Yeah, I'm not sure why it is nec when a default is already set. Tbh, I think the default color being set here is inaccurate? We don't want the default to be `text.primary`? @edgarkhanzadian that is incorrect? Icons by default are `action-primary-default` which is already applied to all icons.

PR #506, opened 2024-10-03
feat: use lingui custom ids

fbwoolf:
This PR implements custom ids for our translations strings.

Best practice naming for ids used:
https://lokalise.com/blog/translation-keys-naming-and-organizing/

Helpful info for the changes made here refactoring from generated to explicit lingui ids:
https://lingui.dev/guides/explicit-vs-generated-ids

### DIFF
--- a/apps/mobile/src/app/(home)/_layout.tsx
+++ b/apps/mobile/src/app/(home)/_layout.tsx
@@ -44,7 +44,12 @@ export default function StackLayout() {
     />
   );
 
-  function NavigationSettings(title: string = t`Settings`) {
+  function NavigationSettings(
+    title: string = t({
+      id: 'settings.header_title',
+      message: 'Settings',
+    })
+  ) {
     return (
       <SimpleHeader
         insets={insets}
@@ -54,7 +59,12 @@ export default function StackLayout() {
     );
   }
 
-  function WalletNavigationSettings(title: string = t`Settings`) {
+  function WalletNavigationSettings(
+    title: string = t({
+      id: 'settings.header_title',
+      message: 'Settings',
+    })
+  ) {
     return (
       <SimpleHeader
         insets={insets}
@@ -76,7 +86,10 @@ export default function StackLayout() {
               borderRadius="xs"
             >
               <Text variant="label03" color="ink.text-subdued">
-                {t`Testnet`}
+                {t({
+                  id: 'settings.header_network',
+                  message: 'Testnet',
+                })}
               </Text>
             </Box>
           </TouchableOpacity>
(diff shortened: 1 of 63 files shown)


markmhendrickson:
This format looks directionally good to me. If we're basically expressing a route that contains nested [views](https://www.notion.so/trustmachines/87ce4a60cb0c4f0f8722fd8821e55dfe?v=027ba9a8a14f46e5b529263dc787ce43), perhaps we can start tracking IDs for views that are used here and possibly elsewhere as well.

E.g. `settings` for [this view](https://www.notion.so/trustmachines/Configure-app-b70b1bf87c7148e8b73bb2c5e0432761).

That way, while the name / titles of the views may change given human-readability purposes (e.g. settings vs. configuration), we treat the ID as immutable.

image

There _is_ a question here as to whether particular views get nested within different routes. In the case of app settings, that's not really something we'd anticipate. But [the tokens view](https://www.notion.so/trustmachines/Tokens-a344fdf90dc248899b1f98ba39f624cd?pvs=4) may be one that we show contextually as nested from the home view (e.g. simply `tokens` as a route) or from an account (`account.tokens`). However, presumably the copy (e.g. `title`) would be the same regardless of the nesting context.

Perhaps the way to avoid this issue is to simply configure Lingui IDs that indicate the single view in which the copy is displayed, and not the parents. E.g. `bitcoin_unit.cell_title` instead of `settings.display.bitcoin_unit.cell_title`.

And I'm good with calling these (copy?) "ids" or "keys" going forward instead of "tokens" etc. "keys" might be a bit clearer than "ids" since a "key" is a concatenation of (view and copy) IDs?

fbwoolf:
> > I wonder why we need to specify the message in the `t` function 🤔 As far as i know it should just read it from the .po file
> 
> Maybe you know something I don't? I am just reading the docs, like here: https://lingui.dev/guides/message-extraction#macro-usages
> 
> Isn't the `.po` file generated from the code?

EDIT: I've also been referring to: https://lingui.dev/guides/explicit-vs-generated-ids

fbwoolf:
> @fbwoolf looking through it, it seems like we can do this with `t` macro:
> 
> ```
> const message = t({
>   id: "welcome_message"  // Explicit ID without a default message
> });
> ```
> 
I am not seeing this, so you are saying the message is the id? Can you link me to the docs where you see this? I don't think we want that? I am always seeing the message in the `t` object, like here? https://lingui.dev/guides/message-extraction#macro-usages

fbwoolf:
Opening this for review.

kyranjamie (approved):

fbwoolf:
@markmhendrickson going to merge this as-is bc I'm not sure of your status. Now that the ids are in place, I can make any changes you want to the id format after merging in another PR. I want to do some experimenting with Crowdin today, so I'm going to merge this to move things fwd.

markmhendrickson:
No worries, I'll review post-hoc with https://github.com/leather-io/issues/issues/409

PR #515, opened 2024-10-09
Revert "fix(mobile): clearing store race condition, closes LEA-1497"

edgarkhanzadian:
This reverts commit 41c6b0e29595d38eceee2d401424688130899b2b.


kyranjamie (changes requested):
There are changes in here we still want. It's only the lock file that needs reverting. I'll make a PR

PR #520, opened 2024-10-10
refactor: inscription queries

kyranjamie:
This PR introduces a partially complete bitcoin balance feature for the mobile wallet.

For summing balances, this approach works quite well. Note that there are _loads_ more requirements. On extension, we have a lot of filtering going on, that needs to also be replicated here. e.g. if I have a UTXO that has an inscription on it, we should not consider it as a bitcoin balance.

This PR uses an experimental [Cloudflare Worker](https://github.com/leather-io/workers). Suggest we replace this with an actually hosted instance on our Leather cloudflare account.

Further, this PR refactors the inscription queries, removing the recursive look up for inscription balances.

@edgarkhanzadian I will follow up with tests when I finish the feature, I promise 🙏🏼 

cc/ @pete-watters @alexp3y


apps/mobile/src/app/(home)/home/home.tsx
```diff
@@ -17,14 +21,22 @@ export function Home() {
...
+  const balanceQuery = useBitcoinAccountBalances();
+
+  console.log({ combiledBalance: balanceQuery });
+
   return (
     <PageLayout>
       <AccountsWidget accounts={accounts.list} wallets={wallets.list} />
       <TokensWidget tokens={getMockTokens()} totalBalance={mockTotalBalance} />
-      <CollectiblesWidget
+      {/* <CollectiblesWidget
```
pete-watters, line 36: Feel free to just delete this `<CollectiblesWidget` and `<TokensWidget` if you want. I'll do it on my next work anyway. 

It seems we don't **have** to do it now on the homepage so lets remove it. I'll get it working on the account page and then we can re-enable once we have more progress on data aggregation. 

That will also get rid of that `source.uri should not be an empty string` message you were seeing

apps/mobile/package.json
```diff
@@ -62,6 +63,7 @@
     "@stacks/common": "6.13.0",
     "@stacks/transactions": "6.15.0",
     "@stacks/wallet-sdk": "6.15.0",
+    "@tanstack/query-core": "5.59.13",
```
pete-watters, line 66: Is this replacing `@tanstack/react-query` here? Should we remove that or are they for different use cases

fbwoolf (approved):

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
```diff
@@ -1,38 +1,41 @@
...
-      }
+      header={<WidgetHeader title={t({ id: 'tokens.header_title', message: 'My tokens' })} />}
     >
-      {tokens.map(
+      {children}
+    </Widget>
+  );
+}
+
+export function AllAccountBalances() {
```
pete-watters, line 33: Should we create a new folder structure for `balances`? I guess we should organise this into smaller components

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
```diff
@@ -41,16 +44,84 @@ export function TokensWidget({ tokens, totalBalance }: TokensWidgetProps) {
...
         )
       )}
-    </Widget>
+    </>
+  );
+}
+
+export function AccountBalances({ fingerprint, accountIndex }: AccountId) {
+  return (
+    <>
```
pete-watters, line 65: No need to the fragments here

apps/mobile/src/store/settings/utils.ts
```diff
@@ -9,7 +9,7 @@ import {
   FiatCurrency,
 } from '@leather.io/models';
 
-export const defaultNetworkPreferences = ['mainnet', 'testnet', 'signet'] as const;
+export const defaultNetworkPreferences = ['mainnet', 'testnet4', 'signet'] as const;
```
pete-watters, line 12: Just double checking is `testnet4` correct here?

apps/mobile/src/queries/balance/bitcoin-balance.query.ts
```diff
@@ -0,0 +1,87 @@
...
+  );
+}
+
+export function useTotalBitcoinBalanceOfDescriptors(descriptors: string[]) {
+  const queries = useCreateBitcoinAccountUtxoQueryOptions(descriptors);
+  const calcUsdAmount = useCalculateBitcoinFiatValue();
+
+  const combine = useCallback(
+    (
+      results: UseQueryResult<
```
pete-watters, line 48: Maybe we can move this type out of here?

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
```diff
@@ -41,16 +44,84 @@ export function TokensWidget({ tokens, totalBalance }: TokensWidgetProps) {
...
+      tokenName={t`Bitcoin`}
+      chain={t`Bitcoin blockchain`}
+      fiatBalance={fiatBalance}
+      showChain={showChain}
+      availableBalance={availableBalance}
+    />
+  );
+}
+
+export function BitcoinBalance() {
```
pete-watters, line 89: I see our preferred style here is to compose nested components for each token type that fetch their own data. 

I wasn't sure if it's better to fetch all the token balance data and cache it on the homepage first, as we will need that for the total balance anyway.

Maybe this is working the same way but on the `tokens-widget` page it should take the data from the cache

pete-watters (approved):
This looks great to me @kyranjamie nice work. We can create some tasks to add more of the filtering options we need to provide similarly to the extension.

apps/mobile/src/store/settings/utils.ts
kyranjamie: yep

apps/mobile/package.json
kyranjamie: Our services layer might rely on `QueryClient`, but we don't want it to be react-specific and undernearth the react impl is query core, so think we'll need this. Not here for now though so can remove

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
kyranjamie: > I see our preferred style here is to compose nested components for each token type that fetch their own data.

I'm not 100% on this, but the benefit is that it makes splitting requests easier. E.g. if this component is used in a single place, we can load bitcoin balances only. Think lazy load feature etc in future.

Happy for this to change, but think we should audit when we have both currencies in

apps/mobile/package.json
pete-watters: OK thanks, I'm just curious 👍

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
pete-watters: Thanks, I'm happy to go with this and we can assess if needed. It's a nice way of grouping things together more neatly too

apps/mobile/src/queries/balance/bitcoin-balance.query.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      return { availableBalance, fiatBalance };
+    },
+    [calcUsdAmount]
+  );
+
+  return useQueries({ queries, combine });
+}
+
+export function useWalletTotalBitcoinBalance() {
```
pete-watters, line 79: @kyranjamie - I forgot to double check that this calculates for multiple wallets. I think it does as `keychains` contains all accounts (across wallets)

Trying to think up a better name as `useWalletTotalBitcoinBalance` is singular. Maybe a simple change to `useWallets` is good enough?

PR #521, opened 2024-10-10
fix: default security level preference [LEA-1517] [LEA-1518] [LEA-1524]

edgarkhanzadian:
- display "Secure your wallet" screen
- update tests that include secure and insecure account setup

https://github.com/user-attachments/assets/a6154a05-958d-41a8-9acc-56019cf842d6

### DIFF
--- a/apps/mobile/src/hooks/create-wallet.ts
+++ b/apps/mobile/src/hooks/create-wallet.ts
@@ -59,7 +59,7 @@ export function useCreateWallet() {
 
   async function navigateAndCreateWallet() {
     switch (securityLevelPreference) {
-      case 'undefined':
+      case 'not-selected':
         router.navigate(AppRoutes.SecureYourWallet);
         return;
       case 'secure':
--- a/apps/mobile/src/store/settings/settings.ts
+++ b/apps/mobile/src/store/settings/settings.ts
@@ -42,7 +42,7 @@ export const defaultThemePreferences = ['light', 'dark', 'system'] as const;
 
 export type ThemePreference = (typeof defaultThemePreferences)[number];
 export type Theme = Exclude<ThemePreference, 'system'>;
-export type SecurityLevelPreference = 'insecure' | 'secure' | 'undefined';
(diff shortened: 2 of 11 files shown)


apps/mobile/src/store/settings/settings.ts
```diff
@@ -67,7 +67,7 @@ export const initialState: SettingsState = {
   fiatCurrencyPreference: 'USD',
   networkPreference: WalletDefaultNetworkConfigurationIds.mainnet,
   privacyModePreference: 'visible',
-  securityLevelPreference: 'insecure',
+  securityLevelPreference: 'undefined',
```
fbwoolf, line 70: I think this is going to break app auth in settings, why does this need to be `undefined`? Can't we just change the code you initially wrote to handle a default of `insecure`?

fbwoolf: If not, please look at the app auth settings and make sure to cover `undefined` there.

fbwoolf: In the app auth setting, we use a toggle so rn it is just handling 2 states. When a user gets to that point it should be defined though as `secure` or `insecure`? So, maybe ok?

fbwoolf (approved):

apps/mobile/src/store/settings/settings.ts
fbwoolf: Thanks for addressing, lgtm!

apps/mobile/src/hooks/create-wallet.ts
```diff
@@ -59,7 +59,7 @@ export function useCreateWallet() {
 
   async function navigateAndCreateWallet() {
     switch (securityLevelPreference) {
-      case 'undefined':
+      case 'not-selected':
```
pete-watters, line 62: This switch should have type `securityLevelPreference: SecurityLevelPreference` right?

pete-watters (approved):

PR #530, opened 2024-10-12
feat: lock screen

edgarkhanzadian:
- [ ] change splash image to match the new designs
- [ ] change Unlock button styles (might need a few button component changes)
- [ ] probably shouldn't show splash animation when device is in insecure mode and user comes back from the background


apps/mobile/src/app/(home)/developer-console/index.tsx
```diff
@@ -79,7 +79,10 @@ export default function DeveloperConsoleScreen() {
         }}
       >
         <PressableListItem
-          title={t`securityLevelPreference:` + ' ' + settings.securityLevelPreference}
+          title={t`appSecurityLevelPreference:` + ' ' + settings.appSecurityLevelPreference}
```
pete-watters, line 82: Can we write this like:
```
t`appSecurityLevelPreference: ${settings.appSecurityLevelPreference} `
```

PR #534, opened 2024-10-15
feat: app layer poc

alexp3y:
WIP and meant for illustrative purposes. Demonstrates adding inversifyJS to manage and organize app-layer services using an IoC container. Implements balance fetching and demos some utxo fetching logic. 

Services are made available to mobile components via react context.

@kyranjamie @pete-watters

### DIFF
--- a/packages/services/src/inversify.config.ts
+++ b/packages/services/src/inversify.config.ts
@@ -1,0 +1,150 @@
+import { Container } from 'inversify';
+
+import { DefaultNetworkConfigurations, defaultNetworksKeyedById } from '@leather.io/models';
+
+import { BlockbookApiClient, createBlockbookApiClient } from './api/blockbook/blockbook.client';
+import {
+  MempoolSpaceApiClient,
+  createMempoolSpaceApiClient,
+} from './api/mempool-space/mempool-space.client';
+import {
+  AccountBalanceService,
+  createAccountBalanceService,
+} from './balances/account-balance.service';
+import { BtcBalanceService, createBtcBalanceService } from './balances/bitcoin/btc-balance.service';
+import { StxBalanceService, createStxBalanceService } from './balances/stacks/stx-balance.service';
+import { QueryClientService } from './cache/query-client.service';
+import { NetworkSettingsService } from './settings/network-settings.service';
+import { UtxoService, createUtxoService } from './utxos/utxo.service';
+
+// Symbols for Service Bindings
+const Services = {
+  QueryClientService: Symbol.for('QueryClientService'),
+  NetworkSettingsService: Symbol.for('NetworkSettingsService'),
+  BlockbookApiClient: Symbol.for('BlockbookApiClient'),
+  MempoolSpaceApiClient: Symbol.for('MempoolSpaceApiClient'),
+  AccountBalanceService: Symbol.for('AccountBalanceService'),
+  BtcBalanceService: Symbol.for('BtcBalanceService'),
+  StxBalanceService: Symbol.for('StxBalanceService'),
+  UtxoService: Symbol.for('UtxoService'),
+};
+
+let currentNetworkContainer: Container;
+type NetworkSpecificContainers = Record<string, Container>;
+const networkSpecificContainers: Partial<NetworkSpecificContainers> = {};
+
+/**
+ * Initializes App Services DI containers
+ */
+export function configureServiceContainers(
+  defaultNetworkSettingsService: NetworkSettingsService,
+  queryClientService: QueryClientService
+): Container {
+  if (!currentNetworkContainer) {
+    currentNetworkContainer = new Container();
+    registerDependencies(
+      currentNetworkContainer,
+      defaultNetworkSettingsService,
+      queryClientService
+    );
+    for (const networkId of Object.keys(defaultNetworksKeyedById)) {
+      const networkSpecificContainer = new Container();
+      registerDependencies(
+        networkSpecificContainer,
+        {
+          getConfig() {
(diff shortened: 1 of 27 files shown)


kyranjamie:
Thanks for putting together Alex, this is great. A really smart outline of how we can structure tonnes of complex logic that is currently very coupled to the React frontend. 

This service architecture set up is something you can predominantly own, so if you're adamant this is the way to go, I'm supportive. Some questions on the choices though:

Invesifyjs implements DI/IoC which are object-oriented patterns. The rest of our codebase largely follows functional patterns: the pure functions of the domain layer, our state management etc. React itself is functional by nature.

While the services layer is encapsulated, and doesn't necessarily expose the OO patterns externally, I wonder if using a mix of patterns across our stack might confuse developers. Here we use classes and DI with inversify, but there we use pure fns and avoid classes etc.

Do you think it would be possible get the architectural benefits of a DI pattern like this, but using a more functional style with HOFs? Skipping inversify, which does seems to be out of maintenance. What are the trade offs of a DI library like this, versus implementing the same thing manually?

These are pretty similar in as far as testability, and ability to control/override dependencies going in to a service. In the former, for example, we take care of instantiating the service, something Inversify would take care of in latter example, but without having to wire up a DI framework.

```ts
// Functional
function createBalanceService({ utxoService }) {}
// Inversify
class BalanceService {
  constructor(public utxoService) {}
}
```

Curious to hear your thoughts. And interested to get input from the rest of team when you're ready.

alexp3y:
Thanks for the feedback @kyranjamie. Definitely not adamant about this pattern or the Inversify framework specifically. The driving force behind the design here is dependency management. IoC container frameworks are generally written using class annotations and, as you mentioned, its usage would be largely isolated here in the service layer, which is why I went with an object-oriented pattern.

I do agree though that mixing in OOP to a largely functional codebase is not ideal for the exact reasons you mentioned. 

Your implementation of `BalanceService` in [API Layer thoughts](https://www.notion.so/trustmachines/API-layer-concerns-11ed74b694c78062b697f6ac6a80df76?pvs=4) does a great job of demonstrating how we can approach service objects using functional patterns.

I could definitely be missing something obvious but the problem I keep coming back to is dependency management.

I've tried to restate the problem a bit more concretely along with some thoughts on how we could approach it using a more functional style [here](https://www.notion.so/p3y/Dependency-Management-1214e431827780c69e63e3eab3677ae6?pvs=4).

alexp3y:
@kyranjamie Here’s another iteration that I think strikes a good balance between the functional style we’ve discussed and the necessary object-oriented (OOP) aspects. I considered approaches like function injection and HOF, but ultimately found that these would require sacrificing some key dependency injection (DI) functionality that we need.

I’m still using InversifyJS, but I’ve worked to minimize reliance on classes as much as possible. By following basic conventions—keeping objects stateless aside from their injected dependencies and extracting logic into pure or utility functions wherever possible—I think we can leverage the benefits of a DI/IoC framework while keeping the codebase aligned with a functional style.

pete-watters:
Nice work @alexp3y 👍 this is coming along nicely. 

I think it's OK to have a slight mixture of OOP/ functional code when we need to. We can endeavour to use functional in the app itself but I think it's OK in isolated cases to use the best fit to make things easier to manage

alexp3y:
Removed the services context in favor of a simplified initialization function. Moves the responsibility of defining the app service API to the services package itself. The DI container(s) and service resolver functions are held in module-scope and can be directly imported by components / hooks.

alexp3y:
Got this working without classes or decorators. All services can now be represented by HOFs - e.g. `createBalanceService(dep1, dep2)`, with dependencies resolved by the DI container.

The HOF and its parameters need to be registered by symbol using `.toDynamicValue`, which requires a bit more boilerplate in `inversify.config.ts`, but the service implementations are now much cleaner and in-line with the functional codebase style. All previous DI functionality is preserved.

packages/services/src/inversify.config.ts
```diff
@@ -0,0 +1,146 @@
...
+import { DefaultNetworkConfigurations, defaultNetworksKeyedById } from '@leather.io/models';
+import {
+  AccountBalanceService,
+  createAccountBalanceService,
+} from 'balances/account-balance.service';
+import { createStxBalanceService, StxBalanceService } from 'balances/stacks/stx-balance.service';
+import { QueryClientService } from 'cache/query-client.service';
+
+// Symbols for Service Bindings
+const TYPES = {
```
kyranjamie, line 21: I know docs call it this way but can we call it `Services` instead

packages/services/src/inversify.config.ts
```diff
@@ -0,0 +1,146 @@
...
+  QueryClientService: Symbol.for('QueryClientService'),
+  NetworkSettingsService: Symbol.for('NetworkSettingsService'),
+  BlockbookApiClient: Symbol.for('BlockbookApiClient'),
+  MempoolSpaceApiClient: Symbol.for('MempoolSpaceApiClient'),
+  AccountBalanceService: Symbol.for('AccountBalanceService'),
+  BtcBalanceService: Symbol.for('BtcBalanceService'),
+  StxBalanceService: Symbol.for('StxBalanceService'),
+  UtxoSerivce: Symbol.for('UtxoSerivce'),
```
kyranjamie, line 29: typo

alexp3y:
Removed all class and annotation-related packages and plugins. Added example unit tests.

fbwoolf:
Really love the changes being proposed here, I think this approach is going to make things so much easier to work with moving forward. 👍

PR #535, opened 2024-10-15
feat: add all web icons story

alter-eggo:
This pr adds all web icons story so we can see what web icons we have in the codebase
Screenshot 2024-10-15 at 17 24 21

### DIFF
--- a/packages/ui/src/icons/all-icons.web.stories.tsx
+++ b/packages/ui/src/icons/all-icons.web.stories.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+
+import { Meta } from '@storybook/react';
+
+import * as Icons from './index.web';
+
+const meta: Meta = {
+  title: 'Icons/All Icons',
+  component: () => <div />,
+  parameters: {
+    controls: { hideNoControlsWarning: true },
+  },
+};
+
+export default meta;
+
+export function AllIcons() {
+  return (
+    <div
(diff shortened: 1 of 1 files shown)


fbwoolf:
A little confused bc this was done a long time ago here: https://github.com/leather-io/mono/tree/dev/packages/ui/src/icons/docs 🤔

pete-watters (approved):
Thanks for doing this @alter-eggo 🎉

packages/ui/src/icons/all-icons.web.stories.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+import { Meta } from '@storybook/react';
+
+import * as Icons from './index.web';
```
pete-watters, line 5: Do you think we could do this in the `extension` instead? 

That storybook gets published so the designers can see it. Since moving icons to the monorepo they don't have a way to see them without building the UI package.

We discussed composing storybooks between `mono` + `extension` but never did it

pete-watters: Maybe we can try to export `all-icons.web.stories.tsx` from `mono`? 

Then we could just load the story into the `extension` storybook? 🤔  

Not something for now but maybe for the future

kyranjamie (approved):

PR #539, opened 2024-10-15
fix: mobile headers, closes LEA-628

fbwoolf:
This PR refactors our mobile headers so they are a bit easy to understand at this point in coordination with designs. Here, we now have headers matching the three designs for:

1. Header: Home => `<HomeHeader />` - has the Leather Logo and icon options on the right
2. Header: Title => `<TitleHeader />` - has a center title and a back button (optional right element)
3. Header: Naked => `<NakedHeader />` - has only the back button
4. `<TabsHeader />` - when tabs are used as the bottom element

I organized our top level _layout Stack, but maybe we can go further at some point and nest _layout Stacks ...I'm not sure if this is necessary or desired?

I renamed some things here for clarity, and removed unused components. ~~I'm not sure why we were just rendering `<Home />` in the initial index route, rather than just moving that code ...it seems fine to remove `<Home />` as it appeared to just be an extra component.~~ Nvm, this is needed, but not 100% sure why it is setup this way. 🤔 

*Note, I will hook up the `NetworkBadge` to our network setting in my next issue: https://linear.app/leather-io/issue/LEA-1531/fix-display-of-network-badge-in-navigation


apps/mobile/src/app/(home)/receive.tsx
```diff
@@ -1,17 +1,17 @@
-import { HeaderAvoidingContainer } from '@/components/headers/containers/header-avoiding-container';
+import { ContainerWithHeaderOffset } from '@/components/container-with-header-offset';
```
pete-watters, line 1: Nice renaming here 👍

packages/ui/src/components/cell/cell.native.tsx
```diff
@@ -26,7 +26,8 @@ interface CellProps extends PressableProps<Theme> {
 export function Root({ caption, icon, title, children, ...props }: CellProps) {
   const theme = useTheme<Theme>();
   const itemLayout = <ItemLayout actionIcon={children} captionLeft={caption} titleLeft={title} />;
-  const defaultColor: string = theme.colors['ink.text-primary'];
+  // This is the default color of all icons, why do we need this?
```
pete-watters, line 29: I'm not sure about this either. We set the color to `'ink.action-primary-default` in `<Icon` so I think thats sufficient?

pete-watters (approved):

apps/mobile/src/components/container-with-header-offset.tsx
```diff
@@ -1,17 +1,18 @@
...
   const contentOffsetTop = top + HEADER_HEIGHT;
+
   return (
     <Box
-      flex={1}
       backgroundColor="ink.background-primary"
+      flex={1}
       style={{ paddingTop: contentOffsetTop }}
       {...props}
     />
```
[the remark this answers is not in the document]
fbwoolf: I was thinking this might be outdated, thanks for flagging! I'll remove it.

packages/ui/src/components/cell/cell.native.tsx
fbwoolf: In removing, I could also remove our use of `cloneElement` here.

pete-watters: I think that's good if we can remove `cloneElement` too as in this case it's not necessary

PR #542, opened 2024-10-16
feat: load stx data in account page

pete-watters:
This PR:

- [x] adds the query package to the mobile app and sets it up
- [x] fetches `stx` token data 
- [x] wires up account level balance data and token widget
- [x] aggregates multi wallet balance data for `stx` 
- [x] TODO - refactor usage of `useGetAccountStxBalance` to not break hooks rules
- [x] integrated with Kyrans BTC balance work 
- [ ] do more work to merge queries 
- [ ] investigate jankiness and incorrect balances  

Here is a demo where I have loaded Leather Dev wallet and one of my own wallets to the mobile app. The data is real but you can observe some jankiness and incorrect data when going into accounts 

https://github.com/user-attachments/assets/a8245830-886e-4ca3-8a00-328c06be8a55


apps/mobile/src/app/(home)/account/[account]/index.tsx
```diff
@@ -12,11 +15,17 @@ const configureAccountParamsSchema = z.object({
 
 export default function AccountScreen() {
   const params = useLocalSearchParams();
+
   const { fingerprint, account: accountIndex } = configureAccountParamsSchema.parse(params);
+  const account = useAccountByIndex(fingerprint, accountIndex);
```
Comment (pete-watters) [inline apps/mobile/src/app/(home)/account/[account]/index.tsx:20]: I need to get `account` here to pass to `useGetTokensList`. 

Should I still use `<AccountLoader` below when passing to `AccountLayout`

apps/mobile/src/app/(home)/account/[account]/account.layout.tsx
```diff
@@ -1,35 +1,35 @@
...
     <PageLayout>
       <AccountOverview
         Icon={getAvatarIcon(account.icon)}
-        heading={<Balance balance={mockTotalBalance} variant="heading02" />}
+        heading={<Balance balance={balance} variant="heading02" />}
         caption={account.name}
       />
-      <TokensWidget tokens={getMockTokens()} totalBalance={mockTotalBalance} />
+      <TokensWidget tokens={tokens} totalBalance={balance} />
+      {/* TODO - re-enable widgets when real data available */}
```
Comment (pete-watters) [inline apps/mobile/src/app/(home)/account/[account]/account.layout.tsx:27]: @fbwoolf - if you want to help and start working on this, coming soon is [adding stacks NFTs](https://linear.app/leather-io/issue/LEA-1526/load-account-level-stacks-nfts)

We will need to load them with a query and pass them through `serializeCollectibles`

apps/mobile/src/components/widgets/accounts/accounts-widget.tsx
```diff
@@ -53,21 +52,18 @@ export function AccountsWidget({ accounts, wallets }: AccountsWidgetProps) {
               sheetRef={sheetRef}
               sheet={<AccountSelectorSheet sheetRef={sheetRef} />}
             />
-            {hasWallets && <Balance balance={balance} variant="heading03" />}
+            {hasWallets && <Balance balance={totalBalance} variant="heading03" />}
           </Box>
         }
       >
+        {/* FIXME  stop using mock accounts and use the real ones
```
pete-watters, line 59: I'm still using `mockAccounts` as I need to update the actual accounts types. 

I am extending `AccountStore` to add extra data we need

```
export interface MockedAccount extends AccountStore {
  type: WalletStore['type'];
  balance: Money;
  fingerprint: string;
  accountIndex: number;
}
```

apps/mobile/src/app/(home)/account/[account]/account.layout.tsx
Comment (fbwoolf) [reply apps/mobile/src/app/(home)/account/[account]/account.layout.tsx:27]: Think I'm assigned to work on push notifications atm. I'll jump in here if that work gets put on hold.

fbwoolf:
Lgtm so far here @pete-watters 👍

apps/mobile/src/hooks/balances/use-stx-balances.ts
```diff
@@ -0,0 +1,41 @@
...
+  } = useStxCryptoAssetBalance(stxAddress ?? '');
+
+  return balance ? balance.totalBalance : createMoney(0, 'STX');
+}
+
+function useGetAccountsStxBalance(accounts: MockedAccount[]): Money {
+  if (!accounts.length) return createMoney(0, 'STX');
+
+  return accounts.reduce(
+    (total, account) => sumMoney([total, useGetAccountStxBalance(account)]),
```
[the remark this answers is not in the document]
pete-watters: I changed this to avoid the error: ` "useGetAccountStxBalance" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks` 

We have a lot of different hooks, calling each other and without refactoring a lot of code I can solve this issue by simply using a React function component - AKA calling the same function `GetAccountStxBalance` instead of `useGetAccountStxBalance`. 

I know we have rules about naming so I'm not sure if the team agrees on this but it seems like a reasonable way to avoid this error

packages/query/package.json
```diff
@@ -52,8 +52,9 @@
     "axios": "1.7.4",
     "bignumber.js": "9.1.2",
     "lodash.get": "4.4.2",
-    "p-queue": "8.0.1",
+    "p-queue": "7.4.1",
     "url-join": "5.0.0",
+    "uuid": "10.0.0",
```
pete-watters, line 57: I need to use `uuid` as `crypto.randomID` seems to be not working on React Native (I tried polyfills but this is easier) 

If I don't do this queries throw the error:
```
"failureReason": [TypeError: crypto.randomUUID is not a function (it is undefined)]
```

packages/query/package.json
```diff
@@ -81,6 +83,11 @@
   "peerDependencies": {
     "react": "*"
   },
+  "pnpm": {
+    "overrides": {
+      "p-queue": "7.4.1"
```
pete-watters, line 88: I am rolling back `p-queue` a version and setting it to `7.4.1` as otherwise queries will throw an error for undefined `options.signal?.throwIfAborted` in React Native code. 

I tried to fix this to no avail and discovered the queries worked OK if I bypass the priority queue. 

I tracked back their releases and found they introduced this change [here](https://github.com/sindresorhus/p-queue/compare/v%20...v8.0.0) . On their README it says the project is feature complete and that they don't plan any further development / offer support so I think it's just easier to roll back for now.

TBD - I need to test that this will work in the `extension`

apps/mobile/src/app/(home)/account/[account]/index.tsx
```diff
@@ -1,26 +1,26 @@
...
-  account: z.string().transform(value => Number(value)),
+  accountId: z.string(),
 });
 
 export default function AccountScreen() {
   const params = useLocalSearchParams();
 
-  const { fingerprint, account: accountIndex } = configureAccountParamsSchema.parse(params);
+  const { accountId } = configureAccountParamsSchema.parse(params);
+  const { fingerprint, accountIndex } = destructAccountIdentifier(accountId);
```
Comment (pete-watters) [inline apps/mobile/src/app/(home)/account/[account]/index.tsx:19]: @edgarkhanzadian - I changed my code to use the real `AccountStore` and started using `destructAccountIdentifier`. Checking back to your [original PR](https://github.com/leather-io/mono/pull/407/files) and I see that we increment `accountId` when we  use `destructAccountIdentifier` but I couldn't be sure that we increment it on creation.

Do we need to double check that it's working? I just want to make sure we aren't choosing the wrong account.

apps/mobile/src/components/widgets/accounts/accounts-widget.tsx
```diff
@@ -66,8 +63,7 @@ export function AccountsWidget({ accounts, wallets, totalBalance }: AccountsWidg
               router.navigate({
                 pathname: AppRoutes.Account,
                 params: {
-                  fingerprint: account.fingerprint,
-                  account: account.accountIndex,
+                  accountId: account.id,
```
pete-watters, line 66: @edgarkhanzadian : I changed this code to pass the `accountId` and then use `destructAccountIdentifer` to get the `fingerprint` + `accountIndex` on arrival at `AppRoutes.Account`. 

Should I refactor `[account-selector-sheet.tsx](https://github.com/leather-io/mono/pull/407/files#diff-c51bb2ab3f10a009531e45d8e5b7fd372b2e54434dafa8925304560c03352c63)` to do that also? 

It destructs the identifier then passes them as separate params so I'm not sure whats best
```
 const onAccountPress = useCallback(
    (accountId: string) => {
      const { fingerprint, accountIndex } = destructAccountIdentifer(accountId);
      modalRef.current?.close();
      router.navigate({
        pathname: APP_ROUTES.WalletWalletsSettingsConfigureAccount,
        params: { wallet: fingerprint, account: accountIndex },
      });
    },
    [router, modalRef]
  );
```

apps/mobile/src/hooks/balances/use-stx-balances.ts
```diff
@@ -0,0 +1,41 @@
...
+
+interface StxBalance {
+  stxBalance: Money;
+  stxBalanceUsd: Money;
+}
+
+function useGetAccountStxBalance(account: MockedAccount): Money {
+  const signers = useStacksSigners().fromAccountIndex(account.fingerprint, account.accountIndex);
+  // TODO - ask Kyran if the magic number 0 is correct - signers always returns an array
+  const stxAddress = signers[0]?.address;
```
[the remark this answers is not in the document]
kyranjamie: There will only ever be 0 or 1 Stacks addresses per account.

Because this is filtering an array, it returns an array of a single item, but yeah it isn't nice to consume this way. Could this component be wrapped in a loader, such that it accepts the stacks details as a prop?

apps/mobile/src/hooks/balances/use-stx-balances.spec.ts
```diff
@@ -0,0 +1,61 @@
+import { AccountStore } from '@/store/accounts/utils';
+
+import { createMoney } from '@leather.io/utils';
+
+import { useStxBalance } from './use-stx-balances';
+
+// Mock the dependencies
+jest.mock('@leather.io/query');
```
kyranjamie, line 8: Why are we using jest? We're using vitest everywhere else

apps/mobile/src/components/widgets/accounts/accounts-widget.tsx
pete-watters: OK so maybe I should revert my other code to keep passing `fingerprint` + `accountIndex`? And ignore `destructAccountIdentifer`

apps/mobile/src/app/app-router.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export function AppRouter() {
+  const { whenTheme } = useSettings();
+  return (
+    <Box backgroundColor="ink.background-secondary" flex={1}>
+      <StatusBar barStyle={whenTheme({ dark: 'light-content', light: 'dark-content' } as const)} />
+      <Slot />
+    </Box>
+  );
+}
```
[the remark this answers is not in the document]
pete-watters: OK thanks, I'll move them back where they were into `app/_layout`

apps/mobile/src/components/widgets/tokens/get-chain-name.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  it('should handle empty input', () => {
+    expect(getChainName('')).toBe('');
+  });
+
+  it('should handle undefined input', () => {
+    expect(getChainName(undefined as unknown as string)).toBe(undefined);
+  });
+});
```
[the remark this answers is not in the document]
pete-watters: AI 😂

apps/mobile/src/hooks/balances/use-stx-balances.spec.ts
pete-watters: I just generated these tests and hadn't seen. I will refactor to use `vitest` 👍

apps/mobile/src/hooks/balances/use-stx-balances.ts
```diff
@@ -0,0 +1,42 @@
...
+  useStxCryptoAssetBalance,
+} from '@leather.io/query';
+import { baseCurrencyAmountInQuote, createMoney, sumMoney } from '@leather.io/utils';
+
+interface StxBalance {
+  stxBalance: Money;
+  stxBalanceUsd: Money;
+}
+
+function GetAccountStxBalance(accountId: string): Money {
```
kyranjamie, line 17: I thought this was a component but indeed think we shouldn't do this, as it's not really how we're supposed to use hooks and will probably cause headaches later.

What's the challenge you're facing here @pete-watters?

pete-watters: See comment here: https://github.com/leather-io/mono/pull/542#discussion_r1806573517 

Basically we kind of use hooks for everything and this is proving difficult in this instance where I want to:
- map over the accounts
- fetch a singular account balance using account ID 

I need to use hooks to get account balance - `useStacksSigners` and `useStxCryptoAssetBalance`. 

If I try and do that and use a `hook` for this function AKA `useGetAccountStxBalance`, I get a lint error
```
"useGetAccountStxBalance" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks
```

The functionality works fine but I hit this error. 

By simply changing the name of the function to be `GetAccountStxBalance` it technically becomes a 'React Function component` which returns a `Money` object rather than JSX and that means I can use it inside of `reduce` as I wish

```
function useGetAccountsStxBalance(accounts: AccountStore[]): Money {
  if (!accounts.length) return createMoney(0, 'STX');

  return accounts.reduce(
    (total, account) => sumMoney([total, GetAccountStxBalance(account.id)]),
    createMoney(0, 'STX')
  );
}
```

I am open to doing further work and refactors to try and avoid doing this but I had tried starting down that path and ended up just adding more and more hooks and still having the same issue.

apps/mobile/src/app/app-router.tsx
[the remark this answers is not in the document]
pete-watters: I decided to move them into `src/common`, does that work? I based it on the fact we keep `ThemeProvider` in `src/app/common` in the extension. 

IMHO `components/` folder should really only be for dumb shared components, not connected stuff.

apps/mobile/src/components/widgets/tokens/get-chain-name.spec.ts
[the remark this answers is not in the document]
pete-watters: Credit really goes to @fbwoolf for putting me on to Cursor IDE. 

It's like having a personal assistant, but one who is better at coding than me

apps/mobile/src/hooks/balances/use-stx-balances.ts
pete-watters: Thanks @edgarkhanzadian 👍 I can take a deeper look into more refactors there. 

In the interest of getting some real balance data on screen I was compromising for a fake functional component

apps/mobile/src/hooks/balances/use-stx-balances.ts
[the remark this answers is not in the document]
pete-watters: Looking into this and I see now that `useStacksSigners` shares `descriptorKeychainSelectors` with `useBitcoinAccounts` (maybe that should be `useBitcoinSigners`?). 

Rather than refactor that I think it's better to get the address from stacks signers in a cleaner way so I have changed it to this:
```
  const [stacksSigner] = signers;
  const stacksAddress = stacksSigner?.address;
```

That way it can get the first element of the array (in which there is always 0 or 1) without using a magic number

apps/mobile/src/hooks/balances/use-stx-balances.ts
kyranjamie: Yep, exactly. `useQueries` is how you're able to dynamically create an array of individual queries that can then be summed with `combine`

packages/query/src/stacks/balance/account-balance.hooks.ts
```diff
@@ -43,6 +43,81 @@ export function createStxCryptoAssetBalance(
   };
 }
 
+export function useStxBalancesQueries(addresses: string[]) {
+  const client = useStacksClient();
+  const network = useCurrentNetworkState();
+
+  // to get these correctly more refactoring and combining of queries is needed
+  // this time in useMempoolTxsInboundBalance + useMempoolTxsOutboundBalance
+  const inboundBalance = createMoney(0, 'STX');
```
pete-watters, line 52: To implement this properly it seems like I also will need to refactor `useMempoolTxsInboundBalance` and `useMempoolTxsOutboundBalance` in a similar way so that they can run `useQueries` and return combined totals if needed

pete-watters:
Moving this back to `DRAFT` as I will have some further work and cleanup to do here. 

The good news is that [I got a POC working](https://github.com/leather-io/mono/pull/542/commits/70eba83b42c3acc60a5c3a6902a26eeddf7b09f3) that uses `useQueries` to fetch multiple address balance data and calculate the totals.

Looking at my logs and I am getting matching data for both approaches:
```
 LOG  totalBalance {"amount": "112230.08114112387", "decimals": 2, "symbol": "USD"}
 LOG  combinedBalancesUsd {"amount": "112230.08114112386443284170870124", "decimals": 2, "symbol": "USD"}
```

I am finding the query files a bit tricky to work with so will perhaps do some refactoring there to split things up. 

Remaining work here:
- refactor more hooks - `useMempoolTxsInboundBalance`  + `useMempoolTxsOutboundBalance` to use `useQueries` so we can take that into account when calculating `StxCryptoAssetBalance` to match the previous use case in `useStxCryptoAssetBalance`
- clean up current implementation 
- investigate deprecating singular `useStxCryptoAssetBalance` and just using `useStxCryptoAssetsBalance` with a single element array

apps/mobile/src/components/widgets/tokens/types.ts
```diff
@@ -0,0 +1,9 @@
+import { Money } from '@leather.io/models';
+
+export interface Token {
+  availableBalance: Record<string, Money>;
```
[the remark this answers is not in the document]
pete-watters: I was just stubbing together something based on the UI. Maybe we should rename the whole thing to `Balance`?

apps/mobile/src/components/widgets/tokens/use-get-tokens-list.ts
```diff
@@ -0,0 +1,29 @@
...
+  return [
+    {
+      chain: t`Stacks blockchain`,
+      availableBalance: {
+        availableBalance: stxBalance,
+      },
+      fiatBalance: stxBalanceUsd,
+      ticker: 'stx',
+      tokenName: t`Stacks`,
+    },
```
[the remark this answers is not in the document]
pete-watters: Open to totally changing this. I was just making some kind of object to render something in the UI. 

So you are thinking that `<TokensWidget` should accept children that are unique to their data? And that those components should be responsible for fetching their own data?

For example:
```export function TokensWidget({ tokens, totalBalance }: TokensWidgetProps) {
  return (
    <Widget...        />
      <BtcBalance /> 
       <Sip10Balance />
   </Widget>
      }
    >
   ... 

```

apps/mobile/src/hooks/balances/use-stx-balances.ts
pete-watters: Changed this to `useQueries` but more work to follow. See here - https://github.com/leather-io/mono/pull/542#issuecomment-2431713706

apps/mobile/src/hooks/balances/use-stx-balances.ts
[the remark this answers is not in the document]
pete-watters: This is outdated and replaced by `useQueries` - https://github.com/leather-io/mono/pull/542#issuecomment-2431713706

apps/mobile/src/components/widgets/tokens/use-get-tokens-list.ts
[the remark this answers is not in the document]
fbwoolf: Has a decision been made to move away from using the models we already have for this? Is that part of the work Alex is doing? We did a lot of work in the extension to handle mapping data to our shared balance models.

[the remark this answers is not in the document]
pete-watters: I just made this type for `TokensWidget` based on what we are actually showing in the UI. We can use the same types created for models in balance if we want. 

I didn't work on any of that stuff so I didn't know they were even there. It seems like we will need to mix these types anyway and also add `chain` which needs to be in the UI. 

What I've done here is basically remove `apps/mobile/src/mocks/tokens.mocks.tsx` and trim it down a bit to feed it actual balance data. 

I was hoping to get advice on my approach here: https://github.com/leather-io/mono/pull/542#issuecomment-2431713706 which I also [asked on Slack](https://trustmachines.slack.com/archives/C05LAP952E7/p1729696101955829) but all instead we are bikeshedding about types and component structure of already existing code that is not yet completed anyway (as data not connected)

apps/mobile/src/app/(home)/account/[account]/account.layout.tsx
```diff
@@ -1,35 +1,35 @@
...
   return (
     <PageLayout>
       <AccountOverview
         Icon={getAvatarIcon(account.icon)}
-        heading={<Balance balance={mockTotalBalance} variant="heading02" />}
+        heading={<Balance balance={balance} variant="heading02" />}
         caption={account.name}
       />
-      <TokensWidget tokens={getMockTokens()} totalBalance={mockTotalBalance} />
+      <TokensWidget tokens={tokens} totalBalance={balance} />
```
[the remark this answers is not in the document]
Comment (pete-watters) [reply apps/mobile/src/app/(home)/account/[account]/account.layout.tsx:26]: All I am changing here is the source of props provided

kyranjamie:
To your questions above, I think refactoring the hooks is a good move. In that refactoring, let's just make sure that the mobile app only imports the query object creator from the query package, and we do the "tying together" with RQ in the mobile app.

apps/mobile/src/components/widgets/tokens/use-get-tokens-list.ts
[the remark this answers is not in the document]
kyranjamie: These are valid questions we shouldn't pass off as bike shedding quite so quickly.

I'm pushing for us to make progress faster on the wallet by cutting scope, but when it comes to implementation, this is our chance to build something properly. Discussions very well might relate to how we compose components, and how we reuse data models from the extension.

IMO, what we're doing right now is a kind of unique situation. We're working on greenfield code, of something we've already built. We're able to take our experience building it first time around, and not make the same mistakes.

What are those mistakes, and how does this implementation improve on them?

This PR has already gone from a React hack, naming a function as a component, to a reputable implementation using useQueries. I think that was time well spent.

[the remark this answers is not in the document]
pete-watters: Understood 👍 These points are important but the most urgent thing is getting the correct data then improving on it which my original hack achieved. 

I am happier knowing I can keep going in [this direction](https://github.com/leather-io/mono/pull/542#issuecomment-2434470770) now and then fix the types at the end when I re-open for proper review.

kyranjamie:
> If you think it's better I can do the query re-work inside of the mobile app and only import the query object creator.

Thinks that's a good move. Then, for example, we can create a `useQueries` hook, to create a mix of both STX and BTX balances, that we compose with combine & the market data hooks to get the total?

pete-watters:
> > If you think it's better I can do the query re-work inside of the mobile app and only import the query object creator.
> 
> Thinks that's a good move. Then, for example, we can create a `useQueries` hook, to create a mix of both STX and BTX balances, that we compose with combine & the market data hooks to get the total?

🫡  I'll add `useQueries` hooks inside the mobile app and we can use them to create a better version of `useTotalBalance(` hook. 

Once we are happy with it we can consider moving it to the query package if we want it on the `extension`.

kyranjamie:
It's a WIP still haven't organised folders, but see this PR for bitcoin balance hooks/structure. Thinking through how to have a reusable TokenWidget list than compiles the total total or account-specific total

apps/mobile/src/store/accounts/utils.ts
```diff
@@ -8,10 +9,16 @@ export const accountStoreSchema = z.object({
...
+  accountIndex: z.number().optional(),
 });
 export interface AccountStore {
   id: string;
   icon: string | AvatarIcon;
   name: string;
   status: AccountStatus;
+  type?: WalletStore['type'];
+  fingerprint?: string;
+  accountIndex?: number;
```
kyranjamie, line 23: This data is encoded in the ID, so we don't need to actually add write it to the store, we can infer it when we read from the store.

This is already done in `initalizeAccount`

pete-watters: Thanks, I'll remove. 

I had been adding this to account to lookup the address but have since replaced it with `signers.list`

apps/mobile/src/components/widgets/tokens/use-get-tokens-list.ts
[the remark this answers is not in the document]
fbwoolf: Fwiw, my question was to make your life easier, not harder. I was suggesting that there is possibly existing code to ref/reuse from the extension ...but I wasn't sure if you were intentionally moving away from it.

kyranjamie:
For the sake of scoping down this task to get it merged, perhaps you could focus only on getting total stacks account balances summed, and not worrying about `useTotalBalance` hook etc here. Doing this later when the bitcoin balances are merged too.

apps/mobile/src/features/accounts/balances/hooks/mempool/use-mempool-txs-balance.ts
```diff
@@ -0,0 +1,44 @@
...
+function getPendingTxs(confirmedTxs: Transaction[], pendingTxs: MempoolTransaction[]) {
+  return pendingTxs.filter(tx => {
+    if (confirmedTxs.some(confirmedTx => confirmedTx.nonce === tx.nonce)) {
+      return false;
+    }
+    return tx.tx_type === 'token_transfer';
+  }) as unknown as MempoolTokenTransferTransaction[];
+}
+
+export function useMempoolTxsBalance(addresses: string[]) {
```
pete-watters, line 31: I did some recreation to refactor the code we already had where `inbound` and `outbound` transactions separated. 

Here we use the same code for both but apply filtering to the `pendingTxs` in the same place to provide `inboundBalance` + `outboundBalance` for all supplied addresses

apps/mobile/src/features/accounts/balances/hooks/use-stx-balances.ts
```diff
@@ -0,0 +1,36 @@
...
+
+function useGetStxBalanceByAddresses(addresses: string[]): StxBalances {
+  const { totalData, queries } = useStxBalancesQueries(addresses);
+
+  if (!addresses || addresses.length === 0)
+    return { totalStxBalance: createMoney(0, 'STX'), accounts: [] };
+
+  return {
+    totalStxBalance: totalData,
+    accounts: queries, // TODO as yet unused but perhaps useful
```
pete-watters, line 27: This `accounts` contains the pre-aggregated individual account balance data. 

I'm not sure yet if this is useful here. Perhaps it is to show account specific data on account cards.

We can also use the same query to call `useGetStxBalanceByAddresses` passing a single address if we want instead

apps/mobile/src/features/accounts/balances/hooks/stacks/use-stacks-pending-transactions.ts
```diff
@@ -0,0 +1,66 @@
...
+    () => ({
+      totalData: queries
+        .map(result => result.data)
+        .reduce((acc, result) => {
+          if (result?.results) {
+            return [...acc, ...result.results];
+          }
+          return acc;
+        }, [] as MempoolTransaction[]),
+      pending: queries.some(result => result.isPending),
```
pete-watters, line 30: I will rename `pending:` here as `isPending` as it's a bit confusing. This is to indicate if querys are pending

packages/query/src/stacks/balance/index.ts
```diff
@@ -0,0 +1,11 @@
...
+import { createStxCryptoAssetBalance } from './create-stx-crypto-asset-balance';
+import { createStxMoney } from './create-stx-money';
+import { useStxBalanceQuery } from './use-stx-balance-query.hooks';
+
+export {
+  createGetStacksAccountBalanceQueryOptions,
+  createStxCryptoAssetBalance,
+  createStxMoney,
+  useStxBalanceQuery,
+};
```
kyranjamie, line 11: Is this file necessary, why not just export at the root level like the rest of the exports?

pete-watters: I will get rid of it when cleaning up 👍

apps/mobile/src/store/accounts/utils.ts
```diff
@@ -8,10 +9,14 @@ export const accountStoreSchema = z.object({
   icon: z.string(),
   name: z.string(),
   status: z.string(),
+  type: z.string().optional(),
+  fingerprint: z.string().optional(),
+  accountIndex: z.number().optional(),
```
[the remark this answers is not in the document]
pete-watters: Thanks, these are from earlier in the work and I didn't notice them again being optional

apps/mobile/src/store/accounts/utils.ts
```diff
@@ -14,4 +15,5 @@ export interface AccountStore {
   icon: AvatarIconName;
   name: string;
   status: AccountStatus;
+  type?: WalletStore['type'];
```
pete-watters, line 18: Adding this as we have some code in `AccountCardLayout` to show `LedgerBadge` if ` type === 'ledger'`. 

We aren't using this now though so I am inclined to remove it until we add ledger support to mobile

apps/mobile/src/store/accounts/utils.ts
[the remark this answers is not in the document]
pete-watters: I've removed these but have kept `type` for now. Maybe I should also remove that?

 It's for ledger but we don't support that yet [see here](https://github.com/leather-io/mono/pull/542#discussion_r1822403566)

packages/query/src/stacks/balance/index.ts
pete-watters: Removed

apps/mobile/src/queries/balance/use-total-balance.ts
```diff
@@ -0,0 +1,65 @@
...
+    return signers.list.map(signer => signer.address);
+  }
+
+  const { fingerprint } = accountId;
+  return signers.list
+    .filter(signer => signer.descriptor.includes(fingerprint))
+    .map(signer => signer.address);
+}
+
+export function useWalletTotalBalance(): TotalBalance {
```
pete-watters, line 33: I  think maybe this should change to plural `useWalletsTotalBalance` as it should support multiple wallets? 

Or else if we have a different name (like portfolio) to signify the users total portfolio balance - all wallets added to the app. 

@kyranjamie I asked a [Q here](https://github.com/leather-io/mono/pull/520/files#r1822656476) about `useWalletTotalBitcoinBalance` . 

I assume that includes all wallets the user has added as it's using their keychains

apps/mobile/src/queries/stacks/use-stacks-confirmed-transactions.ts
```diff
@@ -0,0 +1,11 @@
+import { useMemo } from 'react';
+
+import { useGetAccountTransactionsWithTransfersQueries } from './transactions-with-transfers.query';
+
+/* @deprecated should do this in the query instead */
```
pete-watters, line 5: I'm thinking I should move this logic into `useMempoolTxsBalance` instead

apps/mobile/src/components/widgets/tokens/use-get-tokens-list.ts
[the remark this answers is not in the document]
pete-watters: This file has been totally removed now we have banished the mock data so no need for any of these types here anymore

packages/query/src/stacks/stacks-client.ts
```diff
@@ -71,7 +72,10 @@ export function stacksClient(basePath: string) {
       );
       return resp.data;
     },
+
     // TODO: Need to replace, this endpoint has been deprecated
+    /** @deprecated use getEventsForAddressTransaction
```
pete-watters, line 77: I started to look into removing this and added a comment here: https://github.com/leather-io/mono/pull/542#discussion_r1820903384 

`getAccountTransactionsWithTransfers` uses [this endpoint](https://docs.hiro.so/stacks/api/accounts/transactions-with-transfers) which seems deprecated but no direct alternative is offered.

packages/query/src/stacks/stacks-client.ts
```diff
@@ -87,6 +91,36 @@ export function stacksClient(basePath: string) {
       );
       return resp.data;
     },
+    async getEventsForAddressTransaction(address: string, txId: string, signal: AbortSignal) {
```
pete-watters, line 94: I added these as potential replacements to `getAccountTransactionsWithTransfers` but haven't spent long enough figuring out how to use them yet

packages/query/src/stacks/balance/account-balance.hooks.ts
```diff
@@ -1,61 +1,17 @@
...
-  return useQuery({
-    ...createGetStacksAccountBalanceQueryOptions({
-      address,
-      client,
-      network: network.chain.stacks.url,
-    }),
-    select: resp => createStxMoney(resp),
-  });
-}
+import { createStxCryptoAssetBalance } from './create-stx-crypto-asset-balance';
```
pete-watters, line 12: I just split up this file to help me understand what was going on. I can revert if desired

apps/mobile/src/queries/stacks/use-mempool-txs-balance.ts
```diff
@@ -0,0 +1,44 @@
...
+function getPendingTxs(confirmedTxs: Transaction[], pendingTxs: MempoolTransaction[]) {
+  return pendingTxs.filter(tx => {
+    if (confirmedTxs.some(confirmedTx => confirmedTx.nonce === tx.nonce)) {
+      return false;
+    }
+    return tx.tx_type === 'token_transfer';
+  }) as unknown as MempoolTokenTransferTransaction[];
+}
+
+export function useMempoolTxsBalance(addresses: string[]) {
```
pete-watters, line 31: I did some recreation to refactor the code we already had where inbound and outbound transactions separated.

Here we use the same code for both but apply filtering to the pendingTxs in the same place to provide inboundBalance + outboundBalance for all supplied addresses

[original comment](https://github.com/leather-io/mono/pull/542#discussion_r1820877966)

apps/mobile/src/queries/balance/use-stx-balances.query.ts
```diff
@@ -0,0 +1,51 @@
...
+  useCurrentNetworkState,
+  useStacksClient,
+} from '@leather.io/query';
+import { createMoney, sumMoney } from '@leather.io/utils';
+
+export function useStxBalancesQueries(addresses: string[]) {
+  const client = useStacksClient();
+  const network = useCurrentNetworkState();
+
+  const { inboundBalance, outboundBalance } = useMempoolTxsBalance(addresses);
```
pete-watters, line 20: I think that maybe I should keep going and move `useMempoolTxsBalance` into this `useQueries` also

apps/mobile/src/queries/balance/stacks-balance.query.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!addresses || addresses.length === 0)
+    return { totalStxBalance: createMoney(0, 'STX'), accounts: [] };
+
+  return {
+    totalStxBalance: totalData,
+    accounts: queries, // TODO as yet unused but perhaps useful
+  };
+}
+
+export function useStxBalance(addresses: string[]) {
```
pete-watters, line 27: I am using this single function `useStxBalance` that accepts addresses. 

I can refactor it to provide `useWalletTotalStacksBalance` and `useStacksAccountTotalStacksBalance` to match what we do for bitcoin if desired

apps/mobile/src/queries/balance/stacks-balance.query.ts
```diff
@@ -0,0 +1,44 @@
...
+
+function useGetStxBalanceByAddresses(addresses: string[]): StxBalances {
+  const { totalData, queries } = useStxBalancesQueries(addresses);
+
+  if (!addresses || addresses.length === 0)
+    return { totalStxBalance: createMoney(0, 'STX'), accounts: [] };
+
+  return {
+    totalStxBalance: totalData,
+    accounts: queries, // TODO as yet unused but perhaps useful
```
pete-watters, line 23: I'm not sure if I should return the `accounts` here as maybe we will use this data elsewhere so it could be useful to cache if we are getting it for balances anyway

packages/query/src/stacks/balance/use-stx-balance.query.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+// Mock the query hook response
+vi.mock('@tanstack/react-query', () => ({
+  useQuery: () => ({
+    data: createMoney(1000000, 'STX'),
+    isLoading: false,
+    error: null,
+  }),
+}));
+
```
pete-watters, line 29: I wanted to add some coverage here for the stats:

Screenshot 2024-10-31 at 12 39 18

But this test seems a bit pointless. I am just mocking `vi.mock('@tanstack/react-query'` so it's kind of just testing the mocking works when called through the hook. 

Mentioning as I'd like if we could add coverage to these hooks but not sure if it's superfluous

pete-watters:
This PR got out of hand and is way too big to review so I am going to close it off and split up the work into more manageable chunks. 

Thanks to everyone for the feedback so far 👍

PR #543, opened 2024-10-16
fix: fonts [LEA-1442]

edgarkhanzadian:
diatype
marche


fbwoolf (approved):

kyranjamie (approved):
On the asset storage, I wonder if it'd be nicer to just group by font, not where they're used. 

The `otf`, `woff` etc could live together, agnostic of platform, and the platforms just pick the formats they need.

PR #550, opened 2024-10-17
feat: reset store if zod validation fails [LEA-1474]

edgarkhanzadian:


kyranjamie (approved):
Smart idea, love going zod validator-first

PR #552, opened 2024-10-17
fix: wallet sheet icon colors [LEA-1534]

edgarkhanzadian:
fixIconColors

### DIFF
--- a/apps/mobile/src/components/sheets/warning-sheet.layout.tsx
+++ b/apps/mobile/src/components/sheets/warning-sheet.layout.tsx
@@ -2,6 +2,7 @@ import { RefObject } from 'react';
 
 import { useSettings } from '@/store/settings/settings';
 import { t } from '@lingui/macro';
+import { useTheme } from '@shopify/restyle';
 
 import {
   Box,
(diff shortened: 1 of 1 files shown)


fbwoolf (approved):

apps/mobile/src/components/sheets/warning-sheet.layout.tsx
```diff
@@ -50,7 +53,9 @@ export function WarningSheet({
                 >
                   <ErrorTriangleIcon
                     color={
-                      variant === 'critical' ? 'red.action-primary-default' : 'ink.text-primary'
+                      theme.colors[
```
pete-watters, line 56: Maybe we could move this into a const a bit higher up to make it a bit easier to read? 

Having a ternary inside an array lookup is a bit tricky.

This could be nicer:
```
const iconColor = variant === 'critical' ? theme.colors['red.action-primary-default'] : theme.colors['ink.text-primary'];
...
  <ErrorTriangleIcon
                    color={iconColor}
```

This icon could probably even be in a separate wrapped component that loads the `useTheme` hook there. 

Looking at this code and I see the `Box` has this color settings:
```
<Box
                  bg={
                    variant === 'critical' ? 'red.background-primary' : 'ink.background-secondary'
                  }
```

1. Do we only need the `theme[...]` syntax for the icon?
2. Maybe we can just have this `Box` as a component accepting `variant`

PR #554, opened 2024-10-17
fix: numbering of wallets and accounts [LEA-1533]

edgarkhanzadian:
https://github.com/user-attachments/assets/e5a464b9-32fe-4968-a979-7aaeb6665085


fbwoolf (approved):
🚀

apps/mobile/src/store/accounts/accounts.write.ts
```diff
@@ -73,9 +75,24 @@ export const accountsSlice = createSlice({
       })
 
       .addCase(userAddsAccount, (state, action) => {
+        const { fingerprint: thisWalletFingerprint } = destructAccountIdentifier(
```
pete-watters, line 78: Maybe we could have this in its own function so we could add a unit test on it?

pete-watters (approved):

PR #555, opened 2024-10-17
ci: fix tests privacy button

edgarkhanzadian:
Should be only one `homeSettingsButton` id


apps/mobile/src/components/headers/components/header-options.tsx
```diff
@@ -25,7 +25,7 @@ export function HeaderOptions() {
       <TouchableOpacity
         p="2"
         onPress={() => onUpdatePrivacyMode()}
-        testID={TestId.homeSettingsButton}
+        testID={TestId.homePrivacyButton}
```
fbwoolf, line 28: Oops, thanks!

fbwoolf (approved):

PR #556, opened 2024-10-18
fix(models): testnet4 schema updates

kyranjamie:
I went back and forward on the best schema for this, and ended up adding the separate prop, even though it can be inferred.


pete-watters (approved):

pete-watters (approved):
Code looks good (apart from CI issues). 

I wonder should `bitcoinNetworkModesSchema` be plural `bitcoinNetworkModes`? 🤔

kyranjamie:
good point @pete-watters, think it shouldn't be plural in first place as it validates a single mode at a time, changed

PR #558, opened 2024-10-18
fix: add lock timeout, closes LEA-1588

edgarkhanzadian:
- Don't lock the app if user left the app less than 1 minute ago

NOTE: for testing purposes, timeout is set to 4 seconds here

https://github.com/user-attachments/assets/5076df0a-73ba-4a3d-9f8b-71c267242f65

### DIFF
--- a/apps/mobile/src/components/splash-screen-guard/splash-screen-guard.tsx
+++ b/apps/mobile/src/components/splash-screen-guard/splash-screen-guard.tsx
@@ -2,12 +2,10 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 import { useColorScheme } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
-import { useAppState } from '@/hooks/app-state';
 import { useSettings } from '@/store/settings/settings';
 import { HasChildren } from '@/utils/types';
 import { t } from '@lingui/macro';
 import { useTheme } from '@shopify/restyle';
-import * as LocalAuthentication from 'expo-local-authentication';
 import * as SplashScreen from 'expo-splash-screen';
 import * as SystemUI from 'expo-system-ui';
 import LottieView from 'lottie-react-native';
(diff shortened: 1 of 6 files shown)


apps/mobile/src/store/settings/settings.ts
```diff
@@ -111,6 +117,9 @@ export function useSettings() {
     changeThemePreference(theme: ThemePreference) {
       dispatch(userChangedThemePreference(theme));
     },
+    changeLastUnlocked(timestamp: LastUnlockedTimestamp) {
+      dispatch(userChangedLastUnlocked(timestamp));
+    },
```
kyranjamie, line 122: [Modelling after a user event](https://redux.js.org/style-guide/#model-actions-as-events-not-setters), this would be along the lines of `userUnlocksWallet`, and the result that happening is that we update the last unlocked time

apps/mobile/src/components/splash-screen-guard/splash-screen-guard.tsx
```diff
@@ -18,15 +18,16 @@ import { Box, Button, Theme } from '@leather.io/ui/native';
 import { LeatherLockedSplash } from '../animations/leather-locked-splash';
 import { LeatherSplash } from '../animations/leather-splash';
 
-const DEFAULT_ANIMATION_FINISHED = __DEV__;
-type AuthState = 'started' | 'failed' | 'passedOnFirst' | 'passedAfterwards';
+const DEFAULT_ANIMATION_FINISHED = false;
+const UNLOCK_TIMEOUT = 60 * 1000;
+type AuthState = 'cold-start' | 'started' | 'failed' | 'passedOnFirst' | 'passedAfterwards';
```
kyranjamie, line 23: Mixed casing
```suggestion
type AuthState = 'cold-start' | 'started' | 'failed' | 'passed-on-first' | 'passed-afterwards';
```

apps/mobile/src/store/settings/settings.ts
kyranjamie: Yep that sounds good

apps/mobile/src/components/splash-screen-guard/splash-screen-guard.tsx
```diff
@@ -41,12 +42,25 @@ export function SplashScreenGuard({ children }: HasChildren) {
...
         return;
       }
+      // if in secure mode, skip checks only if unlock time is not exceeding the timeout
+      // and this is not a cold start of the app
+      if (securityLevelPreference === 'secure' && authState !== 'cold-start' && checkUnlockTime()) {
+        setAuthState('passedOnFirst');
+        setAnimationFinished(true);
+        playSplash();
+        return;
+      }
```
kyranjamie, line 63: This timeout functionality seems kinda coupled specifically to the splash screen. Is there a way we could decouple better, perhaps with a hook that handles all timeout/auth concerns?

apps/mobile/src/components/splash-screen-guard/use-auth-state.ts
```diff
@@ -0,0 +1,76 @@
+import { useCallback, useState } from 'react';
+
+import { useAppState } from '@/hooks/app-state';
+import { useSettings } from '@/store/settings/settings';
+import * as LocalAuthentication from 'expo-local-authentication';
+
+const UNLOCK_TIMEOUT = 60 * 1000;
```
kyranjamie, line 7: This is a local const, so should be lowercase

kyranjamie (approved):

PR #559, opened 2024-10-22
fix: lock files

edgarkhanzadian:
We were getting this error on `dev`: 

```
 ERROR  Error: PanGestureHandler must be used as a descendant of GestureHandlerRootView. Otherwise the gestures will not be recognized. See https://docs.swmansion.com/react-native-gesture-handler/docs/installation for more details.
```


pete-watters (approved):
👍 this worked for me and fixed my issue

See here https://trustmachines.slack.com/archives/C05LAP952E7/p1729587735740909

PR #563, opened 2024-10-23
feat: lock app button, closes LEA-1543

edgarkhanzadian:
https://github.com/user-attachments/assets/7d3e85b4-0fde-4969-a846-0535c40ce62e

### DIFF
--- a/apps/mobile/src/components/splash-screen-guard/use-auth-context.ts
+++ b/apps/mobile/src/components/splash-screen-guard/use-auth-context.ts
@@ -1,0 +1,11 @@
+import { createContext, useContext } from 'react';
+
+interface AuthContextArgs {
+  lockApp(): void;
+}
+
+export const AuthContext = createContext<AuthContextArgs>({
+  lockApp: () => {},
+});
+
+export const useAuthContext = () => useContext(AuthContext);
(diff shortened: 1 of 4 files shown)


apps/mobile/src/components/splash-screen-guard/use-auth-state.tsx
```diff
@@ -60,9 +60,17 @@ export function useAuthState({
     if (securityLevelPreference === 'secure') {
       setAnimationFinished(false);
       setAuthState('started');
-      userLeavesApp(+new Date());
+      // add latest active timestamp only if the app was actually unlocked
+      if (authState === 'passed-on-first' || authState === 'passed-afterwards')
```
pete-watters, line 64: How about replacing the comment with a const? 

```suggestion
      const appUnlocked = authState === 'passed-on-first' || authState === 'passed-afterwards';
      if (appUnlocked)
```

pete-watters (approved):
👍

fbwoolf (approved):

apps/mobile/src/components/splash-screen-guard/use-auth-context.ts
```diff
@@ -0,0 +1,11 @@
+import { createContext, useContext } from 'react';
+
+interface AuthContextT {
```
kyranjamie, line 3:
```suggestion
interface AuthContextArgs {
```

apps/mobile/src/components/splash-screen-guard/use-auth-context.ts
```diff
@@ -0,0 +1,11 @@
...
+  lockApp(): void;
+}
+
+export const AuthContext = createContext<AuthContextT>({
```
kyranjamie, line 7:
```suggestion
export const AuthContext = createContext<AuthContextArgs>({
```

not big on the T? That stands for type? Would avoid this in same way we don't use `IAuthContext` naming

PR #564, opened 2024-10-23
feat: header title animation, closes LEA-647 and LEA-1547

fbwoolf:
This PR refactors many of our screens to use a new `Animated.ScrollView` w/ an animated header. I've created two layout components, with and w/out a `KeyboardAwareScrollView`:

1. `AnimatedHeaderScreenLayout`
2. `AnimatedHeaderScreenWithKeyboardLayout`

What happens is this, we detect if there is enough content on the screen to scroll. If there is, then scroll is enabled and the title animates in the header. If not, then the scroll is disabled and the large title remains static. In addition, with the keyboard, the scroll becomes enabled when the keyboard shows and disables when it hides. ~~So, the title animates in the header only when the keyboard is open (though now I'm thinking it needs both, so I'll add the content height check back to that layout too).~~ Updated

Also, I tackled the Cell padding issue here, but I don't think we want to add it directly to the Cell component, right? I did add it to every use, but I can move it to the Cell if we think it won't need to change? Or, I can set `py="3"` as a default?

https://github.com/user-attachments/assets/fde1415c-32ca-4a40-9a0b-c4fa761737fb


fbwoolf:
cc @mica000 for design review.

apps/mobile/src/app/(home)/recover-wallet.tsx
```diff
@@ -190,7 +175,7 @@ export default function RecoverWallet() {
               <ChevronDownIcon color={theme.colors['ink.text-primary']} variant="small" />
             )}
           </TouchableOpacity>
-          {!showAdvancedOptions ? null : (
+          {showAdvancedOptions ? (
```
pete-watters, line 178: Maybe this can be just `{showAdvancedOptions &&`

pete-watters (approved):
Nice work 👍  

I think it's OK passing `py="3"` as is now but it could be good to default it if that's what most `<Cell.Root`s set it to

apps/mobile/src/components/headers/animated-header/animated-header-screen-with-keyboard.layout.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const scrollY = useSharedValue(0);
+
+  const onScrollHandler = useAnimatedScrollHandler({
+    onScroll: event => {
+      scrollY.value = event.contentOffset.y;
+    },
+  });
+
+  const animatedHeaderStyle = useAnimatedStyle(() => {
+    const opacity = withTiming(scrollY.value >= 68 ? 1 : 0, { duration: 300 });
```
kyranjamie, line 44: could the height 68 be dynamic? what if height changes?

apps/mobile/src/components/headers/animated-header/animated-header-screen.layout.tsx
```diff
@@ -0,0 +1,82 @@
...
+      scrollY.value = event.contentOffset.y;
+    },
+  });
+
+  const animatedHeaderStyle = useAnimatedStyle(() => {
+    const opacity = withTiming(scrollY.value >= 68 ? 1 : 0, { duration: 300 });
+    return {
+      opacity,
+    };
+  });
```
kyranjamie, line 43: this looks repeated, could we have a reusable hook?

apps/mobile/src/app/(home)/settings/wallet/index.tsx
```diff
@@ -1,73 +1,66 @@
...
+        title={t({
+          id: 'wallets.header_title',
+          message: 'Wallets',
+        })}
+      >
+        <WalletsList variant="active" />
+        <Divider />
+        <Box gap="3" paddingTop="3">
+          <Cell.Root
+            py="3"
```
kyranjamie, line 33: Wouldn't a VStack avoid all the repeated paddings?

kyranjamie (approved):
Great work @fbwoolf—awesome to see some more animations

mica000:
Nice!!! This is looking awesome!! 🤩
Thanks @fbwoolf!!

apps/mobile/src/components/divider.tsx
```diff
@@ -1,7 +1,23 @@
...
+  useEffect(() => {
+    const subscription = Dimensions.addEventListener('change', ({ window }) => {
+      setScreenWidth(window.width);
+    });
+    return () => subscription?.remove();
+  });
+
+  return (
+    <Box alignSelf="center" bg="ink.border-transparent" height={1} width={screenWidth} {...props} />
+  );
```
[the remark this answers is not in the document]
fbwoolf: In the docs, it suggests this so if the user rotates the screen it updates, is that not something we want? Is ours locked in portrait?

apps/mobile/src/app/(home)/settings/notifications/_layout.tsx
```diff
@@ -1,44 +1,52 @@
...
-    <TabsHeader
-      onGoBack={() => router.navigate(AppRoutes.Settings)}
-      rightElement={<NetworkBadge />}
-      tabs={[
-        {
-          isActive: pathname === AppRoutes.SettingsNotifications,
-          onPress() {
-            router.navigate(AppRoutes.SettingsNotifications);
+    <>
+      <NakedHeader rightElement={<NetworkBadge />} />
```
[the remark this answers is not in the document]
fbwoolf: I believe I had to do it this way bc of the tabs used, but let me check. I did change some things with this so maybe I can move it back.

apps/mobile/src/app/(home)/recover-wallet.tsx
fbwoolf: Good point, prefer that too. Will change.

apps/mobile/src/app/(home)/settings/wallet/index.tsx
kyranjamie: Mmm, I was thinking of the generated ones like panda, that do space elements. Isn't that the whole point of H/VStack?

apps/mobile/src/components/headers/animated-header/animated-title-header.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { HeaderLayout } from '../header.layout';
+
+const AnimatedBox = Animated.createAnimatedComponent(Box);
+
+interface AnimatedTitleHeaderProps {
+  animatedStyle: ReturnType<typeof useAnimatedStyle>;
+  title: string;
+  rightElement?: ReactNode;
+}
+export function AnimatedTitleHeader({
```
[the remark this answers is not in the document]
fbwoolf: Nice idea, I might leave for another PR just bc I want to get this one merged bc I broke translations.

apps/mobile/src/app/(home)/settings/notifications/_layout.tsx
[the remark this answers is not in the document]
fbwoolf: Yep, worked to move it back, thx!

PR #565, opened 2024-10-24
feat: market data service

alexp3y:
Applying new services design concept (application layer servcies + infrastructure layer services) to a market data service.


kyranjamie (commented):
Great work @alexp3y

These are some pretty extensive changes to please forgive the thorough review. Biggest question relates to use of the caching layer. The way I see it, creating a wrapper adds a layer of complexity, and ultimately exposes the same thing?

A lot of code in this PR has been copied from elsewhere in the monorepo, can we delete the source logic in the market data hooks now it lives in the services package?

The extension has uses market data too, so to test out this set up, it'd be great to see an extension PR to go with it using the new services layer as part of this work.

Perhaps for a follow up PR task, but now we have API calls spread between both the query and services packages. I believe in future all of these should live in `@leather.io/api` that both these packages consume.

apps/mobile/src/services/mobile-cache.service.ts
```diff
@@ -0,0 +1,22 @@
...
+      return queryClient.fetchQuery({
+        queryKey: key as QueryKey,
+        queryFn: fetchFn,
+        staleTime: options.ttl ?? 0,
+        gcTime: options.ttl ?? 0,
+        retry: false,
+      });
+    },
+  };
+}
```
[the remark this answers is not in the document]
alexp3y: The broader idea is to use dependency inversion to decouple from implementations and keep things within `@leather.io/services` as platform-agnostic possible. 

The `CacheService` specifically is meant to be a simplified infrastructure-level cache that performs only the kind of caching behavior required by that layer, which IMO should be pretty straightforward (could be wrong here though?).

So, the interface (`fetchWithCache`, `ttl`) isn't meant to be a wrapper for `QueryClient`, but rather a simple interface for usage within API clients that supports any alternative implementation we want to use in the future (e.g. `AsyncStorage` in mobile, `localStorage` in extension, etc..) w/out changing the dependent clients.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -1,68 +1,114 @@
...
-export interface BtcCryptoAssetInfo extends BaseCryptoAssetInfo {
+export interface BtcCryptoAssetInfo extends BaseFungibleCryptoAssetInfo {
+  readonly assetType: 'nativeBtc';
   readonly name: 'bitcoin';
   readonly symbol: 'BTC';
 }
-
-export interface StxCryptoAssetInfo extends BaseCryptoAssetInfo {
+export interface StxCryptoAssetInfo extends BaseFungibleCryptoAssetInfo {
+  readonly assetType: 'nativeStx';
```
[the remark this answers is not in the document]
alexp3y: This is a bit of a side-enhancement - But I wanted to build out the `CryptoAssetInfo` model structure a little bit, which does help within the `MarketDataService`... In general, by adding additional context to types via `assetCategory` and `assetType` fields, we can build out a more robust inheritance structure and leverage discriminated unions for collective types.

Using those types we can then distinguish between different assets and categories (e.g. collectibles vs fungible) and make assumptions in both UI Components and Services (e.g. rendering CollectibleAsset vs FungibleAsset). 

The [`MarketDataService` shows how](https://github.com/leather-io/mono/pull/565/files#diff-f9fa9619f38b7054111a03f1505088fbdb37d7d30ec848bbe8f1aa8f55fba70d) we can leverage these discriminated types in interesting ways, including the top-level `getMarketData` call, which accepts a more generic `FungibleCryptoAssetInfo` asset and the subsequent switch statement, which handles different fungible asset types separately.

@fbwoolf Interested to hear your thoughts!

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
```diff
@@ -11,6 +11,10 @@ import { createGetStx20BalancesQueryOptions } from './stx20-tokens.query';
 
 function createStx20CryptoAssetInfo(stx20Balance: Stx20Balance): Stx20CryptoAssetInfo {
   return {
+    hasMemo: undefined as any,
+    decimals: undefined as any,
```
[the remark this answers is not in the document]
alexp3y: Yeah this is an artifact of moving around some fields when reworking the `CryptoAssetInfo` model. If these fields aren't shared by all fungible assets then we'll just need to put them individually on those that require them.

packages/services/src/infrastructure/api/binance/binance-api.client.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface BinanceApiClient {
+  fetchMarketData(currency: CryptoCurrency, signal?: AbortSignal): Promise<BinanceMarketData>;
+  fetchMarketDataPrice(currency: CryptoCurrency, signal?: AbortSignal): Promise<number | undefined>;
+}
+
+export function createBinanceApiClient(cacheService: CacheService) {
+  async function fetchMarketDataPrice(currency: CryptoCurrency, signal?: AbortSignal) {
+    const marketData = await fetchMarketData(currency, signal);
+    return readBinanceMarketDataPrice(marketData);
```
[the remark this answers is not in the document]
alexp3y: I don't think that should be the responsibility of client, but it should be an option offered via an associated client utility. 

If we tried to convert all client response to internal type by default, it would require a lot of assumptions about how it should be used - i.e. formatting, what data gets included, what data gets excluded, etc... By being faithful to the API return type, we separate concerns and allow consuming services to decide how the response data should be used. 

For things like converting a particular API response to an internal type like `Money`, I think that would be a great example of a utility that can live alongside the client and used selectively in services when required.

packages/services/src/market-data/market-data.service.ts
```diff
@@ -0,0 +1,148 @@
...
+        throw Error('Market data not supported for asset type: ' + asset.assetType);
+    }
+  }
+
+  async function getBtcMarketData(signal?: AbortSignal) {
+    return await _getNativeAssetMarketData({ symbol: 'BTC' } as BtcCryptoAssetInfo, signal);
+  }
+
+  async function getStxMarketData(signal?: AbortSignal) {
+    return await _getNativeAssetMarketData({ symbol: 'STX' } as StxCryptoAssetInfo, signal);
```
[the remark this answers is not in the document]
alexp3y: I built the `MarketDataService` around the `CryptoAssetInfo` model, under the assumption that MarketData would almost always be requested in the context of an asset, which in our application _should_ be represented by the model. I realized that native assets `BTC` and `STX` are likely the most common exceptions so provided convenvience calls within the service to get them directly, which requires passing an object "as" the native asset type.

apps/mobile/src/app/(home)/home/use-market-data.query.ts
```diff
@@ -0,0 +1,42 @@
...
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      getMarketDataService().getMarketData(token, signal),
+    refetchOnReconnect: false,
+    refetchOnWindowFocus: false,
+    refetchOnMount: false,
+    retryOnMount: false,
+    staleTime: 1000 * 60,
+    gcTime: 1000 * 60,
+  } as const;
+}
```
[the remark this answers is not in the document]
alexp3y: Moved it over but FWIW I put this in more as an illustration of how `BalanceService` can be used not necessarily how it _should_ be used.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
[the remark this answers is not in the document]
fbwoolf: I def wouldn't introduce the category `collectible` here? We have typ always used `fungible` and `non-fungible` if it is nec to add it. Also, I don't think I feel it is nec to use `native...` instead of just `btc` and `stx`? We know those are the chain cryptocurrencies.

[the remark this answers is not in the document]
fbwoolf: `Collectible` is just a nice UI word imo.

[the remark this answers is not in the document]
fbwoolf: Tbh, a lot of these changes we just refactored out, so curious to understand why they are nec again? 🤔 Maybe it is helpful for your work, so not opposed if helpful.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -1,68 +1,114 @@
...
+export interface Brc20CryptoAssetInfo extends BaseFungibleCryptoAssetInfo {
+  readonly assetType: 'brc20';
   readonly name: 'brc-20';
   readonly symbol: string;
 }
-
-export interface InscriptionCryptoAssetInfo {
+export interface Src20CryptoAssetInfo extends BaseFungibleCryptoAssetInfo {
+  readonly assetType: 'src20';
+  readonly name: 'src-20';
```
fbwoolf, line 52: Do we need both `assetType` and `name`? I believe they might be being used in the same way?

packages/models/src/crypto-assets/crypto-asset-info.model.ts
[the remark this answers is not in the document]
alexp3y: @fbwoolf Thanks for the feedback.

IMO "collectible" and "non-fungible" are pretty interchangeable. I went with "collectible" here just for consistency with other parts of the codebase.

This is probably a bigger discussion than market data, but the goal of restructuring the `CryptoAssetInfo` model is to build out a more descriptive type hierarchy and more accurately reflect the actual domain. 

It's not exactly necessary, however it does help in the market data service (as well as potentially many other services & UI components) by allowing us to leverage discriminated unions and organize / manage assets based on shared characteristics at different levels of the hierarchy (and not just rely on individual asset types + separate service calls for each).

For example, [`MarketDataService.getMarketData()`](https://github.com/leather-io/mono/pull/565/files#diff-f9fa9619f38b7054111a03f1505088fbdb37d7d30ec848bbe8f1aa8f55fba70d) can take any `FungibleCryptoAssetInfo` and handle it appropriately, using discriminators for BTC, Runes, and other fungible assets.

I could also see this setup enabling higher-level asset handling in UI components for categories like CollectibleAsset, which could render various types (e.g., stamp, sip9, inscription) without hard-coded handling for each individual asset.

Or the account balance work (cc @pete-watters  @kyranjamie) having a full list of account assets in a single array (or `fungibles[]` and `collectibles[]`) that can be handled collectively or further discriminated using the type field.

It also seems like we have a lot of one-off types throughout the codebase, which makes it more difficult to maintain consistency. By improving our models we could make them more useful and extensible - reducing the need for components and services to adapt types independently.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
alexp3y: `assetType` is stictly used as a discriminator, whereas `name` was already being used differently in different contexts and cant be used for this purpose. Most assets have a similar "asset type" pre-assigned in the `name` field, but others type this field as a `string` and fill it differently (e.g. in sip-10 tokens `name` is populated with "ALEX token", etc..)

packages/services/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './inversify.config';
+export * from './infrastructure/settings/network-settings.service';
+export * from './infrastructure/api/alex-sdk/alex-sdk.client';
+export * from './infrastructure/api/coingecko/coingecko-api.client';
+export * from './market-data/market-data.service';
```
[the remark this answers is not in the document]
alexp3y: Infrastructure services are those that exist in parallel to and in support of (i.e. below) business logic (application-layer servcies). Anything that interfaces with a specific technology i.e. network requests, caching, logging, config, storage, etc... By abstracting those interactions out we can swap out implementations (e.g. change caching provider) without disrupting the application layer, which should be agnostic of what's happening below it within the infrastructure services it relies on.

I think a good rule of thumb is to think about the dependency flow. Application-layer services sit on top of and depend on infrastructure, not the other way around. So you wouldn't inject an app-layer service into an infrastructure service. 

The network settings service is a bit of a weird construct but it represents config / some environment setting, and is used across many other infrastructure services (e.g. API clients). So I think this is safely considered infrastructure.

If we built a more robust `UserSettingsService` that managed business rules for user setttings then that would bubble up to the application-layer. However, that would still sit on top of a lower-level infrastructure service that managed the actual implementing technology.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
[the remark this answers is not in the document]
pete-watters: I think `collectible` is a nice word and good for codebase consistency and keeping a similar language in services and UI. I don't think it's a big issue to use `collectible` or `non-fungible` but I like this classification of things.

Overall, I think this kind of type level categorisation is a good idea as it will help eliminate some of the hardcoded logic we need to add for handling individual assets as you mention. 

`assetType: 'nativeStx'` will help in the UI for example where we will need to show the asset chain and another sub avatar in non assets that aren't BTC / STX.

Screenshot 2024-10-31 at 10 11 05

Initially, I would have liked a list of assets in a single array `fungibles[]` / `collectibles[]` as my idea was to fetch this and then feed it into dumb widgets which then map over items in dumb UI components. 

It seems an alternative approach we want to take is to instead have more granular smart components responsible for fetching their own data. 

For example, I was getting the data then making a list of `tokens` with a serialiser:
```
const tokens : [{
 ticker: 'BTC', 
 balance: 0
},{
 ticker: 'STX', 
 balance: 0
},
]

.... 

<TokensWidget tokens={tokens} />
```

The pattern we are going with instead is something like:
```
function BtcBalances() {
  const {balance, fiatBalance} = useGetBtcBalance(); 
 return (<Balance balance={balance} fiatBalance={fiatBalance} />)
}

....
   <TokensWidget> 
      <BtcBalances />
      <StxBalances />
   </TokensWidget>
```

packages/services/src/infrastructure/api/coincap/coincap-api.utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function mapToCoincapCurrency(currency: CryptoCurrency): string {
+  return coincapCurrencyMap[currency];
+}
+
+export function readCoincapAssetPrice(assetResponse: CoincapAssetResponse): number | undefined {
+  try {
+    const priceUsd = assetResponse.data.priceUsd;
+    const price = parseFloat(priceUsd);
+    if (isNaN(price)) return;
```
pete-watters, line 18: Should we throw errors for `undefined` data / add analytics? 

It could be a bit overkill for this type of thing but I'm curious to know your thoughts

packages/services/src/infrastructure/api/coingecko/coingecko-api.utils.ts
```diff
@@ -0,0 +1,27 @@
+import { CryptoCurrency, FiatCurrency } from '@leather.io/models';
+
+import { CoinGeckoMarketData } from './coingecko-api.client';
+
+export const coinGeckoCurrencyMap: Record<CryptoCurrency, string> = {
+  BTC: 'bitcoin',
+  STX: 'blockstack',
```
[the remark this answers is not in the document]
pete-watters: Throws their whole price accuracy into question!

apps/mobile/src/queries/use-market-data.query.ts
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    queryKey: ['get-market-data', network, token],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      getMarketDataService().getMarketData(token, signal),
+    refetchOnReconnect: false,
+    refetchOnWindowFocus: false,
+    refetchOnMount: false,
+    retryOnMount: false,
+    staleTime: 1000 * 60,
+    gcTime: 1000 * 60,
```
pete-watters, line 18: Maybe we can use const's for the `gcTime` and `staleTime` and add some defensive code, maybe a test that will fail if it gets set to an invalid value like Kyran found yesterday? 

https://github.com/leather-io/mono/pull/582/files

pete-watters (commented):
This work looks great to me so far @alexp3y, nice job! 

Maybe you could try consuming the market data in `useTotalBitcoinBalanceOfDescriptors` now that it is merged? That would be a good way of also adding an example use case of the service

packages/services/src/infrastructure/api/coincap/coincap-api.utils.ts
alexp3y: That's a great point. This utility is trying to handle bad responses, which isn't its responsibility. It should expect that the response object is what it says it is and any errors should be thrown upstream of this.

We should probably have Zod validation for all API response types, so that schema failures are detected immediately. Then we could try to centralize analytics directly within Zod to track validation errors. Looks like they support a custom [error interceptor](https://zod.dev/ERROR_HANDLING?id=customizing-errors-with-zoderrormap) where we could send analytics events.

pete-watters: Sounds good 👍 my thinking is more for us to have a first warning that APIs are failing so we can expect people having UI issues as a result and avoid time chasing down this type of intermittent issue

alexp3y: Oh yeah, we can probably centralize all API error tracking in general within the API layer.

We could use something like a shared Axios instance with a custom error interceptor to track failures. By having tracking a bit closer to source we could have much more granular event information (i.e. url, API name, http status, etc..)  instead of the top-level query key.

I just saw [this work](https://github.com/leather-io/extension/pull/5925) from @alter-eggo in `extension` - If we do any analytics here in `@leather.io/services` it may duplicate these existing API events. But the value prop seems high of getting more granular events data (e.g. 429 failures), so maybe we could use different event names for the query-level failures already in `extension` vs request-level failures. @alter-eggo @camerow thoughts?

apps/mobile/src/queries/balance/bitcoin-balance.query.ts
alexp3y, line 1: @pete-watters FYI updated this to use market-data service query.

alexp3y:
@kyranjamie I added the market-data service to the new bitcoin balance logic in `mobile`. Agree that a follow-on PR in `extension` is a good idea, both out of necessity to adapt to the updated `CryptoAssetInfo` model (if we stick with that) and to demo market-data service usage.

I do think completely replacing market-data across the board would be difficult as its a lower-level service - i.e. its used in many different hooks in both `extension` and `leather.io/query`. Something we could also think about as a next step could be to now build out a higher-level service that consumes market-data, maybe either adapting the new BTC balance hooks into a service (and moving towards a top-level `BalanceService`), or building out a swaps service.

fbwoolf:
This all looks great @alexp3y, good to see work on this and updates to the asset models to accommodate making the code easier to work with. 👍

packages/services/src/infrastructure/api/coincap/coincap-api.utils.ts
alexp3y: @alter-eggo sorry, that was bad phrasing on my part. I meant where the error is handled and analytics event sent rather than the where it originates from (i.e. RQ `onError  vs an Axios interceptor). My thought was that because `useQuery` can wrap higher-level calls that encapsulate >1 API request, and requests can be made imperatively outside of RQ, at some point it might make sense to move analytics into an Axios interceptor.

Anyway I see now that the error object in RQ `onError` already does have access to the undelrying response + status code, so its not a huge difference.

apps/mobile/src/services/mobile-cache.service.ts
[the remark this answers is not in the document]
kyranjamie: We can leave as is if you think this is best, but the counter argument would be

> but rather a simple interface for usage within API 
`QueryClient` already is a simple interface. You could even only expose certain methods from it, while retaining the benefits of a uniform API.

> that supports any alternative implementation we want to use in the future
When using QueryClient directly, you still have the flexibility to swap out storage methods. We're already doing that.

I think the contention here is that, the intention of this API is to keep things simple and flexible, but by abstracting it, we add another layer, which I'd argue has the opposite effect, versus using a single QueryClient API globally both in service layer and the apps.

packages/services/src/market-data/market-data.service.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  async function _getNativeAssetMarketData(asset: NativeCryptoAssetInfo, signal?: AbortSignal) {
```
kyranjamie, line 71: What's the meaning of the _ naming?

packages/models/src/crypto-assets/crypto-asset-info.model.ts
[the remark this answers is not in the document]
kyranjamie: I'd like to understand how the new types assist discriminating unions. These types already use string literals for name, such that they can discriminate.

Where does `assetType` differ from `name`? The name effectively describes the type here. So why do we need both?

apps/mobile/src/queries/market-data/market-data.query.ts
```diff
@@ -0,0 +1,46 @@
...
+import { FungibleCryptoAssetInfo, NetworkModes } from '@leather.io/models';
+import { getMarketDataService } from '@leather.io/services';
+import { oneMinInMs } from '@leather.io/utils';
+
+export function createMarketDataQueryOptions(
+  token: FungibleCryptoAssetInfo,
+  network: NetworkModes
+) {
+  return {
+    queryKey: ['market-data-service-get-market-data', network, token],
```
pete-watters, line 14: It doesn't need to be for this PR but sometimes we use `enum`s for the keys and sometimes not. e.g. `BitcoinQueryPrefixes` .

We should probably standardise that

apps/mobile/src/queries/balance/bitcoin-balance.query.ts
```diff
@@ -54,7 +56,7 @@ type TotalBalanceCombineFn = UseQueryResult<
 
 export function useTotalBitcoinBalanceOfDescriptors(descriptors: string[]) {
   const queries = useCreateBitcoinAccountUtxoQueryOptions(descriptors);
-  const calcUsdAmount = useCalculateBitcoinFiatValue();
+  const { data: btcMarketData } = useBtcMarketDataQuery();
```
pete-watters, line 59: Nice work integrating this 👍 I can use it in the STX balance too. That way the service becomes integrated and used already

packages/models/src/crypto-assets/crypto-asset-info.model.ts
kyranjamie: Following from other comment: these interfaces describe asset types, so the `name` refers to the `assetType`, which is why this feels like duplication. 

If ALEX token is being passed to `name` in these asset then therein lies the problem. Do we need a larger refactor here to fix this. That'd be a better solution than adding a new type to deal with existing tech debt. Maybe this is something that'd be good to discuss live?

apps/mobile/src/queries/balance/bitcoin-balance.query.ts
```diff
@@ -66,11 +68,13 @@ export function useTotalBitcoinBalanceOfDescriptors(descriptors: string[]) {
       );
 
       const availableBalance = createMoney(amount, 'BTC');
-      const fiatBalance = calcUsdAmount(availableBalance);
+      const fiatBalance = btcMarketData
+        ? baseCurrencyAmountInQuote(availableBalance, btcMarketData)
+        : createMoney(0, 'USD');
```
pete-watters, line 73: Not for this PR but we have to use this `createMoney(0, 'USD')` fallback sometimes which is technically not correct to show a `0` balance. 

We need to get [this issue](https://linear.app/leather-io/issue/LEA-505/loading-state-for-balances-on-home-screen) decided upon soon so we can show something instead of `0` when the service can't get the balance cc @camerow

packages/models/src/crypto-assets/crypto-asset-info.model.ts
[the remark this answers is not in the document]
pete-watters: I think he is using both due to this https://github.com/leather-io/mono/pull/565/files#r1823194310

packages/services/src/market-data/market-data.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    const alexSdkPrices = await alexSdkClient.getLatestPrices();
+    const alexSdkAssetMatch = alexSdkCurrencies
+      .filter(isDefined)
+      .find(
+        token =>
+          getPrincipalFromContractId(token.underlyingToken) ===
+          getPrincipalFromContractId(asset.contractId)
+      );
+    if (!alexSdkPrices || !alexSdkAssetMatch)
+      return createMarketData(createMarketPair(asset.symbol, 'USD'), createMoney(0, 'USD'));
```
pete-watters, line 100: I think we need to be careful again here with `createMoney(0, 'USD')`

pete-watters (approved):
Great work here @alexp3y 👍  

I'm happy to give my approval here but as its so big and important lets wait for a consensus before merge

packages/services/src/market-data/market-data.service.ts
alexp3y: The `_` prefix is a convention that indicates a function (or variable) is private within the service - i.e. it's an internal helper and not part of the public interface.

Makes it clear that its not meant to be exported and is just an implementation 
detail (and should only called by other functions in the service).

apps/mobile/src/services/mobile-cache.service.ts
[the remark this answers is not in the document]
alexp3y: In the short-run I totally agree that `QueryClient` is more familiar and that establishing these abstractions adds overhead.

But longer-term, as requests migrate to the API-layer and `useQuery` migrates up the stack into higher-level UI/app queries/hooks, I think we will benefit a lot from having a clear boundary between React Query and lower-level API request caching.

Let's discuss more. My inclination would be to try out infrastructure abstractions like `CacheService` first and if we find that its not useful, the overhead is too much or that additional QC-specific functionality is required then we lets swap it out and just use QC directly.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
alexp3y: @kyranjamie I agree we can & should refactor `name` out of assets that duplicate `assetType`. I didn't want to do that here to keep scope reasonable (will take quite some work to support in `extension`) and to get feedback first on the changes.

The model rework does go beyond just adding a discriminable `assetType` field, and establishes more collective types and a clearer hierarchy. Utilizing discrimination and shared properties can be useful in many contexts.

fbwoolf: I know that `name` was setup to be a discriminating union type, so if it has been implemented in correctly somewhere, then it should prob be fixed rather than adding to the model? Also, we know these are assets, so why is it nec to include `asset` in the prop?

fbwoolf: My comment here is just meant to say, we should make sure to make an issue to refactor these so these props aren't duplicated. I am fine if you change them in a way that makes more sense with your work here.

alexp3y: @fbwoolf The `asset` prefix is a bit redundant, but I've found that generic field names like `type` tend to lose context and can benefit from extra clarity. For example child, interfaces inherit all parent fields, so if we add another "type" somewhere down the line, like on an NFT asset, the benefit of the prefix becomes more obvious:
```
interface Sip9CryptoAssetInfo extends BaseCollectibleCryptoAssetInfo {
  assetType: 'sip9';
  collectionType: 'generative' | 'series';
  contractType: 'basic' | 'wrapped';
  mintType: ...
}
```

Totally agree on fixing `name`. It seems to be used mostly as like a display name in `extension` but it's hard to tell exactly where and how it's used. Definitely deserves its own issue / PR.

fbwoolf: Yeah, good point, I remembered `mimeType` after I made the comment ...makes sense to keep it. 👍

fbwoolf (approved):
Looks good, going to approve on my end.

camerow (commented):
> Not for this PR but we have to use this `createMoney(0, 'USD')` fallback sometimes which is technically not correct to show a `0` balance. 
> 
> We need to get [this issue](https://linear.app/leather-io/issue/LEA-505/loading-state-for-balances-on-home-screen) decided upon soon so we can show something instead of `0` when the service can't get the balance cc @camerow 
> 
>  

Yeah, we are deciding on the end design, however I think we could get started on that pretty dang soon. Michelly has two options that are fairly similar and a lot of the work is the underlying mechanism to check network connectivity / capture the errors and ensure we show _something_

packages/services/src/inversify.config.ts
```diff
@@ -0,0 +1,152 @@
...
+        networkSpecificContainer,
+        {
+          getConfig() {
+            return defaultNetworksKeyedById[networkId as WalletDefaultNetworkConfigurationIds];
+          },
+        },
+        cacheService
+      );
+      networkSpecificContainers[networkId] = networkSpecificContainer;
+    }
```
kyranjamie, line 57: Should we leave this out of the initial implementation? The requirement of looking up the same data of different chains was something worth considering, but not something we need now

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.ts
```diff
@@ -0,0 +1,78 @@
...
+  ]);
+
+  function getLimiterKey(type: RateLimiterType, network: NetworkModes) {
+    return `${type}_${network}`;
+  }
+
+  function getLimiter(type: RateLimiterType, mode: NetworkModes) {
+    const key = getLimiterKey(type, mode);
+    return limiters.get(key)!;
+  }
```
kyranjamie, line 57: These are also private, but not `_` prefixed. Is just not returning the function, clear enough that a fn is intended to be private?

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
```diff
@@ -11,9 +11,11 @@ import { createGetStx20BalancesQueryOptions } from './stx20-tokens.query';
 
 function createStx20CryptoAssetInfo(stx20Balance: Stx20Balance): Stx20CryptoAssetInfo {
   return {
+    assetCategory: 'fungible',
+    assetType: 'stx20',
     name: 'stx-20',
     symbol: stx20Balance.ticker,
-  };
+  } as Stx20CryptoAssetInfo;
```
kyranjamie, line 18:
```suggestion
  } satisfies Stx20CryptoAssetInfo;
```

?

kyranjamie: although why is this cast needed in first place when return type is declared

packages/services/src/market-data/market-data.service.ts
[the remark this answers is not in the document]
kyranjamie: Not confident this is the best solution, as it's just tricking the compilor. It's super confusing to come across an object that is typed as X but is actually a subset of X. 

`_getNativeAssetMarketData` only uses `symbol` from `NativeCryptoAssetInfo`, so the function is asking for more data than it needs. Suggest either passing only asset, or if we need dynamic return type 

```ts
  async function _getNativeAssetMarketData<T extends { asset: string }>(asset: T, signal?: AbortSignal) {
```

packages/services/src/infrastructure/cache/cache.service.ts
```diff
@@ -0,0 +1,9 @@
+export type CacheKey = readonly unknown[];
```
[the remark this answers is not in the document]
alexp3y: I used `unknown[]` to allow us to pass objects / variables like in RQ, which supports it.

That said, we might not ever need it and could probably switch to `string[]` and just enforce that, although we would lose some flexibility.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -1,68 +1,114 @@
...
+  sip9: 'sip9',
+  inscription: 'inscription',
+} as const;
+export type CryptoAssetCategory = keyof typeof CryptoAssetCategories;
+export type CryptoAssetType = keyof typeof CryptoAssetTypes;
+
 export interface BaseCryptoAssetInfo {
+  readonly assetCategory: CryptoAssetCategory;
+  readonly assetType: CryptoAssetType;
+  readonly name: string;
```
kyranjamie, line 25: Remove name, as not all elements have names

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -1,68 +1,114 @@
...
+  stamp: 'stamp',
+  sip9: 'sip9',
+  inscription: 'inscription',
+} as const;
+export type CryptoAssetCategory = keyof typeof CryptoAssetCategories;
+export type CryptoAssetType = keyof typeof CryptoAssetTypes;
+
 export interface BaseCryptoAssetInfo {
+  readonly assetCategory: CryptoAssetCategory;
+  readonly assetType: CryptoAssetType;
```
kyranjamie, line 24: rename protocol?

packages/models/src/crypto-assets/crypto-asset-info.model.ts
kyranjamie, line 52: Should either be `string` or removed

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.ts
alexp3y: Forgot to apply the underscores here. But yes, I think we could probably do without this convention. It should already be pretty clear from the exports what's public/private and I guess the utility of the underscore is probably eclipsed by the risk that we don't apply it consistently (like here) and end up adding more confusion. I'll remove it elsewhere.

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
alexp3y: This type assertion is needed to ignore the shared fungible base fields `hasMemo` and `decimals`. `Stx20CryptoAssetInfo` was previously not extending any base types, so this is a workaround to get it into the heirarchy. As a follow-up issue we should try to extract the shared fields to more specific types as I suspect only a couple of assets actually use them.

packages/services/src/market-data/market-data.service.ts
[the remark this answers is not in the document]
alexp3y: Totally agree. For this native asset market data we can make the signature super focused and just use the `CryptoCurrency` type ("STX" or "BTC") as a parameter, there's no need to pass the entire asset.

packages/services/src/inversify.config.ts
alexp3y: Agreed, let's YAGNI this out and keep it in our back pockets until we have a concerete use case. Good to know that initialization will be lazy - i.e. the containers register upfront but the actual network-specific services won't instantiate until called for the first time.

packages/models/src/crypto-assets/crypto-asset-info.model.ts
```diff
@@ -1,68 +1,122 @@
 import { InscriptionMimeType } from './bitcoin/inscription.model';
 
+export const CryptoAssetChains = {
+  bitcoin: 'bitcoin',
+  stacks: 'stacks',
+} as const;
```
kyranjamie, line 6:
```suggestion
export const CryptoAssetChains = ['bitcoin', 'stacks'] as const;
export type CryptoAssetChain = (typeof CryptoAssetChains)[number];
```

Any reason these are objects vs just an array?

packages/query/src/bitcoin/stamps/stamps-by-address.hooks.ts
```diff
@@ -15,10 +21,12 @@ export function useStampsByAddress(address: string) {
 
 function createSrc20CryptoAssetInfo(src20: Src20Token): Src20CryptoAssetInfo {
   return {
+    chain: CryptoAssetChains.bitcoin,
```
kyranjamie, line 24: ignore earlier comment, see you're using as obj here

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
kyranjamie: this can go with your latest changes though right as you removed from the base class?

packages/models/src/crypto-assets/crypto-asset-info.model.ts
alexp3y: When you define it with array or a literal union like that then it's difficult to satisfy the type when instantiating an asset object - Just using string `"bitcoin"` doesn't satisfy `CryptoAssetChain`. You need to use something like `asset = { chain: "bitcoin" as const, ... }`.

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
alexp3y: I didn't remove `decimals` and `hasMemo`, I just moved them from the base asset to the base fungible asset. It existed previously on 6 of 7 fungible assets, with this as the lone exception (`Stx20CryptoAssetInfo`), so I figured a temporary type assertion was preferable than redefining these fields 6 times on each fungible model type. But I think this definitely warrants a follow-up issue to figure out on which asset types exactly these fields are relevant/required and make them much more specific.

alexp3y:
@fbwoolf @pete-watters @kyranjamie Think I covered all comments but lmk if I missed or haven't addressed anything. 

I'll follow-up once this is merged with additional issues to integrate the updated models in `extension` and look into `decimals` / `hasMemo` usage.

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
kyranjamie: Had a look into this. There's actually no need to satisfy the return type of `Stx20CryptoAssetInfo`.

Typing something as X when it isn't actually X, is a crime demanding the most severe punishment lol. It defeats the point of having rich type information if you can't trust it.

Nothing is expecting the full `Stx20CryptoAssetInfo` here, so typing this function as returning `Partial<Stx20CryptoAssetInfo>` resp is totally fine. 

```suggestion
  };
```

packages/query/src/stacks/stx20/stx20-tokens.hooks.ts
```diff
@@ -11,9 +11,11 @@ import { createGetStx20BalancesQueryOptions } from './stx20-tokens.query';
 
 function createStx20CryptoAssetInfo(stx20Balance: Stx20Balance): Stx20CryptoAssetInfo {
```
kyranjamie, line 12:
```suggestion
function createStx20CryptoAssetInfo(stx20Balance: Stx20Balance): Partial<Stx20CryptoAssetInfo> {
```

kyranjamie (approved):
Great work Alex, looking forward to adding some code into the service architecture 💪🏼

Please make the fix that avoids the type casting, otherwise good to merge.

PR #577, opened 2024-10-27
fix: prep refactoring for send flow, closes LEA-1641

fbwoolf:
This PR preps some code for starting the send flow. I did a bunch of refactoring here renaming some layout components and others where the files and component names were not matching. Also, as mentioned in Slack, I refactored all `Icon` props to be `icon` and passed as ReactNodes. I relocated the avatar-icon svgs to the UI lib and created exported icons for them.

Also, I created the full height sheet here and a temp placeholder for the send flow. I have a a working branch for how I'll propose using this with React context to manage passing state through the form sheets. I'd like to get this work merged first since I'm renaming components here.

In addition, I refactored the account list to be list items rather than cards and located it as a reusable feature so we can access it in the send and receive flows.


apps/mobile/src/components/headers/components/header-title-with-subtitle.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export function HeaderTitleWithSubtitle({
+  title,
+  subtitle,
+  testID,
+}: {
+  title: string;
+  subtitle: string;
+  testID?: string;
+}) {
```
kyranjamie, line 11: 👀 separate props interface?

packages/ui/src/icons/account-avatars/bank.native.tsx
```diff
@@ -0,0 +1,10 @@
+import Bank from '../../assets/icons/account-avatars/bank-24-24.svg';
+import { Icon, IconProps } from '../icon/icon.native';
+
+export function BankIcon({ variant, ...props }: IconProps) {
+  return (
+    <Icon {...props}>
+      <Bank />
+    </Icon>
+  );
+}
```
kyranjamie, line 10: Icons should all be `forwardRef'd`

kyranjamie (approved):

apps/mobile/src/components/headers/components/header-title-with-subtitle.tsx
fbwoolf: Not sure how I missed, thx. I'll change all the headers to use interfaces so this mistake can't be made again. Seeing lots that need interfaces added, so will try to do here.

apps/mobile/src/features/account-list/account-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+      iconTestID={defaultIconTestId(account.icon)}
+      key={account.id}
+      onPress={() => {
+        router.navigate({
+          pathname: navigateToPath,
+          params: { fingerprint: walletFingerprint, account: account.accountIndex },
+        });
+      }}
+      testID={TestId.walletListAccountCard}
+      walletName={walletName ?? ' '}
```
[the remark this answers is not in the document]
fbwoolf: If I don't do this then the layout changes, this keeps the title in place, but maybe this should be changed in `ItemLayout`? I'll look, I would assume we always want the title to stay on aligned to top rather than centered if caption is undefined.

PR #578, opened 2024-10-28
feat(bitcoin): recursive address look up helper

kyranjamie:
Pulling this out of some WIP work in another branch to merge separately. 

This function takes an address and tries to find the corresponding path. I don't really want to need this function, it's a very expensive look up, but, given the xpub API response doesn't include the derivation path, in order to support sends on non-zero index, we need this check.

### DIFF
--- a/packages/bitcoin/src/bitcoin.utils.ts
+++ b/packages/bitcoin/src/bitcoin.utils.ts
@@ -15,6 +15,7 @@ import { defaultWalletKeyId, isDefined, whenNetwork } from '@leather.io/utils';
 
 import { BtcSignerNetwork, getBtcSignerLibNetworkConfigByMode } from './bitcoin.network';
 import { getTaprootPayment } from './p2tr-address-gen';
+import { getNativeSegwitPaymentFromAddressIndex } from './p2wpkh-address-gen';
 
 export interface BitcoinAccount {
   type: PaymentTypes;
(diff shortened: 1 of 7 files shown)


packages/bitcoin/src/bitcoin.utils.ts
```diff
@@ -277,6 +279,22 @@ export function getTaprootAddress({ index, keychain, network }: GetTaprootAddres
...
+  if (!keychain) throw new Error('Expected keychain to be provided');
+
+  if (keychain.depth !== DerivationPathDepth.Account)
+    throw new Error('Expects keychain to be on the account index');
+
+  const addressIndex = deriveAddressIndexKeychainFromAccount(keychain)(index);
+
+  if (!addressIndex.publicKey) throw new Error('Expected publicKey to be defined');
+
+  const payment = getNativeSegWitPaymentFromAddressIndex(addressIndex, network);
```
fbwoolf, line 292: Did you mean to do `SegWit` in name here, just flagging bc we don't typ ...thou maybe more common to?

fbwoolf (approved):

packages/bitcoin/src/bitcoin.utils.ts
kyranjamie: That's a mistake, should stick to Segwit as one word, will change

PR #580, opened 2024-10-29
fix: remove redundant, verbose in query cache

kyranjamie:
This PR creates a method that removes huge amounts of unused data from one of the API calls, reducing the amount of data cached by about 13x


pete-watters (approved):
Great work 👍 lets keep an eye on these query results to keep control of this as I saw some stacks contracts and other excess data myself earlier when working on mempool pending transactions

packages/query/src/stacks/transactions/transactions-with-transfers.query.ts
```diff
@@ -25,8 +46,10 @@ export function createGetAccountTransactionsWithTransfersQueryOptions({
   return {
     enabled: !!address && !!network,
     queryKey: [StacksQueryPrefixes.GetAccountTxsWithTransfers, address, network],
-    queryFn: ({ signal }: QueryFunctionContext) =>
-      client.getAccountTransactionsWithTransfers(address, signal),
+    queryFn: async ({ signal }: QueryFunctionContext) => {
+      const resp = await client.getAccountTransactionsWithTransfers(address, signal);
```
fbwoolf, line 50: This needs to be completely removed, right? This is the deprecated endpoint?

kyranjamie: I believe so, but believe that it's likely the new endpoint will have similar properties that need removing, I'll add a comment

pete-watters: I did some investigations on removing this in my PR and added some notes [here](https://github.com/leather-io/mono/pull/542#discussion_r1820903384)

- for `getAccountTransactionsWithTransfers` we call [this endpoint](https://docs.hiro.so/stacks/api/accounts/transactions-with-transfers) which seems deprecated but no alternative is offered
- There is a [similar endpoint](https://docs.hiro.so/stacks/api/accounts/transaction-with-transfers) - also deprecated, that accepts a tx_id and offers a new alternative [Get events for an address transaction](https://docs.hiro.so/stacks/api/transactions/events-for-an-address-transaction) but that is on a per tx basis
- I think we need to use [Get address transactions](https://docs.hiro.so/stacks/api/transactions/address-transactions) and then get `"tx_type": "token_transfer",`

PR #581, opened 2024-10-29
feat: send context and select account, closes #LEA-1642

fbwoolf:
This PR sets up the send context to handle navigation using the full height sheets and sharing state. In this work, the sheet to select an account has been implemented and the sheet to select an asset has a temp placeholder for the token list. I asked a question in Figma regarding the subtitle in the `select account` header, bc it shows the account name, but at that point no account would have been selected. I made the prop conditional for now, but there might be other into we want to show there in the future.

I am using a WIP state in the send context atm just to keep track of the selected account, asset, and the `step` to render the correct sheet content.


fbwoolf:
cc @mica000 to answer this question in Figma at some point: https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0?node-id=22304-272664#1005392394

apps/mobile/src/features/send/send-sheet-content.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  switch (step) {
+    case 'select-account':
+      return <SelectAccount />;
+    case 'select-asset':
+      return <SelectAsset />;
+    case 'send-form':
+      return <SendForm />;
+    case 'approver':
+      return <></>;
```
pete-watters, line 17: You can skip this `return <></>;` and have:

```
case 'approver':
default:
      return <></>;

```

apps/mobile/src/features/send/send-sheet-content.tsx
```diff
@@ -0,0 +1,21 @@
+import { useSendContext } from './send-context';
+import { SelectAccount } from './send-steps/select-account';
+import { SelectAsset } from './send-steps/select-asset';
+import { SendForm } from './send-steps/send-form';
+
+export function SendSheetContent() {
+  const { step } = useSendContext();
+
+  switch (step) {
```
pete-watters, line 9: This should have type `SendStep` right?

apps/mobile/src/features/send/send-sheet.layout.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+// TODO: Add translations to Crowdin
+export function SendSheetLayout({ children }: HasChildren) {
+  const { account, step } = useSendContext();
+  const { bottom } = useSafeAreaInsets();
+  const theme = useTheme<Theme>();
+
+  return (
+    <>
```
pete-watters, line 33: Unneeded fragment here

apps/mobile/src/features/send/send-container.tsx
```diff
@@ -0,0 +1,25 @@
+import { useState } from 'react';
+
+import { Account } from '@/store/accounts/accounts';
+import { HasChildren } from '@/utils/types';
+
+import { CryptoCurrency } from '@leather.io/models';
+
+import { SendContext, SendProvider, SendStep } from './send-context';
+
+export function SendContainer({ children }: HasChildren) {
```
pete-watters, line 10: Maybe we can consider using `useReducer` here instead of `useState` ? 

When there are a few state things to keep track of sometimes its a nicer way of managing things. We can use `dispatch` and `actions` then to update an object then rather than a few separate state items

pete-watters (approved):
Nice work here @fbwoolf 👍  

I just mentioned a few ideas but all good to merge as is 🚀

apps/mobile/src/features/send/send-container.tsx
fbwoolf: Good point, even if I don't refactor it here, I will as I start to expand on this approach.

apps/mobile/src/features/send/send-sheet-content.tsx
fbwoolf: This is already typed here.

apps/mobile/src/features/send/send-sheet-content.tsx
fbwoolf: Not too sure if I'll need this, but if I do there will be something diff from the default for the approver content. Sorry, I prob should have left comments that some of this is just setup.

apps/mobile/src/features/send/send-sheet.layout.tsx
fbwoolf: This is actually nec 👀

apps/mobile/src/features/send/send-container.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const [asset, setAsset] = useState<CryptoCurrency>('STX');
+  const [step, setStep] = useState<SendStep>('select-account');
+
+  const sendContextValue: SendContext = {
+    account,
+    asset,
+    step,
+    onSetAccount: (account: Account) => setAccount(account),
+    onSetAsset: (asset: CryptoCurrency) => setAsset(asset),
+    onSetStep: (step: SendStep) => setStep(step),
```
kyranjamie, line 21: Can `step` be inferred by the state of `account` and `asset`?

apps/mobile/src/features/send/send-sheet-content.tsx
kyranjamie, line 9: Could we look into using the router here? Using a context, we should be able to do this without passing state between steps.

fbwoolf: When using sheets, I didn't see anyone suggesting using the router for this ...this was the recommended approach. @edgarkhanzadian am I wrong here? Happy to change, but not sure why this is a bad approach? I thought we were trying to avoid route state with the forms? Didn't we go that direction in the extension and wanted to avoid it this time?

apps/mobile/src/features/send/send-sheet.layout.tsx
pete-watters: My bad 😊 sorry!

apps/mobile/src/features/send/send-sheet-content.tsx
pete-watters: OK, I wasn't sure if it needed to be re-added in the switch. Thanks! 🙇

apps/mobile/src/features/send/send-sheet-content.tsx
pete-watters: No problem 👍 just mentioned some stuff as I was going through it. 

PR good to go for me as is

apps/mobile/src/features/send/send-sheet-content.tsx
fbwoolf: @kyranjamie I think I misunderstood your comment here? You mean just for the `account` and `asset`, not for the form state? Let me look bc this is what I thought was suggested to do, but maybe I can use a nested Stack.

kyranjamie: >  I thought we were trying to avoid route state with the forms?

In the extension there was a gross overuse of form state. Passing everything from tx details, and pre-formatted sums of money etc. Really bad. IMO routes should be used for state that represents _a location within the app_. 

Maybe there are complexities with the sheet I'm unclear on, that mean we have to do it this way. 

But, I would typically argue this is a "bad approach" because it reimplements a much less capable router. One example: What if we need to navigate straight into the form, because from where the user initiates it the send, `account` and `asset` are already known. That's easy if we know the routing scheme is `/send/[fingerprint][accountId]/stx` or something like that. @edgarkhanzadian is already using the router to pass wallet/account data in some places.

fbwoolf:
@edgarkhanzadian curious your thoughts here on my implementation following the example from bottom sheet docs:
https://github.com/gorhom/react-native-bottom-sheet/blob/398e7114d7577c37361e6e24c9180b13de2c08ed/example/src/screens/integrations/navigation/NavigatorExample.tsx

The bottom sheet is working with the `independent` react-navigation stack, but I think maybe our theme isn't working? Ever seen this before? I can kinda see the `sparkles` icon and these are the account list items being rendered. I must have something minor off with setting up the `NavigationContainer`?

fbwoolf:
@kyranjamie curious your thoughts here now, the react-navigation/stack navigator is now working inside the bottom sheet. But, the `NavigationContainer` has to be flagged as `independent` to work inside the expo-router project already using a built-in `NavigationContainer`. So, the two stacks are independent from one another. Everything is working with this last push. Bc the `NavigationContainer` expects to be the first component in the app, I couldn't find another way except to use a RN hook to get the window height and set the stack navigator to that height to see the screen content.

PR #582, opened 2024-10-30
fix(query): gcTime using too big of a number for javascript

kyranjamie:
Some requests can be cached near-indefinitely, such as FT metadata which very infrequently changes. This endpoint is fetched each SIP10 token a user may have, which could be 10s or 100s of requests for well-used wallets.

We had efforts to cache this for 30 days with a staleTime of a week. I noticed in debugging extension that we would always refetch token metadata, resulting in a huge amount of wasteful requests being made to the Hiro API.

Because the `gcTime` was set to a number greater than the maximum of 2,147,483,647 supported by `setTimeout`, the value fails silently. Setting a lower gcTime, or setting it to `Infinity` (to disable it) fixes this issue. This data can be cached indefinitely, so I've set it to `Infinity`.

We've had so many discussions talking about ideas like cycling API keys, hosting our own API infra, and building a throttling system to manage Hiro API requests and avoid 429s by increasing complexity. If we run into 429s again, let's begin by establishing whether we're being efficient in our use of the API. Better to cure the disease with a balanced lifestyle, than treat it with expensive pharmaceuticals.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -62,7 +62,7 @@
     "@stacks/common": "6.13.0",
     "@stacks/transactions": "6.15.0",
     "@stacks/wallet-sdk": "6.15.0",
-    "@tanstack/react-query": "5.51.23",
+    "@tanstack/react-query": "5.59.16",
     "bignumber.js": "9.1.2",
     "buffer": "6.0.3",
     "dayjs": "1.11.13",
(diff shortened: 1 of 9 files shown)


pete-watters (approved):
Great work! And wise words about not seeking a magic pill - diet and exercise is the way

PR #583, opened 2024-10-30
Add @leather.io/analytics to mobile

camerow:
Updates the analytics package with more original events, segments them, adds the package to mobile in `utils`. 

A follow up step to this PR will be to publish this package and add all the existing events in `extension`.

Screenshot 2024-10-30 at 3 54 05 PM


pete-watters (approved):
Nice work @camerow 👍 

Code LGTM but I don't know much about the intricacies of the analytics.

Further to that point I wouldn't be averse to making a clean break from the old logging format for 2025 and just implement a new way for the new Leather. Old data is useful but it could be nice to have an easier to consume format for the future

fbwoolf (approved):

camerow:
@pete-watters I advocated for that in the initial project pitch but @kyranjamie felt strongly that we should not lose historical data. I do still feel that a hard fork is better but I am not willing to rock the boat on this particular issue.

kyranjamie (approved):
Thanks @camerow. 

Yeah we discussed in Linear. As I understand, we can follow all the new practices, _and keep all the existing data_, which seems close enough the best of both worlds. 

It's not quite as clean as starting fresh, but we don't lose any analytics data which I think it highly valuable. The beginning of our analytics history is when we launched the wallet, not some arbitrary date when we decided to refactor our analytics library.

PR #584, opened 2024-10-31
fix(mobile): remove fontWeight from MarchePro font

camerow:
MarchePro-Super has a single font weight. Setting the fontWeight value too high prevented Android renderer from finding a fallback value, and applying fontWeight to the font has no real effect. 

I have ensured this has no effect on web fonts by inspecting MarchePro-Super fonts in the extension and removing the `font-weight` css property. It has no effect.


camerow:
@edgarkhanzadian as far as I can tell none of those cases are for `MarchePro-Super` fonts so I think we're ok.

PR #586, opened 2024-10-31
feat: secure store utils, closes LEA-1576

edgarkhanzadian:
Created an Expo module that handles additional secure store functionality. In this case, updating biometrics security settings in iOS on an array of keys.

https://github.com/user-attachments/assets/4bb571d6-e22b-4324-bdb7-9257376ddb88


fbwoolf (approved):
🚀

kyranjamie (commented):
This looks great @edgarkhanzadian, awesome to see some Swift code being written 🚀 

Why a new package though? We won't be using this elsewhere, so why wouldn't we just keep it in the mobile folder?

apps/mobile/src/features/settings/app-authentication-sheet.tsx
```diff
@@ -13,42 +15,35 @@ interface AppAuthenticationSheetProps {
...
       );
+      changeSecurityLevelPreference(isSettingSecure ? 'secure' : 'insecure');
+      displayToast({
+        title: t({
+          id: 'app_auth.toast_title_success',
+          message: 'App authorization updated',
+        }),
+        type: 'success',
+      });
+    } catch (_) {
```
pete-watters, line 38: Can you remove this `_`?

pete-watters (approved):
Great work! 🎉

apps/mobile/src/features/settings/app-authentication-sheet.tsx
kyranjamie: catch brackets can't be empty

packages/secure-store-utils/android/src/main/java/expo/modules/securestoreutils/SecureStoreUtils.kt
```diff
@@ -0,0 +1,21 @@
+package expo.modules.securestoreutils
+
+import expo.modules.kotlin.modules.Module
+import expo.modules.kotlin.modules.ModuleDefinition
+
+class SecureStoreUtils : Module() {
```
camerow, line 6: Do you think we should write a separate issue for the Android implementation piece?

packages/secure-store-utils/ios/SecureStoreUtils.swift
```diff
@@ -0,0 +1,244 @@
...
+                } catch {
+                    // Think of a better way to relay exception from self.set function
+                    promise.reject(InvalidKeyException())
+                }
+            }
+        }
+
+    }
+
+    private func get(
```
camerow, line 59: Perhaps a more descriptive name than `get`? What is this returning? Looks like maybe a stored document that is either signed/unsigned with they users keychain?

patches/expo-module-scripts@3.5.2.patch
```diff
@@ -0,0 +1,44 @@
...
+index 7c556d528ecb92351da7aa4bcaefe665d7b59cc3..9c0bbf8b42cc6845359f1ed51615c2978795ff44 100755
+--- a/bin/expo-module-build
++++ b/bin/expo-module-build
+@@ -9,8 +9,7 @@ args=("$@")
+ # If the command is used like `yarn build plugin`, set the --build option to point to
+ # plugin/tsconfig.json
+ extra_module_build_types=("plugin" "cli" "utils" "scripts")
+-for i in "${extra_module_build_types[@]}"
+-do
++for i in "${extra_module_build_types[@]}"; do
```
camerow, line 11: this looks like it is just a formatting change - am I missing something?

patches/expo-module-scripts@3.5.2.patch
camerow, line 1: These looks like formatting change - am I missing something?

packages/secure-store-utils/android/src/main/java/expo/modules/securestoreutils/SecureStoreUtils.kt
camerow: Sorry, I guess the implied question is, are you planning on doing the Kotlin work in this PR? If not I'll create an issue to capture the Android side.

camerow:
This is a really great solution to this problem Edgar, nice one!

packages/secure-store-utils/ios/SecureStoreUtils.swift
```diff
@@ -0,0 +1,244 @@
...
+                localizedReason: options.authenticationPrompt ?? "Placeholder"
+            ) { success, error in
+
+                if let error = error as? NSError {
+                    promise.reject(LocalAuthenticationException(error))
+                    return
+                }
+                do {
+                    try keys.forEach { key in
+                        let item = try self.get(
```
camerow, line 37: I don't think you need `self.get` and can just use `get`, right?

kyranjamie:
@edgarkhanzadian is it correct that this PR is to stop the multi-face id in a row issue, because we store mnemonics by fingerprint? This is impressive work, but if another solution would be to use a single store, the complexity involved there would be orders of magnitude less than adding a new native code based module to our stack.

PR #587, opened 2024-10-31
Release @leather.io/analytics via release-please

camerow:
- Improve the API interface: remove need for consuming package to import any types, improve property type with rest param tuple type so `undefined` is an optional param but events with properties require them.
- Set up CI to deploy @leather.io/analytics

### DIFF
--- a/packages/analytics/src/types.ts
+++ b/packages/analytics/src/types.ts
@@ -1,104 +1,38 @@
+export * from './events';
 export interface DefaultProperties {
   platform: 'web' | 'extension' | 'mobile';
 }
 
-export interface ExternalAnalyticsClientInterface {
-  track: (event: string, properties?: JsonMap) => Promise<void>;
-  screen: (name: string, properties?: JsonMap) => Promise<void>;
-  group: (groupId: string, traits?: JsonMap) => Promise<void>;
-  identify: (userId: string, traits?: JsonMap) => Promise<void>;
+export interface AnalyticsClientInterface {
+  screen: (name: string, ...args: any[]) => Promise<any>;
+  track: (event: string, ...args: any[]) => Promise<any>;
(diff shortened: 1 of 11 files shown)


fbwoolf (approved):

packages/analytics/src/types.ts
```diff
@@ -2,15 +2,15 @@ export interface DefaultProperties {
...
-export interface ExternalAnalyticsClientInterface {
-  track: (event: string, properties?: JsonMap) => Promise<void>;
-  screen: (name: string, properties?: JsonMap) => Promise<void>;
-  group: (groupId: string, traits?: JsonMap) => Promise<void>;
-  identify: (userId: string, traits?: JsonMap) => Promise<void>;
+export interface AnalyticsClientInterface {
+  track: Function;
+  screen: Function;
+  group: Function;
+  identify: Function;
```
kyranjamie, line 9: Can you explain this change? Why the loosening of types? https://www.totaltypescript.com/dont-use-function-keyword-in-typescript#alternatives

camerow: I let this slip, I can try to enhance these types a bit but was struggling to find the right balance between the two Segment clients (browser vs. react-native). I will give it a bit more effort today.

camerow: Ironically I felt I was over typing them and did this to test the package `track` type, meaning to come back for this. I think for now I can keep them relatively loose types.

kyranjamie: All good, just trying to understand the change and problem that triggered it. 

With this interface, you get type errors because of the difference between the clients on each platform?

camerow: The client returned from `configureAnalyticsClient` has them properly typed the way we'd want them. I just pushed a change that makes the interface quite a bit more clear I think.

PR #588, opened 2024-11-01
Updates to support query package in React Native

pete-watters:
This PR [originates from](https://github.com/leather-io/mono/pull/542/files) and includes two important steps for the query package to work on React Native:
- [chore: use uuid instead of crypto.randomID to support RN](https://github.com/leather-io/mono/commit/7125dbc303ead3e90a475f3a64c07c6cec54ae8f) 
- [chore: downgrade p-queue and fix version to prevent throwIfAborted error](https://github.com/leather-io/mono/commit/76d5ea9e9d9d9d31e0c891497eb0dbac6b0159e3)

### UUID
I need to use uuid as crypto.randomID seems to be not working on React Native (I tried polyfills but this is easier)

If I don't do this queries throw the error: `"failureReason": [TypeError: crypto.randomUUID is not a function (it is undefined)]`

### p-queue
I am rolling back p-queue a version and setting it to 7.4.1 as otherwise queries will throw an error for undefined `options.signal?.throwIfAborted` in React Native code.

I tried to fix this to no avail and discovered the queries worked OK if I bypass the priority queue.

I tracked back their releases and found they introduced this change [here](https://github.com/leather-io/mono/pull/542/files#:~:text=introduced%20this%20change-,here,-.%20On%20their%20README) . On their README it says the project is feature complete and that they don't plan any further development / offer support so I think it's just easier to roll back for now.

I tested that this will work in the _extension_ by testing the new query package locally but will need to double check the published package

### DIFF
--- a/apps/mobile/src/scripts/injected-provider.js
+++ b/apps/mobile/src/scripts/injected-provider.js
@@ -1,3 +1,5 @@
+import { v4 as uuidv4 } from 'uuid';
+
 function createEventListeners() {
   const eventListeners = {};
   return {
@@ -24,7 +26,7 @@ const provider = {
     // window.ReactNativeWebView.postMessage('getURL');
   },
   async request(method, params) {
-    const id = crypto.randomUUID();
+    const id = uuidv4();
     const rpcRequest = {
       jsonrpc: '2.0',
       id,
--- a/apps/mobile/src/scripts/injected-provider.js
+++ b/apps/mobile/src/scripts/injected-provider.js
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from 'uuid';
+import { v4 as uuid } from 'uuid';
 
 function createEventListeners() {
   const eventListeners = {};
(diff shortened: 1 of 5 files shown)


apps/mobile/src/scripts/injected-provider.js
```diff
@@ -1,3 +1,5 @@
+import { v4 as uuidv4 } from 'uuid';
```
kyranjamie, line 1:
```suggestion
import { v4 as uuid } from 'uuid';
```

kyranjamie: is the v4 part relevant?

kyranjamie (approved):

fbwoolf (approved):

apps/mobile/src/scripts/injected-provider.js
pete-watters: In our usage, no. Thanks, I'll update

PR #589, opened 2024-11-01
feat: add numpad component, closes LEA-1647

tigranpetrossian:
Adds an initial version of the numeric pad.

* Currently supports two layouts: numeric and decimal, but is built with extensibility in mind, in case we need more variations in the future
* Locale-aware: needs to be supplied with a locale identifier string to display the appropriate decimal separator.
* Fully controlled: I'm on the fence about supporting uncontrolled mode for special use-case components like this, but open to suggestions.

https://github.com/user-attachments/assets/a5440e40-eea7-4d31-940a-3c7a91801b65

Some WIP improvements that will come after this PR, some need discussion, marked as TBD:

* Use Pressable internally, add less generic pressed state (to be designed)
* Use SVGs for visual labels for perfect optical alignment
* Use haptic feedback **(TBD)**
* Add disabled state **(TBD)**

Please do scrutinize code, naming, file structure, etc. as much as possible, this will also help me write down some of the patterns we prefer as conventions in more detail.

### DIFF
--- a/packages/ui/src/components/numpad/numpad-key.native.tsx
+++ b/packages/ui/src/components/numpad/numpad-key.native.tsx
@@ -1,0 +1,45 @@
+import { ReactNode } from 'react';
+
+import { isString } from '@leather.io/utils';
+
+import { HasChildren } from '../../utils/has-children.shared';
+import { Box } from '../box/box.native';
+import { TouchableOpacity } from '../button/touchable-opacity.native';
+import { Text } from '../text/text.native';
+
+interface NumpadKeyProps {
+  label: string;
+  element: ReactNode;
+  onPress(): void;
+  onLongPress?(): void;
+}
+
+export function NumpadKey({ element, label, onPress, onLongPress }: NumpadKeyProps) {
+  return (
+    <TouchableOpacity
+      flexDirection="row"
+      alignItems="center"
+      justifyContent="center"
+      height="100%"
+      accessibilityLabel={label}
+      onPress={onPress}
+      onLongPress={onLongPress}
+    >
+      {isString(element) ? (
+        <Text variant="heading04" allowFontScaling={false}>
+          {element}
+        </Text>
+      ) : (
+        element
+      )}
+    </TouchableOpacity>
+  );
+}
+
+export function NumpadKeySlot({ children }: HasChildren) {
+  return (
+    <Box height={48} flexGrow={1} flexShrink={1} flexBasis="30%">
+      {children}
+    </Box>
+  );
+}
--- a/packages/ui/src/components/numpad/numpad.native.tsx
+++ b/packages/ui/src/components/numpad/numpad.native.tsx
@@ -1,0 +1,159 @@
+import { ReactNode } from 'react';
+
+import { ArrowLeftIcon } from '../../icons/arrow-left-icon.native';
+import { Box } from '../box/box.native';
+import { NumpadKey, NumpadKeySlot } from './numpad-key.native';
+
+const FALLBACK_LOCALE_IDENTIFIER = 'en';
+
+type Mode = 'numeric' | 'decimal';
+type DecimalSeparator = '.' | ',';
+type KeyId =
+  | '1'
+  | '2'
+  | '3'
+  | '4'
+  | '5'
+  | '6'
+  | '7'
+  | '8'
+  | '9'
+  | '0'
+  | 'delete'
+  | 'decimalSeparator';
+type Layout = [KeyId | null, KeyId | null, KeyId | null][];
+
+interface Key {
+  id: KeyId;
+  accessibilityLabel: string;
+  element: ReactNode;
+}
+
+function getAllKeys(locale: string): Record<KeyId, Key> {
+  const decimalSeparator = getDecimalSeparator(locale);
+
+  return {
+    '1': { id: '1', accessibilityLabel: '1', element: '1' },
+    '2': { id: '2', accessibilityLabel: '2', element: '2' },
+    '3': { id: '3', accessibilityLabel: '3', element: '3' },
+    '4': { id: '4', accessibilityLabel: '4', element: '4' },
+    '5': { id: '5', accessibilityLabel: '5', element: '5' },
+    '6': { id: '6', accessibilityLabel: '6', element: '6' },
+    '7': { id: '7', accessibilityLabel: '7', element: '7' },
+    '8': { id: '8', accessibilityLabel: '8', element: '8' },
+    '9': { id: '9', accessibilityLabel: '9', element: '9' },
+    '0': { id: '0', accessibilityLabel: '0', element: '0' },
+    decimalSeparator: {
+      id: 'decimalSeparator',
+      accessibilityLabel: decimalSeparator,
+      element: decimalSeparator,
+    },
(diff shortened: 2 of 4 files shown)


packages/ui/src/components/numpad/numpad-key.native.tsx
```diff
@@ -0,0 +1,48 @@
+import { ReactNode } from 'react';
+
+import { Box } from '../box/box.native';
+import { TouchableOpacity } from '../button/touchable-opacity.native';
+import { Text } from '../text/text.native';
+
+function normalizeKeyElement(element: ReactNode) {
+  if (typeof element === 'string') {
```
kyranjamie, line 8: we have an `isString` helper util (same for all other primitives)

packages/ui/src/components/numpad/numpad.native.tsx
```diff
@@ -0,0 +1,157 @@
...
+  value,
+  onChange,
+  locale = FALLBACK_LOCALE_IDENTIFIER,
+  mode = 'decimal',
+}: NumpadProps) {
+  const decimalSeparator = getDecimalSeparator(locale);
+
+  const handlePress = (key: Key) => () => {
+    onChange(updateValue(value, key.id, decimalSeparator));
+  };
```
kyranjamie, line 131:
```suggestion
  function handlePress (key: Key) {
    return () => onChange(updateValue(value, key.id, decimalSeparator));
  }
```

generally followed convention is to use fn declarations

packages/ui/src/components/numpad/numpad.native.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const handleLongPress = (key: Key) => () => {
+    if (key.id === 'delete') {
+      onChange('0');
+    }
+  };
```
kyranjamie, line 137: Per normal keyboard behaviours, I'd imagine holding down return key to incrementally delete input (vs this behaviour of clearing the field)

kyranjamie (commented):
Great start @tigranpetrossian. 

What made you implement your own vs using a package?

packages/ui/src/components/numpad/numpad.native.tsx
tigranpetrossian: Misunderstood what top-level meant in the conventions 🙃

fbwoolf (approved):

packages/ui/src/components/numpad/numpad.native.tsx
tigranpetrossian: I've debated doing this, but decided to skip it in the initial version for two related reasons:
1. native pad widgets/real keyboards are more general purpose, those don't typically know what kind of input data they're dealing with
2. Most custom numeric pads I've tried (especially those in send/swap flows) had long press to clear, I'm guessing because the use-case is typically dealing with few digits at once, so you either tap few times to correct an erroneous input or clear altogether without waiting.

I could easily revisit this in the next round, we could potentially make it optional too.

tigranpetrossian:
> Great start @tigranpetrossian.
> 
> What made you implement your own vs using a package?

@kyranjamie The few I found seemed to be very general purpose, with an excessively large API surface, and/or opinions on styling. It felt like “configuring” those would still take quite a bit of time. Unless there’s a huge amount of functionality and no unopinionated/headless libraries like Radix to build on top of, it’s probably a safer bet to roll out our own, at least for relatively small things like this.

packages/ui/src/components/numpad/numpad-key.native.tsx
```diff
@@ -0,0 +1,50 @@
+import { ReactNode } from 'react';
+
+import { isString } from '@leather.io/utils';
+
+import { Box } from '../box/box.native';
+import { TouchableOpacity } from '../button/touchable-opacity.native';
+import { Text } from '../text/text.native';
+
+function normalizeKeyElement(element: ReactNode) {
```
pete-watters, line 9: Maybe this can just be a react component - `KeyElement` with `HasChildren` type?

packages/ui/src/components/numpad/numpad-key.native.tsx
```diff
@@ -0,0 +1,50 @@
...
+      accessibilityLabel={label}
+      onPress={onPress}
+      onLongPress={onLongPress}
+    >
+      {normalizeKeyElement(element)}
+    </TouchableOpacity>
+  );
+}
+
+export function NumpadKeySlot({ children }: { children?: ReactNode }) {
```
pete-watters, line 44: We have a `HasChildren` type you can use here

packages/ui/src/components/numpad/numpad.native.tsx
```diff
@@ -0,0 +1,159 @@
+import { ReactNode } from 'react';
```
pete-watters, line 1: This file is quite a monster. Maybe it could be split up a bit? Separating logic from presentation

pete-watters (approved):
Nice work here @tigranpetrossian 👍 I added a few suggestions

packages/ui/src/components/numpad/numpad.native.tsx
```diff
@@ -0,0 +1,159 @@
...
+      return null;
+    }
+
+    return getAllKeys(locale)[id];
+  });
+}
+
+function getDecimalSeparator(locale: string): DecimalSeparator {
+  try {
+    const formatter = new Intl.NumberFormat(locale);
```
camerow, line 82: An open question here, but do we want to allow for different formats? Will we have to normalize them once we start signing transactions and does that add complexity for us? @fbwoolf/@kyranjamie curious on your thoughts.

packages/ui/src/components/numpad/numpad.native.tsx
```diff
@@ -0,0 +1,159 @@
...
+    return formatter.format(1.1).includes(',') ? ',' : '.';
+  } catch (e) {
```
camerow, line 84:
```suggestion
  } catch {
```

camerow (commented):
Nice work overall Tigran!

packages/ui/src/components/numpad/numpad.native.tsx
tigranpetrossian: Great question. The numpad, and the display component that will feed off of it, deals with the string representation of the number in the correct format for based on the user's locale: period for English, comma for most non-English speaking users. This is the visual aspect of the interaction, it really only has to be a string because of intermediate trailing decimal zeros, i.e. the user is typing "12.00...123". Transforming this to a number further down the line regardless of the format is likely going to be trivial.

packages/ui/src/components/numpad/numpad-key.native.tsx
tigranpetrossian: Good call.  Honestly, for an component this small extracting this was an overkill to begin with. I'll inline this for now, as it's going to be removed in favor of SVGs anyway. 
I anticipate the **text or not** to be a recurring pattern though, so I'll look into making a more generic version of KeyElement as a helper component at some point.

packages/ui/src/components/numpad/numpad.native.tsx
tigranpetrossian: Agreed. Making a note to extract the logic in the upcoming iteration, along with other tweaks.

PR #590, opened 2024-11-01
feat: integrate stacks total balances, ref LEA-1500

pete-watters:
This PR:
- takes stacks balance queries into the mobile app 
- setups `LeatherQueryProvider` in mobile
- replaces the mocked token and account data with real data 

Here is a demo where I have loaded Leather Dev wallet and one of my own wallets to the mobile app. The data is real but you can observe some jankiness and incorrect data when going into accounts which I can investigate

https://github.com/user-attachments/assets/970f2d84-6855-421a-988a-f3de5ffaba95


apps/mobile/src/queries/balance/stacks-balance.query.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (!addresses || addresses.length === 0)
+    return { totalStxBalance: createMoney(0, 'STX'), accounts: [] };
+
+  return {
+    totalStxBalance: totalData,
+    accounts: queries, // TODO as yet unused but perhaps useful
+  };
+}
+
+function useStxBalancesQueries(addresses: string[]) {
```
pete-watters, line 35: This function is the workhorse to tie together all of the queries. 

I chose to pass in an array of addresses as we can likely refactor this out to the queries package if we want and use it in the extension by passing an array of a single address. 

Ideally I think we can merge the queries it calls into one bigger `useQueries` call but I was unsure to go ahead with that yet or keep things separate. 

Here we need to:
- load mempool data using `useMempoolTxsBalance` 
- `useMempoolTxsBalance` in turn has queries to load confimed transactions `useStacksConfirmedTransactions`

apps/mobile/src/queries/balance/stacks-balance.query.ts
```diff
@@ -0,0 +1,76 @@
...
+interface StxBalances {
+  totalStxBalance: Money;
+  accounts: QueryObserverResult<StxCryptoAssetBalance, unknown>[];
+}
+
+function useGetStxBalanceByAddresses(addresses: string[]): StxBalances {
+  const { totalData, queries } = useStxBalancesQueries(addresses);
+
+  if (!addresses || addresses.length === 0)
+    return { totalStxBalance: createMoney(0, 'STX'), accounts: [] };
```
pete-watters, line 27: Defaulting to `0` here is wrong. We need to prioritise [this issue](https://linear.app/leather-io/issue/LEA-505/loading-state-for-balances-on-home-screen)

apps/mobile/src/queries/stacks/use-transactions-with-transfers.query.ts
```diff
@@ -0,0 +1,41 @@
...
+  useCurrentNetworkState,
+  useStacksClient,
+} from '@leather.io/query';
+
+export function useGetAccountTransactionsWithTransfersQueries(addresses: string[]) {
+  const network = useCurrentNetworkState();
+  const client = useStacksClient();
+  const queries = useQueries({
+    queries: addresses.map(address => ({
+      ...createGetAccountTransactionsWithTransfersQueryOptions({
```
fbwoolf, line 17: We will refactor away from this deprecated endpoint later?

fbwoolf (approved):
Nice work here @pete-watters, lots of nice updates to our balance hooks.

apps/mobile/src/queries/balance/total-balance.query.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Money } from '@leather.io/models';
+import { sumMoney } from '@leather.io/utils';
+
+interface TotalBalance {
+  btcBalance: Money;
+  btcBalanceUsd: Money;
+  stxBalance: Money;
+  stxBalanceUsd: Money;
+  totalBalance: Money;
+}
```
kyranjamie, line 18: Maybe we should use naming `balanceFiat` to more generically apply to users preference

apps/mobile/src/store/accounts/utils.ts
```diff
@@ -14,4 +15,5 @@ export interface AccountStore {
   icon: AvatarIconName;
   name: string;
   status: AccountStatus;
+  type?: WalletStore['type'];
```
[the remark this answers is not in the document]
kyranjamie: Think we should just remove this, and the badge for now. 

Even when we do, this would better be read via selector by referencing parent wallet

kyranjamie (approved):

apps/mobile/src/queries/balance/stacks-balance.query.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+
+  return combinedResult();
+}
+
+export function useStxBalance(addresses: string[]) {
+  const { totalStxBalance } = useGetStxBalanceByAddresses(addresses);
+  const stxMarketData = useCryptoCurrencyMarketDataMeanAverage('STX');
+  const stxBalanceUsd = baseCurrencyAmountInQuote(totalStxBalance, stxMarketData);
+  return { availableBalance: totalStxBalance, fiatBalance: stxBalanceUsd };
```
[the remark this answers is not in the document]
pete-watters: It shouldn't and should update based on the chosen currency but when I had dug into the code I found that `i18nFormatCurrency` is hardcoded to use USD and will throw an error now if not USD

```
// @leather-io/utils/src/money/format-money.ts
export function i18nFormatCurrency(quantity: Money, decimals: number = 2) {
  if (quantity.symbol !== 'USD') throw new Error('Cannot format non-USD amounts');
  const currencyFormatter = new Intl.NumberFormat('en-US', {
    style: 'currency',
    currency: 'USD',
    maximumFractionDigits: decimals,
  });
```

I can update the naming so it's not USD but my intention was to make it more obvious that it's USD only right now

apps/mobile/src/queries/balance/total-balance.query.ts
```diff
@@ -0,0 +1,52 @@
...
+    totalBalance,
+  };
+}
+
+export function useAccountTotalBalance(accountId: AccountId): TotalBalance {
+  const stacksAddresses = useGetStacksAddresses(accountId);
+  const { availableBalance: stxBalance, fiatBalance: stxBalanceUsd } =
+    useStxBalance(stacksAddresses);
+  const { availableBalance: btcBalance, fiatBalance: btcBalanceUsd } =
+    useBitcoinAccountTotalBitcoinBalance(accountId);
```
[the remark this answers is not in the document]
pete-watters: The `useBitcoinAccountTotalBitcoinBalance`? That's from @kyranjamie in [this PR](https://github.com/leather-io/mono/pull/520)

We have slightly different approaches in that I call it:
- `useStxBalance` and it can return the balance of one or many stacks addresses

Kyran is using:
- `useWalletTotalBitcoinBalance` for multiple accounts
- `useBitcoinAccountTotalBitcoinBalance` for singular accounts 

We can standardise the naming? I think it's named that way for a reason though now. 

As an aside to this, I have [an open Q on that PR](https://github.com/leather-io/mono/pull/520/files#r1822656476) post merge  about naming of `useWalletTotalBitcoinBalance` 

> I forgot to double check that this calculates for multiple wallets. I think it does as keychains contains all accounts (across wallets)
> 
> Trying to think up a better name as `useWalletTotalBitcoinBalance` is singular. Maybe a simple change to useWallets is good enough?

apps/mobile/src/features/balances/stacks/stacks-balance.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface StacksTokenBalanceProps {
+  availableBalance: Money;
+  fiatBalance: Money;
+}
+export function StacksTokenBalance({ availableBalance, fiatBalance }: StacksTokenBalanceProps) {
+  return (
+    <TokenBalance
+      ticker="STX"
+      icon={<TokenIcon ticker="STX" />}
+      tokenName={t`Stacks`}
```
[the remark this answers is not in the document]
pete-watters: I started doing this but then stopped to think - do we even want to translate currency names? 

Probably not so we can hardcode it as `Stacks`

apps/mobile/src/queries/balance/total-balance.query.ts
pete-watters: I was explicitly calling it USD as `i18nFormatCurrency ` [is hardcoded to only accept USD](https://github.com/leather-io/mono/pull/590#discussion_r1827469125). 

I agree it's bad naming here though so maybe I should change it to `balanceFiat` now and create a task / project for adding multiple currency support?

apps/mobile/src/queries/stacks/use-mempool-txs-balance.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import { useMempoolTxsBalance } from './use-mempool-txs-balance';
+import { useStacksConfirmedTransactions } from './use-stacks-confirmed-transactions';
+import { useStacksPendingTransactions } from './use-stacks-pending-transactions';
+
+vi.mock('../stacks/use-stacks-confirmed-transactions');
+
+vi.mock('../stacks/use-stacks-pending-transactions');
+const mockAddresses = ['SP2417H88DQFN7FNDMSKM9N0B3Q6GNGEM40W7ZAZW'];
+
+const mockPendingTx = {
```
pete-watters, line 18: This adds some tests for inbound / outbound `tx`s

apps/mobile/src/queries/balance/stacks-balance.query.ts
pete-watters: I did test manually and it is working but I think I should take some extra time and add more unit tests to make sure. I added some unit tests [here](https://github.com/leather-io/mono/pull/590/files#r1827515351) but I can improve on them

We have no unit tests in the query package for the old code now so my plan was:
- add unit tests for the old code (single account based)
- obtain mocks for `pending` / more transaction types
- make sure the same tests pass with the new code (multi account code)
- further refactor the new code to merge the multiple `useQueries` into one
- either refactor the old code or just migrate `useStxBalancesQueries` to the query package and replace it in the extension also

To get up to speed with this code I had performed [some refactoring](https://github.com/leather-io/mono/pull/542/files#r1822790109) in the query package - in [this PR](https://github.com/leather-io/mono/pull/542) but left it out of here for now as it was bloating the PR and is not actually used at all on mobile. 

I can get working on this task right away, I opened a new linear issue - https://linear.app/leather-io/issue/LEA-1695/increase-test-coverage-for-stx-balances 

Ideally I should have done all of this here but the PR was already really big

apps/mobile/src/queries/stacks/use-transactions-with-transfers.query.ts
pete-watters: We will and I started looking into it but [it's a non-trivial task](https://github.com/leather-io/mono/pull/542#discussion_r1820903384). 

 In `createGetAccountTransactionsWithTransfersQueryOptions` we call `client.getAccountTransactionsWithTransfers` which uses `transactions_with_transfers` and [the API](https://docs.hiro.so/stacks/api/accounts/transactions-with-transfers) is deprecated now with no direct replacement provided. 
 
 I thought I had found a replacement but that was for a different API `transaction-with-transfers` where you need to provide the transaction. 
 
I think we need to use [Get address transactions](https://docs.hiro.so/stacks/api/transactions/address-transactions) and then get filter based on `"tx_type": "token_transfer",`.  I started some work on this in the larger [PR here](https://github.com/leather-io/mono/pull/542/files#r1822787679) but stopped as it was too much to change in one go. 

It's something I can add to https://linear.app/leather-io/issue/LEA-1695/increase-test-coverage-for-stx-balances and tackle it once we have more solid tests in place

pete-watters:
@kyranjamie @fbwoolf - thanks for the review!

I'll merge this now as it's adding some good value. There are some on-going threads I will keep an eye on and can update as part of subsequent work

PR #594, opened 2024-11-04
chore: upgrade stacks/transactions to v6.17 for non-seq multisig support

alexp3y:
Upgrading @stacks/transactions to v6.17 to support [SIP-027](https://www.hiro.so/blog/breaking-down-the-multisig-improvements-coming-to-stacks-with-sip-027).

Slack discussion [here](https://trustmachines.slack.com/archives/C06RADSM2TX/p1730581514660509).

Corresponding [extension PR](https://github.com/leather-io/extension/pull/5935).


kyranjamie (approved):
Note that a `chore` commit won't trigger publishes to republish

fbwoolf (approved):

PR #595, opened 2024-11-04
feat: architecture diagram

kyranjamie:
I've been trying out [d2](https://d2lang.com/) to present our architecture. This turns [Alex's architecture diagram](https://www.notion.so/trustmachines/Architecture-Proposal-118d74b694c780389ddfd24c8454924b?pvs=4#118d74b694c78030bbadea7635010c0d) into code we can commit. Describing our architecture with code is nice as we can see it evolve, and discuss proposals alongside PRs that illustrate the changes. The output isn't as beautiful as a hand made one, but it conveys the point well enough, I think.

When changed, CI builds the diagram, and commits it to the `architecture` branch, so we can reference it in the main README.md.

architecture diagram


architecture/leather-architecture.d2
```diff
@@ -0,0 +1,155 @@
...
+  style: {
+    font-size: 44
+    italic: true
+    font-color: '#12100F'
+  }
+}
+
+mono: {
+  label: ''
+  entity layer: {
```
alexp3y, line 38: Don't think we need this extra "entity layer". Models already kind of implicitly define and form the core of the domain.

alexp3y (commented):
Looks great. Sounds like a solid way to track architecture going forward.

architecture/leather-architecture.d2
kyranjamie: I separated this just to make absolutely clear that at the core we have models that simply define the structure of the data 

Many (not all) onion style diagrams have entity at the centre, or divide domain services and models

image

image

alexp3y (approved):

fbwoolf (approved):

PR #598, opened 2024-11-04
feat: receive sheet navigator, closes LEA-1699

fbwoolf:
This PR refactors a bit of what I just merged with adding the send sheet navigator so that we can scale it with other sheet flows. In addition, this adds the receive navigator and sets up that folder/file structure so we can build on it in tandem with the send flow. Right now, I left a bit of redundancy with `SelectAccount` sheets sharing an accounts list feature, but we might find it is best just to share a sheet for selecting an account later.


apps/mobile/src/app/_layout.tsx
```diff
@@ -51,12 +53,13 @@ export default function RootLayout() {
                   <SplashScreenGuard>
                     <GestureHandlerRootView style={{ flex: 1 }}>
                       <SheetProvider>
-                        <SendSheetProvider value={{ sendSheetRef }}>
+                        <SheetNavigatorProvider value={{ sendSheetRef, receiveSheetRef }}>
```
camerow, line 56: Is the idea for having different named refs to allow for popping up different sheets over one another? Just wondering if it makes sense to have a single ref `fullPageSheetRef`?

camerow: Probably does make sense to create refs per sheet for maximum flexibility.

camerow: wdyt about moving the logic of creating the provider and the refs into a separate component so that `_layout` doesn't know anything about how that all works, it just wraps the page in another "layer" - something like `<FullsheetWrapper>`? where that looks like 

```
export default function FullsheetWrapper({ children }) {
  const sendSheetRef = useRef<SheetRef>(null);
  const receiveSheetRef = useRef<SheetRef>(null);
  
  return (<SheetNavigatorProvider value={{ sendSheetRef, receiveSheetRef }}> { children }</<SheetNavigatorProvider>);
  }```

camerow:
Nice Fara! 🚀

camerow (approved):

apps/mobile/src/app/_layout.tsx
fbwoolf: Yep, I can move into a wrapper, np. Prob a good idea so the _layout doesn't become bloated overtime.

pete-watters: In case it helps I put `LeatherProvider` into `src/common/` recently and plan to move `ThemeProvider` + `AppRouter` into there too. 

Originally I create files directly under `app/` but Edgar said that would expose them to the router so not a good idea

pete-watters (approved):
Nice work Fara 🚀

kyranjamie (approved):
No comments, tidy work Fara

camerow:
FYI just labeling this with a build to show Michelly how we can run PR builds.

PR #600, opened 2024-11-05
fix(query): stx token specific fee setting

kyranjamie:
Possible quick win fix Re discussion on fees


kyranjamie (commented):
LGTM thanks @alter-eggo

PR #602, opened 2024-11-05
refactor(analytics)!: prepare analytics client for use in extension

camerow:
Some small compatibility and type refactors for using this client in the extension codebase.

- Made some events a bit more permissive
- Made the generic interface types more permissive for better compatibility between browser/native segment client
- Pass through function calling for methods we don't need types on right now (screen, page, identify
- Surfaced underlying `client` in the analytics for flexibility
- Updated tests

### DIFF
--- a/packages/analytics/src/events.ts
+++ b/packages/analytics/src/events.ts
@@ -12,10 +12,10 @@ interface HistoricalEvents {
   broadcast_ordinal_error: { error: any };
   broadcast_transaction: {
     symbol: string;
-    amount: number;
-    fee: number;
-    inputs: number;
-    outputs: number;
+    amount?: number;
+    fee?: number;
+    inputs?: number;
+    outputs?: number;
   };
   broadcast_btc_error: { error: any };
   copy_btc_address_to_clipboard: { type: string };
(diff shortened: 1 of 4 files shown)


packages/analytics/src/events.ts
```diff
@@ -11,11 +11,11 @@ interface HistoricalEvents {
   broadcast_ordinal_transaction: undefined;
   broadcast_ordinal_error: { error: any };
   broadcast_transaction: {
-    symbol: string;
-    amount: number;
-    fee: number;
-    inputs: number;
-    outputs: number;
+    symbol?: string;
```
pete-watters, line 14: This looks a bit unusual. It seems like we do always need `symbol` for this? 

As it's for analytics maybe we can just have a more generic `Record` here?

We call this 3 times and only once does it have only `symbol` so maybe instead we should feed more parameters in that case? 

```
// use-submit-stx-transactions
 void analytics.track('broadcast_transaction', { symbol: 'stx' });
```

pete-watters (approved):
NICE WORK 🚀 

(capitals was on and I liked it better then)

packages/analytics/src/events.ts
```diff
@@ -89,6 +95,9 @@ interface HistoricalEvents {
   ledger_message_signed_approved: undefined;
   ledger_message_signed_rejected: undefined;
   ledger_public_keys_pulled_from_device: undefined;
+  user_clicked_requested_by_link: { endpoint: string };
+  user_approved_get_addresses: { origin?: string };
+  user_approved_message_signing: { origin?: string };
```
kyranjamie, line 100: There should always be an origin on these

kyranjamie (approved):

packages/analytics/src/events.ts
camerow: Trying to just inline replace and not change much yet.

Issue #72, opened 2024-02-26
Design system - Typography update


mica000:
**This ticket is to update the typography scale to cover @fbwoolf comment:**
"We used to have a Title component in stacks/ui, so when we did the refactor we did make one using label01 atm to make the refactor easier. We could remove it and just style each piece of text separately, or we could update that component to share across the wallet still. It seems it should be label02 though and checked where used. Also, currently our label02 doesn NOT have a fontWeight of 500, so that needs to be changed. My thought is that with a separate coordination issue typography overall should be revisited again llke we did for colors."

https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=1257-1463&mode=design

![image](https://github.com/leather-wallet/mono/assets/119523541/2246445e-00f0-4c96-9fdd-619de73d7308)

kyranjamie:
This is an important one for us to get done. Is it ready for development @mica000 ?

mica000:
@fabric-8 curious on your take in regards to Caption02 - it's not very crispy but I added it here to cover to the old typography system we had to avoid things breaking when launching the rebrand - it's not particularly in use by the new designs system, shall we remove it?

fabric-8:
Yes I think it's fine to remove 👍

mica000:
There is on more change we need to make which is to remove Caption 01. 

**Context:**
The old designs system had these in the scale so I preserved them during rebranding to avoid breaking the layout however body02 and caption 01 were the same. This wasn't good so they are now merged. This means that now we have only 1 size for Body and 1 size for Caption. [See design](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=5833-45783&mode=design&t=cKYocv4FIHdeUZmM-4)

On the design system in Figma I already replaced the old Body02 for caption 01 (now new Caption). Components Where it was replaced:
• Call out 
• Input
• Tooltip

**On development, Kyran suggested:**
[x] Finalise the typography in Figma, removing the duplicate entry. First principals: what typography do we want? No legacy design system decisions should influence our new design system. 
[ ] Remove all fontWeight definitions across the whole product in one find replace.
[ ] Dev and designer live pair together going through all UI flows, fixing what's broken on the spot


**Here is the new scale**
[...]

PR #603, opened 2024-11-06
feat: Add AddressDisplayer for mobile, closes LEA-1638

tigranpetrossian:
Adds native version of AddressDisplayer, adds Fira Code and addresses some minor font/styling issues as a prerequisite:

1. Added Fira Code for mobile, new and correct weights for the web (had to compile woff2 and otf myself from source, as they weren't present in the official build 😬), updated references everywhere.
2. Update storybook for native, update the build to make sure all fonts are displayed correctly.
3. Add `address` textStyle to tokens, as `code` that's currently used for the web version of the AddressDisplayer doesn't match the new designs. I'm not extremely happy with the text style name this specific, [brought this up](https://www.figma.com/design/2MLHeIeL6XPVi3Tc2DfFCr?node-id=11410-32620&m=dev#1014894187) with the design team to see if it's worth using something more generic.
4. Refactored the AddressDisplayer for web to match the current designs: 
	1. Used the new text style
	2. Added a flexbox wrapper to correct line height (didn't work with inline context), replaced margin with `1ch` wide gaps to avoid trailing spaces.
	3. Had to merge into a single component, the layout pattern didn't make sense after doing the above, thankfully this is a fairly small component.
5.	Added native version of the component

@edgarkhanzadian I would really appreciate your help double checking the config and all the artifacts expo generated after running `prebuild` to bundle fonts. I'm way out of my depth when it comes to this kind of thing.


apps/mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,12 +1,12 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```
kyranjamie, line 2: 🔤

tigranpetrossian:
@kyranjamie Missed an important part: extension will likely need a small PR to address the changes to the component here. Two questions:
1. Is there script that pulls the [base.css](https://github.com/leather-io/mono/blob/dev/packages/ui/src/assets-web/base.css) and font files from the UI package automatically, or is it done manually?
2. How do we typically orchestrate releases when there's a potentially breaking change in the monorepo that affects the extension?

packages/ui/src/components/address-displayer/address-displayer.web.tsx
```diff
@@ -1,21 +1,26 @@
...
-import { AddressDisplayerLayout } from './address-displayer.layout.web';
 import { groupByFour } from './address-displayer.utils.shared';
 
 interface AddressDisplayerProps extends HTMLStyledProps<'span'> {
   address: string;
 }
 export function AddressDisplayer({ address, ...props }: AddressDisplayerProps) {
   return (
-    <>
+    <Flex direction="row" columnGap="1ch" flexWrap="wrap">
```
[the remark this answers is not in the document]
tigranpetrossian: Definitely. If we ever want to expose text-related options, that should be done via more constrained prop variants.

kyranjamie:
> Is there script that pulls the [base.css](https://github.com/leather-io/mono/blob/dev/packages/ui/src/assets-web/base.css) and font files from the UI package automatically, or is it done manually?

No, and I didn't realise they were copied. We shouldn't have duplicates of this file, ideally the base.css in the extension reads this one in the mono.

> How do we typically orchestrate releases when there's a potentially breaking change in the monorepo that affects the extension?

This is a mostly manual process currently. Suggest linking the packages locally, making a PR in the extension (with a broken build because the package hasn't been released yet), then updating the PR when you've released the monorepo changes.

packages/ui/package.json
```diff
@@ -17,7 +17,9 @@
     "storybook:android": "expo start --android",
     "storybook:ios": "expo start --ios",
     "storybook:web": "storybook dev -p 6006 -c ./src/.storybook-web",
-    "typecheck": "tsc --noEmit --project ./tsconfig.json"
+    "typecheck": "tsc --noEmit --project ./tsconfig.json",
+    "android": "expo run:android",
+    "ios": "expo run:ios"
```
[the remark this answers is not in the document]
tigranpetrossian: I did run `prebuild` on both the app and the storybook yeah, thinking that the way to go to integrate fonts, this also incidentally fixed the storybook developoment not working. Happy to scrape this in favor of runtime loading in storybook, going to give it a shot and see if I can make it work.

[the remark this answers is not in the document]
tigranpetrossian: Dropped that commit in favor of `useFonts` in storybook, this is much better. Thanks @edgarkhanzadian!

packages/tokens/src/typography.ts
```diff
@@ -219,6 +227,11 @@ export function getWebTextVariants() {
       description: 'code',
       value: textVariants.code,
     },
+
+    address: {
```
pete-watters, line 231: Usually we match these exactly to the design system but I think this is a nice place to put this also. I wonder does it need to be in the global typography though as it's only really used in address-displayer?

We have this other long standing task to clean up typography on mobile, maybe you'd like to investigate that as you were working on this? 

https://github.com/leather-io/mono/issues/72

It's unclear to me what exactly needs to be done there or if it's done already but it would be nice to close it off if we can

packages/ui/src/components/address-displayer/address-displayer.web.tsx
```diff
@@ -1,21 +1,24 @@
...
 
-interface AddressDisplayerProps extends HTMLStyledProps<'span'> {
-  address: string;
-}
+type AddressDisplayerProps = AddressDisplayerBaseProps & FlexProps;
+
 export function AddressDisplayer({ address, ...props }: AddressDisplayerProps) {
   return (
-    <>
+    <Flex direction="row" columnGap="1ch" flexWrap="wrap" {...props}>
```
pete-watters, line 12: Nice work here, inspiring me to learn about `ch` 👍

pete-watters (approved):
Nice work here @tigranpetrossian , this is a really clean and valuable PR 🚀

pete-watters:
> @kyranjamie Missed an important part: extension will likely need a small PR to address the changes to the component here. Two questions:
> 
> 1. Is there script that pulls the [base.css](https://github.com/leather-io/mono/blob/dev/packages/ui/src/assets-web/base.css) and font files from the UI package automatically, or is it done manually?
> 2. How do we typically orchestrate releases when there's a potentially breaking change in the monorepo that affects the extension?

When originally building the UI package we decided against hosting the font files and static assets in there and instead keeping them local to the `extension`. That's the reason for this `base.css` duplication.

It could be better to remove that duplication and keep everything in the UI library but we would need to do some work on bundling the font files for export with the UI package. We also have other static assets in `extension/public` anyway so I think it's OK to have small duplication here to avoid more complication.

packages/tokens/src/typography.ts
tigranpetrossian: I've been initially reluctant about adding this to the typography tokens too, but hardcoding in two places felt even messier. Already asked the design team to see if there are any other use-cases for this so we can come up with a more generic style name.

Re: #72, I've actually set out to refactor the typography a little bit, but it started leaking out of the scope of this task very quickly, so I'm planning to do that separately some time soon. Might be a good place to figure out the `address` as a style conundrum too.

tigranpetrossian:
> Nice work here @tigranpetrossian , this is a really clean and valuable PR 🚀
> 
> If you run ` pnpm syncpack:format` it should fix your lint error

Thanks! Did this and it seems fine. I suspect the errors are either because I either modified the history of the branch too much or we need to add a `pnpm install` prior to running these scripts. Will investigate.

tigranpetrossian:
> When originally building the UI package we decided against hosting the font files and static assets in there and instead keeping them local to the `extension`. That's the reason for this `base.css` duplication.
> 
> It could be better to remove that duplication and keep everything in the UI library but we would need to do some work on bundling the font files for export with the UI package. We also have other static assets in `extension/public` anyway so I think it's OK to have small duplication here to avoid more complication.

Agreed. I'm looking into this. If it turns out to be too complicated we'll tackle it separately or worst case scenario stick with duplication.

packages/tokens/src/typography.ts
pete-watters: Sounds great! I just mentioned https://github.com/leather-io/mono/issues/72 as it would be good to keep that in mind and find out what exactly else needs to be done there if you are refactoring the typography. 

I agree lets not do it here 👍

apps/mobile/android/app/src/main/AndroidManifest.xml
tigranpetrossian: I guess expo is going to keep doing this later if I roll back now, might as well keep it as is.

packages/ui/src/components/address-displayer/address-displayer.native.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  },
+};
+
+export const Stacks: Story = {
+  args: {
+    address: 'SP3XKZE32KZ925AAAGWPG1W66YP5BM2RD73T6AJHS',
+  },
+};
+
+export const WithBTCIcon: Story = {
```
fbwoolf, line 29: Typ we've been using `Btc` and `Stx` in naming.

fbwoolf (approved):
Lgtm @tigranpetrossian 👍

packages/ui/src/components/address-displayer/address-displayer.native.stories.tsx
kyranjamie: https://www.notion.so/trustmachines/Naming-Camelcase-acronyms-119d74b694c78080974aee5e3bc18bc9?pvs=4

camerow:
I saw this was sitting in the merge queue for ~6 hours. Not sure what is going on but tried removing/readding it.

fbwoolf:
Trying again.

PR #604, opened 2024-11-06
draft: testing biometric auth on android

camerow:


markmhendrickson:
@camerow let's get this into Linear?

PR #605, opened 2024-11-07
fix: import all accounts for wallet

pete-watters:
This PR contains a working draft of importing multiple accounts from a seed. 

It is based on previous extension code but some more work is needed.

When importing the `dev` wallet account it takes quite a while and the resulting UI is very laggy (maybe as we have no controls in place on the max limits of the accounts)


apps/mobile/package.json
```diff
@@ -61,14 +61,15 @@
     "@rnx-kit/metro-config": "1.3.14",
     "@rnx-kit/metro-resolver-symlinks": "0.1.35",
     "@scure/bip32": "1.5.0",
+    "@scure/bip39": "1.4.0",
```
pete-watters, line 64: Added this for `mnemonicToSeedSync` in `mnemonicToRootNode` but maybe we have a different way on mobile?

apps/mobile/src/store/key-store.ts
```diff
@@ -109,6 +127,59 @@ export function useKeyStore() {
...
+            // const hasBtcBalance = await doesBitcoinAddressHaveBalance(btcAddress.address!);
+            // TODO: add inscription check here also?
+            return hasStxBalance; // || hasNames || hasBtcBalance;
+          },
+        }).then(recursiveActivityIndex => {
+          if (recursiveActivityIndex <= legacyAccountActivityLookup) return;
+          console.info('Found account activity at higher index', { recursiveActivityIndex });
+          // dispatch(stxChainSlice.actions.restoreAccountIndex(recursiveActivityIndex));
+          console.log('recursiveActivityIndex', recursiveActivityIndex);
+          this.createNewAccountOfWallet(fingerprint, recursiveActivityIndex);
```
pete-watters, line 169: In the extension `stx-chain.slice.ts` for `stxChainSlice.actions.restoreAccountIndex(recursiveActivityIndex)` it is using `highestAccountIndex` which gets incremented in `createNewAccount` so maybe that's where I am going wrong.

apps/mobile/package.json
kyranjamie: both of these methods exist in `@leather.io/crypto`

apps/mobile/package.json
```diff
@@ -142,7 +143,7 @@
     "babel-plugin-macros": "3.1.0",
     "babel-preset-expo": "11.0.6",
     "concurrently": "8.2.2",
-    "eslint": "8.53.0",
+    "eslint": "9.14.0",
```
kyranjamie, line 146: unintended I imagine

apps/mobile/src/store/accounts/recurse-accounts.ts
```diff
@@ -0,0 +1,79 @@
+// from extension src/app/common/account-restoration/account-restore.ts
```
kyranjamie, line 1: this shouldn't be added directly to mobile, instead though of as a distinct module that I think should like in `/crypto` pkg, the refactored out to extension

apps/mobile/src/store/key-store.ts
```diff
@@ -109,6 +127,60 @@ export function useKeyStore() {
...
+      const secretKey = mnemonic;
+
+      const legacyAccountActivityLookupGaia =
+        await checkForLegacyGaiaConfigWithKnownGeneratedAccountIndex(secretKey);
+
+      // hardcoding this to match extension count for dev wallet
+      // checkForLegacyGaiaConfigWithKnownGeneratedAccountIndex fails with
+      // [TypeError: _this.nodePbkdf2 is not a function (it is undefined)]
+      const legacyAccountActivityLookup =
+        legacyAccountActivityLookupGaia === 0 ? 123 : legacyAccountActivityLookupGaia;
```
kyranjamie, line 140: suggest we don't add any of the gaia checks here, resolving the import issue, check stacks/bitcoin balances only

apps/mobile/src/store/key-store.ts
kyranjamie: There should not be side effects in this callback.

The idea of this recursive look up is that we give it a callback to determine whether there's activity or not. Then, the response (which isn't used here because of `void recurseAccountsForActivity`) tell you the highest account index that has activity. 

Later, you use that index to know how many accounts should be created.

apps/mobile/package.json
pete-watters: I saw some warnings about it being out of date so updated it. I should have removed from here

apps/mobile/src/store/accounts/recurse-accounts.ts
pete-watters: I will do this once I can get it to work properly

kyranjamie:
pseudo code we need here

```ts
// user restores mnemonic
const mnemonic = 'abandon …';

async function doesStacksAddressHaveBalance (index) {
  // derive address
  // look up api for balances, return true if found
}
async  function doesNativeSegwitAddressHaveBalance (index) {
  // derive address
  // look up api for balances, return true if found
}

const highestAccountIndex = recurseAccountsForActivity({
  async doesAddressHaveActivityFn(index) {
    const hasStacksBal = await doesStacksAddressHaveBalance(index);
    const hasNsBal = await doesNativeSegwitAddressHaveBalance(index)
    return hasStacksBal ?? hasNsBal;
  }
})

// now we know that `highestAccountIndex` is the account with the farthest known balance

const newAccountsToAdd = createNullArrayOf(highestAccountIndex).map(index => {
  return theAccountDetailsWeNeedToPersistToStore();
})
```

apps/mobile/src/store/key-store.ts
kyranjamie: I se here you're try to recurse as well, but that isn't needed because all the recursion is handled internally to this function, it's way easier that way

pete-watters: I based this code on the extension code commented out above it. That was using `dispatch(stxChainSlice.actions.restoreAccountIndex(recursiveActivityIndex))` to restore accounts. 

I could be mistaken, but when I debug this, it seems like in `.then` the `recursiveActivityIndex` is always `0` so this code is never reached (on extension or copied version). See extension [draft](https://github.com/leather-io/extension/pull/5939/files) . 

I think accounts are actually created further on by:
```
 if (legacyAccountActivityLookup !== 0)
        // dispatch(stxChainSlice.actions.restoreAccountIndex(legacyAccountActivityLookup)); -> extension
        this.createNewAccountOfWallet(fingerprint, legacyAccountActivityLookup);
```

`legacyAccountActivityLookup` is determined by `checkForLegacyGaiaConfigWithKnownGeneratedAccountIndex` which I think is actually what is determining how many accounts to create. 

What I think I can get to work is to run `this.createNewAccountOfWallet` inside of the recursive function if there is a balance

apps/mobile/src/store/key-store.ts
```diff
@@ -109,6 +127,60 @@ export function useKeyStore() {
...
+          // dispatch(stxChainSlice.actions.restoreAccountIndex(recursiveActivityIndex));
+          console.log('recursiveActivityIndex', recursiveActivityIndex);
+          this.createNewAccountOfWallet(fingerprint, recursiveActivityIndex);
+        });
+      } catch {
+        // Errors during account restore are non-critical and can fail silently
+      }
+
+      // on the extension dev wallet this is where most accounts are created
+      if (legacyAccountActivityLookup !== 0)
```
pete-watters, line 177: If I try restore one of my personal wallets on the `extenstion` the accounts also get generated here. 

That account is from 2023 at the earliest so I don't think it should be considered legacy?

apps/mobile/src/store/key-store.ts
kyranjamie: Why does it always return 0? Does the account have activity you're testing with? It's quite slow, so are you waiting until the end? 

image

> What I think I can get to work is to run this.createNewAccountOfWallet inside of the recursive function if there is a balance

I really don't think this is the way to do it. The purpose of `recurseAccountsForActivity` is to handle the recursion of the look up. If you have to do additional recursion outside that fn, then either your implementation misses the point of the function, or the implementation of `recurseAccountsForActivity` doesn't work and needs fixing.

Happy to run through in more detail the gist of this function, but really its just a look up function that stops after _n_ false values and returns a number.

Also, let's remove all the gaia stuff so we don't need to worry about legacy account look up code.

pete-watters: Thanks Kyran. I will do more testing and then implement the best solution using your recommendations. 

I'm not sure if it's working as we think in the extension right now. I am testing with my personal wallet and the leather dev wallet and in both cases accounts are created only if `if (legacyAccountActivityLookup !== 0)`.

We have two calls to add accounts:
- in `recurseAccountsForActivity.then` 
`dispatch(stxChainSlice.actions.restoreAccountIndex(recursiveActivityIndex));`
- after a check for `if (legacyAccountActivityLookup !== 0)`:
`dispatch(stxChainSlice.actions.restoreAccountIndex(legacyAccountActivityLookup));`


- If I remove the `dispatch` after the `legacyAccountActivityLookup` check, no accounts are created apart from the first one. 
- If I add a debugger in `recurseAccountsForActivity.then`  it never gets hit. 

Here's the full code. 
```
 try {
      void recurseAccountsForActivity({
        async doesAddressHaveActivityFn(index) {
          ....
          return hasStxBalance || hasNames || hasBtcBalance;
        },
      }).then(recursiveActivityIndex => {
        if (recursiveActivityIndex <= legacyAccountActivityLookup) return;
        logger.info('Found account activity at higher index', { recursiveActivityIndex });
        dispatch(stxChainSlice.actions.restoreAccountIndex(recursiveActivityIndex));
      });
    } catch (e) {
      // Errors during account restore are non-critical and can fail silently
    }

 ....
    if (legacyAccountActivityLookup !== 0)
      dispatch(stxChainSlice.actions.restoreAccountIndex(legacyAccountActivityLookup));
  };
```

If I change this code to be:
```
const accountsWithBalances = recurseAccountsForActivity({
        async doesAddressHaveActivityFn(index) {
          ....
          return hasStxBalance || hasNames || hasBtcBalance;
        },
      })
```

`accountsWithBalances` is always `0`. If I add logs inside of there I can see it finding accounts with balances but I think it's not working as we expect.

pete-watters: Upon further investigation into this, it seems like the recursion we are using does not generate any accounts. They are in fact all generated initially based on the count returned from `checkForLegacyGaiaConfigWithKnownGeneratedAccountIndex` 

Here is a demo PR from the extension showing that: https://github.com/leather-io/extension/pull/5945

apps/mobile/src/store/key-store.ts
```diff
@@ -89,6 +105,26 @@ export function useKeyStore() {
...
+        await this.deriveNextAccountKeychainsFrom(fingerprint);
+
+      dispatch(
+        userAddsAccounts(
+          Array.from({ length: activeAccounts }, (_, i) => ({
+            account: {
+              id: makeAccountIdentifer(fingerprint, i),
+            },
+            // not sure why this is needed?
+            // don't think I need to pass in the same keychain for each account
```
kyranjamie, line 119: the keychain is different for every account

pete-watters: Thanks @kyranjamie 👍  

I realised I have more work to do as I was hitting issues in `bitcoin-keychains.read.ts` with undefined `nativeSegwit  taproot`.

apps/mobile/src/store/keychains/bitcoin/bitcoin-keychains.write.ts
```diff
@@ -32,6 +32,12 @@ export const bitcoinKeychainSlice = createSlice({
         handleEntityActionWith(adapter.addMany, payload => payload.withKeychains.bitcoin)
       )
 
+      // NOTE: need to adapt this to add multiple accounts
```
kyranjamie, line 35: Curious about this comment. This adapter already uses `addMany`, not `addOne`

pete-watters: I added a new action `userAddsAccounts` to add many accounts at once so I think I need to adapt this to also handle their keychains as that's not working fully and the accounts seem incomplete. 

I am hitting issues with undefined `keyOrigin` so I think I missed some things during account creation

pete-watters: Solved this. I'm passing an array of account creation data so needed a `flatMap` here

packages/crypto/src/recurse-accounts.ts
```diff
@@ -0,0 +1,77 @@
...
+      const hasActivity = await doesAddressHaveActivityFn(indexCounter.getValue());
+      activity.push({ index: indexCounter.getValue(), hasActivity });
+      indexCounter.increment();
+      yield;
+    }
+    return returnHighestIndex(activity);
+  }
+  const mostRecentAccount = await recurseUntilGeneratorDone(checkForMostRecentAccount());
+  const accountsToRestore = mostRecentAccount;
+  return accountsToRestore;
```
kyranjamie, line 76: odd reassignments here

```suggestion
  const accountsToRestore = await recurseUntilGeneratorDone(checkForMostRecentAccount());
  return accountsToRestore;
```

apps/mobile/src/store/keychains/stacks/utils.ts
```diff
@@ -1,6 +1,16 @@
...
+  AddressVersion,
+  createStacksPrivateKey,
+  getPublicKey,
+  publicKeyToAddress,
+} from '@stacks/transactions';
+import { deriveStxPrivateKey } from '@stacks/wallet-sdk';
 import z from 'zod';
 
+// FIXME fix this deprecated import
+import { mnemonicToRootNode } from '@leather.io/bitcoin';
```
kyranjamie, line 12: this is in crypto pkg

pete-watters: I'll be removing this whole file in favour of the newer balance queries we have

kyranjamie:
I wonder how we're going to manage this functionality, because in the case of lots of accounts, it's going to take ages

pete-watters:
I'm closing this off for now as I want to start work on loading activity. Right now to determine if an address is active we are checking it's balance but I think if we are to change this approach anyway we should first investigate if it's better to check using actual activity rather than balance

PR #606, opened 2024-11-07
feat: get bns name from bns v2 api

alter-eggo:
This pr adds fetching bns names from bns api

### DIFF
--- a/packages/query/src/stacks/bns/bns.utils.ts
+++ b/packages/query/src/stacks/bns/bns.utils.ts
@@ -1,37 +1,63 @@
 import { parseZoneFile } from '@fungible-systems/zone-file';
 import { BnsNamesOwnByAddressResponse } from '@stacks/stacks-blockchain-api-types';
+import axios from 'axios';
+import { getPrimaryName } from 'bns-v2-sdk';
+import { z } from 'zod';
 
+import { BNS_V2_API_BASE_URL, NetworkModes } from '@leather.io/models';
 import { isString, isUndefined } from '@leather.io/utils';
 
 import { StacksClient } from '../stacks-client';
+import { bnsV2NamesByAddressSchema } from './bns.schemas';
 
 /**
  * Fetch names owned by an address.
  */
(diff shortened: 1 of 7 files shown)


packages/query/src/stacks/bns/bns.utils.ts
```diff
@@ -1,37 +1,64 @@
...
   client: StacksClient;
   address: string;
-  isTestnet: boolean;
+  network: NetworkModes;
   signal: AbortSignal;
 }
+
+type BnsV2NamesByAddressResponse = z.infer<typeof bnsV2NamesByAddressSchema>;
+
+async function fetchPrimaryName(address: string, network: NetworkModes) {
```
camerow, line 25: Worth adding tests for these? `fetchNamesForAddress` is doing some filtering so a unit test there may be nice to validate that behavior.

Perhaps codecov is a little aggressive, eg we don't really want to test `schema.ts` files - maybe for now we want to opt certain directories in?

camerow (approved):

pete-watters (approved):
Nice work @alter-eggo 👍

packages/query/src/stacks/bns/bns.utils.ts
```diff
@@ -1,37 +1,63 @@
...
-  isTestnet,
   signal,
+  network,
 }: FetchNamesForAddressArgs): Promise<BnsNamesOwnByAddressResponse> {
-  const fetchFromApi = async () => {
-    return client.getNamesOwnedByAddress(address, signal);
-  };
-  if (isTestnet) {
-    return fetchFromApi();
+  const res = await axios.get<BnsV2NamesByAddressResponse>(
```
pete-watters, line 39: A _nit_ suggestion here would be to:
- call `res` => `namesResponse`
- call `namesResponse` => `bnsNames`
- use a constant name `const addressHasMultipleNames = namesResponse.length > 1`
    - that way you can add your comment but it's in code not a comment

PR #609, opened 2024-11-08
feat: approver ux UI

edgarkhanzadian:
Well probably too many changes for 1 PR but oh well. 😅
- Implement Approver UX ui trying to keep the same Approval structure as on web
- Swap NavigationContainer with BottomSheetScrollView to unblock the scroll view on send sheet navigation
- Move Prism highlighting from extension to monorepo, unblock https://github.com/leather-io/extension/pull/5943
- Implement Prism highlighting in mobile

Initial route changed here to showcase the approver ux

https://github.com/user-attachments/assets/46e68aae-83e5-4171-bab0-6d5fe4e443b8


apps/mobile/src/features/approval-ux/approval.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { Box, Text, Theme } from '@leather.io/ui/native';
+
+import { AdvancedDetailsButton } from './components/advanced-details-button';
+import { AccountCard } from './components/cards/account-card';
+import { CodeCard } from './components/cards/code-card';
+import { FeeCard } from './components/cards/fee-card';
+import { InputsAndOutputsCard } from './components/cards/inputs-outputs-card';
+import { OutcomesCard } from './components/cards/outcomes-card';
+import { Footer } from './components/footer';
```
kyranjamie, line 16: The core styles for approver ux should be in UI package `src/components/approver`. 

Also approver should not be "aware" of any details like fees/inputs etc. `signPsbt` should be an instance of `<Approver />`

apps/mobile/src/features/approver/approver.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { ApproverFooter } from './components/approver-footer';
+import { ApproverHeader } from './components/approver-header';
+import { ApproverSection } from './components/approver-section';
+import { AccountCard } from './components/cards/account-card';
+import { CodeCard } from './components/cards/code-card';
+import { FeeCard } from './components/cards/fee-card';
+import { InputsAndOutputsCard } from './components/cards/inputs-outputs-card';
+import { OutcomesCard } from './components/cards/outcomes-card';
+
+export function Approver({ children }: HasChildren) {
```
kyranjamie, line 21: Can we colocate this in `packages/ui/src/components/approver` though, or are you just working in mobile first and will migrate later for ease of dev?

kyranjamie: for example, I think we can even reuse the same approver context for both implementations I think? As there's no platform specific behaviour inside (at least there shouldn't be)

apps/mobile/src/components/full-height-sheet/full-height-sheet.tsx
```diff
@@ -12,7 +12,12 @@ export function FullHeightSheet({ children, sheetRef }: FullHeightSheetProps) {
   const { themeDerivedFromThemePreference } = useSettings();
 
   return (
-    <Sheet isFullHeight isScrollView ref={sheetRef} themeVariant={themeDerivedFromThemePreference}>
+    <Sheet
+      isFullHeight
+      shouldHaveContainer={false}
```
fbwoolf, line 17: I made this change too locally, so came across the same issue. 👍

apps/mobile/src/features/approver/components/approver-account-card.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export function ApproverAccountCard() {
+  const theme = useTheme<Theme>();
+  const { list: walletsList } = useWallets();
+  const wallet = walletsList[0];
+  /* eslint-disable-next-line lingui/no-unlocalized-strings  */
+  const testAccountName = 'testName';
+  return (
+    <>
+      <Text variant="label01">{t`With account`}</Text>
```
fbwoolf, line 17: Add these thru Crowdin?

apps/mobile/src/features/approver/components/approver-account-card.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <AccountListItem
+        accountName={testAccountName}
+        address={t`Address`}
+        balance={t`$1234`}
```
fbwoolf, line 21: I made an `AccountBalance` and `AccountAddress` component in my PR that can go here at some point.

fbwoolf (approved):
This looks great @edgarkhanzadian. I may have created some conflicts for you bc I didn't realize you were moving stuff arnd with the send form routes. Let me know if you need help resolving.

packages/ui/src/components/approver/approver.native.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { ApproverHeader } from './components/approver-header.native';
+import { ApproverSection } from './components/approver-section.native';
+
+interface ApproverProps extends HasChildren {
+  requester: string;
+}
+function Approver({ requester, children }: ApproverProps) {
+  const [isDisplayingAdvancedView, setIsDisplayingAdvancedView] = useState(false);
+  const [actionBarHeight, setActionBarHeight] = useState(100);
+  const childRegister = useRegisterApproverChildren();
```
kyranjamie, line 19: Could this be all wraped into its own hook so we share it between web/native?

kyranjamie (approved):

apps/mobile/src/features/approver/components/code-card.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Highlighter, Prism, Text } from '@leather.io/ui/native';
+
+/* eslint-disable-next-line lingui/no-unlocalized-strings  */
+const sampleCode = `
+;; hello-world contract
+
+(define-constant sender 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR)
+(define-constant recipient 'SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G)
+
+(define-fungible-token novel-token-19)
```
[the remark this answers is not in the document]
pete-watters: I think it would be good to put it in the UI library

apps/mobile/ios/leatherwalletmobile.xcodeproj/project.pbxproj
```diff
@@ -11,17 +11,17 @@
 		13B07FBC1A68108700A75B9A /* AppDelegate.mm in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB01A68108700A75B9A /* AppDelegate.mm */; };
 		13B07FBF1A68108700A75B9A /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		13B07FC11A68108700A75B9A /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
+		1B12FD95F6504898B7BE8160 /* FiraCode-Medium.otf in Resources */ = {isa = PBXBuildFile; fileRef = B66C2888E3BB47F7A5A88DEB /* FiraCode-Medium.otf */; };
```
pete-watters, line 14: I kept seeing this diff in my PR too, maybe it will go away once we merge this one?

apps/mobile/package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "@crowdin/ota-client": "2.0.0",
     "@expo/vector-icons": "14.0.0",
+    "@gorhom/bottom-sheet": "4.6.3",
```
pete-watters, line 39: I moved `@gorhom/bottom-sheet` dependancy into the UI library. 

Is there some reason not to use standard `Sheet`? Maybe we can export `BottomSheetScrollView` from the UI library and import here to avoid extra dependancies?

The same for `prism-react-renderer`, if we can move that component into the library then we could save the import here

apps/mobile/src/components/badges/badge.tsx
```diff
@@ -0,0 +1,56 @@
+import { ResponsiveValue } from '@shopify/restyle';
```
pete-watters, line 1: Can this go into the UI library?

apps/mobile/src/components/badges/network-badge.tsx
```diff
@@ -0,0 +1,25 @@
+import { AppRoutes } from '@/routes';
```
pete-watters, line 1: I'm not sure about having state connected components under `src/components` 

I would think they should be a feature?

apps/mobile/src/features/approver/components/approver-account-card.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useTheme } from '@shopify/restyle';
+
+import { Text, Theme } from '@leather.io/ui/native';
+
+export function ApproverAccountCard() {
+  const theme = useTheme<Theme>();
+  const { list: walletsList } = useWallets();
+  const wallet = walletsList[0];
+  /* eslint-disable-next-line lingui/no-unlocalized-strings  */
+  const testAccountName = 'testName';
```
pete-watters, line 14: Maybe just load the first account from wallets here? Then it will be more realistic

apps/mobile/src/features/approver/components/inputs-outputs-card.tsx
```diff
@@ -0,0 +1,41 @@
+import { t } from '@lingui/macro';
+
+import { Avatar, Box, Flag, ItemLayout, LockIcon, Text, UnlockIcon } from '@leather.io/ui/native';
+
+function UtxoRow({ isLocked }: { isLocked: boolean }) {
```
pete-watters, line 5: How about just using the `<Flag` inline instead of having `UtxoRow`? 

Or else move it into it's own component

apps/mobile/src/features/send/send-sheets/sign-psbt.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { t } from '@lingui/macro';
+import { useTheme } from '@shopify/restyle';
+
+import { FeeTypes } from '@leather.io/models';
+import { Approver, ArrowLeftIcon, Box, Button, Text, Theme } from '@leather.io/ui/native';
+
+export function SignPsbt() {
+  const theme = useTheme<Theme>();
+
+  function getButtons(buttonState: 'start' | 'approving' | 'success') {
```
pete-watters, line 15: This could be it's own `ActionButton` component rather than a function?

pete-watters (approved):
Great work here @edgarkhanzadian 👍 

I added some organisation suggestions but really solid PR overall

apps/mobile/src/components/badges/network-badge.tsx
fbwoolf: Yeah, we have some dep cruiser rules for this in the extension we need to add to mobile...

Screenshot 2024-11-15 at 1 36 49 PM

pete-watters: My understanding is similar but that if a shared component is connected to state, it then becomes a `feature` rather than a `component`. That's just how I have seen things done in the past generally. 

I'm not sure if we can import one feature to another with those cruiser rules? Maybe they are enforcing a different setup we prefer. 

Usually only dumb, non state connected components would go to `src/components/` then state connected ones into `src/features`. If a component is used in multiple features:
- it could be made more dumb and receive props rather than hook into state
- if it needs to fetch state be promoted to a feature

PR #611, opened 2024-11-09
feat: send select asset sheet, closes LEA-1643 and LEA-1628

fbwoolf:
This PR adds sheets to select an asset for both the send and receive flows. To do this, I've added a hook to use the account display preference address (needed for the account list). I've also created a couple hooks to get the STX and BTC addresses. Hoping to confirm I've done this correctly?

https://github.com/user-attachments/assets/6d4a592a-c89c-42ab-9541-a9b5fbc485f5


apps/mobile/src/features/balances/bitcoin/bitcoin-balance.tsx
```diff
@@ -13,17 +12,22 @@ import { TokenBalance } from '../token-balance';
...
+  fiatBalance,
+  onPress,
+}: BitcoinTokenBalanceProps) {
   return (
     <TokenBalance
       ticker="BTC"
       icon={<BtcAvatarIcon />}
       tokenName={t`Bitcoin`}
-      chain={t`Bitcoin blockchain`}
+      chain={t`Layer 1`}
```
pete-watters, line 27: Do you need to show `Layer 1` ? 

I think we can actually even remove this field as I have used `showChain={false}` to not show it in `<TokenBalance` 

In the design we are supposed to show `Bitcoin blockchain` / `Stacks blockchain` for all non BTC / STX tokens

apps/mobile/src/features/account-list/account-list.tsx
```diff
@@ -23,13 +24,17 @@ export function AccountList({ accounts, onPress, showWalletInfo }: AccountListPr
       {wallet => (
         <AccountListItem
           accountName={account.name}
-          address={t`Address`}
-          balance={t`$1234`}
+          address={
```
pete-watters, line 27: Thanks for updating this 👏  

Not for this PR and aside from your work, I was looking at this file and I wonder could we improve how we do things. 

We map over `accounts` and then load `wallet` for each account so we can show the `wallet.name`. Maybe we could structure this better? Or add a selector to get the `wallet.name` based on the account

```
return accounts.map(account => (
    <WalletLoader fingerprint={account.fingerprint} key={account.id}>
      {wallet => (
        <AccountListItem
          accountName={account.name}
          ...
          walletName={showWalletInfo ? wallet.name : undefined}
        />
      )}
    </WalletLoader>
  ));
```

pete-watters (approved):

apps/mobile/src/components/address-type-badge.tsx
```diff
@@ -0,0 +1,25 @@
+import { useLingui } from '@lingui/react';
+
+import { Box, Text } from '@leather.io/ui/native';
+
+export function AddressTypeBadge({ type }: { type: string }) {
```
kyranjamie, line 5: separate props interface

apps/mobile/src/features/account-list/components/account-address.tsx
```diff
@@ -0,0 +1,17 @@
+import { useAccountDisplayAddress } from '@/store/settings/settings.read';
+
+import { Text } from '@leather.io/ui/native';
+import { truncateMiddle } from '@leather.io/utils';
+
+interface AccountAddressProps {
+  accountIndex: number;
+  fingerprint: string;
+}
```
kyranjamie, line 9:
```suggestion
type AccountAddressProps = AccountId;
```

can reuse types

kyranjamie (approved):

apps/mobile/src/features/balances/bitcoin/bitcoin-balance.tsx
fbwoolf: I did confirm in Slack that designs have changed to `Layer 1`: https://trustmachines.slack.com/archives/C05LRS7G44R/p1731024428631009

apps/mobile/src/features/account-list/account-list.tsx
fbwoolf: Agree here, been thinking on this too bc I'm just now starting to interact with code dealing with wallets and accounts.

PR #613, opened 2024-11-11
fix: bns-v2-sdk patch

edgarkhanzadian:
mobile app imports @leather-io/query package that imports bns-v2-sdk package that imports dotenv and @stacks/connect packages. Those last 2 packages break the mobile app because those are written for the web environment.

This PR deletes bns-v2-sdk package from dependencies and copies a small subset of functionality that is actually used in query package


packages/query/src/stacks/bns/bns-v2-sdk/callersHelper.ts
```diff
@@ -0,0 +1,119 @@
+import { StacksMainnet, StacksTestnet } from '@stacks/network';
```
kyranjamie, line 1: file name casing

kyranjamie (approved):

PR #616, opened 2024-11-13
feat: send form, closes LEA-1649 and LEA-1651

fbwoolf:
Opening this for review so I can get this portion merged as an incremental step. I addressed the comments so far, but make new ones if there are still concerns. I ended up passing `defaultValues` and a `schema` with each form so that we have more control over each form separately. I'm sure more refinements will come up, but I think this is a good starting point. 🤞 

*Using `react-hook-form` w/ zod validation and a pattern similar to the Approver setup

Updated video...

https://github.com/user-attachments/assets/d0b74bae-03ea-44ec-b8f3-ecacfcac64de


pete-watters (commented):
This looks great to me so far, really clean and organised code 👍 nice job!

apps/mobile/src/features/send/send-form/components/send-form-numpad.tsx
```diff
@@ -0,0 +1,17 @@
+import { useFormContext } from 'react-hook-form';
+
+import { Box, Numpad } from '@leather.io/ui/native';
+
+import { SendFormSchema } from '../send-form.schema';
+
+export function SendFormNumpad() {
+  const { setValue, watch } = useFormContext<SendFormSchema>();
+
+  const amount = watch('amount');
```
kyranjamie, line 10: not super clear what `watch` does

apps/mobile/src/features/send/send-form/send-form.schema.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from 'zod';
+
+export const sendFormSchema = z.object({
+  amount: z.string().refine(val => !isNaN(Number(val)) && Number(val) > 0, {
+    message: '',
+  }),
+  recipient: z.string(),
+  memo: z.string().optional(),
+  fee: z.string(),
+});
```
kyranjamie, line 10: Parts of this will certainly need to be modified and composed for different chains/tokens. How might we go about doing that?

apps/mobile/src/features/send/send-form/send-form.schema.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type SendFormSchema = z.infer<typeof sendFormSchema>;
+
+export const defaultSendFormValues: SendFormSchema = {
+  amount: '',
+  recipient: '',
+  memo: '',
+  fee: '',
+};
```
kyranjamie, line 19: How might we auto-set a recipient if there's ever a feature like "send again" somewhere in the UI?

apps/mobile/src/features/send/send-sheets/send-form-btc-sheet.tsx
```diff
@@ -35,7 +41,19 @@ export function SendForm() {
...
+        availableBalance={availableBalance}
+        fiatBalance={fiatBalance}
+      >
+        <SendForm.Asset assetName={t`Bitcoin`} chain={t`Layer 1`} icon={<BtcAvatarIcon />} />
+        <SendForm.AmountField />
+        <SendForm.RecipientField />
+        <SendForm.Memo />
+        <SendForm.Numpad />
+        <SendForm.Button />
+      </SendForm>
```
kyranjamie, line 56: I like this approach, it gives us the flexibility to replace certain elements if some chain as a bespoke requirements.

I think we might also want to compose together, and pass in the, the schema for each instance of SendForm too. All the balance validation, decimals logic etc differ greatly.

If, as we build out more forms, we realise sendform needs tens more props to configure behaviour, I think that might be a sign we need to build out a context for each form type too.

kyranjamie (commented):
Great start @fbwoolf

apps/mobile/src/features/send/send-form/components/send-form-numpad.tsx
fbwoolf: Docs for `watch`: https://react-hook-form.com/docs/useform/watch

PR #617, opened 2024-11-13
feat: add account card pressed state animations

tigranpetrossian:
Part of [LEA-659](https://linear.app/leather-io/issue/LEA-659/define-tapon-press-effect-for-interactive-elements).

### Pressable updates

`pressable.native` is now an `Animated` component, and supports specifying basic state transitions. This is a decent enough temporary implementation, as the better, more declarative solution with using Moti was hit by a myriad of third-party tooling/typings issues that are not worth being stuck on at the moment. Will likely revisit this as part of establishing a consistent animation "framework" across apps.

Along the way made some minor changes to the Pressable:
- Moved pressable.native.tsx into pressable folder
- Removed typography-related props
- Improved types and made PressableProps a non-generic
- Only exposed shorthand props for background and spacing

### Account card
Added a pressed state transition to home screen account cards. Remaining transitions will be quite chonky as they require some layout refactoring, so I'll submit a separate PR.


kyranjamie (commented):
Code lgtm, but can you share a screen recording of the animation too?

tigranpetrossian:
> Code lgtm, but can you share a screen recording of the animation too?

Sure. Updated the description.

apps/mobile/src/components/widgets/accounts/components/cards/account-card.layout.tsx
```diff
@@ -16,9 +20,16 @@ export function AccountCardLayout({
   width = 200,
   testID,
 }: AccountCardLayoutProps) {
+  const { pressed, onPressIn, onPressOut } = usePressedState();
+  const animatedStyle = useAnimatedStyle(() => ({
+    transform: [{ scale: withSpring(pressed ? 0.95 : 1) }],
+  }));
```
kyranjamie, line 26: Should simple transitions like this be put into a shared folder and reused? Can imagine elsewhere another dev does 0.85 etc instead and we have a mash of slight variants.

apps/mobile/src/hooks/use-pressed-state.tsx
```diff
@@ -0,0 +1,39 @@
...
+ *   return <Pressable onPressIn={onPressIn} onPressOut={onPressOut}/>
+ * }
+ * */
+export function usePressedState(props: UsePressedStateInputProps = {}) {
+  const [pressed, setPressed] = useState(false);
+
+  const onPressIn = (event: GestureResponderEvent) => {
+    setPressed(true);
+    props.onPressIn?.(event);
+  };
```
kyranjamie, line 27:
```suggestion
  function onPressIn(event: GestureResponderEvent) {
    setPressed(true);
    props.onPressIn?.(event);
  }
```

kyranjamie (approved):
Great work @tigranpetrossian

apps/mobile/src/hooks/use-pressed-state.tsx
tigranpetrossian: D'uh, old habits. It's my todos to make this lintable.

apps/mobile/src/hooks/use-pressed-state.tsx
```diff
@@ -0,0 +1,39 @@
+import { useState } from 'react';
```
pete-watters, line 1: This is a hook so I don't think it should be a `.tsx` file?

Maybe this could be moved to the `ui` package, located alongside `Pressable`. 

Then you could also remove the large comment into a story file instead to act more as documentation. We don't have any stories for `pressable.native` now

packages/ui/src/theme-native/theme.tsx
```diff
@@ -27,6 +27,8 @@ export function generateTheme(colorScheme: ColorSchemeName) {
       round: 9999,
     },
     textVariants,
+    // Temporary workaround for type errors. TODO: Add a zIndex scale.
```
pete-watters, line 30: Can we do this now?

apps/mobile/src/components/widgets/accounts/components/cards/account-card.layout.tsx
tigranpetrossian: 100%. I should've elaborated: this is what I meant by consistent animation "framework". This should stem from design & prototyping tools of choice as well, so we can define standardizes spring properties to make the animations feel the same. I'll re-visit this along with the remaining pressed transitions.

packages/ui/src/components/pressable/pressable.native.tsx
```diff
@@ -0,0 +1,55 @@
...
+  buttonRestyleFunctions,
+  RNPressable
+);
+
+export const Pressable = Animated.createAnimatedComponent(PressableBase);
+export interface PressableProps extends ComponentPropsWithoutRef<typeof Pressable> {
+  // Incorrect typings in: https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/helperTypes.ts#L48
+  // resulting in inferred key prop of createAnimatedComponent(X) to be incompatible with itself.
+  // TODO: Needs further investigation/raising an issue.
+  key?: any;
```
pete-watters, line 54: Do you get an error if you don't have this key set to `any`?

apps/mobile/src/components/widgets/accounts/components/cards/account-card.layout.tsx
pete-watters: I would put this `animatedStyle` and `usePressedState` with the pressable component in the UI library. 

I guess we will want to apply this same animation to most instances of `Pressable`, at least as a default

packages/ui/src/components/pressable/pressable.native.tsx
tigranpetrossian: A very nasty one, yes. This is a typings issue stemming from react-native-reanimated that isn't straightforward (if at all possible) to fix in userland. It's low priority, but in my list try to (reluctantly 😂) take another look.

tigranpetrossian:
> Nice work @tigranpetrossian . I added a few suggestions.
> 
> I have another task somewhat related to this [linear.app/leather-io/issue/LEA-1544/disable-opacity-feedback-on-all-buttons-on-home](https://linear.app/leather-io/issue/LEA-1544/disable-opacity-feedback-on-all-buttons-on-home)
> 
> My understanding is we want to use `Pressable` instead of `TouchableOpacity` on all buttons on the homepage. Maybe that's something you'd like to do to further use this new code?

Exactly. Should've touched on this in description. I didn't know that ticket existed, happy to take it off your hands as a part of this whole Pressable saga if you want, @pete-watters!

apps/mobile/src/hooks/use-pressed-state.tsx
kyranjamie: There are loads of new linting features in eslint 9 we haven't gotten around to using yet.

apps/mobile/src/components/widgets/accounts/components/cards/account-card.layout.tsx
tigranpetrossian: Agreed. Let me do this after I'm done with the remaining pressed transitions. Might need to do some thinking in terms of naming/structure of those as well. 
The initial idea was to add an ability for defining those more declaratively, through serializable config presets: `const animationX = { type: 'spring', damping: 5, stiffness: 20, ...etc  }` passed right into the pressable, but there are too many tooling obstacles in the way of doing this quickly, so settling with RNR API for the time being.
Will move the `usePressableState` into `ui` within this PR.

pete-watters:
> > Nice work @tigranpetrossian . I added a few suggestions.
> > I have another task somewhat related to this [linear.app/leather-io/issue/LEA-1544/disable-opacity-feedback-on-all-buttons-on-home](https://linear.app/leather-io/issue/LEA-1544/disable-opacity-feedback-on-all-buttons-on-home)
> > My understanding is we want to use `Pressable` instead of `TouchableOpacity` on all buttons on the homepage. Maybe that's something you'd like to do to further use this new code?
> 
> Exactly. Should've touched on this in description. I didn't know that ticket existed, happy to take it off your hands as a part of this whole Pressable saga if you want, @pete-watters!

That would be great and to get all of the `Pressable` stuff done in one batch, thanks!

packages/ui/src/components/pressable/pressable.native.tsx
pete-watters: OK, that's fair enough to leave it as so 👍  thanks!

apps/mobile/src/components/widgets/accounts/components/cards/account-card.layout.tsx
pete-watters: Sounds great 👍

fbwoolf (approved):

PR #622, opened 2024-11-15
feat: add sheet animation

alter-eggo:
This pr adds
- sheet animations in full and extension mode
- close on click esc

https://github.com/user-attachments/assets/880b641e-d04c-4577-841f-f3196e952d66

### DIFF
--- a/packages/ui/src/components/sheet/sheet.web.stories.tsx
+++ b/packages/ui/src/components/sheet/sheet.web.stories.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useRef } from 'react';
 
 import type { Meta } from '@storybook/react';
 import { Box, Flex } from 'leather-styles/jsx';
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/sheet/sheet.web.stories.tsx
```diff
@@ -23,14 +23,17 @@ const meta: Meta<typeof Component> = {
 export default meta;
 
 export function Sheet() {
-  const [isShowing, setIsShowing] = useState(false);
+  const triggerRef = useRef<HTMLButtonElement>(null);
```
camerow, line 26: This may be un-wise, but what do you think about `Sheet` exporting a hook like `useSheet` which returns functions around the refs, eg `{ triggerSheet, closeSheet } = useSheet()`. Feels to me like `Sheet` should better encapsulate it's behaviors. Just a thought.

Really loving the animation!

alter-eggo:
@edgarkhanzadian yeah, let's close for now, will return to this later

PR #627, opened 2024-11-17
feat: migrate stacks generate txs, closes LEA-1732

fbwoolf:
This PR migrates code from the extension to generate the unsigned stacks tx for the STX send flow.

Approach here now uses a separate Provider for each chain/protocol form so that we can handle loading data in the provider. I also did a version using our loader pattern, which seemed ok too, but felt using this approach might be more straightforward with using React context for the forms. We can add a loading state later, etc.

I also removed any helper functions I wasn't directly using with generating the unsigned tx. I realized I had ported over some unnec functions for this work. I got rid of all the stacks connect types.

I did add two new money formatters that can eventually help to replace `stacksValue`, thought I didn't end up needing them anywhere for now.


apps/mobile/src/features/balances/token-balance.tsx
```diff
@@ -1,16 +1,29 @@
 import { ReactNode } from 'react';
 
 import { Balance } from '@/components/balance/balance';
+import { t } from '@lingui/macro';
 
-import { Money } from '@leather.io/models';
+import { CryptoAssetProtocol, Money } from '@leather.io/models';
 import { Flag, ItemLayout, Pressable } from '@leather.io/ui/native';
 
+export function getChainLayerFromAssetProtocol(protocol: CryptoAssetProtocol) {
```
pete-watters, line 9: Nice work! I guess this changed again in designs? 

Before we had to say nothing for `BTC` / `STX` then always say `X blockchain` for others

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return {
+    address: addresses[0] ?? '',
+    availableBalance,
+    fiatBalance,
+    formMethods,
+    onGoBack() {
+      navigation.navigate('send-select-asset', { account: route.params.account });
+    },
+    async onSubmit(data: SendFormStxSchema) {
```
pete-watters, line 51: Some rogue logs here but I guess more work to follow so it's OK

apps/mobile/src/features/send/send-sheets/send-form-stx-sheet.tsx
```diff
@@ -55,34 +44,28 @@ export function SendFormStxSheet() {
...
-            })}
-          />
-          <SendForm.AmountField />
-          <SendForm.RecipientField />
-          <SendForm.Memo />
-          <SendForm.Footer>
-            <SendForm.Numpad />
-            <SendForm.Button />
-          </SendForm.Footer>
+          <SendForm.StacksSetter address={address}>
```
pete-watters, line 59: Great refactor here 👍

packages/stacks/src/stacks.utils.ts
```diff
@@ -71,3 +73,110 @@ export function getStacksBurnAddress(chainId: ChainID): string {
...
+export const getStacksContractName = (contractId: string): string => {
+  if (contractId.includes('.')) {
+    const parts = contractId?.split('.');
+    if (contractId.includes('::')) {
+      return parts[1].split('::')[0];
+    }
+    return parts[1];
+  }
+  // eslint-disable-next-line no-console
+  console.warn(
```
pete-watters, line 124: I think the logs here need to go so they don't get published. Unless we will lock versions until all work is finished which is OK too

packages/stacks/src/transactions/generate-unsigned-tx.ts
```diff
@@ -0,0 +1,137 @@
...
+    publicKey,
+    anchorMode: anchorMode ?? AnchorMode.Any,
+    postConditionMode: postConditionMode,
+    postConditions: getPostConditions(postConditions),
+    network,
+  };
+  return makeUnsignedContractDeploy(options);
+}
+
+// type STXTransferPayload = Omit<ConnectSTXTransferPayload, 'network'> &
```
pete-watters, line 97: Some commented out code here

pete-watters (approved):
Nice work @fbwoolf 👍  

There's some console logs + commented code to remove. 

For the `stacks` package work, maybe we could add some of the functions into smaller independent files and generate some unit tests for them also?

apps/mobile/src/features/send/send-form/components/send-form-stacks-setter.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+export function SendFormStacksSetter({ address, children }: SendFormStacksSetterProps) {
+  const { schema } = useSendFormContext();
+  const { setValue, formState, getValues } = useFormContext<z.infer<typeof schema>>();
+  const { data: nextNonce } = useNextNonce(address);
+
+  // Temporary default fee
+  const defaultFee =
+    defaultStacksFees.estimates[0]?.fee.amount ?? new BigNumber(defaultFeeFallback);
+
```
kyranjamie, line 26: Dislike this inline "setter" component, it's a very side-effecty way of setting state.

What other ways could we do this? Why does it need to be a component? Think it'd be worth trying to find a way to avoid this, as to not copy over bad design patterns from extension.

apps/mobile/src/common/transactions/stacks-transactions.hooks.ts
```diff
@@ -0,0 +1,51 @@
...
+  generateUnsignedTransaction,
+} from '@leather.io/stacks';
+import { stxToMicroStx } from '@leather.io/utils';
+
+export function useGenerateStxTokenTransferUnsignedTx({ fingerprint, accountIndex }: AccountId) {
+  const network = useNetworkPreferenceStacksNetwork();
+  const account = useAccountByIndex(fingerprint, accountIndex);
+  const stxSigner = useStacksSigners().fromAccountIndex(fingerprint, accountIndex)[0];
+  const stxAddress = stxSigner?.address ?? '';
+  const { data: nextNonce } = useNextNonce(stxAddress);
```
kyranjamie, line 23: I think it might be interesting to decouple this logic. This hook is essentially responsible for 1) preparing account specific tx details, and 2) feeding them through to the `generateUnsignedTransaction` fn.

What if I want to modify these values in another tx signing use case that isn't the form specifically? I can't compose this logic well, because it assumes I always need to do both at same time.

If these were separate, I could do something like:

```ts
// send form use case
function useGenerateStxTokenTransferUnsignedTx(account: AccountId) {
  const stxAccountSignerDetails = useStxTokenTransferDetails();
  // spread details to fn
  return (formValues) => generateUnsignedTransaction({ 
    ...stxAccountSignerDetails,
    ...parsedFormValues
  })
}
```

apps/mobile/src/common/transactions/stacks-transactions.hooks.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  return useCallback(
+    async (values?: FieldValues) => {
```
kyranjamie, line 26: This fn is coupled to the form library. Is it accurate that we only ever need to create transactions from a user's input?

Consider increase fee/cancel transaction.

packages/stacks/src/stacks.utils.ts
```diff
@@ -71,3 +73,110 @@ export function getStacksBurnAddress(chainId: ChainID): string {
...
+export function hexToBuff(hex: string): Buffer {
+  return Buffer.from(cleanHex(hex), 'hex');
+}
+
+export const stacksValue = ({
+  value,
+  fixedDecimals = true,
+  withTicker = true,
+  abbreviate = false,
+}: {
```
kyranjamie, line 93: This really old code that'd be best not copied across. Can we refactor it out entirely?

packages/stacks/src/stacks.utils.ts
```diff
@@ -71,3 +73,110 @@ export function getStacksBurnAddress(chainId: ChainID): string {
...
+ * @param contractId - the source string: [principal].[contract-name] or [principal].[contract-name]::[asset-name]
+ */
+export const getStacksContractName = (contractId: string): string => {
+  if (contractId.includes('.')) {
+    const parts = contractId?.split('.');
+    if (contractId.includes('::')) {
+      return parts[1].split('::')[0];
+    }
+    return parts[1];
+  }
```
kyranjamie, line 122: Let's make a function declaration for consistency

packages/stacks/src/stacks.utils.ts
```diff
@@ -71,3 +73,110 @@ export function getStacksBurnAddress(chainId: ChainID): string {
...
+  );
+  return contractId;
+};
+
+/**
+ * Gets the asset name from a a fully qualified name of an asset.
+ *
+ * @param contractId - the fully qualified name of the asset: [principal].[contract-name]::[asset-name]
+ */
+const getStacksContractAssetName = (contractId: string): string => {
```
kyranjamie, line 136: same here

packages/stacks/src/transactions/stacks-connect.types.ts
```diff
@@ -0,0 +1,142 @@
+import { StacksNetwork } from '@stacks/network';
+import {
+  AnchorMode,
+  ClarityValue,
+  PostCondition,
+  PostConditionMode,
+  StacksTransaction,
+} from '@stacks/transactions';
+
+// Types are copied from @stacks/connect to avoid installing the package
```
kyranjamie, line 10: Do we need all these types though? Any way we can get rid of them entirely?

kyranjamie (commented):
Great work @fbwoolf

Lots to discuss here though. I think there's some code, and general patterns here, we might want to reconsider bringing over if possible

kyranjamie:
> I used an approach where for each form we have a SendForm.StacksSetter (or appropriate chain/prototool) to set nec form data. We do this in the extension with NonceSetter. I think we can use this later to load/set fee values, etc. Open to other ideas here too.

I would investigate options where we do not even render the form components until we have this data loaded from the external source. Either the loader pattern, or some kind of route loader if Expo has those.

fbwoolf:
> Great work @fbwoolf
> 
> Lots to discuss here though. I think there's some code, and general patterns here, we might want to reconsider bringing over if possible

Yep, agree, I was debating abt just doing a draft PR but wasn't sure how quickly we wanted to push the send form forward. I'll change this to draft and tackle some of the refactoring you pointed out. I mostly wanted to make sure I got feedback on effort here so far. 👍

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useSendSheetRoute,
+} from '../../send-form.utils';
+import { SendFormStxSchema } from '../schemas/send-form-stx.schema';
+
+export type CurrentRoute = CreateCurrentSendRoute<'send-form-stx'>;
+
+function parseSendFormValues(values: SendFormStxSchema) {
+  return {
+    amount: values.amount ? stxToMicroStx(values.amount).toString(10) : '0',
+    fee: stxToMicroStx(values.fee || 0).toNumber(),
```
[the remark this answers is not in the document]
fbwoolf: Yep, will fix those and check others.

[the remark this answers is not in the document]
fbwoolf: Actually thinking I shouldn't use the types from `@stacks/transactions` the way I am atm, so going to refactor those out so we can pass `Money` type as args.

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
```diff
@@ -0,0 +1,24 @@
+import {
+  CreateCurrentSendRoute,
+  useSendSheetNavigation,
+  useSendSheetRoute,
+} from '../../send-form.utils';
+import { SendFormStxSchema } from '../schemas/send-form-stx.schema';
+
+export type CurrentRoute = CreateCurrentSendRoute<'send-form-stx'>;
```
camerow, line 8:
```suggestion
export type CurrentRoute = CreateCurrentSendRoute<'send-form-btc'>;
```

fbwoolf:
I am going to open this open for review again. Open to feedback on using the separate providers here vs our loader pattern. I'm happy to switch if people prefer the loader pattern. Expo router doesn't offer loaders.

fbwoolf:
Working a bit more here to split up the providers bc currently they are named incorrectly. They need to be protocol specific.

fbwoolf:
Looking at this closer, it might be that we want separate providers per protocol for flexibility, but use the loader pattern per chain so we don't duplicate efforts with loading chain data like fees and nonces. Might try adding that as a next step.

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -0,0 +1,52 @@
...
+    accountIndex: route.params.account.accountIndex,
+    fingerprint: route.params.account.fingerprint,
+  });
+
+  return {
+    onGoBack() {
+      navigation.navigate('send-select-asset', { account: route.params.account });
+    },
+    // Temporary logs until we can hook up to approver flow
+    async onSubmit(values: SendFormStxSchema) {
```
kyranjamie, line 39: Wonder if we could use a slightly more descriptive name for the action intended by the form, rather than form specific terminology.

`onInitiateSend`
`onUserInitTransfer`
`onSignTransaction` etc

packages/models/src/types.utils.ts
```diff
@@ -11,3 +11,10 @@ export type LiteralUnion<LiteralType, BaseType extends Primitive> =
   | (BaseType & Record<never, never>);
 
 export type Entries<T> = { [K in keyof T]: [K, T[K]] }[keyof T][];
+
+export type ReplaceTypes<T, Replacements extends { [K in keyof T]?: any }> = Omit<
+  T,
+  keyof Replacements
+> & {
+  [K in keyof Replacements]: Replacements[K];
+};
```
kyranjamie, line 20: What does this do?

packages/stacks/src/stacks.utils.ts
```diff
@@ -71,3 +71,69 @@ export function getStacksBurnAddress(chainId: ChainID): string {
...
+function cleanHex(hexWithMaybePrefix: string): string {
+  if (hexWithMaybePrefix !== 'string') return hexWithMaybePrefix;
+  return hexWithMaybePrefix.startsWith('0x')
+    ? hexWithMaybePrefix.replace('0x', '')
+    : hexWithMaybePrefix;
+}
+
+export function hexToBuff(hex: string): Buffer {
+  return Buffer.from(cleanHex(hex), 'hex');
+}
```
kyranjamie, line 84: Do we still need this `Buffer` specific method versus a `UInt8Array` alternative?

packages/stacks/src/transactions/generate-unsigned-transaction.ts
```diff
@@ -0,0 +1,76 @@
...
+import {
+  StacksUnsignedContractCallOptions,
+  StacksUnsignedContractDeployOptions,
+  StacksUnsignedTokenTransferOptions,
+  StacksUnsignedTransactionOptions,
+  TransactionTypes,
+  isTransactionTypeSupported,
+} from './transaction.types';
+
+function initNonce(nonce?: number | string) {
```
kyranjamie, line 20: What issues do we run into if we remove the optional param?

packages/stacks/src/transactions/post-condition.utils.ts
```diff
@@ -0,0 +1,19 @@
+import { hexToBytes } from '@noble/hashes/utils';
+import { BytesReader, PostCondition, deserializePostCondition } from '@stacks/transactions';
+
+export function getPostConditionFromString(postCondition: string): PostCondition {
+  const reader = new BytesReader(hexToBytes(postCondition));
+  return deserializePostCondition(reader);
+}
+
+export function getPostCondition(postCondition: string | PostCondition): PostCondition {
+  return typeof postCondition === 'string'
```
kyranjamie, line 10:
```suggestion
  return isString(postCondition)
```

apps/mobile/src/common/transactions/stacks-transactions.hooks.ts
```diff
@@ -0,0 +1,53 @@
...
+import { createMoney } from '@leather.io/utils';
+
+export function useStxTokenTransferDetails({ fingerprint, accountIndex }: AccountId) {
+  const account = useAccountByIndex(fingerprint, accountIndex);
+  const network = useNetworkPreferenceStacksNetwork();
+  const stxSigner = useStacksSigners().fromAccountIndex(fingerprint, accountIndex)[0];
+  const stxAddress = stxSigner?.address ?? '';
+  const { data: nextNonce } = useNextNonce(stxAddress);
+
+  if (!account || !stxSigner) return;
```
kyranjamie, line 23: Any way to handle these missing values differently, such that by the time we get to this hook, we're sure they're defined?

apps/mobile/src/features/send/send-form/providers/send-form-stx-provider.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const address = addresses[0] ?? '';
+  const { data: nextNonce } = useNextNonce(address);
+
+  // Load fees here, temp default fee
+  const defaultFeeFallbackAsMoney = createMoney(defaultFeeFallback, 'STX');
+  const defaultFee = convertAmountToBaseUnit(
+    defaultStacksFees.estimates[0]?.fee ?? defaultFeeFallbackAsMoney
+  );
+
+  if (!nextNonce) return null;
```
kyranjamie, line 38: Should we have a `nonce` loader component, or even a loader component specific to this send form? 

I could see the Loader pattern living side by side use of context here.

```ts
<StxSendFormLoader fallback={<LoadingScreen />}>
  {nonceEtc => <SendFormProvider {...nonceEtc} />}
</StxSendFormLoader/>
```

kyranjamie (commented):
Curious to hear thoughts on feedback I've left, but very impressed with this solution.  Think you've struck a great balance between reusability of the form, and compatibility of all the elements with the context.

apps/mobile/src/features/send/send-sheets/send-form-btc-sheet.tsx
```diff
@@ -1,38 +1,22 @@
...
-  const formMethods = useForm<SendFormBtcSchema>({
-    defaultValues: defaultSendFormBtcValues,
-    resolver: zodResolver(sendFormBtcSchema),
-  });
 
-  const { availableBalance, fiatBalance } = useBitcoinAccountTotalBitcoinBalance({
-    accountIndex: route.params.account.accountIndex,
-    fingerprint: route.params.account.fingerprint,
-  });
+  const { onGoBack } = useSendFormBtc();
```
kyranjamie, line 19: Alternative to the loader, wonder if async dependencies & loading states of SendFormBtcProvider could be handled here?

packages/models/src/types.utils.ts
fbwoolf: I wanted to reuse part of the @stacks/transactions types but replace their use of `IntegerType` from @stacks/common with our `Money` type and narrow the other types rather than installing another lib to handle their `IntegerType`. I could completely create our own types copied from @stacks/transactions, but I looked up this utility to replace types. Now, we can pass our Money type all the way to the stacks pkg and then just convert it to a string their before using their typed function. Look at use in `transaction.types.ts`.

apps/mobile/src/features/send/send-form/providers/send-form-stx-provider.tsx
fbwoolf: Yep, agree, was refactoring to this when you were reviewing. I was thinking maybe we only need loaders per chain, but maybe we do need per protocol. 🤔 Stacks would load the same fees and nonce for stx and sip10, so there would be some duplication there if do separately.

fbwoolf: Thinking should be separate, so separating. I think the loader pattern makes more sense to me rather than using the hook for loading, but open if others feel strongly either way.

apps/mobile/src/common/transactions/stacks-transactions.hooks.ts
fbwoolf: I ended up creating a loader for the StacksSigner and passing the signer address and publicKey hex thru the route params. Hoping this is a good solution here?

apps/mobile/src/features/send/send-sheets/send-form-btc-sheet.tsx
fbwoolf: Went with the loader pattern, but happy to revisit.

fbwoolf:
@kyranjamie hoping to get your feedback on how I've addressed your questions. I am now handling all undefined values thru a form loader component, which I think is working well.

kyranjamie (approved):
Looks great, thanks Fara

PR #630, opened 2024-11-18
refactor: convert widget into compound components

tigranpetrossian:
Part of [LEA-659](https://linear.app/leather-io/issue/LEA-659/define-tapon-press-effect-for-interactive-elements)

Converts the Widget into a Radix-style compound component in preparation for list item animations, but also in an attempt to standardize with other similar components like the [Approver](https://github.com/leather-io/mono/blob/dev/packages/ui/src/components/approver/approver.web.tsx)

#### Before
```
<Widget
  header={
    <WidgetHeader />
  }
>
  ...
</Widget>
```

#### After
```
<Widget>
  <Widget.Header>
    <Widget.Title />
  </Widget.Header>
  <Widget.Body>
    ...
  </Widget.Body>
</Widget>
```

This significantly reduces the prop drilling and subsequently the need to expose more props, de-couples it from ScrollView, some spacing opinions (padding, gaps, etc.), and replaces the dependency on `sheet` with a generic action.

### DIFF
--- a/apps/mobile/src/components/widgets/accounts/accounts-widget.tsx
+++ b/apps/mobile/src/components/widgets/accounts/accounts-widget.tsx
@@ -1,4 +1,5 @@
 import { useRef } from 'react';
+import { ScrollView } from 'react-native-gesture-handler';
 
 import { AddWalletSheet } from '@/components/add-wallet/';
 import { AccountSelectorSheet } from '@/features/account-selector-sheet';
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):
This is a great instance of when compound components are great, thanks @tigranpetrossian 

I wonder if you'd fancy writing a [Coding Convention](https://www.notion.so/trustmachines/041fab7adc84487db0058e64f27d7064?v=175af98264b348e69cee48f17c1f6353&pvs=4) page on how we should use this pattern (and how we shouldn't).

pete-watters:
Great work @tigranpetrossian 👏

PR #633, opened 2024-11-18
feat(models): sbtc testnet network

kyranjamie:
Updates networks and adds a new live testnet instance for sBTC testing

### DIFF
--- a/packages/models/src/network/network.model.ts
+++ b/packages/models/src/network/network.model.ts
@@ -34,6 +34,7 @@ export enum WalletDefaultNetworkConfigurationIds {
   testnet = 'testnet',
   testnet4 = 'testnet4',
   signet = 'signet',
+  sbtcTestnet = 'sbtcTestnet',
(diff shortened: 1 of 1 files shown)


packages/models/src/network/network.model.ts
```diff
@@ -177,7 +196,7 @@ const networkSbtcDevenv: NetworkConfiguration = {
       blockchain: 'bitcoin',
       bitcoinNetwork: 'regtest',
       mode: 'regtest',
-      bitcoinUrl: 'http://localhost:8083/api',
+      bitcoinUrl: 'http://localhost:3000/api/proxy',
```
camerow, line 199: Thought: Would we want a potential option to configure this?

kyranjamie: Yes, totally. 

And users can add networks in extension, _but_ we're missing feature to target those via the API, so for now just an accurate set up will unblock sbtc teams

PR #634, opened 2024-11-19
chore: use market data service for STX balances

pete-watters:
This PR refactors STX balances to use the new market data service


apps/mobile/src/queries/market-data/stx-market-data.query.ts
```diff
@@ -0,0 +1,21 @@
+import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
+
+import { getMarketDataService } from '@leather.io/services';
+import { oneMinInMs } from '@leather.io/utils';
+
+export function createStxMarketDataQueryOptions() {
```
tigranpetrossian, line 6: I'm guessing this was a factory accepting arguments during the implementation, that were later deleted?
Also curious why `as const` – is the object being consumed elsewhere?

tigranpetrossian (approved):

PR #638, opened 2024-11-20
chore: LEA-1696 account balance fixes

pete-watters:


apps/mobile/src/features/balances/stacks/mock-signers.ts
```diff
@@ -0,0 +1,62 @@
+const multiWalletSigners = [
```
pete-watters, line 1: This is a log of the signers list for:
- Leather dev wallet: `fingerprint: 'efd01538'`
- my wallet : `fingerprint: '558efa7e'`

I imported the dev wallet and had one new account created with `accountIndex: 0`. 

On adding a new account to that wallet in the UI I can see it added with a new derivation path but the `accountIndex` is always `0`.

apps/mobile/src/features/balances/stacks/use-get-stacks-addresses.ts
```diff
@@ -8,8 +8,14 @@ export function useGetStacksAddresses(accountId?: AccountId) {
     return signers.list.map(signer => signer.address);
   }
 
-  const { fingerprint } = accountId;
-  return signers.list
-    .filter(signer => signer.descriptor.includes(fingerprint))
-    .map(signer => signer.address);
+  const { accountIndex } = accountId;
+  console.log('signers.list', signers.list);
+  // is this wrong? all signer descriptors underneath the same wallet have the same fingerprint
```
pete-watters, line 13: I was naively using `fingerprint` to get stacks addresses to use to check balance. 

That was returning all addresses associated with that fingerprint. 

I think I need to change this to use the fingerprint in combination with `accountIndex` but account index is not incrementing in signers.

The code I added here - `signers.list[accountIndex]` loads the correct balances but by coincidence of the order of the signers array. 

 I think the code should be more like:
 ```
return signers.list
    .filter(signer => signer.descriptor.includes(fingerprint) && signer.accountIndex === accountIndex)
    .map(signer => signer.address);
```

PR #640, opened 2024-11-20
fix: ensure correct balance per account is shown

pete-watters:
This PR addresses some issues with incorrect STX account balances being shown
- add filtering to signers by account index
- ensure we update the accountIndex for stacks when initialising stacks signers
- roll back update to new market data service for STX as price is incorrect precision 

I will add unit tests for account generation but opening for initial review 

Here is a demo of loading the dev wallet and creating some accounts:
https://github.com/user-attachments/assets/c737e6b3-d480-48ce-9e69-50d0c52e324e

**NOTE**
There is still some weird janky and false positive balances appearing when switching accounts which seems to be related to BTC balances. Here you can see an empty account having a `$0.06` cent balance when another account is entered

https://github.com/user-attachments/assets/fc69b963-f402-4095-ba44-c7934ea55652


fbwoolf (approved):

apps/mobile/src/features/balances/stacks/use-get-stacks-addresses.ts
```diff
@@ -8,8 +8,10 @@ export function useGetStacksAddresses(accountId?: AccountId) {
     return signers.list.map(signer => signer.address);
   }
 
-  const { fingerprint } = accountId;
+  const { accountIndex, fingerprint } = accountId;
   return signers.list
-    .filter(signer => signer.descriptor.includes(fingerprint))
+    .filter(
+      signer => signer.descriptor.includes(fingerprint) && signer.accountIndex === accountIndex
+    )
```
kyranjamie, line 15: It looks like you're repeating logic here, `useStacksSigners` has a `.fromAccountIndex` method to do the filtering?

kyranjamie (changes requested):

apps/mobile/src/features/balances/stacks/use-get-stacks-addresses.ts
pete-watters: Good catch 👍

pete-watters:
Tagging this with `needs:build` to create a demo for QA. Ideally I will first merge https://github.com/leather-io/mono/pull/645 then this PR and have a demo of both available

apps/mobile/src/features/balances/stacks/stacks-balance.tsx
```diff
@@ -52,8 +55,11 @@ export function StacksBalanceByAccount({
   fingerprint,
   onPress,
 }: StacksBalanceByAccountProps) {
-  const addresses = useGetStacksAddresses({ accountIndex, fingerprint });
-  const { availableBalance, fiatBalance } = useStxBalance(addresses);
+  const address = useStacksSignerAddressFromAccountIndex(fingerprint, accountIndex);
+  if (!address) {
```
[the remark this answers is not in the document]
kyranjamie: Throwing error doesn't break rules of hooks.

Though, for safety, I would consider using a loader component such that you pass it in with a props, and handle the null case outside this component

kyranjamie (approved):

PR #652, opened 2024-11-25
feat: add amount field component

tigranpetrossian:
Part of [LEA-1720](https://linear.app/leather-io/issue/LEA-1720/send-form-amount-field)

Adds the AmountField component for the Send form

https://github.com/user-attachments/assets/eb5ef453-cdb2-457e-97bb-95c9ea82081e

@fbwoolf I was initially intending to plug this into the form myself, but decided against it for two reasons:
1. Still behind the curve of doing calculations correctly based on market data, chosen currency, etc., it will probably take me much longer
2. Not sure what stage the form development is in, as I've noticed some things like error handling are still in progress.

I've made the component domain-agnostic enough to be easily integrated. Formatting & currency conversion when swapping values is done via inversion of control. Here's a basic example I've put together based on testing this against an actual simple demo form. Notice  using `useController` instead of `Controller` with render prop, as this is technically not a field anymore.

```
// Assumes a `currencyMode` field added to the form values for tracking the current mode user has selected (not using Zod for brevity):
//
// interface FormValues {
//   amount: string;
//   currencyInputMode: 'crypto' | 'fiat';
//   ...etc
// }
//
// useForm<FormValues>(...)

function SendFormAmountField() {
  const { watch, setValue } = useFormContext<FormValues>();
  const { field, fieldState } = useController<FormValues>({
    name: 'amount',
    rules: {
      required: t({
        id: 'send-form.amount-field.error.amount-required',
        message: 'Amount is required',
      }),
    },
  });
  const { value, onChange } = field;
  const { invalid, isTouched } = fieldState;
  const currencyMode = watch('currencyInputMode');

  const handleCurrencyModeChange = (currencyMode: CurrencyMode, newInputValue: string) => {
    setValue('currencyInputMode', currencyMode);
    setValue('amount', newInputValue);
  };

  return (
    <AmountField
      inputValue={value}
      invalid={isTouched && invalid}
      currencyMode={currencyMode}
      onCurrencyModeChange={handleCurrencyModeChange}
      onSetIsSendingMax={() => {
        // set input value to available balance in `currencyMode`
      }}
      formatValue={(value, currencyMode) => {
        // format value based on currencyMode, i.e. prepend $ for fiat, append BTC to crypto 
      }}
      calculateSecondaryValue={(value, currencyMode) => {
        // convert primary value from fiat to crypto or vice versa depending on current mode
      }}
    />
  );
}
```

#### Up next
I'll add a separate PR adding character animations and possibly few more transitions. There's currently a noticeable flicker when the font size changes, will be gone by itself once the text is animated.

### DIFF
--- a/apps/mobile/src/components/amount-field/amount-field-primary-value.tsx
+++ b/apps/mobile/src/components/amount-field/amount-field-primary-value.tsx
@@ -1,0 +1,85 @@
+import { useState } from 'react';
+import { NativeSyntheticEvent, TextLayoutEventData } from 'react-native';
+
+import { Box, Text, TextProps, Theme } from '@leather.io/ui/native';
+
+const maxFontSize = 44;
+const lineHeightRatio = 1;
+
+const commonTextProps: TextProps = {
+  variant: 'heading02',
+  fontVariant: ['tabular-nums'],
+  letterSpacing: 1,
+};
+
+interface TextMeasurementProxyProps {
+  value: string;
+  textProps: TextProps;
+  onFontSizeChange(fontSize: number): void;
+}
+
+// The Text component's `adjustFontSizeTofit` is not animatable. It also causes text to shift
+// due to dynamic positioning issue: https://github.com/facebook/react-native/issues/29507
+// Use an invisible placeholder to extract the dynamic font size set by `adjustFontSizeToFit`
+// as the text changes during typing.
+function TextMeasurementProxy({ value, textProps, onFontSizeChange }: TextMeasurementProxyProps) {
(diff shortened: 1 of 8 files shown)


fbwoolf (approved):
Lgtm 🚀

kyranjamie (approved):
Great work @tigranpetrossian

pete-watters:
> LEA-1720

The UI was added here but it's not yet integrated to the app yet. I added a note [here](https://linear.app/leather-io/issue/LEA-1873/send-max-option-to-send-form) to try get it added

PR #654, opened 2024-11-25
Draft: Icon with activity sub-icon

camerow:
Work in progress on adding an icon with activity sub icon. 

- Need to fix some shadow issues, but overall it is nearly done. 

Screenshot 2024-11-25 at 2 40 48 PM
Screenshot 2024-11-25 at 2 41 13 PM
Screenshot 2024-11-25 at 2 41 28 PM


apps/mobile/src/components/widgets/accounts/components/cards/add-account-card.tsx
```diff
@@ -9,10 +10,12 @@ export interface AddAccountCardProps {
 }
 
 export function AddAccountCard({ onPress }: AddAccountCardProps) {
+  const theme = useTheme<Theme>();
+
   return (
     <AccountCardLayout
       onPress={onPress}
-      icon={<PlusIcon />}
+      icon={<PlusIcon color={theme.colors['ink.background-primary']} />}
```
camerow, line 18: Noticed this color bug while i was exploring icons.

apps/mobile/src/components/widgets/tokens/token-icon.tsx
```diff
@@ -7,7 +12,8 @@ interface TokenIconProps {
 export function TokenIcon({ ticker }: TokenIconProps) {
   switch (ticker) {
     case 'STX':
-      return <StxAvatarIcon />;
+      return <IconWithActivity avatar={<StxAvatarIcon />} activity="error" />;
```
camerow, line 15: TODO: Revert this change, just added for quick testing.

packages/ui/src/icons/icon/icon-with-activity.native.tsx
```diff
@@ -0,0 +1,19 @@
+import { View } from 'react-native';
```
pete-watters, line 1: Sometimes we need to show an activity type icon on top of a protocol item, and sometimes we need to show a protocol icon on top for example here:
Screenshot 2024-11-26 at 09 23 58

**NOTE** the design is wrong for the first two items - Stacks + Btc. 

For subsequent tokens like `sBTC`, `ALEX Lab` etc. we need to show a base protocol icon on top of their specific item. 

Maybe we can extend this component to be more broad and handle both activity and protocol sub-icons

pete-watters (approved):
Nice work so far here @camerow , I added a couple of suggestions but it's looking good overall

packages/ui/src/icons/arrow-down-icon.native.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const variants: IconVariantMap = {
+  small: <ArrowDownSmall />,
+  default: <ArrowDown />,
+};
+
+export const ArrowDownIcon = forwardRef<Component, IconProps>(({ variant, ...props }, ref) => {
+  return (
+    <Icon ref={ref} {...props}>
+      {variants[variant ?? 'default']}
```
fbwoolf, line 17: Is it nec to make this change when all other icons use a different code pattern?

packages/ui/src/icons/icon/icon-with-activity.native.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { ActivityIcon } from '../activity-icons/activity-icon.native';
+import { ActivityIconType } from '../index.native';
+
+export function IconWithActivity({
+  avatar,
+  activity,
+}: {
+  avatar: React.JSX.Element;
+  activity?: ActivityIconType;
```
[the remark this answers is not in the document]
fbwoolf: Agree here, why would we default to `send`?

packages/ui/src/icons/icon/icon.native.tsx
```diff
@@ -16,7 +16,7 @@ export const Icon = forwardRef<Component, IconProps>(({ children, ...props }, re
   return (
     <>
       {React.cloneElement(child, {
-        color: theme.colors['ink.action-primary-default'],
+        color: props.color ?? theme.colors['ink.action-primary-default'],
```
fbwoolf, line 19: Is this change nec? We have always been able to override the icon color w/out this. This should just set the default color?

packages/ui/src/icons/arrow-down-icon.native.tsx
camerow: No, was just playing with patterns and thinking about suggesting a different way of doing this down the line.

Actually as I was looking at this I was wondering why we don't just have a single component that can render any icon, eg `<LeatherIcon icon="icon-name" variant="small" {...props} />` - I find that pattern makes adding icons much easier as you just have a lookup table / function and all new icons can just be added there. 

Was the idea with this implementation that it allows for better tree-shaking? 

I can revert it to how the others do it for now.

camerow: eg [FontAwesome](https://docs.fontawesome.com/v5/web/use-with/react)

packages/ui/src/icons/icon/icon.native.tsx
camerow: It does make the fallback color more explicit and thus easier to debug. 

Of course it can always be overriden without this, but the underlying `svg` usually has something like `stroke="currentColor"` and it's not obvious to me what that default is - having something explicitly set rather than a side effect seems like an improvement.

packages/ui/src/icons/icon/icon-with-activity.native.tsx
camerow: Yeah that sounds good, should be somewhat easy to achieve. 

To that point I notice we are hard coding our protocol icons - I would love to see us use something more generic like a `<ProtocolIcon symbol="btc | stx | ..." />`

packages/ui/src/icons/icon/icon-with-activity.native.tsx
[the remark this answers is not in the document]
camerow: Still a draft, not intentionally left to be that way. I'll update to omit the icon if not provided as a optional prop.

packages/ui/src/icons/icon/icon.native.tsx
fbwoolf: I would have to find the resource I used when implementing this with `[react-native-svg-transformer](https://github.com/kristerkari/react-native-svg-transformer)`. I know it was the suggested way to set the svg `currentColor` using a theme color. I'm not totally following your explanation tbh.

packages/ui/src/icons/arrow-down-icon.native.tsx
fbwoolf: Most of our components are based off Radix. https://www.radix-ui.com/icons

packages/ui/src/icons/icon/icon.native.tsx
camerow: > Is this change nec? We have always been able to override the icon color w/out this. This should just set the default color?

I think I misunderstood your original comment - I thought it was regarding `theme.colors['ink.action-primary-default']` which had already been there, you're right this isn't necessary given that the props are spread as the final argument.

packages/ui/src/icons/arrow-down-icon.native.tsx
fbwoolf: Was thinking on this, might be best to get @tigranpetrossian opinion here bc he is mostly owning the UI library? I don't have a strong opinion if others prefer a diff setup. It would just be quite a chunk of work to refactor as they are currently used this way in both web and mobile.

tigranpetrossian (approved):

camerow:
Closing this because @tigranpetrossian took it on.

PR #660, opened 2024-11-26
fix: hide the scrollbar on accounts and collectibles widgets, closes LEA-1432

tigranpetrossian:
#### Hides the horizontal scrollbar on scrollabe widgets.

The scrollability is obvious enough from the designs. Unlike with browsers, I can't think of any obvious gotchas related to not having scrollbars. Will revisit with a less destructive approach if anything turns up.

https://github.com/user-attachments/assets/90865f09-bb6b-49f3-9e1b-b1e8d6d1c1d3


pete-watters (approved):
Thanks @tigranpetrossian 🚀

PR #666, opened 2024-11-27
Will/mobile-segment-events

camerow:
adds some basic event tracking for mobile

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import { StatusBar } from 'react-native';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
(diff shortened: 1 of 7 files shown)


apps/mobile/src/app/_layout.tsx
```diff
@@ -41,6 +43,15 @@ initAppServices();
...
 
 export default function RootLayout() {
+  const pathname = usePathname();
+  const params = useGlobalSearchParams();
+
+  useEffect(() => {
+    analytics.screen(pathname, {
+      params,
+    });
+  }, [pathname, params]);
```
kyranjamie, line 53: Even though a few LOC, this'd be great as its own hook with api like

`useOnRouteChange(route => analytics.screen(pathname, { params }))`

packages/analytics/src/events.ts
```diff
@@ -102,3 +106,28 @@ interface HistoricalEvents {
...
+  | { fiat_currency: string }
+  | {
+      network:
+        | 'mainnet'
+        | 'testnet'
+        | 'testnet4'
+        | 'signet'
+        | 'sbtcTestnet'
+        | 'sbtcDevenv'
+        | 'devnet';
```
kyranjamie, line 124: We could pull values from models such that this'll update when we add new networks

camerow: Done

pete-watters:
@camerow just re-visiting this as I think we aren't respecting the users preferences here, is that correct? 

We have a setting for it `analyticsPreference` but it seems like we don't use it here and always track? 

Screenshot 2024-12-09 at 10 56 55

PR #669, opened 2024-11-29
feat: add banner component, closes LEA-1801

tigranpetrossian:
Part of [LEA-505](https://linear.app/leather-io/issue/LEA-505/loading-state-for-balances-on-home-screen)

Adds a banner component for displaying warning notifications.
Intentionally not following radix-style pattern as the it has a faily small surface, and I'm hoping that this will be used rather sparingly, and in future replaced with a more subtle yet prominent design that can cover different warning/alert scenarios (once we gather enough).

#### Web
image

#### Mobile

image


tigranpetrossian:
Somehow didn't notice we already had a Callout component. Will adjust that one, add a native version and re-open.

PR #670, opened 2024-12-02
refactor: rewrite Callout for web, add portalling

tigranpetrossian:
Adds a callout for mobile, updates the web version.

### Web
image

* Added ability to display global callouts in the extension using portalling. `<Callout portal />` will render the callout into a corresponding `div`.
* Added `maxContentWidth` prop, to be able to center and limit the content width of global callouts.
* Refactored the component, added a relevant `role`.

### Mobile

https://github.com/user-attachments/assets/e19c40ab-ed51-4a90-ad5d-6f3a64d1bd65


packages/ui/src/components/callout/callout.web.tsx
```diff
@@ -31,38 +31,71 @@ const calloutRecipe = cva({
...
+    <Box className={calloutRecipe({ variant })} role="status" {...rest}>
+      <Flex
+        justifyContent="space-between"
+        maxWidth={maxContentWidth}
+        mx="auto"
         p="space.05"
-        reverse
-        spacing="space.03"
+        gap="space.03"
+        color="ink.text-primary"
```
pete-watters, line 82: Do we always need this color or should it be over-writable?

packages/ui/src/components/callout/callout.web.tsx
```diff
@@ -31,38 +31,71 @@ const calloutRecipe = cva({
...
+  success: <CheckmarkCircleIcon />,
+  warning: <ErrorTriangleIcon />,
+  error: <ErrorTriangleIcon />,
+  default: null,
+};
 
-interface CalloutProps extends BoxProps {
-  icon?: ReactNode;
+interface CalloutOwnProps {
+  /** Title of the callout.  */
```
pete-watters, line 45: I don't think all of these comments are necessary

pete-watters (approved):
Nice work Tigran 👍  I added some questions / suggestions. 

Generally we like using the `Flag` pattern to keep things more consistent so I'm not sure if removing that is an improvement here

packages/ui/panda.config.ts
```diff
@@ -17,6 +17,16 @@ export default defineConfig({
 
   strictTokens: false,
 
+  staticCss: {
+    css: [
+      {
+        properties: {
```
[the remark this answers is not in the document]
tigranpetrossian: Prop renaming [isn't supported by panda](https://panda-css.com/docs/concepts/style-props#property-renaming) unfortunately, so they added a way to specify custom properties. It's a bit of a crutch, I don't really like the disconnect of having to setup something like this, but decided to chose DX (avoiding confusion over what `maxWidth` is and not overriding the actual `maxWidth`) over generating few bytes of CSS.

packages/ui/src/components/callout/callout.web.tsx
tigranpetrossian: This was carried over from before the refactor (just different location). We do need it to display the color correctly. Happens to over-writable too but incidentally.

tigranpetrossian:
> Nice work Tigran 👍 I added some questions / suggestions.
> 
> Generally we like using the `Flag` pattern to keep things more consistent so I'm not sure if removing that is an improvement here

@pete-watters The usage of `Flag` here was problematic, in that it used `reverse` for purely visual reasons, which [can be problematic](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction#accessibility)

packages/ui/panda.config.ts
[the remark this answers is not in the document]
tigranpetrossian: Oh and this totally should be in the preset, thanks! I actually have no idea how this worked without being in the preset 😳

packages/ui/src/components/callout/callout.web.tsx
pete-watters: I think that here the only useful comment is:
```
Limit the content width to the specified value. Suitable for full-width callouts displayed at the top of the page.
```

The rest are self explanatory. If you want to add docs I would add them into the `.story` file instead, maybe in the docs description?

```
const meta: Meta<typeof Component> = {
  component: Component,
  tags: ['autodocs'],
  title: 'Callout',
  parameters: {
    docs: {
      description: { 
          component: '...',
      },
```

PR #673, opened 2024-12-02
fix: hardcode broken translations for leatherhood release, ref LEA-1780

pete-watters:
In our current Testflight build, some of our CrowdIn translations are not loaded and instead we are showing the key.

I'm still investigating why they aren't loaded properly build but thought it could be prudent to hardcode the text for now so the correct text will appear now and we can do some pre-release tests.

This is an interim solution to make what we have now more releasable pending fixing the underlying problem

### DIFF
--- a/apps/mobile/src/app/(home)/create-new-wallet.tsx
+++ b/apps/mobile/src/app/(home)/create-new-wallet.tsx
@@ -37,6 +37,12 @@ export default function CreateNewWallet() {
     void tempMnemonicStore.setTemporaryMnemonic(tempMnemonic);
   }, []);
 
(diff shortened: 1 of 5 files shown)


pete-watters:
@camerow , I assume you are OK with temporarily hardcoding these values based on [this thread](https://trustmachines.slack.com/archives/C05LAP952E7/p1733164437012019?thread_ts=1733148496.430359&cid=C05LAP952E7)? 

@alexp3y - can you please help me get this merged? I'm hopeful we can get this + balances on to a test build soon

alexp3y (approved):

PR #677, opened 2024-12-04
fix: replace barrel imports with relative paths

tigranpetrossian:
Switch out `import {} from 'native'` with relative imports to avoid Storybook Native crashing from unresolved tsconfig paths.

### DIFF
--- a/packages/ui/src/components/approver/components/approver-actions.native.tsx
+++ b/packages/ui/src/components/approver/components/approver-actions.native.tsx
@@ -1,5 +1,6 @@
-import { Box } from 'native';
 import { HasChildren } from 'src/utils/has-children.shared';
+
+import { Box } from '../../box/box.native';
 
 export function ApproverActions({ children }: HasChildren) {
   return (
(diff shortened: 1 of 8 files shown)


pete-watters (approved):
Nice work @tigranpetrossian 👍  

Is this clash as we are exporting things from `native.ts`? 

Maybe we could rename that slightly instead and keep the barrel imports?

tigranpetrossian:
> Nice work @tigranpetrossian 👍
> 
> Is this clash as we are exporting things from `native.ts`?
> 
> Maybe we could rename that slightly instead and keep the barrel imports?

Two separate issues:
1. Importing X from Y via 'native' (barrel exports), while X is being imported and exported from 'native'. So that's a circular reference.
2. Storybook native doesn't resolve absolute imports.

These were the only few imports used this way in the entire library, the rest of the files use relative paths. 
I think we should use aliases in the packages in the same way we use in `mobile`, but apparently `tsup` doesn't currently work with `import X from '@/components/X'` either, so we'll need to look into other bundling libraries first.

PR #678, opened 2024-12-04
feat: add success btn recipe

alter-eggo:
This pr adds success btn


pete-watters (approved):
Thanks @alter-eggo 👍 hopefully oncer released we'll see this showing up.

packages/panda-preset/src/recipes/button-recipe.ts
```diff
@@ -33,11 +33,11 @@ export const buttonRecipe = defineRecipe({
     size: {
       sm: {
         px: 'space.02',
-        py: 'space.01',
+        height: '40px',
```
pete-watters, line 36: Can you use `space.07` + `space.08` here?

pete-watters: Sorry, I thought you were changing `py` 👍

PR #680, opened 2024-12-04
feat: stx approver

edgarkhanzadian:
https://github.com/user-attachments/assets/c1d8ec6b-209c-4c22-852f-e73e6701226e

### DIFF
--- a/apps/mobile/src/features/stacks-tx-signer/stacks-tx-signer.tsx
+++ b/apps/mobile/src/features/stacks-tx-signer/stacks-tx-signer.tsx
@@ -1,0 +1,147 @@
+import { useState } from 'react';
+
+import { formatBalance } from '@/components/balance/balance';
+import { useStxMarketDataQuery } from '@/queries/market-data/stx-market-data.query';
+import { useAccountByIndex } from '@/store/accounts/accounts.read';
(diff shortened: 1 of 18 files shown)


apps/mobile/src/features/stacks-tx-signer/stacks-tx-signer.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { ApproverButtons } from '../approver/components/approver-buttons';
+import { OutcomeAddressesCard } from '../approver/components/outcome-addresses-card';
+import { StacksFeeCard } from '../approver/components/stacks-fee-card';
+import { StacksOutcome } from '../approver/components/stacks-outcome';
+import { ApproverState } from '../approver/utils';
+import { useBroadcastStxTransaction } from './use-broadcast-stx-transaction';
+
+function formReviewTxSummary({
+  tx,
+  symbol = 'STX',
```
kyranjamie, line 34: Not sure on setting default here, seems like a behaviour that might cause bugs. E.g. missing symbol and not realising because there's no type error. Could we just be explicit every time?

kyranjamie (approved):
Great work @edgarkhanzadian, very tidy implementation

fbwoolf (approved):
Nice work @edgarkhanzadian, great to see this all come together! 🚀

PR #682, opened 2024-12-05
fix: gate features not ready for production release

pete-watters:
This PR adds a basic feature flag based on the users `NODE_ENV` to hide some things in the mobile app that aren't ready. 

It works off [this notion doc](https://www.notion.so/trustmachines/Leatherhood-First-Release-151d74b694c780cc859ecf4e2a979dd5) which I made [this issue](https://linear.app/leather-io/issue/LEA-1818/leatherhood-cleanup) for. 

On that document it says to hide some things and disable others, however when disabling there is currently no extra visual state to signify it to users so I decided to just hide everything. I can change this as desired but thought it best to first hide. 

You can see a demos here of

### Disabling notify me
https://github.com/user-attachments/assets/611c97a3-8640-4e88-812c-bdb47401aab3

### Disabling buttons
https://github.com/user-attachments/assets/f8049a04-eaa5-446f-939a-a95484576a90

Here is a demo of the APP with some features hidden:

https://github.com/user-attachments/assets/73683f63-bfd5-4aac-b6dc-322534c0b2ca

### DIFF
--- a/apps/mobile/src/app/(home)/settings/index.tsx
+++ b/apps/mobile/src/app/(home)/settings/index.tsx
@@ -7,6 +7,7 @@ import { useAuthContext } from '@/components/splash-screen-guard/use-auth-contex
 import { NetworkBadge } from '@/features/settings/network-badge';
 import { AppRoutes } from '@/routes';
 import { TestId } from '@/shared/test-id';
+import { isFeatureEnabled } from '@/utils/feature-flag';
 import { t } from '@lingui/macro';
 import * as Application from 'expo-application';
 import { useRouter } from 'expo-router';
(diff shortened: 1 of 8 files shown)


apps/mobile/src/app/(home)/settings/index.tsx
```diff
@@ -64,10 +65,14 @@ export default function SettingsScreen() {
             id: 'settings.display.cell_title',
             message: 'Display',
           })}
-          caption={t({
-            id: 'settings.display.cell_caption',
-            message: 'Theme, bitcoin unit and more',
-          })}
+          caption={
```
pete-watters, line 68: I added this handling of text as it says `Theme, bitcoin unit and more` but that's not correct if we hide the other settings

kyranjamie (approved):

pete-watters:
@kyranjamie @edgarkhanzadian : I pushed [one more commit](https://github.com/leather-io/mono/pull/682/commits/e8395a7711b1f7766c964aec1edc92e7e09f1c17) to keep the `ActionBar` enabled for now based on [this discussion](https://trustmachines.slack.com/archives/C05LRS7G44R/p1733381141338449)

We can easily hide it again at a later date if needs be

PR #687, opened 2024-12-06
fix: lingui explicit ids

edgarkhanzadian:
Apparently we didn't have a default set for explicit ids and thus lingui sometimes couldn't tell if the msgid should be considered an explicit id or a generated id. (and it mostly tried to default us to generated id).

Setting this default to explicit ids should fix the issue.

### DIFF
--- a/apps/mobile/src/features/psbt-signer/psbt-signer.tsx
+++ b/apps/mobile/src/features/psbt-signer/psbt-signer.tsx
@@ -98,7 +98,7 @@ export function PsbtSigner({ psbtHex, onEdit, onSuccess }: PsbtSignerProps) {
             message: 'Show advanced options',
           })}
           titleOpened={t({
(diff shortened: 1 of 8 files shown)


pete-watters (approved):

apps/mobile/src/features/psbt-signer/psbt-signer.tsx
```diff
@@ -98,7 +98,7 @@ export function PsbtSigner({ psbtHex, onEdit, onSuccess }: PsbtSignerProps) {
             message: 'Show advanced options',
           })}
           titleOpened={t({
-            id: 'approver.advanced.show',
+            id: 'approver.advanced.hide',
```
pete-watters, line 101: Just double checking this is intentional?

fbwoolf (approved):

kyranjamie (approved):

PR #690, opened 2024-12-06
fix: add pull to refresh to home and accounts, ref LEA-1814

pete-watters:
This PR adds a pull to refresh to `Home` + the `Account` page. 

On refresh I invalidate all queries: 

https://github.com/user-attachments/assets/a338f1f9-caa3-422b-89ea-d78719890604

### DIFF
--- a/apps/mobile/src/components/page/page.layout.tsx
+++ b/apps/mobile/src/components/page/page.layout.tsx
@@ -1,4 +1,5 @@
-import { useRef } from 'react';
+import { useCallback, useRef, useState } from 'react';
+import { RefreshControl } from 'react-native';
 import { ScrollView } from 'react-native-gesture-handler';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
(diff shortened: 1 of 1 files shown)


apps/mobile/src/components/page/page.layout.tsx
```diff
@@ -18,6 +20,15 @@ export function PageLayout({ children }: HasChildren) {
   const actionBarRef = useRef<ActionBarMethods>(null);
   const contentOffsetBottom = bottom + ACTION_BAR_TOTAL_HEIGHT;
   const contentOffsetTop = top;
+  const [refreshing, setRefreshing] = useState(false);
+
+  const onRefresh = useCallback(() => {
+    setRefreshing(true);
+    queryClient.invalidateQueries();
```
pete-watters, line 27: Perhaps I should be more nuanced here and invalidate specific balance queries only?

This is a sledgehammer approach to invalidate all queries but we only use wallet/ balance related ones now so I believe thats OK.

fbwoolf: Yeah, seems ok for now and we can refine it later.

fbwoolf (approved):

PR #691, opened 2024-12-06
feat: add new version of the Cell

tigranpetrossian:
Refactors the Cell component to support migrating various list items to Cell instead of Pressable + ItemLayout combination.

### Cell changes
* Cell is now full-width with its own horizontal padding to support background transitions
* Only optionally pressable. We have few places where it's used as a non-interactive display component.
* Added background color transition on press
* Supports all possible permutations [in the design](https://www.figma.com/design/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?node-id=17780-13392&t=BNJJ7JCoenGLd5ga-4), vertically centers single labels.
* Removed the Switch & Radio from within the Cell in favor of direct usage where necessary. The new `Cell.Aside` accepts arbitrary JSX that can display anything, see the implementation of `SettingsListItem`

### Settings changes
Only migrated settings pages, both to use as a testing ground as it already used Cell directly, and avoid this PR growing out of proportions. Remaining list items not using Cell will need to be updated separately.

* Extracted direct usages of the `Cell` within settings pages into a `SettingsListItem` component to minimize the signature, and keep any future changes contained within a single file, as this is a really widely used component.
* Wrapped all groups of setting list items into a SettingList, to both provide logical & semantic grouping, and some negative margins to support full-width Cells without updating the containing layouts within this PR.

### Transitions preview

https://github.com/user-attachments/assets/9d703478-6f92-4d44-af1b-5ff9e7bb8b3f

### DIFF
--- a/packages/ui/src/components/cell/components/cell-icon.native.tsx
+++ b/packages/ui/src/components/cell/components/cell-icon.native.tsx
@@ -1,0 +1,7 @@
+import { Box, BoxProps } from '../../box/box.native';
+
+export interface CellIconProps extends BoxProps {}
+
+export function CellIcon(props: CellIconProps) {
+  return props.children ? <Box {...props} /> : null;
+}
--- a/apps/mobile/src/app/(home)/hardware-wallets.tsx
+++ b/apps/mobile/src/app/(home)/hardware-wallets.tsx
@@ -1,6 +1,8 @@
 import { useRef, useState } from 'react';
 
 import { AnimatedHeaderScreenLayout } from '@/components/headers/animated-header/animated-header-screen.layout';
+import { SettingsList } from '@/components/settings/settings-list';
+import { SettingsListItem } from '@/components/settings/settings-list-item';
 import {
   NotifyUserSheetData,
   NotifyUserSheetLayout,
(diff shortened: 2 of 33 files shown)


packages/ui/src/components/cell/components/cell-icon.native.tsx
```diff
@@ -0,0 +1,7 @@
+import { Box, BoxProps } from '../../box/box.native';
+
+export interface CellIconProps extends BoxProps {}
+
+export function CellIcon(props: CellIconProps) {
+  return <Box {...props} />;
+}
```
kyranjamie, line 7: is this component really needed?

kyranjamie (approved):
Great work @tigranpetrossian

packages/ui/src/components/cell/components/cell-icon.native.tsx
tigranpetrossian: I considered removing this as it's right now noop and holds no opinions on its contents, but thought to keep around as we'll likely need to tweak the spacing for the re-designed double icon Avatars that leak out of the square bounding box by design.

Screenshot 2024-12-06 at 18 57 57

fbwoolf (approved):
Lgtm @tigranpetrossian, nice refactor. 👍

PR #696, opened 2024-12-07
fix: strings with placeholders

edgarkhanzadian:
Apparently we also had to parse strings that have placeholders in them.

### DIFF
--- a/apps/mobile/src/locales/index.ts
+++ b/apps/mobile/src/locales/index.ts
@@ -5,12 +5,36 @@ import { formatter } from '@lingui/format-po';
 
 const prodHash = 'adcc836a66272410c0b94e9twcj'; // with po file format
 const devHash = 'a6b025ebb570b783a20df09twcj'; // with po file format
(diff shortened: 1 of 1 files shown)


apps/mobile/src/locales/index.ts
```diff
@@ -5,14 +5,38 @@ import { formatter } from '@lingui/format-po';
...
 let LOCALES: string[] = [];
 export function getAvailableLocales() {
   return LOCALES;
 }
 
+const placeholderRegex = /[.]*{([^{}]*)}[.]*/g;
+
+type StringWithPlaceholders = (string | string[])[];
+// TODO: really not something we should be doing. We are practically parsing po files ourselves
+// and i don't want it to be our job. Maybe we need to just ditch po format altogether and just use json.
```
camerow, line 20: I think using JSON is probably going to be much more simple as a future enhancement as well. `po`

camerow (approved):

PR #698, opened 2024-12-09
leatherhood: pre release cleanup

pete-watters:
I thought I had merged this work last week but I actually lost it in a rebase so re-did it:
- hide action bar in production version
- only show network badge if non mainnet and adjust position on home screen

### DIFF
--- a/apps/mobile/src/components/headers/home-header.tsx
+++ b/apps/mobile/src/components/headers/home-header.tsx
@@ -1,7 +1,21 @@
+import { NetworkBadge } from '@/features/settings/network-badge';
+
+import { Box } from '@leather.io/ui/native';
+
 import { HeaderLeatherLogo } from './components/header-leather-logo';
 import { HeaderOptions } from './components/header-options';
 import { HeaderLayout } from './header.layout';
 
 export function HomeHeader() {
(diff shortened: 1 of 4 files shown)


apps/mobile/src/components/headers/home-header.tsx
```diff
@@ -1,7 +1,21 @@
...
 import { HeaderLayout } from './header.layout';
 
 export function HomeHeader() {
-  return <HeaderLayout leftElement={<HeaderLeatherLogo />} rightElement={<HeaderOptions />} />;
+  return (
+    <HeaderLayout
+      leftElement={
+        <Box flexDirection="row" alignItems="center">
+          <HeaderLeatherLogo />
+          <NetworkBadge px="0" ml="-1" />
```
pete-watters, line 15: I added this working off [this design](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%AA-App-Explorations-2024-H2?node-id=10194-400055&node-type=instance&m=dev).

In there the logo has less padding but I don't want to risk breaking something so giving this negative margin to move the badge into place:
Screenshot 2024-12-09 at 06 53 59

PR #701, opened 2024-12-09
fix: add support links and text updates, ref #LEA-1835

pete-watters:


apps/mobile/src/shared/constants.ts
```diff
@@ -1,4 +1,16 @@
...
 export const HEADER_HEIGHT = 64;
 
 export const GITHUB_ORG = 'leather-io';
 export const GITHUB_REPO = 'mono';
+
+export const LEATHER_GUIDES_MOBILE_ANALYTICS = `${LEATHER_GUIDES_URL}/mobile-analytics`;
+export const LEATHER_GUIDES_MOBILE_ACCOUNT_IDENTIFIER = `${LEATHER_GUIDES_URL}/mobile-account-identifier`;
+export const LEATHER_GUIDES_MOBILE_BITCOIN_UNIT = `${LEATHER_GUIDES_URL}/mobile-bitcoin-unit`;
+// TODO: Add this to Leather Guides on Framer
+export const LEATHER_GUIDES_MOBILE_APP_AUTHENTICATION = `${LEATHER_GUIDES_URL}/mobile-app-authentication`;
```
pete-watters, line 12: I added this URL for `LEATHER_GUIDES_MOBILE_APP_AUTHENTICATION` but the link is not live yet. I passed on the URL so it can be added later

apps/mobile/src/shared/constants.ts
```diff
@@ -1,4 +1,16 @@
...
+
+// TODO: On Framer but no direct link from App
```
pete-watters, line 14: I just added these guides as we have them but no where links to them yet

packages/ui/src/components/sheet/components/sheet-header.native.tsx
```diff
@@ -19,11 +19,13 @@ export function SheetHeader({ title, icon, onPressSupport }: SheetHeaderProps) {
           </Text>
         </Box>
       </Box>
-      <Box alignItems="flex-end" flex={1}>
-        <TouchableOpacity onPress={onPressSupport} zIndex="10">
-          <QuestionCircleIcon variant="small" />
-        </TouchableOpacity>
-      </Box>
+      {onPressSupport && (
```
pete-watters, line 22: Sometimes we were showing a `?` icon when there was no link so I have changed that

apps/mobile/src/components/sheets/warning-sheet.layout.tsx
```diff
@@ -59,9 +58,6 @@ export function WarningSheetLayout({
                   />
                 </Box>
               }
-              onPressSupport={() => {
```
pete-watters, line 62: I got rid of this as we are just showing a `?` in `Remove wallet` with no link. 

We give the user a lot of information there so I'm not sure we need the link but will double check cc @markmhendrickson 

Screenshot 2024-12-09 at 10 20 57

apps/mobile/src/app/(home)/settings/help/index.tsx
```diff
@@ -23,10 +26,12 @@ export default function SettingsHelpScreen() {
           })}
           caption={t({
             id: 'help.support.cell_caption',
-            message: 'Placeholder',
+            message: 'Contact our support team',
```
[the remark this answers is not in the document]
pete-watters: Yeah lets think about it and see 👍 with your recent fixes it seems to be working well and publishing OTA. It could be we just need to document it a bit / have a stricter process.

I think it would still be nice to have an emergency fallback text though

apps/mobile/src/components/sheets/warning-sheet.layout.tsx
pete-watters: I have added a placeholder URL here [as per](https://trustmachines.slack.com/archives/C05LRS7G44R/p1733740921786479?thread_ts=1733381141.338449&cid=C05LRS7G44R)

apps/mobile/src/shared/constants.ts
pete-watters: `LEATHER_GUIDES_MOBILE_REMOVE_WALLET` is also not yet live but I added it

tigranpetrossian (approved):

apps/mobile/src/components/sheets/warning-sheet.layout.tsx
markmhendrickson: I'd recommend we keep the link but cut down on the text. We can do this later via Crowdin though, so it's not imperative for this week's release

PR #705, opened 2024-12-09
fix: remove number conversion in sumMoney utility

alexp3y:
This PR updates the `sumMoney` utility to avoid an unnecessary type conversion from `BigDecimal` to `number`, which was leading to a loss of precision and inaccurate calculation.

This fixes a bug in mobile where some zero balances are displaying as "-$0.00" or  "<$0.01". 

The bug originated in the sub-balance calculations of the new balance helper function `createBtcCryptoAssetBalance` in `@leather.io/utils`, which utilizes the `sumMoney` function. The rounded balance returned by `sumMoney` is subsequently subtracted from the original total, causing very small positive or negative differences (depending on the direction of the rounding).

image


pete-watters (approved):
Thanks @alexp3y 🚀

PR #706, opened 2024-12-09
fix: only use analytics if consented

pete-watters:


apps/mobile/src/utils/analytics.ts
```diff
@@ -13,3 +14,24 @@ export const analytics = configureAnalyticsClient<SegmentClient>({
     platform: 'mobile',
   },
 });
+export function useAnalytics() {
```
pete-watters, line 17: @camerow I think we need to respect this setting. Tagging you as an FYI as I look at it 

See https://github.com/leather-io/mono/pull/666#issuecomment-2527590291

pete-watters:
This was fixed in https://github.com/leather-io/mono/pull/713

PR #713, opened 2024-12-09
Fix/analytics-opt-out

camerow:


pete-watters (approved):
Great work @camerow ! thanks for fixing this

PR #721, opened 2024-12-11
refactor: Add press effect support to Pressable, remove TouchableOpacity closes LEA-659 LEA-1743 LEA-1544

tigranpetrossian:
This is hopefully the last sizeable PR addressing and closing [LEA-659](https://linear.app/leather-io/issue/LEA-659/define-tapon-press-effect-for-interactive-elements) and its blocker issues.

## Pressable updates

Pressable now supports declarative transitions through a prop:

#### Before
```tsx
const { pressed, onPressIn, onPressOut } = usePressedState();

const animatedStyle(() => ({
  backgroundColor: withSpring(pressed ? theme.colors['ink.background-secondary'] : theme.colors['ink.background-primary'])
}))

return (
  <Pressable
    onPressIn={onPressIn}
    onPressOut={onPressOut}
    style={animatedStyle}
  />
)
```
#### After
```tsx
<Pressable
  pressEffects={
    backgroundColor: { 
      from: 'ink.background-primary', 
      to: 'ink.background-primary'
    }
  }
/>
```
This adds a compatibility layer between our tokens and `reanimated` by converting theme values to raw styles under the hood, so we can avoid long and repetitive `useAnimatedStyle`definitions and manually referencing the theme values through `theme.key['token']` format. We have about a dozen buttons using TouchableOpacity that need a custom pressed state style (with more to come), so this hopefully saves us some time down the line.

### More related updates
* replaced all occurrences of the TouchableOpacity with Pressable + a custom temporary preset mimicking TouchableOpacity's pressed state. All these components will need to be passed to the design team for styles to be gradually updated.
* Migrated remaining variations of Pressable + ItemLayout + Flag to the new Cell. 

### Future improvements to Pressable 
Descoping these to get this PR out, will address in the near future:
* transform transitions: currently unsupported – restyle doesn't provide a transform prop out of the box
* ability to specify global animation settings for all specified props (delay, spring config, etc.)

## Haptics
Since this has been part of [LEA-659](https://linear.app/leather-io/issue/LEA-659/define-tapon-press-effect-for-interactive-elements)
* Added haptic feedback to accound cards, and the wallet creation flow with a success feedback once a wallet is created.
* Replaced the haptic icon in settings.
This is still experimental and will likely requirement few more iterations in the future. `expo-haptics` presets are very limiting, and the softest feedback feels quite strong compared to other apps. On top of that we need to more carefully consider which flows/elements need to have haptic feedback.

### DIFF
--- a/apps/mobile/src/features/approver/components/inputs-outputs-card.tsx
+++ b/apps/mobile/src/features/approver/components/inputs-outputs-card.tsx
@@ -29,48 +29,52 @@ export function InputsAndOutputsCard({ inputs, outputs }: InputsAndOutputsCardPr
 
   const outputsWithBalance = outputs.map(addBalance);
   return (
-    <>
-      <Text variant="label01">
+    <Box gap="5">
+      <Text variant="label02">
         {t({
           id: 'approver.inputs-outputs.title',
           message: 'Inputs and Outputs',
         })}
       </Text>
-      <Text variant="label01">
-        {t({
-          id: 'approver.inputs-outputs.input.title',
-          message: 'Input',
-        })}
-      </Text>
-      <Box gap="4">
-        {inputsWithBalance.map(input => (
-          <UtxoRow
-            key={input.txid + input.address + input.btcBalance.amount}
-            txid={input.txid}
-            address={input.address}
-            btcBalance={input.btcBalance}
-            usdBalance={input.usdBalance}
-            isLocked
-          />
-        ))}
+      <Box gap="1">
+        <Text variant="label02">
+          {t({
+            id: 'approver.inputs-outputs.input.title',
+            message: 'Input',
+          })}
+        </Text>
+        <Box mx="-5">
+          {inputsWithBalance.map(input => (
+            <UtxoRow
+              key={input.txid + input.address + input.btcBalance.amount}
+              txid={input.txid}
+              address={input.address}
+              btcBalance={input.btcBalance}
+              usdBalance={input.usdBalance}
+              isLocked
+            />
+          ))}
+        </Box>
       </Box>
-      <Text variant="label01">
-        {t({
-          id: 'approver.inputs-outputs.output.title',
-          message: 'Output',
-        })}
-      </Text>
-      <Box gap="4">
-        {outputsWithBalance.map(output => (
-          <UtxoRow
-            key={output.address + output.btcBalance.amount}
-            address={output.address}
-            btcBalance={output.btcBalance}
-            usdBalance={output.usdBalance}
-            isLocked
-          />
-        ))}
+      <Box gap="1">
+        <Text variant="label02">
+          {t({
+            id: 'approver.inputs-outputs.output.title',
+            message: 'Output',
+          })}
+        </Text>
+        <Box mx="-5">
+          {outputsWithBalance.map(output => (
+            <UtxoRow
+              key={output.address + output.btcBalance.amount}
+              address={output.address}
+              btcBalance={output.btcBalance}
+              usdBalance={output.usdBalance}
+              isLocked
+            />
+          ))}
+        </Box>
       </Box>
-    </>
+    </Box>
   );
 }
--- a/packages/ui/src/components/pressable/pressable.native.tsx
+++ b/packages/ui/src/components/pressable/pressable.native.tsx
@@ -4,7 +4,10 @@ import { type GestureResponderEvent, Pressable as RNPressable } from 'react-nati
 import { isString } from '@leather.io/utils';
 
 import { useHaptics } from '../../hooks/use-haptics.native';
-import { PressableCore, PressableCoreProps, PressableRestyleProps } from './pressable-core.native';
+import { usePressedState } from '../../hooks/use-pressed-state.native';
+import { PressableCore, PressableCoreProps } from './pressable-core.native';
+import { PressEffects } from './pressable.types.native';
+import { usePressEffectStyle } from './pressable.utils.native';
 
 type PressableHapticFeedbackType = 'soft' | 'light' | 'medium' | 'heavy' | 'rigid';
 
@@ -14,6 +17,7 @@ interface HapticConfig {
 }
 
 type PressableElement = ElementRef<typeof RNPressable>;
+
 interface PressableOwnProps {
   /**
    * Configure haptic feedback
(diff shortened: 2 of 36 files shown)


packages/ui/src/components/pressable/pressable.native.tsx
```diff
@@ -95,9 +80,35 @@ const PressableBase = forwardRef<PressableElement, PressableBaseProps>(
...
+        style={[pressEffectStyle, style]}
+        {...rest}
+      />
+    );
   }
 );
 
-export const Pressable = Animated.createAnimatedComponent(PressableBase);
-export type PressableProps = AnimatedProps<PressableBaseProps>;
+Pressable.displayName = 'Pressbale';
```
kyranjamie, line 97:
```suggestion
Pressable.displayName = 'Pressable';
```

apps/mobile/src/features/approver/components/fee-card.tsx
```diff
@@ -52,31 +50,39 @@ export function FeeCard({ feeType, amount }: FeeCardProps) {
         </Text>
         <FeeBadge type="normal" />
       </Box>
-      <Pressable flexDirection="row" onPress={() => {}} py="3">
-        <Flag
-          img={
-            <Avatar bg="ink.background-secondary" p="1">
+
+      <Box mx="-5">
+        <Cell.Root pressable={true} onPress={() => {}}>
```
pete-watters, line 55: Maybe this shouldn't be pressable with no `onPress` defined? 🤔

apps/mobile/src/features/approver/components/inputs-outputs-card.tsx
```diff
@@ -31,48 +31,52 @@ export function InputsAndOutputsCard({ inputs, outputs }: InputsAndOutputsCardPr
...
-          />
-        ))}
+      <Box gap="1">
+        <Text variant="label02">
+          {t({
+            id: 'approver.inputs-outputs.output.title',
+            message: 'Output',
+          })}
+        </Text>
+        <Box mx="-5">
```
pete-watters, line 68: Are all these `mx="-5"` to override the child? Can we avoid that? 🤔

packages/ui/src/components/pressable/pressable.native.tsx
```diff
@@ -62,24 +30,41 @@ interface PressableBaseProps extends RNPressableProps, RestyleProps {
    * <Pressable haptics={{ onPress: 'light', onLongPress: 'rigid' }} />
    */
   haptics?: PressableHapticFeedbackType | HapticConfig;
+  /**
```
pete-watters, line 33: Do we need these comments? Maybe we could replace them with storybook examples

packages/ui/src/components/pressable/pressable.native.tsx
```diff
@@ -95,9 +80,35 @@ const PressableBase = forwardRef<PressableElement, PressableBaseProps>(
...
   }
 );
 
-export const Pressable = Animated.createAnimatedComponent(PressableBase);
-export type PressableProps = AnimatedProps<PressableBaseProps>;
+Pressable.displayName = 'Pressable';
+
+/**
+ * Press effect preset to mimic react-native Touchable transition.
+ * This preset is up for removal. It's only used for components that previously used Touchable, and require a new design for the pressed state.
```
pete-watters, line 101: Is there a task open for this new design work we can link here?

pete-watters (approved):
Nice work @tigranpetrossian 🎉

packages/ui/src/components/pressable/pressable.native.tsx
tigranpetrossian: Good call. None at the moment, we have few scattered conversations here and there, I'll make one and follow up.

apps/mobile/src/features/approver/components/fee-card.tsx
tigranpetrossian: Not sure what's the intent there, I just moved over the existing logic. Perhaps this is still being worked on by Edgar or Fara.

apps/mobile/src/features/approver/components/inputs-outputs-card.tsx
tigranpetrossian: Good question. Many (if not all) of our layouts currently have a horizontal padding applied by default. This is a bit too opinionated and is defined in way too many places for me to be able to address within this PR. Cells and few other components on the other hand are full-width–something that was missed during initial implementations.
I'm hoping to address this at some point with adding few layout primitives, but this isn't a trivial task unfortunately.

packages/ui/src/components/pressable/pressable.native.tsx
tigranpetrossian: These are JSDocs

packages/ui/src/components/pressable/pressable.native.tsx
pete-watters: If there isn't an updated one yet I'd stop calling it `legacy`

apps/mobile/src/features/approver/components/inputs-outputs-card.tsx
pete-watters: Thanks! Yes it seems its a SMELL of an underlying issue we should solve but agreed too much for this PR.

packages/ui/src/components/pressable/pressable.native.tsx
tigranpetrossian: Thanks @pete-watters, wrote a [linear issue](https://linear.app/leather-io/issue/LEA-1859/press-effects-for-legacy-touchableopacity-elements) and linked in the comment. Re: `legacy`, would marking it as `@deprecated` be more appropriate you think?

pete-watters: great idea!

packages/ui/src/components/pressable/pressable.native.tsx
pete-watters: I get that but often comments become stale but in storybook it will break if props change. 

We had this chat before though and it's not a big deal for me, just _nit_-ing 😄

apps/mobile/src/features/approver/components/fee-card.tsx
fbwoolf: @edgarkhanzadian is working on this.

fbwoolf (approved):

apps/mobile/src/features/approver/components/fee-card.tsx
pete-watters: 👍 now we have a release out, albeit a beta one we should tighten up on stuff like this

PR #726, opened 2024-12-12
CI: mobile maestro E2E tests

pete-watters:
This PR covers adding some basic E2E tests to cover some of the flows we added for the Leatherhood release. 

- Creating / restoring wallets
- Wallet balances for STX/BTC
- Multiple wallet functionality (create / restore multiple wallets, manage and configure those wallets)
- themes
- change networks
- Links to web guides and updated guides for mobile features (mobile guides currently not publicly shown)

We need to add better tests for Biometrics. 

When running tests locally I see we have some fails and I can't see the tests being run in CI (Expo or Github). @edgarkhanzadian do you know where they should be ran?

I had these failures running locally:
```
[Failed] rename-account (1m 7s) (Assertion is false: "testAccount" is visible)
[Failed] change-account-icon (1m 4s) (Element not found: Text matching regex: Confirm)
```

I believe we should spend some time:
- organising the test files and [introducing page objects](https://maestro.mobile.dev/examples/page-object-model)
- getting tests to run in CI 
- add mocking of real responses to test balances better (all tests now are for empty wallets)

### DIFF
--- a/apps/mobile/maestro/flows/change-theme.yaml
+++ b/apps/mobile/maestro/flows/change-theme.yaml
@@ -1,0 +1,21 @@
+appId: io.leather.mobilewallet
+---
+- launchApp
+- runFlow: ../shared/clean-up.yaml
+- tapOn:
+    id: 'homeSettingsButton'
+- tapOn: Display Theme and account identifier
+- tapOn: Theme System
+- tapOn: Light
+- tapOn: DISPLAY
+- assertVisible: Theme Light
+- assertNotVisible: Theme System
+- assertNotVisible: Theme Dark
+- tapOn: Theme Light
+- tapOn: DISPLAY
+- tapOn: Theme Light
+- tapOn: Dark
+- tapOn: DISPLAY
+- assertVisible: Theme Dark
+- assertNotVisible: Theme System
+- assertNotVisible: Theme Light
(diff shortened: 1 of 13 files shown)


apps/mobile/maestro/flows/wallet-management.yaml
camerow, line 1: Needs some whitespace fixes maybe?

apps/mobile/maestro/flows/change-theme.yaml
camerow, line 1: whitespace off?

apps/mobile/maestro/flows/wallet-management.yaml
pete-watters: 🤷  I used `maestro studio` to stub out a lot of this from in app clicks. I think it looks OK compared to other tests we had already. 

I can add whitespace but I'm nit sure where?

kyranjamie (approved):
Thanks @pete-watters — I haven't written any maestro tests but they look pretty straightforward. Curious how you'd use variables and write more complex tests though?

pete-watters:
> Thanks @pete-watters — I haven't written any maestro tests but they look pretty straightforward. Curious how you'd use variables and write more complex tests though?

Thanks! 👍  I will look into it and keep you posted on it. For now adding the basics to cover what we have but we can improve this over time.

They have a great tool called `maestro studio` that lets you generate test code clicking in the app which is handy

We will need to extend the setup to:
- do more realistic tests with mocks (now we are checking new empty wallets)
- get it running in CI properly (its not running in Github and I found some local failures so I guess not in Expo either)

PR #732, opened 2024-12-13
feat: udpate types to conform with extension provider

camerow:
This PR updates the types in our `rpc` package to conform the the _actual_ API running in the `LeatherProvider`

### DIFF
--- a/packages/rpc/src/methods/send-transfer.ts
+++ b/packages/rpc/src/methods/send-transfer.ts
@@ -1,16 +1,46 @@
-import { DefineRpcMethod, RpcParameterByName, RpcRequest, RpcResponse } from '../rpc/schemas';
+import { z } from 'zod';
 
-export interface SendTransferRequestParams extends RpcParameterByName {
-  account?: number;
-  address: string;
-  amount: string;
-}
+import { DefineRpcMethod, RpcRequest, RpcResponse } from '../rpc/schemas';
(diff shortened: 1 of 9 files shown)


packages/rpc/src/methods/send-transfer.ts
```diff
@@ -6,11 +6,29 @@ export interface SendTransferRequestParams extends RpcParameterByName {
...
+interface TransferRecipientParam {
+  address: string;
+  amount: string;
+}
+
+export interface RpcSendTransferParams extends RpcParameterByName {
+  account?: number;
+  recipients: TransferRecipientParam[];
+  network: string;
+}
```
kyranjamie, line 22: Can all these new types be describes as zod schemas, and the types inferred with `z.infer`?

camerow: Yeah, I will conform to the patterns elsewhere. 

What is the advantage of just using zod for types though? I guess down the line validators can be exported?

camerow: Would you be happy with a fast follow on the zod stuff? Would ideally like to keep this scope down - it should be 1:1 replacements but I feel is different in scope than this effort, which is trying to move `extension` rpc types to our package.

kyranjamie (approved):

PR #736, opened 2024-12-13
fix: LEA-1789 hide keyboard after users paste

pete-watters:
This PR:
- adds a fix to hide the keyboard after users paste to the recover wallet screen
- improves tests for restore wallet


apps/mobile/src/app/(home)/recover-wallet.tsx
```diff
@@ -78,10 +78,14 @@ export default function RecoverWallet() {
     }
   }
 
-  function onChangeText(text: string) {
+  async function onChangeText(text: string) {
     checkMnemonic(text);
     setRecoveryMnemonic(text);
     setInputState('default');
+    const copiedString = await Clipboard.getStringAsync();
```
pete-watters, line 85: Logic here is that if they change the text and the change is the same as they clipboard contents then we should dismiss the keyboard

PR #747, opened 2024-12-16
fix: still show BTC send form for zero balance accounts, ref #LEA-1889

pete-watters:
This PR allows us to show the send BTC form for zero balance accounts by adding mocks for zero fees. 

I think more work could be needed here but I just wanted to get the form to appear for Leatherhood as previously we are showing an empty screen for zero balance accounts 

https://github.com/user-attachments/assets/18e40a52-3c74-4fa0-9373-95f0a7144842

### DIFF
--- a/apps/mobile/src/features/send/send-form/loaders/send-form-btc-loader.tsx
+++ b/apps/mobile/src/features/send/send-form/loaders/send-form-btc-loader.tsx
@@ -3,6 +3,7 @@ import {
   useBitcoinAccountTotalBitcoinBalance,
   useBitcoinAccountUtxos,
 } from '@/queries/balance/bitcoin-balance.query';
+import BigNumber from 'bignumber.js';
 
 import { AverageBitcoinFeeRates, Money } from '@leather.io/models';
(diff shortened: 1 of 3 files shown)


apps/mobile/src/features/send/send-form/loaders/send-form-btc-loader.tsx
```diff
@@ -25,7 +26,19 @@ export function SendFormBtcLoader({ account, children }: SendFormBtcLoaderProps)
   const { availableBalance, fiatBalance } = useBitcoinAccountTotalBitcoinBalance(accountId);
 
   // Handle loading and error states
-  if (!utxos.length || !feeRates) return null;
+  // if (!utxos.length || !feeRates) return null;
```
pete-watters, line 29: We will need to improve on this but for now we can at least allow the form to appear and mock out 0 fees

fbwoolf: Prob fine but will prob want to show a loading state or something later? So, fees just aren't loaded yet but eventually do?

fbwoolf (approved):

PR #757, opened 2024-12-17
feat: don't block large UTXO transactions, LEA-1897

camerow:


packages/query/src/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -16,14 +16,18 @@ import { createGetUtxosByAddressQueryOptions } from './utxos-by-address.query';
...
   return <T extends UtxoIdentifier>(utxo: T) => {
-    return !inscriptions.some(
-      inscription =>
-        `${utxo.txid}:${utxo.vout.toString()}` === `${inscription.txid}:${inscription.output}`
-    );
+    return !inscriptions.some(inscription => {
+      // TODO: This is a temporary fix to avoid blocking people from spending large amounts that contain inscriptions.
+      return (
+        `${utxo.txid}:${utxo.vout.toString()}` === `${inscription.txid}:${inscription.output}` &&
+        utxo.value < 1
```
fbwoolf, line 28: I don't think this is a good solution, but I know not coming from you. We would have to apply this to runes too. I think this is just a bandaid and likely to anger another user. If we are going to filter, then we need to communicate these choices as a wallet, or be a wallet that doesn't filter at all? I just think this makes a very arbitrary decision to satisfy as edge case.

camerow: Yeah, I agree, posted that for the team as well. I think @kyranjamie has he best proposed solution here.

PR #759, opened 2024-12-18
fix: wait for utxos to load before creating context [LEA-1900]

edgarkhanzadian:
bug: 

https://github.com/user-attachments/assets/d545d1b3-0dc5-4773-92d0-a413bc875ed5

fix:

https://github.com/user-attachments/assets/42827010-0bc5-4daf-be14-f919b2ce6ca7

### DIFF
--- a/apps/mobile/src/features/send/send-form/loaders/send-form-btc-loader.tsx
+++ b/apps/mobile/src/features/send/send-form/loaders/send-form-btc-loader.tsx
@@ -21,12 +21,12 @@ interface SendFormBtcLoaderProps {
 }
 export function SendFormBtcLoader({ account, children }: SendFormBtcLoaderProps) {
   const accountId = { fingerprint: account.fingerprint, accountIndex: account.accountIndex };
-  const { data: feeRates } = useAverageBitcoinFeeRates();
-  const { data: utxos = [] } = useBitcoinAccountUtxos(accountId);
+  const { data: feeRates, isLoading: isFeeRatesLoading } = useAverageBitcoinFeeRates();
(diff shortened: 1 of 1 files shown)


apps/mobile/src/features/send/send-form/loaders/send-form-btc-loader.tsx
```diff
@@ -21,12 +21,12 @@ interface SendFormBtcLoaderProps {
...
   const accountId = { fingerprint: account.fingerprint, accountIndex: account.accountIndex };
-  const { data: feeRates } = useAverageBitcoinFeeRates();
-  const { data: utxos = [] } = useBitcoinAccountUtxos(accountId);
+  const { data: feeRates, isLoading: isFeeRatesLoading } = useAverageBitcoinFeeRates();
+  const { data: utxos = [], isLoading: isUtxosLoading } = useBitcoinAccountUtxos(accountId);
   const { availableBalance, fiatBalance } = useBitcoinAccountTotalBitcoinBalance(accountId);
 
   // Handle loading and error states
-  // if (!utxos.length || !feeRates) return null;
+  if (isUtxosLoading || isFeeRatesLoading) return null;
```
pete-watters, line 29: Did you check that this will work for empty accounts? 

I had to add [this workaround ](https://github.com/leather-io/mono/pull/747) we need to cleanup as otherwise we show an empty screen if the BTC balance is `0` and users press `Send`. 

We need to handle this more elegantly anyway and show a loading state https://github.com/leather-io/mono/pull/747/files#r1887179417

PR #762, opened 2024-12-18
ci: add new tests to CI, ref LEA-1858

pete-watters:
This PR adds new maestro tests to the Expo CI


docs/e2e-tests.md
```diff
@@ -1,4 +1,4 @@
-# E2E tests with Maestro
+# Mobile App E2E tests with Maestro
```
pete-watters, line 1: I'm thinking maybe I should move this into `maestro/flows/` ?

PR #767, opened 2024-12-19
Disable segment client console logging

tigranpetrossian:
Segment event logs in console are enabled by default. Disabling them to declutter the console a little bit. Let me know if there's value in having those on, I can look into trying the setting to an env variable.


pete-watters (approved):
Thanks @tigranpetrossian 👍 this was cluttering up my console too so happy to clear as I don't think it's needed if it's tracking in segment itself

PR #773, opened 2024-12-19
feat: fee modal for BTC

edgarkhanzadian:
Implemented fee modal for BTC (STX is on the way)

https://github.com/user-attachments/assets/1b9289ad-09d5-4859-8e26-3e1be8dfd077

### DIFF
--- a/apps/mobile/src/features/approver/components/bitcoin-fee-card.tsx
+++ b/apps/mobile/src/features/approver/components/bitcoin-fee-card.tsx
@@ -1,40 +1,23 @@
-import { ReactNode } from 'react';
-
 import { Balance } from '@/components/balance/balance';
 import { FeeBadge } from '@/features/send/fee-badge';
 import { useBtcMarketDataQuery } from '@/queries/market-data/btc-market-data.query';
 import { t } from '@lingui/macro';
 
 import { FeeTypes, Money } from '@leather.io/models';
-import {
-  AnimalChameleonIcon,
-  AnimalEagleIcon,
-  AnimalRabbitIcon,
-  AnimalSnailIcon,
-  Avatar,
-  Box,
-  Cell,
-  ChevronRightIcon,
-  Text,
-} from '@leather.io/ui/native';
-import { baseCurrencyAmountInQuoteWithFallback, match } from '@leather.io/utils';
+import { Avatar, Box, Cell, ChevronRightIcon, Text } from '@leather.io/ui/native';
+import { baseCurrencyAmountInQuoteWithFallback } from '@leather.io/utils';
+
+import { getFeeData } from '../utils';
 
 interface FeeCardProps {
   feeType: FeeTypes;
   amount: Money;
+  onPress(): void;
 }
 
-export function BitcoinFeeCard({ feeType, amount }: FeeCardProps) {
-  const matchFeeType = match<FeeTypes>();
-  const feeIcon = matchFeeType<ReactNode>(feeType, {
-    [FeeTypes.Low]: <AnimalSnailIcon />,
-    [FeeTypes.Middle]: <AnimalRabbitIcon />,
-    [FeeTypes.High]: <AnimalEagleIcon />,
-    [FeeTypes.Custom]: <AnimalChameleonIcon />,
-    [FeeTypes.Unknown]: <AnimalChameleonIcon />,
-  });
-
+export function BitcoinFeeCard({ feeType, amount, onPress }: FeeCardProps) {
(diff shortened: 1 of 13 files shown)


apps/mobile/src/features/approver/components/bitcoin-fee-card.tsx
```diff
@@ -48,27 +30,16 @@ export function BitcoinFeeCard({ feeType, amount }: FeeCardProps) {
...
         <FeeBadge type="normal" />
       </Box>
-
       <Box mx="-5">
-        <Cell.Root pressable={true} onPress={() => {}}>
+        <Cell.Root pressable onPress={onPress}>
           <Cell.Icon>
             <Avatar bg="ink.background-secondary" p="2">
-              {feeIcon}
+              {getFeeData(feeType).icon}
```
pete-watters, line 37: Can you just call `getFeeData(feeType)` once instead? 

```
const { icon, title, time } = getFeeData(feeType);
```

apps/mobile/src/features/psbt-signer/psbt-signer.tsx
```diff
@@ -1,41 +1,114 @@
...
   onEdit(): void;
   onSuccess(): void;
 }
 
-export function PsbtSigner({ psbtHex, onEdit, onSuccess }: PsbtSignerProps) {
-  const network = useCurrentNetworkState();
+function getFeeType({
+  psbtFee,
+  fees,
+}: {
```
pete-watters, line 61: Can you move this type out to an `interface`?

apps/mobile/src/features/psbt-signer/psbt-signer.tsx
```diff
@@ -71,71 +198,88 @@ export function PsbtSigner({ psbtHex, onEdit, onSuccess }: PsbtSignerProps) {
...
-          titleOpened={t({
-            id: 'approver.advanced.hide',
-            message: 'Hide advanced options',
-          })}
-        >
-          <Approver.Section noTopPadding>
-            <InputsAndOutputsCard
-              inputs={psbtDetails.psbtInputs}
-              outputs={psbtDetails.psbtOutputs}
+    <>
```
pete-watters, line 201: Maybe you can move this to a `.layout` file?

apps/mobile/src/features/psbt-signer/psbt-signer.tsx
```diff
@@ -71,71 +198,88 @@ export function PsbtSigner({ psbtHex, onEdit, onSuccess }: PsbtSignerProps) {
...
+      <Approver requester="https://leather.io">
```
pete-watters, line 202: I know it was already like this but should `requester` be hardcoded?

apps/mobile/src/features/approver/utils.tsx
```diff
@@ -0,0 +1,72 @@
+import { t } from '@lingui/macro';
```
pete-watters, line 1: Maybe you can just call this file `get-fee-data.ts`?

apps/mobile/src/features/approver/components/fees-sheet.tsx
```diff
@@ -0,0 +1,91 @@
...
+  function getUsd(fee: number) {
+    if (!fee) return createMoney(0, 'USD');
+    const btcBalance = createMoney(fee, 'BTC');
+    const usdBalance = baseCurrencyAmountInQuoteWithFallback(btcBalance, btcMarketData);
+    return usdBalance;
+  }
+
+  function getFee(feeType: FeeTypes) {
+    const feeRate = match<FeeTypes>()(feeType, {
+      [FeeTypes.Low]: fees?.hourFee ?? BigNumber(0),
```
pete-watters, line 50: Maybe we can just call `BigNumber(0)` once?

`const bigZero = BigNumber(0);`

pete-watters:
Nice work @edgarkhanzadian 👍 I left a few comments, mostly minor things. 

I tried it out and its working quite well but I have a couple of questions:
1. `custom` fee is disabled but it doesn't look any different than the others. 
    - Maybe we should just hide it instead as it seems broken?
    - we probably need to add a `disabled` style for `Pressable`
2. maybe a seperate bug but on the `Send` form 
    - the `Review` button doesn't appear active if you enter recipient and then amount
    - you can still click `Review` but the button is greyed out
    
Screenshot 2024-12-20 at 12 00 45

3. if you try and send funds from an empty account it doesn't show any errors
4. the drawers for recipient + fee stay open even after you select / populate so you need to tap again to close them

fbwoolf (approved):

PR #775, opened 2024-12-20
fix: ensure consistent drag down behaviour, ref LEA-1855

pete-watters:
This PR fixes [LEA-1855](https://linear.app/leather-io/issue/LEA-1855/dragging-down-list-behavior-is-different-on-select-account-and-select). 

The drag behaviour seems different but it is in fact because the `select-asset` sheets have a `ScrollView` for their content.

In the original bug, it seems like there is different behaviour but this is due to the `ScrollView`:
https://github.com/user-attachments/assets/113372c8-703b-4445-8c4a-d403d84fe338

I added `ScrollView` to `select-account` as otherwise if you have more than 10 accounts you can't reach them

### DIFF
--- a/apps/mobile/src/features/receive/receive-sheets/select-account.tsx
+++ b/apps/mobile/src/features/receive/receive-sheets/select-account.tsx
@@ -1,3 +1,5 @@
+import { ScrollView } from 'react-native-gesture-handler';
+
 import { FullHeightSheetHeader } from '@/components/full-height-sheet/full-height-sheet-header';
 import { FullHeightSheetLayout } from '@/components/full-height-sheet/full-height-sheet.layout';
 import { AccountList } from '@/features/accounts/account-list/account-list';
@@ -30,7 +32,9 @@ export function SelectAccount() {
         />
       }
     >
-      <AccountList accounts={accounts.list} onPress={onSelectAccount} showWalletInfo />
+      <ScrollView>
+        <AccountList accounts={accounts.list} onPress={onSelectAccount} showWalletInfo />
+      </ScrollView>
     </FullHeightSheetLayout>
   );
 }
--- a/apps/mobile/src/features/receive/receive-sheets/select-asset.tsx
+++ b/apps/mobile/src/features/receive/receive-sheets/select-asset.tsx
@@ -93,7 +93,7 @@ export function SelectAsset() {
           />
         }
       >
-        <ScrollView>
+        <ScrollView scrollEnabled={assets.length > 9}>
           {assets.map(asset => (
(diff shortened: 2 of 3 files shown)


apps/mobile/src/features/receive/receive-sheets/select-asset.tsx
```diff
@@ -93,7 +93,7 @@ export function SelectAsset() {
           />
         }
       >
-        <ScrollView>
+        <ScrollView scrollEnabled={assets.length > 9}>
```
pete-watters, line 96: I added this check so it only scrolls if the screen is full. It's easy to check on this screen but on `SelectAssetSheet` it's more difficult as the `ScrollView` wraps loaders that load account assets. 

```
  <ScrollView>
        <BitcoinPayerLoader fingerprint={account.fingerprint} accountIndex={account.accountIndex}>
          {({ nativeSegwitPayer }) => (
            <BitcoinBalanceByAccount
    ...
```

Maybe there's a better way of using `ScrollView` to only enable scroll if the screen is full? I was looking at [alwaysBounceVertical](https://reactnative.dev/docs/scrollview#alwaysbouncevertical-ios) but it seems to be iOS only

fbwoolf: @pete-watters is this helpful? I set this up to detect the height and enable/disable based on that...
https://github.com/leather-io/mono/blob/3068cea3905eae77aafbe890d802aee60d3ef919/apps/mobile/src/components/headers/animated-header/animated-header-screen.layout.tsx#L43

apps/mobile/src/features/receive/receive-sheets/select-account.tsx
```diff
@@ -30,7 +32,9 @@ export function SelectAccount() {
         />
       }
     >
-      <AccountList accounts={accounts.list} onPress={onSelectAccount} showWalletInfo />
+      <ScrollView>
+        <AccountList accounts={accounts.list} onPress={onSelectAccount} showWalletInfo />
```
camerow, line 36: Just a thought, but should the `ScrollView` perhaps be in the render of `AccountList`? If i see `SomeList` I kind of assume it's a scrollable element in the view.

apps/mobile/src/features/receive/receive-sheets/select-asset.tsx
pete-watters: Thanks, I'll check it out

pete-watters:
> This one is fixed already here: #782

Thanks!

PR #778, opened 2024-12-20
ci: setup fastlane to deploy to firebase

pete-watters:


.github/actions/eas-deploy-android/action.yml
```diff
@@ -87,4 +88,10 @@ runs:
         cd apps/mobile
         # Release version of the app, we should wait for it to see the result
         eas build --platform android --profile=production --non-interactive --auto-submit
+        fastlane run distribute_android
       shell: bash
+
+    - name: Firebase App Distribution
+      run: |
+        echo "${{ secrets.FIREBASE_TOKEN }}" > $HOME/.firebase_token
```
pete-watters, line 96: @camerow I generated a token but I don't have permissions to add it as a secret in this repo

apps/mobile/ios/fastlane/Fastfile
```diff
@@ -0,0 +1,27 @@
+platform :ios do
+  desc "Build and distribute iOS app"
+  lane :distribute_ios do
+    build_app(
+      scheme: "YourAppScheme" # Specify your app's scheme
```
pete-watters, line 5: I'm not sure what to put for `scheme` ?

camerow: Screenshot 2024-12-20 at 7 56 01 AM

it should be this:

```
<string>myapp</string>
<string>io.leather.mobilewallet</string>
```

we should change `myapp` to `leather` and you can specify the full scheme as `leather://`. That will allow links to `leather://` to resolve to the application on iOS when followed by a user.

.github/actions/eas-deploy-android/action.yml
pete-watters: Did you add this secret @camerow ?

apps/mobile/ios/leatherwalletmobile/AppDelegate.mm
```diff
@@ -32,7 +32,11 @@ - (NSURL *)bundleURL
 
 // Linking API
 - (BOOL)application:(UIApplication *)application openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {
-  return [super application:application openURL:url options:options] || [RCTLinkingManager application:application openURL:url options:options];
+  if ([url.scheme isEqualToString:@"leather"]) {
```
pete-watters, line 35: I'm not sure if I need to do this to support `leather://` links?

pete-watters:
Closing this off as I will setup firebase deploy from EAS instead

PR #792, opened 2025-01-07
Standardize Badge component, closes LEA-1830

tigranpetrossian:
Several small updates to the Tag/Badge component to standardize the naming and the signature between web and mobile versions.

The web version lived under `Tag`, while the mobile was called `Badge`. Going with `Badge` since "tag" carries a slightly different meaning. Will request for the component to be renamed in Figma as well.

### Web updates
* Correct height
* Renamed transparent -> outline

### Mobile
* Correct height
* Added outlined variant
* Made default variant the default
* Removed Pressable. Interactive badges are rare, it's better to add the interactivity at the usage site as necessary.
* Added a story

Pending items to be addressed in a future PR:
* Per-variant Icon colors aren't applied in the web version
* Mobile version is still missing icon support
These depend on the WIP icon system revamp.

### DIFF
--- a/apps/mobile/src/components/headers/home-header.tsx
+++ b/apps/mobile/src/components/headers/home-header.tsx
@@ -12,7 +12,7 @@ export function HomeHeader() {
       leftElement={
         <Box flexDirection="row" alignItems="center">
           <HeaderLeatherLogo />
-          <NetworkBadge px="0" ml="-1" />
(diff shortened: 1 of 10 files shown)


pete-watters (approved):
Nice work 👍

kyranjamie (approved):
Great work @tigranpetrossian

PR #793, opened 2025-01-07
fix: only show hidden accounts if accounts hidden, ref LEA-1913

pete-watters:
This PR makes a change to only show the hidden accounts setting if there are hidden accounts 

https://github.com/user-attachments/assets/cd68830b-93bb-4092-877d-1edc578ff972

### DIFF
--- a/apps/mobile/maestro/flows/hide-account.yaml
+++ b/apps/mobile/maestro/flows/hide-account.yaml
@@ -14,7 +14,7 @@ appId: io.leather.mobilewallet
     id: 'walletListAccountCard'
 - tapOn:
     text: 'Hide account'
-- assertVisible: 'Unhide account'
+- assertVisible: 'Un-hide account'
 - tapOn:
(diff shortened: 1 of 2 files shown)


apps/mobile/maestro/flows/hide-account.yaml
```diff
@@ -14,7 +14,7 @@ appId: io.leather.mobilewallet
     id: 'walletListAccountCard'
 - tapOn:
     text: 'Hide account'
-- assertVisible: 'Unhide account'
+- assertVisible: 'Un-hide account'
```
pete-watters, line 17: I updated this test text as it was failing due to copy change from `Un-hide account` to `Unhide account`. 

This is a bit of a manual step but I think it's worth it as:
- the tests are easier to read
- it exposed that we aren't running these tests in CI as this should have been failing and blocking merges

tigranpetrossian (approved):

PR #794, opened 2025-01-07
feat: stacks balances service

alexp3y:
This PR adds several new services supporting Stacks-related balance retrieval (STX & SIP10) and continues to migrate functionality from `@leather.io/query` to the `@leather.io/services` package. 

- `StxBalanceService`
- `Sip10BalanceService`
- `Sip10TokensService`

This update moves us slightly away from the idea of extremely high-level and opinionated services encapsulating complex use-cases (e.g. a single "Balance Service") in favor of more granular services that can be composed more flexibly within client applications.

For example, the total balance hook in `mobile` consists of calls to `StxBalanceService`, `BtcBalanceService`, and `Sip10BalanceService` and calculates the total balance. This allows for better UX as the hook can handle 3 independent loading states, versus being blocked entirely by the slowest loading sub-service. 

Additional changes here include infrastructure components related to the HIro Stacks API and associated rate limiting / request queueing, as well as various utilities supporting the migrated data fetching functions.


packages/services/src/assets/sip10-tokens.utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function isTransferableSip10Token(token: Partial<HiroFtMetadataResponse>) {
+  return !isUndefined(token.decimals) && !isUndefined(token.name) && !isUndefined(token.symbol);
+}
+
+export function getAssetNameFromIdentifier(assetIdentifier: string) {
+  return !assetIdentifier.includes('::') ? assetIdentifier : assetIdentifier.split('::')[1];
+}
+
+export function getContractPrincipalFromAssetIdentifier(assetIdentifier: string) {
```
pete-watters, line 19: `getContractPrincipalFromAssetIdentifier` and `getAddressFromAssetIdentifier` seem to have the same logic?

packages/services/src/infrastructure/api/alex-sdk/alex-sdk.client.ts
```diff
@@ -1,21 +1,44 @@
...
 
 const alex = new AlexSDK();
 
 export function createAlexSdkClient(cacheService: HttpCacheService): AlexSdkClient {
+  async function getApiTokenPrices(signal?: AbortSignal) {
+    return await cacheService.fetchWithCache(
+      ['alex-api-client-token-prices'],
+      async () => {
+        const res = await axios.get<AlexApiTokenPricesResponse>(
+          `https://api.alexgo.io/v2/public/token-prices`,
```
pete-watters, line 33: Maybe this URL can be a constant?

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -1,38 +1,145 @@
...
+            throwOnTimeout: true,
+          }
+        );
+        return res.data;
+      },
+      { ttl: HttpCacheTimeMs.twoMinutes }
+    );
+  }
+
+  async function getAddressConfirmedTransactions(address: string, signal?: AbortSignal) {
```
pete-watters, line 66: Great work on this! This gets a lot of the data we will need for `Activity`. I wonder could we also add BTC transaction data to`infrastructure`?

packages/services/src/infrastructure/api/alex-sdk/alex-sdk.client.ts
pete-watters: I see in most other places we have hardcoded URLs so it's probably fine to leave it as is if they are just used once anyway

apps/mobile/src/queries/balance/sip10-balance.query.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { isLoading, data: balance } = useSip10AggregateBalanceQuery(addresses);
+  return !isLoading && balance ? balance.usd.availableBalance : createMoney(0, 'USD');
+}
+
+export function useSip10AggregateBalanceQuery(addresses: string[]) {
+  return useQuery({
+    queryKey: ['sip10-balances-service-get-sip10-aggregate-balance', addresses],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      getSip10BalancesService().getSip10AggregateBalance(addresses, signal),
+    refetchOnReconnect: false,
```
pete-watters, line 16: Maybe we can have a generic query configuration to avoid repeating these props? Then we can over-ride on demand if we need to

apps/mobile/src/queries/balance/sip10-balance.query.ts
```diff
@@ -0,0 +1,37 @@
+import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
+
+import { getSip10BalancesService } from '@leather.io/services';
+import { createMoney } from '@leather.io/utils';
+
+export function useSip10AggregateAvailableBalance(addresses: string[]) {
+  const { isLoading, data: balance } = useSip10AggregateBalanceQuery(addresses);
+  return !isLoading && balance ? balance.usd.availableBalance : createMoney(0, 'USD');
```
pete-watters, line 8: Do you have any ideas what we can do to avoid defaulting to `createMoney(0, 'USD')`? 

Maybe if it has timed out and there is no balance we should set it to `createMoney(-1, 'USD')` / `createMoney(0, 'XXX')` so we can correctly pickup an error? (this isn't ideal but an idea) 

If we can't get their balance we shouldn't show a `0` balance and we [need to show an error state / fallback ](https://linear.app/leather-io/issue/LEA-505/loading-state-for-balances-on-home-screen)

packages/services/src/balances/sip10-balances.service.ts
```diff
@@ -0,0 +1,108 @@
...
+      const totalBalance = createMoney(amount, tokenInfo.symbol, tokenInfo.decimals);
+      const marketData = await marketDataService.getSip10MarketData(tokenInfo);
+      return {
+        info: tokenInfo,
+        sip10: createBaseCryptoAssetBalance(totalBalance),
+        usd: createBaseCryptoAssetBalance(baseCurrencyAmountInQuote(totalBalance, marketData)),
+      };
+    } catch {
+      return;
+    }
```
kyranjamie, line 101: What's the idea behind this baked in error handling? Shouldn't this be handled outside the service? What if I want to show an error msg in the UI with the error details passed from API?

packages/services/src/assets/sip10-tokens.service.ts
```diff
@@ -0,0 +1,27 @@
...
+  createSip10CryptoAssetInfo,
+  getContractPrincipalFromAssetIdentifier,
+} from './sip10-tokens.utils';
+
+export interface Sip10TokensService {
+  getInfo(assetIdentifier: string, signal?: AbortSignal): Promise<Sip10CryptoAssetInfo>;
+}
+
+export function createSip10TokensService(stacksApiClient: HiroStacksApiClient): Sip10TokensService {
+  /**
```
pete-watters, line 14: I'm not sure about the JSDocs comments. In some cases they are useful, and we could do with some more documentation. In other cases we could infer the behaviour from a better named function with a return type

packages/services/src/balances/stx-balances.utils.ts
```diff
@@ -0,0 +1,42 @@
...
+    'STX'
+  );
+}
+
+export function calculateOutboundStxBalance(
+  address: string,
+  pendingTransactions: StacksTx[]
+): Money {
+  // includes transfers and fees when address is the sender
+  const balanceMicroStx = pendingTransactions.reduce((acc, tx) => {
```
kyranjamie, line 26: Would it be cleaner to get the amount first, then use `sumNumbers` helper?

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -1,38 +1,145 @@
-import type { Transaction } from '@stacks/stacks-blockchain-api-types';
+import { FtMetadataResponse } from '@hirosystems/token-metadata-api-client';
+import {
+  AddressBalanceResponse,
+  AddressTransactionsListResponse,
+  MempoolTransactionListResponse,
```
kyranjamie, line 5: Alternative to redefining below


```suggestion
  AddressBalanceResponse as HiroAddressBalanceResponse,
  AddressTransactionsListResponse as HiroAddressBalanceResponse,
  MempoolTransactionListResponse as MempoolTransactionListResponse,
```

kyranjamie: or even `import * as Hiro ....`?

packages/services/src/infrastructure/cache/http-cache.utils.ts
```diff
@@ -3,11 +3,12 @@ export const HttpCacheTimeMs = {
...
   thirtySeconds: 30 * 1000,
-  oneMinute: 1 * 60 * 1000,
+  oneMinute: 60 * 1000,
   twoMinutes: 2 * 60 * 1000,
   fiveMinutes: 5 * 60 * 1000,
   tenMinutes: 10 * 60 * 1000,
   thirtyMinutes: 30 * 60 * 1000,
   oneHour: 60 * 60 * 1000,
   oneDay: 24 * 60 * 60 * 1000,
+  oneYear: 365 * 24 * 60 * 60 * 1000,
```
kyranjamie, line 13: this number is too big to work with `setTimeout` (also used internally by RQ), and including here might lead developers to use and and re-introduce caching bugs like fixed here https://github.com/leather-io/mono/pull/582

do we absolutely need it?

packages/services/src/market-data/market-data.service.ts
```diff
@@ -59,10 +63,20 @@ export function createMarketDataService(
...
+   */
   async function getBtcMarketData(signal?: AbortSignal) {
     return await getNativeAssetMarketData('BTC', signal);
   }
 
+  /**
+   * Get current STX Market Data.
+   *
+   * @returns {Promise<MarketData>}
+   */
```
kyranjamie, line 79: Curious where everyone stands on tsdoc comments that annotate type infomation. Isn't it duplicating info to have this, what if type changes and dev forgets to update `@returns`?

packages/services/src/transactions/stacks-transactions.utils.ts
```diff
@@ -0,0 +1,17 @@
...
+  return mempoolTransactions.filter(tx => !confirmedTxIds.has(tx.tx_id));
+}
+
+export function removeStaleTransactions(
+  mempoolTransactions: StacksTx[],
+  confirmedTransactions: StacksTx[]
+): StacksTx[] {
+  const confirmedNonces = new Set(confirmedTransactions.map(tx => tx.nonce));
+  return mempoolTransactions.filter(tx => !confirmedNonces.has(tx.nonce));
+}
```
kyranjamie, line 17: When touching filtering logic in extension recently, I refactored some filtering code that was similar to this.

This API decision of accepting array directly leads to deep nesting of calls, which gets kinda ugly esp. when line wrapped e.g. 

```ts
const txs = removeStaleTransactions(
  mempoolTxs,
  removeConfirmedTransactions(mempoolTxs, confirmedTxs)
);
```

Returning higher order filter functions helps remedy this, and make it more composable


```ts
const txs = mempoolTransactions
  .filter(filterOutConfirmedTxs(confirmedTransactions))
  .filter(filterOutStaleTransactions(confirmedTransactions))
```

kyranjamie (commented):
Great work @alexp3y — can't wait to see this in action in extension

apps/mobile/src/queries/balance/sip10-balance.query.ts
kyranjamie: +1 what to do when handling loading/pending state is a concern of the UI. Defaulting to zero will lead to bugs and unexpected behaviours

packages/services/src/assets/sip10-tokens.service.ts
alexp3y: My intention with the JSDocs was to highlight exported service functions specifically, which I think can help improve the readability of service files (that can also contain non-exported functions) and highlight the public API of the services package. Agree that it shouldn't substitute for well-named functions / parameters in general (this function probably in particular could be better named), but hoping it can provide additional value, context for external consumers.

packages/services/src/balances/sip10-balances.service.ts
alexp3y: The idea here was to prevent the entire call from failing because of a single SIP10 failure, of which there can be many dozens under a single `getSip10Aggregate/AddressBalance`. Although I agree, this is error swallowing and probably not a great solution. I can remove.

packages/services/src/infrastructure/cache/http-cache.utils.ts
alexp3y: The `Infinity` value from #582 seems to be the right solution here. Just looking for something to support long-lived cache data e.g token metadata.

packages/services/src/market-data/market-data.service.ts
alexp3y: Makes sense to me.

alexp3y:
@pete-watters @kyranjamie Latest push addresses comments and adds unit tests for the utilities & services used here. I'll try to keep coverage up and in-step with development w/in this package going forward.

PR #795, opened 2025-01-07
chore: enforce func-style in eslint

kyranjamie:
Our convention is to [favour function declarations](https://www.notion.so/trustmachines/Function-declarations-22f7055df3124d868a242adb04f0f60d?pvs=4).

This PR enforces this with eslint's [`func-style`](https://eslint.org/docs/latest/rules/func-style) rule.


tigranpetrossian (approved):

tigranpetrossian:
Thanks! This was one of the reasons for the eslint upgrade, yet I compeltely forgot this rule.

PR #796, opened 2025-01-07
refactor: upgrade stacks.js v7

kyranjamie:
Here I'm trying to upgrade Stacks.js to v7. This is necessary work we'll need to undertake eventually, but the most pressing reason is support for Clarity v3 https://github.com/leather-io/extension/issues/5983

There are many breaking changes, discussed in [the migration guide](https://github.com/hirosystems/stacks.js/blob/main/.github/MIGRATION.md#stacksjs-5xx--7xx). Currently I'm just working through type issues. There'll be lots more changes extension side.

I've manually QA'd sending a stacks transaction in app post these changes.

See LEA-2074


packages/stacks/src/transactions/post-condition.utils.spec.ts
```diff
@@ -52,8 +51,8 @@ describe('getPostCondition', () => {
 
   it('should return post condition directly if already a PostCondition object', () => {
     (isString as unknown as ReturnType<typeof vi.fn>).mockReturnValue(false);
-    const result = PostConditionModule.getPostCondition(mockStxPostCondition);
-    expect(isString).toHaveBeenCalledWith(mockStxPostCondition);
-    expect(result).toBe(mockStxPostCondition);
+    const result = PostConditionModule.getPostCondition(postConditionToWire(mockStxPostCondition));
+    // expect(isString).toHaveBeenCalledWith(mockStxPostCondition);
```
fbwoolf, line 55: Do you mean to leave this?

fbwoolf (approved):

packages/stacks/src/transactions/post-condition.utils.spec.ts
kyranjamie: I've removed, I think it's not necessary

kyranjamie:
I'm now trying to upgrade the extension code locally, so this can hold off being merged until I'm done with that

packages/stacks/package.json
```diff
@@ -25,10 +25,12 @@
...
     "@scure/bip32": "1.5.0",
-    "@stacks/encryption": "6.16.1",
-    "@stacks/network": "6.13.0",
+    "@stacks/common": "7.0.2",
+    "@stacks/encryption": "7.0.2",
+    "@stacks/network": "7.0.2",
     "@stacks/stacks-blockchain-api-types": "7.8.2",
-    "@stacks/transactions": "6.17.0",
+    "@stacks/transactions": "7.0.2",
+    "@stacks/transactions-v6": "npm:@stacks/transactions@6.17.0",
```
fbwoolf, line 33: @kyranjamie why are both of these installed here? Should just be `7.0.2`, correct?

kyranjamie: for a while I had both to deal with some transitional stuff, but yeah we should remove this

markmhendrickson:
@kyranjamie can you link to an issue we're tracking on the board?

fbwoolf:
> @kyranjamie can you link to an issue we're tracking on the board?

@markmhendrickson the issue is linked to this PR, we don't want this PR to close the issue: https://github.com/leather-io/extension/pull/6076

PR #799, opened 2025-01-09
CD: fastlane - firebase

pete-watters:


.github/actions/eas-deploy-android/action.yml
```diff
@@ -24,10 +27,11 @@ runs:
         eas-version: latest
         token: ${{ inputs.EXPO_TOKEN }}
 
-    - name: Install jq tool
+    - name: Install jq tool and fastlane
       run: |
         sudo apt-get update
         sudo apt-get install jq
+        sudo gem install fastlane
```
kyranjamie, line 34: Perhaps better to do as separate job? If one or the other fails, it's then easier to debug

pete-watters: Good idea, I'll split it out. Thanks!

PR #800, opened 2025-01-10
feat(rpc): add sip-30 schemas/types

kyranjamie:
Adding the new SIP-30 schemas to RPC


kyranjamie:
@fbwoolf it won't take me much longer to finish this. In mean time, maybe you could figure out how we'll implement the new stx message signing APIs? We already support the feature, but there are now specific APIs you can call for normal/structured  data

fbwoolf:
> @fbwoolf it won't take me much longer to finish this. In mean time, maybe you could figure out how we'll implement the new stx message signing APIs? We already support the feature, but there are now specific APIs you can call for normal/structured data

Ok, I can np, I had started the `stx_callContract` in extension issue and copied your code from here temporarily. But, if better to do the message signing, just let me know. I was mostly just becoming familiar with the changes on Friday.

fbwoolf (approved):

packages/rpc/src/methods/stacks/stx-contract-call.ts
```diff
@@ -0,0 +1,32 @@
+import { z } from 'zod';
+
+import { DefineRpcMethod, RpcRequest, RpcResponse } from '../../rpc/schemas';
+import { stacksTransactionDetailsSchema } from './_stacks-helpers';
+
+export const stxCallContractMethodName = 'stx_callContract';
```
fbwoolf, line 6: I think maybe this file should be rename to match the method? `stx-call-contract.ts`?

PR #802, opened 2025-01-13
fix: unhandled error in is nft asset condition, LEA-1935

alter-eggo:
This pr fixes `Cannot use 'in' operator to search for 'error' in` error

### DIFF
--- a/packages/query/src/stacks/token-metadata/token-metadata.utils.ts
+++ b/packages/query/src/stacks/token-metadata/token-metadata.utils.ts
@@ -13,10 +13,10 @@ function isAssetMetadataNotFoundResponse(
   return 'error' in resp;
 }
 
-export function isFtAsset(resp: FtAssetResponse): resp is FtMetadataResponse {
-  return !isAssetMetadataNotFoundResponse(resp);
+export function isFtAsset(resp: FtAssetResponse | null | undefined): resp is FtMetadataResponse {
+  return !!resp && !isAssetMetadataNotFoundResponse(resp);
 }
 
-export function isNftAsset(resp: NftAssetResponse): resp is NftMetadataResponse {
-  return !isAssetMetadataNotFoundResponse(resp);
+export function isNftAsset(resp: NftAssetResponse | null | undefined): resp is NftMetadataResponse {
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

packages/query/src/stacks/token-metadata/token-metadata.utils.ts
```diff
@@ -10,7 +10,7 @@ export type NftAssetResponse = NftMetadataResponse | NotFoundErrorResponse;
 function isAssetMetadataNotFoundResponse(
   resp: FtAssetResponse | NftAssetResponse
 ): resp is NotFoundErrorResponse {
-  return 'error' in resp;
+  return resp && 'error' in resp;
```
kyranjamie, line 13: Do the types say that `resp` can be undefined?

alter-eggo: from the code perspective there are no cases where we might provide `null` or `undefined` to this fn, look my [comment](https://github.com/leather-io/extension/issues/6053#issuecomment-2587097282)

I'm not sure tbh how user faced this issue, I haven't found problem in the code, that's why just made this change

kyranjamie: I feel like if it can be undefined, it should be typed as so.

Otherwise, developers later looking at this code will ask the question: why is this unnecessary check here (according to the types)? and either be confused, or remove it.

Could we do this check in the query hook itself?

alter-eggo: we already do the check, that's why I'm confused here
Screenshot 2025-01-13 at 21 12 15

kyranjamie: I'm also unclear how this could happen. Could it be that `resp.data` is truthy but not an object?

Either way, I don't think we should merge this PR as it. It sets the precedent that it's okay for us to not trust our type system, and that even though the types say it is value x, actually it might not be.

kyranjamie (approved):
I see you've added the checks to the types are correct. Still not an ideal solution IMO as we haven't gotten to the bottom of understand why this error would've been thrown in the first place

PR #803, opened 2025-01-13
fix: android secure storage

edgarkhanzadian:
Secure mode now works also for PIN, password, or pattern.

Things that need to get merged into this PR:

- [x]  Dark mode errors out on wallet creation lottie animation for some reason #812 
- [x] Blur view is blocking the whole screen instead of just the mnemonic section for some reason, needs investigation #815


apps/mobile/src/store/key-store.ts
```diff
@@ -90,11 +92,45 @@ export function useKeyStore() {
       );
     },
 
-    async deriveNextAccountKeychainsFrom(fingerprint: string) {
-      const { mnemonic, passphrase } = await mnemonicStore(fingerprint).getMnemonic();
-
-      if (!mnemonic) throw new Error('No mnemonic found for fingerprint ' + fingerprint);
+    async deriveNextAccountKeychainsFrom(
+      props:
+        | {
```
pete-watters, line 97: Can you clean up this typing to use an interface?

apps/mobile/src/store/key-store.ts
```diff
@@ -90,11 +92,45 @@ export function useKeyStore() {
       );
     },
 
-    async deriveNextAccountKeychainsFrom(fingerprint: string) {
-      const { mnemonic, passphrase } = await mnemonicStore(fingerprint).getMnemonic();
-
-      if (!mnemonic) throw new Error('No mnemonic found for fingerprint ' + fingerprint);
+    async deriveNextAccountKeychainsFrom(
```
pete-watters, line 95: How about having two functions. 
- `deriveNextAccountKeychainsFromFingerprint` 
- `deriveNextAccountKeychains` - accepts `mnemonic` + `passphrase` and calls `deriveNextAccountKeychainsFromFingerprint` ?

I think that will make it cleaner and easier to follow

kyranjamie: +1

camerow:
@edgarkhanzadian so this isn't quite testable in the current state of the app?

fbwoolf (approved):

camerow:
@edgarkhanzadian tested and this is working for me now! Nice one.

🚀

kyranjamie (approved):
Great work @edgarkhanzadian

PR #805, opened 2025-01-13
fix(rpc): stx_getAddress, exports

kyranjamie:
Some fixes building on my previous PR


fbwoolf (approved):

packages/rpc/src/index.ts
```diff
@@ -25,6 +26,12 @@ export * from './methods/bitcoin/send-transfer';
 export * from './methods/bitcoin/sign-message';
 export * from './methods/stacks/stx-sign-message';
 export * from './methods/stacks/stx-sign-transaction';
+export * from './methods/stacks/stx-contract-call';
```
fbwoolf, line 29: Can you rename this file here? It should match the method name being called? `stx-call-contract.ts`

fbwoolf:
@kyranjamie I think the params to call a contract are wrong? Maybe they were copied but not changed?

https://github.com/janniks/sips/blob/main/sips/sip-030/sip-030-wallet-interface.md#method-stx_callcontract

PR #807, opened 2025-01-13
fix(rpc): incorrect contract call schema

kyranjamie:
Copy paste fail adding some new types

### DIFF
--- a/packages/rpc/src/methods/stacks/_stacks-helpers.ts
+++ b/packages/rpc/src/methods/stacks/_stacks-helpers.ts
@@ -5,93 +5,22 @@ export const stacksTransactionDetailsSchema = z.object({
   transaction: z.string(),
 });
 
-// Clarity values
-export const cvIntSchema = z.object({
-  type: z.literal('int'),
-  value: z.string(), // `bigint` compatible
+export const baseStacksTransactionConfigSchema = z.object({
+  address: z.string().optional(),
+  network: z
+    .union([
+      z.literal('mainnet'),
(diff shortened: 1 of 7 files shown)


fbwoolf (approved):

fbwoolf:
@kyranjamie hitting this, maybe the schemas for Clarity values need to use a schema for ClarityType?

Screenshot 2025-01-13 at 2 42 35 PM

fbwoolf:
Came across `isMultisig` param in the current `stx_signTransaction` method using the auth spendingCondition on the the passed tx, is that ever used for a contract call? 🤔

kyranjamie:
> Came across isMultisig param in the current stx_signTransaction method using the auth spendingCondition on the the passed tx, is that ever used for a contract call? 🤔 

this was added for asignia I think, will have to check. Fixing type errs now

packages/rpc/src/methods/stacks/_stacks-helpers.ts
```diff
@@ -5,93 +5,22 @@ export const stacksTransactionDetailsSchema = z.object({
   transaction: z.string(),
 });
 
-// Clarity values
-export const cvIntSchema = z.object({
-  type: z.literal('int'),
-  value: z.string(), // `bigint` compatible
+export const baseStacksTransactionConfigSchema = z.object({
```
fbwoolf, line 8: @kyranjamie I think this should be exported from the package so available. I am having to copy it over rn. I can do, but just flagging.

PR #810, opened 2025-01-14
feat: utxos service

alexp3y:
This PR brings several additions to the `@leather.io/services` package and updates to service implementation / usage in `mobile`.

**New features**:
- `UtxosService` -> provides categorized UTXO summaries per account or descriptor (UTXO categories map to BTC balance categories)
- `RunesBalancesService` -> account and aggregate balances
- Bitcoin Account Xpub Requests -> allows for matching of returned aggregate sub-balance data to individual accounts during aggregate calls
- Unprotected UTXO support during UTXO / Balance service calls
- Asset Services (sip10 & runes)
- Loadable balances -> mobile app uses predefined constants as indicators of balance loading / error states to prevent zero balance UX issue
  - `-1` = loading
  - `-2` = error
  - `0` = no data (or actually zero balance)
- `SendFormBtcLoader` & `BasePsbtSigner` -> now use available UTXOs via service (instead of total UTXOs) to match available balance
- Attempts to refine naming convention:
  - "Total" -> multi-wallet, all wallets and accounts
  - "Wallet" -> single-wallet (fingerprint), all wallet accounts
  - "Account" -> single account (fingerprint + account index), all payment types and sub-addresses

**TODO / Up next:**
- More unit tests
- Add `RunesBalancesService` to mobile total balance -> Waiting on new BIS runes xpub endpoint implementation
- `BitcoinTransactionsService` Implementation -> Waiting on new Leather API transactions endpoint
  - Will enable BTC Outbound balances
  - Will enables Activity service(s) alongside `StacksTransactionsService`
- `BitcoinFeesService` Implementation
  - Enables accurate BTC Uneconomical balances
- Implement UTXO protection feature in `mobile` and use in service calls


apps/mobile/src/hooks/use-bitcoin-account-requests.ts
```diff
@@ -0,0 +1,123 @@
...
+      return `wpkh(${accountKeychain.xpub})`;
+    default:
+      return undefined;
+  }
+}
+
+function findSegwitDescriptors(descriptors: string[]) {
+  return {
+    trXpubDescriptor: descriptors.find(desc => desc.startsWith('tr(')) ?? '',
+    nsXpubDescriptor: descriptors.find(desc => desc.startsWith('wpkh(')) ?? '',
```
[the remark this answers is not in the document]
kyranjamie: also xpub is the `xpub{hex}` part of the descriptor within the parentheses, so "descriptor" alone is sufficient I think

apps/mobile/src/hooks/use-bitcoin-account-requests.ts
```diff
@@ -0,0 +1,123 @@
...
+            account: accountXpub,
+            unprotectedUtxos: [] as UtxoId[], // TODO: incorporate unprotected UTXOs
+          };
+        });
+      }),
+    [wallets.list, keychains.list]
+  );
+}
+
+export function useWalletBitcoinRequests(fingerprint: string) {
```
kyranjamie, line 73: what do you mean by bitcoin requests here?

packages/services/src/shared/bitcoin.types.ts
```diff
@@ -0,0 +1,13 @@
+import { UtxoId } from '@leather.io/models';
+
+export interface BitcoinAccountXpub {
+  fingerprint: string;
+  accountIndex: number;
+  trXpubDescriptor: string;
+  nsXpubDescriptor: string;
```
kyranjamie, line 7: why not represent this as an array of `descriptors`? They have identifying tr/wpkh strings in them.

packages/services/src/infrastructure/api/leather/leather-api.client.ts
```diff
@@ -1,43 +1,93 @@
...
+
+const leatherApiTxVinLightSchema = z.object({
+  value: z.string(),
+  n: z.number(),
+  addresses: z.array(z.string()),
+  isAddress: z.boolean(),
+  isOwn: z.boolean(),
+});
+
+const leatherApiTxVoutLightSchema = z.object({
```
kyranjamie, line 26: what is a light schema?

kyranjamie (commented):
Looking good @alexp3y 👏🏼 

Left some comments, and would love to jam on some naming. Personally, I'm really thrown by what you mean by `BitcoinAccountRequest`. Other than the fact that under the hood, some HTTP requests are being made, I'm not sure what it means. Request is a very widely used term so if possible another name might be a better option.

packages/services/src/infrastructure/api/leather/leather-api.client.ts
alexp3y: The Blockbook API underlying this request supports a "light" option that returns transaction data with fewer fields than a full Transaction object. This bitcoin tx stuff is WIP.

fbwoolf (approved):
Lots of great work here @alexp3y 🚀

packages/services/src/shared/bitcoin.types.ts
alexp3y: Although more verbose, I favored individual fields because it both centralizes the responsibility for parsing descriptors at the point of object construction (versus script types being parsed everywhere the object is used), and because ultimately the object represents a Bitcoin account as described by its xpubs, which in our system is represented by exactly 1 taproot account xpub descriptor and 1 native segwit account descriptor and nothing more. IMO the purpose and use of an array of descriptors could be misleading.

alexp3y: Although more verbose, I favored individual fields because it both centralizes the responsibility for parsing descriptors at the point of object construction (versus script types being parsed everywhere the object is used), and because ultimately the object represents a Bitcoin account as described by its xpubs, which in our system is represented by exactly 1 taproot account xpub descriptor and 1 native segwit account descriptor and nothing more. IMO the purpose and use of an array of descriptors could be misleading.

apps/mobile/src/queries/balance/btc-balance.query.ts
```diff
@@ -0,0 +1,51 @@
...
+export function useBtcAccountBalanceQuery(request: BitcoinAccountRequest) {
+  return useQuery({
+    queryKey: ['btc-balance-service-get-btc-account-balance', request],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      getBtcBalancesService().getBtcAccountBalance(request, signal),
+    refetchOnReconnect: false,
+    refetchOnWindowFocus: false,
+    refetchOnMount: true,
+    retryOnMount: false,
+    staleTime: 1 * 1000,
```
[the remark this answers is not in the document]
alexp3y: TBH I have not looked much at RQ / cache time params for this or other requests. We should prioritize a systematic review of all of our cache times at both levels (RQ + HTTP cache).

apps/mobile/src/shared/loadable-balance.ts
```diff
@@ -0,0 +1,35 @@
...
+export const zeroMoneyBtc = createMoney(0, 'BTC');
+export const zeroMoneyStx = createMoney(0, 'STX');
+
+export const loadingMoneyUsd = createMoney(-1, 'USD');
+export const loadingMoneyBtc = createMoney(-1, 'BTC');
+export const loadingMoneyStx = createMoney(-1, 'STX');
+
+export const errorMoneyUsd = createMoney(-2, 'USD');
+export const errorMoneyBtc = createMoney(-2, 'BTC');
+export const errorMoneyStx = createMoney(-2, 'STX');
```
kyranjamie, line 14: What are the benefits of this number coded approach? It's easy to consume because it's all `Money` (baby), but seems like it's asking to get `-1 BTC` in the UI.

Not a huge fan, but open to being convinced. Isn't this where discriminating union types are helpful, where the resp is like

```ts
interface BitcoinBalanceLoading {
  status: 'loading';
}
interface BitcoinBalanceError {
  status: 'error';
  error: Error;
}
interface BitcoinBalanceSuccess {
  status: 'success';
  amount: Money;
}
type BitcoinBalance = BitcoinBalanceLoading | BitcoinBalanceError | BitcoinBalanceSuccess;
```

Components can then use early returns to handle the error/loading states, and have the inferred success type after that.

packages/services/src/shared/bitcoin.types.ts
kyranjamie: Sure, happy to keep if you think it's better, but please use full `nativeSegwit` naming

apps/mobile/src/hooks/use-bitcoin-account-requests.ts
alexp3y: The hooks here should be named "...BitcoinAccountRequests" - They refer to the `BitcoinAccountRequest` type that was mentioned above. 

This type is a DTO that communicates a `BitcoinAccountXpub` and related context to service calls.

The "Request / Response" suffix is a frequent DTO naming convention, although I agree its not the most intuitive here. Open to any other suggestions... Perhaps even just extending it to be more specific, like `BitcoinAccountXpubRequest`?

apps/mobile/src/hooks/use-bitcoin-account-requests.ts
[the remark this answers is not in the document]
alexp3y: Yeah, I went back and forth on this several times because descriptors can _techincally_  contain things other than xpubs. You can tell my naming is a bit split between "descriptor" and "xpub"... Perhaps we just agree in our system that in general "descriptor" refers to an xpub descriptor?

apps/mobile/src/queries/balance/btc-balance.query.ts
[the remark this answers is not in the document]
pete-watters: Maybe we can have a shared `query` config which we then can override if needed

apps/mobile/src/queries/balance/total-balance.query.ts
```diff
@@ -1,63 +1,126 @@
...
+      isSip10Error,
+      sip10Balance?.usd.availableBalance
+    ),
+    totalBalance: getLoadableFiatBalance(isTotalLoading, isTotalError, totalBalance),
   };
 }
 
 export function useAccountTotalBalance(accountId: AccountId): TotalBalance {
   const { fingerprint, accountIndex } = accountId;
+
```
pete-watters, line 72: Maybe it's time to split this into a separate file now?

apps/mobile/src/shared/loadable-balance.ts
pete-watters: It's all Money (baby) 😂  

This looks good to me as a data contract between service and UI and clear definition of the states.  Keeping it as `Money` and will help us properly show `loading` and `undefined` in an easy-ish way without a lot of rework. 

I wonder will this scale for other individual SIP-10 balances? Maybe we can use `STX` for those also?

In the UI we can use `-1` and `-2` as clear indicators of the state and not show currency in those cases so maybe a single currency would suffice? To be extra clear we could add a fake currency `XXX` but I don't think thats needed

Using just `USD` we can use: 
- `createMoney(-1, 'USD')`  - `loadingMoney`
- `createMoney(-2, 'USD')` - `errorMoney`

We can even type them simply to be clear:
```
interface BalanceLoading { 
    amount: -1;
    status: 'loading'; 
} 

interface BalanceError { 
    amount: -2;
    status: 'error'; 
    error: Error; 
}
```

Here is the UI to show we don't need the currency for these states, just the state 

Screenshot 2025-01-14 at 17 07 59

packages/services/src/infrastructure/api/best-in-slot/best-in-slot-api.client.ts
```diff
@@ -7,127 +7,151 @@ import { HttpCacheService } from '../../cache/http-cache.service';
...
+      [
+        'best-in-slot-runes-valid-outputs',
+        bitcoinNetworkModeToCoreNetworkMode(networkService.getConfig().chain.bitcoin.mode),
+        xpub,
+      ],
+      // eslint-disable-next-line @typescript-eslint/require-await
+      async () => {
+        // TODO: Replace once new endpoint is ready
+        return [];
+        // const res = await limiter.add(
```
pete-watters, line 136: Maybe you can delete this?

packages/services/src/infrastructure/api/leather/leather-api.client.ts
```diff
@@ -1,43 +1,93 @@
...
+
+  async function fetchTxs(xpub: string, signal?: AbortSignal) {
     const network = networkService.getConfig().chain.bitcoin.bitcoinNetwork;
     return await cacheService.fetchWithCache(
-      ['leather-utxos', network, descriptor],
+      ['leather-txs', network, xpub],
       async () => {
         const res = await axios.get<LeatherApiUtxo[]>(
-          `https://leather-bitcoin-balances.wallet-6d1.workers.dev/${network}/${descriptor}`,
+          `https://leather-bitcoin-balances.wallet-6d1.workers.dev/${network}/${xpub}/txs`,
```
pete-watters, line 80: Should we be more protective of exposing this endpoint?

pete-watters:
Great work here @alexp3y 👏  even adding Rune's support as you go 💪  

I added some suggestions but nothing major. Looks great overall.

apps/mobile/src/hooks/use-bitcoin-account-requests.ts
```diff
@@ -0,0 +1,123 @@
...
+
+          const accountXpub: BitcoinAccountXpub = {
+            fingerprint: wallet.fingerprint,
+            accountIndex,
+            ...findSegwitDescriptors(accountDescriptors),
+          };
+
+          return {
+            account: accountXpub,
+            unprotectedUtxos: [] as UtxoId[], // TODO: incorporate unprotected UTXOs
```
camerow, line 65: Should we capture this TODO in an issue?

apps/mobile/src/queries/balance/btc-balance.query.ts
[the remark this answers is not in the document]
camerow: > prioritize a systematic review of all of our cache times at both levels (RQ + HTTP cache)

Can you write that up as a simple RFC?

apps/mobile/src/hooks/use-bitcoin-account-requests.ts
alexp3y: This array allows us to connect to an Inscription protection feature (like the one that @kyranjamie recently implemented in `extension`) and allows any selectively unprotected UTXOs to be considered when doing things like calculating available balance and fetching UTXOs.

We can capture this in an issue but would depend on / need to be a part of a broader project implementing the Inscription protection feature in `mobile`.

packages/services/src/infrastructure/api/leather/leather-api.client.ts
alexp3y: This specific endpoint actually doesn't exist yet and will change soon to be under our own API. The base URL here is the same worker URL we already expose for the UTXO endpoint so its not sensitive.

apps/mobile/src/shared/loadable-balance.ts
alexp3y: Good point! If the currency isn't needed in those states then it makes sense to just use a single currency. SIP10s and Runes would be difficult otherwise.

I was also returning a currency-specific `zeroMoneyX` when there's no loading or error state and an undefined `Money` object... But now thinking it through, we should just change that case to also return `errorMoney`. The balance services would return a zero balance if / when that's the intention, but an undefined money should be considered an error state. That will simplify it further.

apps/mobile/src/hooks/use-bitcoin-account-requests.ts
pete-watters: Maybe we can get that project on the board and just add a reference here? It's more so we don't forget what is to be done

alexp3y: FYI [LEA-1983](https://linear.app/leather-io/issue/LEA-1983/inscription-protection-feature-mobile)

apps/mobile/src/queries/balance/btc-balance.query.ts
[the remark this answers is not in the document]
alexp3y: FYI [Cache Review Process](https://www.notion.so/trustmachines/Cache-Review-Process-17cd74b694c780bcba13e2b1a82b52b8?pvs=4)

apps/mobile/src/shared/loadable-balance.ts
kyranjamie: I'm concerned that this is a transparent behaviour that encourages bad practice (creating illegal data models). How do new developers joining the team know that `-1 === 'loading'`?

As I see it, one of the values of having data models is that, if you encounter `Money` you know its data representing a monetary value. Reusing the data model to represent async data states, goes beyond the responsibility of the model.

kyranjamie (changes requested):
This is great work, so apologies for rejecting Alex. Let's debate as a team, happy to be overruled if other's agree the trade offs are worthy—but my opinion is that that putting "error values" in `Money` is a really dangerous move we should reconsider.

Currently Money === Money. This PR introduces a way of encoding non-monetary data into our `Money` data model. Developers can no longer be assured about the data type. They'll have to check from list of known-error values (documented where?). Encoding these states means that Money is aware of asynchronous concerns, and that isn't the responsibility of it.

pete-watters:
> This is great work, so apologies for rejecting Alex. Let's debate as a team, happy to be overruled if other's agree the trade offs are worthy—but my opinion is that that putting "error values" in `Money` is a really dangerous move we should reconsider.
> 
> Currently Money === Money. This PR introduces a way of encoding non-monetary data into our `Money` data model. Developers can no longer be assured about the data type. They'll have to check from list of known-error values (documented where?). Encoding these states means that Money is aware of asynchronous concerns, and that isn't the responsibility of it.

I agree that doing it as proposed [here](https://github.com/leather-io/mono/pull/810#discussion_r1916552385) is better however are we willing to spend time doing that right now? 

If not I think this way of doing things is better than always falling back to a default of `0` balance which may not be correct. 

Balance calls are wrapped with `getLoadableBalance` so we could add documentation or typing there to indicate the intention and also have an un-used token like 'XXX' so we could potentially have a `-1 BTC` value. 

If we want to hold off introducing this in favour of refactoring I am happy with that also, and in that case perhaps we can update this PR so that `getLoadableBalance` returns `createMoney(0, 'USD')` in the meantime so we can at least take in all of the other good work

kyranjamie:
> I agree that doing it as proposed https://github.com/leather-io/mono/pull/810#discussion_r1916552385 is better however are we willing to spend time doing that right now?

If not now, when? In building out the services layer, that's our opportunity to remove tech debt.

I don't see the productivity win in a negative value money approach, which seems to be the main argument in its favour. In rolling out this feature to the extension, we're going to need to go through all the components that consume these values going...

```ts
if (result.money.amount.toNumber() === -1) return <Loading />
```

So, if the other way is better, why not just build it the other way and do...

```ts
if (result.state === 'loading') return <Loading />;
// result.money === `Money`
```

I feel obliged to push on this one, because it's a loosening the responsibility of one of most widely used data entities. It's a overstep of responsibilities having `Money` be aware of async states. If this were a quick-win approach done at the presentation layer in a single feature component, sure, what's the harm?

But introduces a complicated, unclear behaviour to one of our most-core data models, and that's a slippery slope.

alexp3y:
@kyranjamie All good and I agree with you (and mr gpt). Ultimately overloading `Money` and encoding numeric values isn't sustainable and leads us down a dangerous path. It was really meant as an intermediate solution to keep things moving but you're right, we should take this opportunity to get it right while we're still setting things up.

Introducing a reusable loading state wrapper that can be used for this and other types is a good plan and implementing it for balances should be relatively contained, so I'll just knock this out now while the discussion is going and the iron is hot.

pete-watters:
> > I agree that doing it as proposed [#810 (comment)](https://github.com/leather-io/mono/pull/810#discussion_r1916552385) is better however are we willing to spend time doing that right now?
> 
> If not now, when? In building out the services layer, that's our opportunity to remove tech debt.
> 
> I don't see the productivity win in a negative value money approach, which seems to be the main argument in its favour. In rolling out this feature to the extension, we're going to need to go through all the components that consume these values going...
> 
> ```ts
> if (result.money.amount.toNumber() === -1) return <Loading />
> ```
> 
> So, if the other way is better, why not just build it the other way and do...
> 
> ```ts
> if (result.state === 'loading') return <Loading />;
> // result.money === `Money`
> ```
> 
> I feel obliged to push on this one, because it's a loosening the responsibility of one of most widely used data entities. It's a overstep of responsibilities having `Money` be aware of async states. If this were a quick-win approach done at the presentation layer in a single feature component, sure, what's the harm?
> 
> But it introduces a complicated, unclear behaviour to one of our most-core data models, and that's a slippery slope.

OK 👍 lets finish it properly now once and for all 🫡

alexp3y:
@camerow @kyranjamie @pete-watters Ok, addressed all previous comments and now added a generic `FetchState<T>` wrapper with a discriminable `.state` field as discussed. 

`FetchState` can be applied to any type and I included a utility `toFetchState` that converts directly from react query responses. Already applied it to balance and UTXO query hooks and did my best to accommodate in the associated components, although my handling mostly consists of early returns if `state !== 'success'`. 

App still runs fine, just a bit of loading wonkiness. Will require follow-on work to handle the loading / error states in components but this should be a good base to work from for all fetchable data. I flagged each state check and early return with `// TODO: handle balance loading & error states` comments.

kyranjamie (approved):
Thanks @alexp3y, LGTM

apps/mobile/src/shared/fetch-state.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!data) {
+    return {
+      state: 'error' as const,
+      errorMessage: '',
+    };
+  }
+  return {
+    state: 'success' as const,
+    value: data,
+  };
```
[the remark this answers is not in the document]
alexp3y: Either seems to work, your version looks cleaner. I'll change it.

PR #814, opened 2025-01-14
fix(query): update stamp schema

kyranjamie:
Prop is now optional


camerow (approved):

markmhendrickson:
Does this PR resolve the broken display of Stamps and SRC-20 tokens?

PR #819, opened 2025-01-17
fix: no wallet balance service calls

alexp3y:
PR fixes an issue that occurs when balance services (BTC, STX, SIP10) are called with fresh app state and/or no wallets loaded (i.e. an empty array of accounts). The balance services should return a default zero balance, instead they throw errors, preventing components from rendering.

The PR also adds logic to skip the BIS xpub inscriptions request when on testnet as it isn't currently supported on the BIS testnet server and is returns a 404. The testnet server will be upgraded end of month and the request can be reinstated.


pete-watters (approved):
Nice work @alexp3y 👍 

So we default to `0` and then can still handle errors appropriately via the new state then I can also filter out tokens with `0` we don't want to see client side

alexp3y:
@pete-watters The balance defaults should only apply at aggregate-level calls, e.g. the aggregate SIP10 balance of `[]` accounts is `0 USD` (instead of an error), but there wouldn't be any zero-balance tokens under that.

PR #820, opened 2025-01-17
feat: SIP-10 balances on mobile

pete-watters:
This PR:
- Integrates the SIP-10 balances service to mobile: LEA-1832
- Adds boiler plate code for Runes: LEA-1981 

I didn't add the icons yet as Tigran will add update icon component to accept URLs LEA-1909

For the `Tokens Widget` we have a hard cap to show only 5 tokens so I added some code to cap it:
- BTC
- STX
- 3 SIP tokens

This is OK for now but we need to revisit this. You can see a demo here:

https://github.com/user-attachments/assets/37cfb782-250d-4015-80ab-3735bb22e830

I did more testing before this cap to make sure we are showing the correct aggregate balances on the homepage:

https://github.com/user-attachments/assets/3bf2cf5e-c6bb-43f8-adc6-21d2c99b7c67

### DIFF
--- a/apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
+++ b/apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
@@ -2,6 +2,8 @@ import {
   BitcoinBalance,
   BitcoinBalanceByAccount,
 } from '@/features/balances/bitcoin/bitcoin-balance';
+import { RunesBalance, RunesBalanceByAccount } from '@/features/balances/bitcoin/runes-balance';
+import { Sip10Balance, Sip10BalanceByAccount } from '@/features/balances/stacks/sip10-balance';
 import { StacksBalance, StacksBalanceByAccount } from '@/features/balances/stacks/stacks-balance';
 import { AccountId } from '@/models/domain.model';
 import { HasChildren } from '@/utils/types';
@@ -25,6 +27,9 @@ export function AllAccountBalances() {
     <>
       <BitcoinBalance />
       <StacksBalance />
+      <Sip10Balance />
+      {/* TODO LEA-1982: add runes balance */}
+      <RunesBalance />
     </>
   );
 }
@@ -34,6 +39,8 @@ export function AccountBalances({ fingerprint, accountIndex }: AccountId) {
     <>
       <BitcoinBalanceByAccount fingerprint={fingerprint} accountIndex={accountIndex} />
       <StacksBalanceByAccount fingerprint={fingerprint} accountIndex={accountIndex} />
+      <Sip10BalanceByAccount fingerprint={fingerprint} accountIndex={accountIndex} />
+      <RunesBalanceByAccount fingerprint={fingerprint} accountIndex={accountIndex} />
     </>
   );
 }
--- a/apps/mobile/src/features/balances/bitcoin/runes-balance.tsx
+++ b/apps/mobile/src/features/balances/bitcoin/runes-balance.tsx
@@ -1,0 +1,77 @@
+import { TokenIcon } from '@/components/widgets/tokens/token-icon';
+import { createBitcoinAccountIdentifier } from '@/hooks/use-bitcoin-account-service-requests';
+import { AccountId } from '@/models/domain.model';
+import {
+  useRunesAccountBalance,
+  useRunesTotalBalance,
+} from '@/queries/balance/runes-balance.query';
+
+import { Money } from '@leather.io/models';
+import { PressableProps } from '@leather.io/ui/native';
+
+import { TokenBalance } from '../token-balance';
+
+interface RunesTokenBalanceProps extends PressableProps {
+  availableBalance: Money;
+  fiatBalance: Money;
+  symbol: string;
+  name: string;
+}
+export function RunesTokenBalance({
+  availableBalance,
+  fiatBalance,
+  name,
+  symbol,
+  ...rest
+}: RunesTokenBalanceProps) {
+  return (
+    <TokenBalance
+      ticker={symbol}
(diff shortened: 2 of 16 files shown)


apps/mobile/src/queries/balance/runes-balance.query.ts
```diff
@@ -0,0 +1,47 @@
+import { useTotalBitcoinAccountServiceRequests } from '@/hooks/use-bitcoin-account-service-requests';
+import { toFetchState } from '@/shared/fetch-state';
+import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
+
+import { BitcoinAccountIdentifier, getRunesBalancesService } from '@leather.io/services';
+
+export function useRunesTotalBalance() {
+  // TODO LEA-1982: check with Alex once we have xpub-based runes endpoint from BIS
```
pete-watters, line 8: @alexp3y I just stubbed out this `Runes` code, we can revisit it once the BIS endpoint is ready and probably refactor this

packages/services/src/balances/sip10-balances.service.ts
```diff
@@ -55,6 +57,7 @@ export function createSip10BalancesService(
     return {
       usd: totalUsdBalance,
       addressBalances,
+      aggregateBalances: getAggregateSip10Balances(addressBalances),
```
pete-watters, line 60: @alexp3y I added `aggregateBalances` as that's what we need on the homepage. 

We don't use `addressBalances` from this service but it could be useful? I'm not sure. I think it has a different query cache than account specific balances anyway?

apps/mobile/src/features/balances/stacks/sip10-balance.tsx
```diff
@@ -0,0 +1,82 @@
...
+  useSip10AccountBalance,
+  useSip10TotalBalance,
+} from '@/queries/balance/sip10-balance.query';
+
+import { Money } from '@leather.io/models';
+import { PressableProps } from '@leather.io/ui/native';
+
+import { TokenBalance } from '../token-balance';
+
+const sip10MaxDisplay = 3;
```
pete-watters, line 12: I'm adding this hard cap of 3 SIP-10 tokens until we figure out what we want to do here

apps/mobile/src/features/balances/bitcoin/runes-balance.tsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+}
+
+export function RunesBalanceByAccount({ fingerprint, accountIndex }: AccountId) {
+  const account = createBitcoinAccountIdentifier(fingerprint, accountIndex, []);
+  const data = useRunesAccountBalance(account);
+
+  // TODO LEA-1726: handle balance loading & error states
+  if (data.state !== 'success') return;
+  return data.value?.runes.map((balance: any, index: any) => (
```
kyranjamie, line 66: isn't it typed such that values is definitely defined when `state !== 'success'`?

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
```diff
@@ -25,6 +27,9 @@ export function AllAccountBalances() {
     <>
       <BitcoinBalance />
       <StacksBalance />
+      <Sip10Balance />
```
pete-watters, line 30: @kyranjamie : do you have some good ideas about how to handle this best? I asked about it [in the design channel](https://trustmachines.slack.com/archives/C05LRS7G44R/p1737023080561729).

In the accounts widget, we only have space for 5 tokens. For now I have prioritised:
- BTX
- STX
- 3 others SIPs

We load tokens from different places so if we only want to show 5 and prioritise some it gets messy quickly. At the moment users will also have no way to see all of their SIP-10s as work on that[ is ongoing](https://www.figma.com/design/gcXpWKFK4cODzlPuauX4Im/%F0%9F%94%8E-Details-views-(Token%2C-Collectible%2C-App)?node-id=2003-45775&t=o1Cwqr5V0mcis0LX-1). 


I [added a hard cap](https://github.com/leather-io/mono/pull/820/files#r1920388678) of 3 SIP-10's in the interim but there is no sorting there yet. This will also not work if there are `0` balances for STX or BTC as we will still only show 3  SIP-10's 😢 

@alexp3y maybe we could use a balance wrapper service to wrap and curate the balances? 

That would move us away from this coding style though of separate balance components which we favour

apps/mobile/src/features/balances/bitcoin/runes-balance.tsx
pete-watters: Yes you're right, I'll change! Old habits eh. 

Also it reinforces the benefit of this well typed error response

apps/mobile/src/components/widgets/tokens/tokens-widget.tsx
pete-watters: I'm merging the initial batch of work on this but we will need to revisit and figure this out in future

kyranjamie: Sorry @pete-watters missed this—but think you got guidance on Slack?

Imagine we'll need some sort of token management similar to how we have in the ext, in mean time I think a hard cap is fine as you've impleemented.

pete-watters: No problem, and thanks! There is a task for token ordering here [LEA-1952](https://linear.app/leather-io/issue/LEA-1952/order-tokens) so we can tackle it properly then. 

I think it could be cleaner to have a `balances` interface wrapper with a single service call to get all of the balances in a predefined manner but we can assess it once the requirements are more defined. Otherwise I see a lot of UI code to show/hide based on token type and holdings or a whole management feature as you suggest.

PR #823, opened 2025-01-22
fix: specify sheet height for android, ref LEA-2023

pete-watters:
This PR solves a problem with the Send Flow on Android whereby the recipient sheet does not open as there is no set height.

### DIFF
--- a/apps/mobile/src/features/send/send-form/sheets/recipient-sheet.tsx
+++ b/apps/mobile/src/features/send/send-form/sheets/recipient-sheet.tsx
@@ -26,7 +26,7 @@ export function RecipientSheet<T extends SendFormBaseContext<T>>({
 
   return (
     <Sheet ref={sheetRef} themeVariant={themeDerivedFromThemePreference}>
-      <Box pt="5" px="5">
+      <Box p="5" gap="1" height={112}>
         <TextInput
           autoCapitalize="none"
           autoComplete="off"


apps/mobile/src/features/send/send-form/sheets/recipient-sheet.tsx
```diff
@@ -26,7 +26,7 @@ export function RecipientSheet<T extends SendFormBaseContext<T>>({
 
   return (
     <Sheet ref={sheetRef} themeVariant={themeDerivedFromThemePreference}>
-      <Box pt="5" px="5">
+      <Box p="5" gap="1" height={112}>
```
pete-watters, line 29: Setting the height here means the sheet can open on Android but now there is too much space below it on iOS:

Screenshot 2025-01-22 at 11 00 21

I didn't want to start adding platform specific code but maybe that's what we should do here?

This flow is being re-vamped anyway so I can ask the design team about this

packages/services/src/balances/sip10-balances.service.spec.ts
```diff
@@ -105,7 +105,7 @@ describe('Sip10BalancesService', () => {
       expect(mockStacksApiClient.getAddressBalances).toHaveBeenCalledTimes(3);
       expect(mockMarketDataService.getSip10MarketData).toHaveBeenCalledTimes(6);
       expect(mockSip10TokensService.getAssetInfo).toHaveBeenCalledTimes(6);
-      expect(aggregateBalance.aggregateBalances.length).toEqual(3);
+      expect(aggregateBalance.aggregateBalances.length).toEqual(2);
```
pete-watters, line 108: I'm not sure why this was failing. Flaky test / poor developer

tigranpetrossian (approved):
Thanks for fixing this @pete-watters. I'll take another look too, while I'm working on the form.

apps/mobile/src/features/send/send-form/sheets/recipient-sheet.tsx
camerow: I think this is fine for the time being - future plans will add more content here so it'll be a non-issue in the midterm.

camerow (approved):

PR #825, opened 2025-01-22
feat: change spam filter logic, LEA-2072

alter-eggo:
This pr changes spam filter logic, [context](https://trustmachines.slack.com/archives/C0520G8J33M/p1737410885630849)


packages/utils/src/spam-filter/spam-filter.spec.ts
```diff
@@ -5,16 +5,21 @@ describe('Spam filter', () => {
     expect(spamFilter('This token name is OK')).not.toEqual(spamReplacement);
   });
   it('should detect spam urls in strings and replace content', () => {
-    expect(spamFilter('www.fake')).toEqual(spamReplacement);
-    expect(spamFilter('https://www.fake.com')).toEqual(spamReplacement);
+    expect(spamFilter('www.fake')).toEqual('www.fake');
```
kyranjamie, line 8: This, and all the other tests changed below, should be removed because they no longer reflect the functionality described on L7

camerow:
> I have the feeling we can drastically simplify this.
> 
> Originally, because well-formed URLs are most often used to link users to scam websites, we blocked all URLs.
> 
> The ask in the issue is that we allow all known TLDs, so here we've added all known TLDs as exceptions.
> 
> In other words, we block all urls, then make an exception for all possible urls... or am I missing something?

@kyranjamie I thought we were doing the opposite with the TLDs, using them as the source of truth for what to block rather than, say, a regex that checks for a well formed URL.

kyranjamie (approved):
> @kyranjamie I thought we were doing the opposite with the TLDs, using them as the source of truth for what to block rather than, say, a regex that checks for a well formed URL.

@camerow Yeah realise this now we've discussed on Slack

I think we should not allow known web protocols, as they only serve as a means to try and deceive users, but don't wish to debate further.

PR #830, opened 2025-01-27
fix: use correct background color on account sheet, closes LEA-2081

tigranpetrossian:
Fixes visual regression on Account selector sheet.

image

Created [an issue](https://linear.app/leather-io/issue/LEA-2086/make-native-sheet-component-composable) to make the native sheet component composable, so we can easily reuse its pre-styled subcomponents in the future.


pete-watters (approved):
🚀

PR #832, opened 2025-01-27
fix: blur view ci

edgarkhanzadian:
Seems to be connected to the jitpack issue that demezis blurview is having on 2.0.3 version. Bumping to 2.0.6 manually should fix it. https://github.com/Dimezis/BlurView/releases/tag/version-2.0.6

Latest expo-blur has it updated to 2.0.6, but latest expo-blur doesn't support expo@51, so decided to do patch instead.

Running demo build to test it as i am unable to reproduce this on local


pete-watters (approved):
Thanks @edgarkhanzadian

tigranpetrossian (approved):

PR #837, opened 2025-01-29
CD: Fastlane setup

pete-watters:


.github/workflows/firebase-deploy-ios.yml
```diff
@@ -0,0 +1,74 @@
+name: iOS Development Build
```
pete-watters, line 1: @camerow you can ignore this action. It's just doing a direct deploy to `firebase`

They could be useful as a reference for `runs-on: macos-latest` etc. as this works with doing `eas build` then deploy to firebase

.github/workflows/firebase-deploy.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Build Android app
+        working-directory: ./apps/mobile
+        run: eas build --platform android --profile=simulator-dev --non-interactive --no-wait --local --output ${{ github.workspace }}/app-release.apk
+
+      - name: Upload APK artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: app-release
+          path: ${{ github.workspace }}/app-release.apk
+      #   - name: Run Release Please
```
pete-watters, line 61: @camerow here I was getting the release notes from `release-please` but its taking them for all the differences between `main` and `dev`, of which there are way too many. This is why lots of tickets got re-opened also

apps/mobile/android/app/build.gradle
```diff
@@ -99,12 +99,17 @@ android {
...
-                storePassword LEATHER_UPLOAD_STORE_PASSWORD
-                keyAlias LEATHER_UPLOAD_KEY_ALIAS
-                keyPassword LEATHER_UPLOAD_KEY_PASSWORD
-            }
+
+            storeFile file('debug.keystore')
+            storePassword 'android'
+            keyAlias 'androiddebugkey'
+            keyPassword 'android'
+            // if (project.hasProperty('LEATHER_UPLOAD_STORE_FILE')) {
```
pete-watters, line 107: @camerow on the hackathon, it wouldn't work until I disabled `if (project.hasProperty('LEATHER_UPLOAD_STORE_FILE')` then I just copied in the `debug.keystore` etc. to get it working locally but I think thats not correct (although working)

pete-watters:
Closing this

PR #838, opened 2025-01-29
feat: firebase notifications

edgarkhanzadian:
- Add support for firebase notifications on the mobile side (both ios and android)
- add notifications sheet that would show up to users when they have at least 1 wallet in the app. (and when the feature flag for them is enabled)
- Add /register endpoint to the packages/services
- Add launchdarkly support
- Add firebase-tools to install the google-services json/plist files without much of a headache for developers
- Add 1password cli script to download the .env file. Currently it only holds the launchdarkly mobile key.

Checkout these docs to install 1password cli https://developer.1password.com/docs/cli/get-started


apps/mobile/ios/leatherwalletmobile/AppDelegate.mm
```diff
@@ -7,6 +8,9 @@ @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+// @generated begin @react-native-firebase/app-didFinishLaunchingWithOptions - expo prebuild (DO NOT MODIFY) sync-ecd111c37e49fdd1ed6354203cd6b1e2a38cccda
```
camerow, line 11: indent just to keep it formatted as per the rest of the doc?

apps/mobile/src/app/(home)/home/home.tsx
```diff
@@ -1,12 +1,45 @@
...
+import { featureFlagKeys } from '@/utils/feature-flag';
+import { useBoolVariation } from '@launchdarkly/react-native-client-sdk';
 import { useLingui } from '@lingui/react';
 
+import { SheetRef } from '@leather.io/ui/native';
+
 export function Home() {
   useLingui();
   const { hasWallets } = useWallets();
+  const notificationSheetRef = useRef<SheetRef>(null);
```
camerow, line 19: Do you think this could be worth abstracting into one / two hooks that encapsulate this logic?

camerow: eg. `useConfigureNotifications(sheetRef)`

apps/mobile/src/utils/feature-flag.ts
```diff
@@ -1,3 +1,57 @@
...
+    debug: process.env.NODE_ENV === 'development',
+    applicationInfo: {
+      id: 'leather-mobile-wallet',
+      version: Application.nativeApplicationVersion ?? '0.0.1',
+    },
+  }
+);
+
+export const featureFlagKeys = {
+  notifications: 'notifications',
```
camerow, line 35: Can you make this name more specific? Eg. `pushNotificationSupportEnabled`

apps/mobile/src/utils/feature-flag.ts
```diff
@@ -1,3 +1,57 @@
...
+};
+
+export type FEATURE_FLAG_KEYS = keyof typeof featureFlagKeys;
+
+async function setupFeatureFlags() {
+  const deviceId = await getDeviceId();
+  if (!deviceId) {
+    // TODO: handle this error properly
```
camerow, line 43: How likely is this to fail? If it fails should we still identify the user somehow?

apps/mobile/src/components/notifications/use-notifications.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  requestUserPermissions(): Promise<void>;
+}
+
+export function useRegisterNotifications(): RegisterNotifications {
+  const keychains = useBitcoinAccounts();
+  const { mutateAsync: registerToken } = useRegisterNotificationMutation();
+  const stacksAddresses = useStacksSignerAddresses();
+  const { displayToast } = useToastContext();
+  const isNotificationsEnabled = useBoolVariation(featureFlagKeys.notifications, false);
```
camerow, line 41: IMO this should be more encapsulated, eg we should export feature flag hooks from a directory so that each view doesn't have to know about `useXVariation`, it just calls `usePushNotificationsEnabled()`. Perhaps even better if the feature flags constant exports the flag name and the default value.

apps/mobile/src/app/(home)/settings/notifications/index.tsx
```diff
@@ -1,22 +1,36 @@
...
           })}
           icon={<PlaceholderIcon />}
           type="switch"
-          switchValue={false}
-          onSwitchValueChange={() => {
-            // TODO: re-enable when ready.
+          switchValue={notificationsPreference === 'enabled'}
+          onSwitchValueChange={isEnabled => {
+            if (isEnabled) {
+              changeNotificationsPreference('enabled');
```
camerow, line 28: Can this logic be encapsulated better so the view doesn't do so much deciding of what happens? Eg a hook that does all this?

apps/mobile/src/hooks/use-create-wallet.ts
```diff
@@ -37,6 +39,7 @@ export function useCreateWallet() {
           }),
         });
         router.navigate(AppRoutes.Home);
+        await registerNotifications();
```
camerow, line 42: Do we really want to registerNotifications every time a wallet is created? Not sure what the right solution is but calling `registerNotifications` all over the place feels like a code smell to me - feels like `createWallet` is a leaky abstraction. Not necessarily asking you to change this now but can we think of a more sustainable approach here so that we don't keep adding side effects to `createWallet` in the future?

apps/mobile/src/store/keychains/stacks/stacks-keychains.read.ts
```diff
@@ -47,7 +47,9 @@ export function useStacksSigners() {
 }
 
 export function useStacksSignerAddresses() {
-  return useStacksSigners().list.map(signer => signer.address);
+  const { list } = useStacksSigners();
+
+  return useMemo(() => list.map(signer => signer.address), [list]);
```
camerow, line 52:
```suggestion
  const { stacksSigners } = useStacksSigners();

  return useMemo(() => stacksSigners.map(signer => signer.address), [stacksSigners]);
```

apps/mobile/src/utils/feature-flag.ts
```diff
@@ -1,3 +1,57 @@
...
+    applicationInfo: {
+      id: 'leather-mobile-wallet',
+      version: Application.nativeApplicationVersion ?? '0.0.1',
+    },
+  }
+);
+
+export const featureFlagKeys = {
+  notifications: 'notifications',
+};
```
camerow, line 36:
```suggestion
export const featureFlagKeys = {
  enablePushNotificationsEnabled: {
    key: 'enablePushNotificationsEnabled',
    defaultValue: false
  }
};
```

apps/mobile/src/app/(home)/home/home.tsx
pete-watters: 👍 to this and for moving it to `features/notifications` along with the other notifications code

apps/mobile/src/store/keychains/stacks/stacks-keychains.read.ts
pete-watters: Do you even need `useMemo` here at all? https://kentcdodds.com/blog/usememo-and-usecallback

apps/mobile/src/utils/feature-flag.ts
```diff
@@ -1,3 +1,57 @@
...
+
+function getDeviceId() {
+  if (Platform.OS === 'android') {
+    return Promise.resolve(Application.getAndroidId());
+  } else {
+    return Application.getIosIdForVendorAsync();
+  }
+}
+
+// TODO: this is used for localization and it should be revised in favor
```
pete-watters, line 15: Maybe we can `TODO` this now? I added it to hide some features from internal release but it could be updated now to use launchdarkly

apps/mobile/src/utils/feature-flag.ts
```diff
@@ -1,3 +1,57 @@
...
+export type FEATURE_FLAG_KEYS = keyof typeof featureFlagKeys;
+
+async function setupFeatureFlags() {
+  const deviceId = await getDeviceId();
+  if (!deviceId) {
+    // TODO: handle this error properly
+    throw new Error('No device id detected');
+  }
+  featureFlagClient.identify({ kind: 'user', key: deviceId }).catch((e: any) => {
+    // TODO: analytics?
```
pete-watters, line 47: Analytics is setup so we can add an event here

packages/services/src/infrastructure/api/leather/leather-api.client.ts
```diff
@@ -125,9 +133,40 @@ export function createLeatherApiClient(
     );
   }
 
+  async function registerAddresses(
```
pete-watters, line 136: I would be nice to add some unit tests here

apps/mobile/README.md
```diff
@@ -33,6 +33,16 @@ Running on Android device
 - Install [Expo app from Google Play](https://play.google.com/store/apps/details?id=host.exp.exponent&pli=1) on your device.
 - Scan the QR code that is going to be shown in the terminal after running the expo server in the next section
 
+#### Credentials
+
```
pete-watters, line 37: I'm not sure about forcing this to everyone as it will stop the app being able to be built by external contributors. 

Can we set it up so that if these credentials are not present we don't include notifications? Maybe it can be done in the `notifications.service`?

apps/mobile/src/components/notifications/use-notifications.ts
tigranpetrossian: `useFeatureFlags` would make a lot of sense and be quite a DX improvement (assuming type safety and default value funcitonality is preserved), but there's likely a reason LD implemented `useFlags` for the React web SDK, but omitted it for RN. They [mention it being deliberate](https://docs.launchdarkly.com/sdk/features/evaluating#react-native), but don't give any rationale. 
I can only assume this has to do with caching, and avoiding too many rerenders as RN is much more sensitive to it.
Looks like the SDK is open source, I can try and read through it later today to understand the rationale.

tigranpetrossian: P.S.
To Will's point, it would make sense to at least extract to hooks like `usePushNotificationsEnabled`, both for reausability and less boilerplate. If anything we can look into having a one-hook-to-rule-it-all later.

apps/mobile/src/utils/feature-flag.ts
pete-watters: We use it in other places too, to hide settings and other things based on `ENV` so they aren't shown to leatherhood users

apps/mobile/README.md
pete-watters: Understood but it's still a big decision to make. It kind of makes us more open book than open source which we need to consider before committing to this path

pete-watters: Before merging can we investigate a way of doing this where we can still run the app, maybe via mocked services as Kyran suggests [here](https://trustmachines.slack.com/archives/C05LAP952E7/p1739365883080339?thread_ts=1739283921.335999&cid=C05LAP952E7)? 

I get that we need to do this for firebase but if we can't still run the app without these credentials maybe we should assess other options

packages/services/src/notifications/notifications.service.ts
```diff
@@ -0,0 +1,35 @@
+import { LeatherApiClient } from '../infrastructure/api/leather/leather-api.client';
+
+export interface NotificationsService {
+  registerAddressNotification(
+    variables: {
+      addresses: string[];
+      notificationToken: string;
+      chain: 'bitcoin' | 'stacks';
```
kyranjamie, line 8:
```suggestion
      chain: SupportedBlockchains;
```

apps/mobile/src/features/feature-flags/use-feature-flags.ts
```diff
@@ -0,0 +1,14 @@
+import { useBoolVariation } from '@launchdarkly/react-native-client-sdk';
+
+import { featureFlagKeys } from './feature-flag';
+
+export function useFeatureFlags() {
```
camerow, line 5: Don't flags update in realtime? I don't think it's wise to use one hook "to rule them all" personally. It provides very little immediate utility and will only get bloated over time. I feel pretty strongly personally that we should use one hook per flag - @tigranpetrossian could you weigh in and be the third perspective here?

camerow (changes requested):

apps/mobile/src/components/notifications/use-notifications.ts
camerow: Just saw this resolved thread after leaving another review. It doesn't sound like a single `useFeatureFlags` hook really does much for us _and_ there is potentially that it is hard to deal with or non-performant so it could be a pain. In that case, since there isn't any real clear advantage and many potential drawbacks I think we go with individual hooks. That's also more inline with every other implementation I've used.

apps/mobile/src/utils/feature-flag.ts
camerow: Shit I didn't realize we don't have sentry.

apps/mobile/src/features/notifications/use-notifications.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  useEffect(() => {
+    if (releasePushNotifications) {
+      if (notificationsPreference === 'enabled') {
+        void registerNotificationsSTX();
+      }
+      if (notificationsPreference === 'disabled') {
+        void unregisterNotifications();
+      }
+    }
```
kyranjamie, line 159:
```suggestion
    if (!releasePushNotifications) return
    if (notificationsPreference === 'enabled') {
      void registerNotificationsSTX();
    }
    if (notificationsPreference === 'disabled') {
      void unregisterNotifications();
    }
```

camerow (approved):

apps/mobile/src/app/(home)/home/home.tsx
```diff
@@ -1,12 +1,29 @@
...
+  const { notificationsPreference } = useSettings();
+  const releasePushNotifications = useReleasePushNotificationsFlag();
+
+  useEffect(() => {
+    if (releasePushNotifications && hasWallets && notificationsPreference === 'not-selected') {
+      notificationSheetRef.current?.present();
+    }
+  }, [hasWallets, notificationsPreference, releasePushNotifications]);
```
kyranjamie, line 25: I would push for a separate hook that handles the _when_ for readability.

```
useOnDetectNoNotiicationPreference(() => notificationSheetRef.current?.present())
```

apps/mobile/src/features/feature-flags/setup-feature-flags.ts
```diff
@@ -0,0 +1,24 @@
...
+    // eslint-disable-next-line no-console
+    console.log(e);
+  });
+}
+
+export function useSetupFeatureFlags() {
+  useEffect(() => {
+    void setupFeatureFlags();
+  }, []);
+}
```
kyranjamie, line 24: does the set up need to be in the react lifecycle?

apps/mobile/src/features/notifications/use-notifications.ts
```diff
@@ -0,0 +1,167 @@
...
+   * 3.Deletes a wallet
+   * 4. Toggles the notification setting
+   *
+   */
+  registerNotifications(): Promise<void>;
+
+  requestUserPermissions(): Promise<void>;
+
+  registerNotificationsBTC(): Promise<void>;
+  registerNotificationsSTX(): Promise<void>;
```
kyranjamie, line 36:
```suggestion
  registerNotificationsBtc(): Promise<void>;
  registerNotificationsStx(): Promise<void>;
```


https://www.notion.so/trustmachines/Naming-Camelcase-acronyms-119d74b694c78080974aee5e3bc18bc9?pvs=4

apps/mobile/src/features/notifications/use-notifications.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export function useRegisterNotifications(): RegisterNotifications {
+  const bitcoinAddresses = useBitcoinAddresses();
+  const { mutateAsync: registerToken } = useRegisterNotificationMutation();
+  const stacksAddresses = useStacksSignerAddresses();
+  const { displayToast } = useToastContext();
+  const releasePushNotifications = useReleasePushNotificationsFlag();
+
+  const registerNotificationsBTC = useCallback(async () => {
```
kyranjamie, line 47: Thoughts on returning a single `useMemo` with object of functions, rather than many `useCallbacks`?

apps/mobile/src/features/notifications/use-notifications.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (releasePushNotifications && notificationsPreference === 'enabled') {
+      void requestUserPermissions();
+      registerInAppNotifications();
+    }
+  }, [
+    releasePushNotifications,
+    notificationsPreference,
+    registerInAppNotifications,
+    requestUserPermissions,
+  ]);
```
kyranjamie, line 132: Perhaps this is just the way to do it, but I get scared seeing all these render-time side effects. Code like this is common in react but it runs this every render? 🤯

kyranjamie (approved):

apps/mobile/src/features/feature-flags/use-feature-flags.ts
tigranpetrossian: @camerow Covered this a little bit [here](https://github.com/leather-io/mono/pull/838#discussion_r1952453476)
Pros:
1. Not tied to LD at usage-level, in case we need to use different means to supply certain flags, or swap out LD with something else altogether.
2. Type-safety & nice DX, assuming the god hook can be implemented in a way that lists all available flags and their types.

Cons:
1. Perf: rerenders can be expensive in RN, I'm unclear of internal workings of the SDK, there's some caching logic at the very least
2. LD deliberately didn't provide a global hook in RN SDK (they did in React web SDK).

While I wouldn't be worried with either solution, because it's trivial to swap one solution with the other, it's probably the safest to just used wrapped hooks a la `usePushNotificationsFlag()` to at least partially address Pro 1. and Con 1. until we have a better understanding.

PR #845, opened 2025-02-04
Add "blur" effect to animated headers, closes LEA-1911

tigranpetrossian:
Adds a blur effect to animated headers, and fixes animated title being invisible for a short amount of time while scrolling up.

### Before
https://github.com/user-attachments/assets/f324e96b-a164-404b-99e2-ce1d264ae19e

### After
https://github.com/user-attachments/assets/f25719ef-a9f3-4016-aeaa-3b2ef45712e7

### DIFF
--- a/apps/mobile/src/components/headers/animated-header/animated-header.hooks.tsx
+++ b/apps/mobile/src/components/headers/animated-header/animated-header.hooks.tsx
@@ -1,6 +1,7 @@
 import { useState } from 'react';
 import { LayoutChangeEvent } from 'react-native';
 import {
+  interpolate,
   useAnimatedScrollHandler,
   useAnimatedStyle,
   useSharedValue,
@@ -9,7 +10,10 @@ import {
 
 import { useScrollViewStyles } from '@/hooks/use-scroll-view-styles';
 
-export function useAnimatedHeader(triggerAnimationYValue = 68) {
+const blurOverlayVisibilityThreshold = 12;
+const secondaryTitleVisibilityThreshold = 26;
+
+export function useAnimatedHeader(triggerAnimationYValue = secondaryTitleVisibilityThreshold) {
   const defaultStyles = useScrollViewStyles();
   const [contentHeight, setContentHeight] = useState(0);
   const [viewHeight, setViewHeight] = useState(0);
@@ -31,6 +35,11 @@ export function useAnimatedHeader(triggerAnimationYValue = 68) {
       });
       return {
         opacity,
+      };
+    }),
+    animatedBlurOverlayStyle: useAnimatedStyle(() => {
+      return {
+        opacity: interpolate(scrollY.value, [0, blurOverlayVisibilityThreshold], [0, 1], 'clamp'),
       };
     }),
     onContentSizeChange(_: number, height: number) {
--- a/apps/mobile/src/components/headers/animated-header/animated-title-header.tsx
+++ b/apps/mobile/src/components/headers/animated-header/animated-title-header.tsx
@@ -2,9 +2,12 @@ import { ReactNode } from 'react';
 import Animated, { useAnimatedStyle } from 'react-native-reanimated';
 
 import { TestId } from '@/shared/test-id';
+import { useSettings } from '@/store/settings/settings';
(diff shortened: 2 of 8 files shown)


apps/mobile/src/app/(home)/settings/index.tsx
```diff
@@ -100,6 +100,58 @@ export default function SettingsScreen() {
             onPress={() => router.navigate(AppRoutes.SettingsNetworks)}
             testID={TestId.settingsNetworkButton}
           />
+          <SettingsListItem
```
pete-watters, line 103: This is a mistake I think?

apps/mobile/src/components/headers/animated-header/animated-header.hooks.tsx
```diff
@@ -42,3 +51,16 @@ export function useAnimatedHeader(triggerAnimationYValue = 68) {
     },
   };
 }
+
+export function normalize(sourceRange: [number, number], targetRange: [number, number]) {
```
pete-watters, line 55: Does this get used anywhere? I can't see it in this PR. 

- If it's not used lets remove it
- If it is used how about:
    - moving it into `./utils.ts`
    - renaming `animated-header.hooks` as `use-animated-header.ts`

There is only one hook in this file so I think it should be renamed anyway

apps/mobile/src/components/headers/animated-header/animated-title-header.tsx
```diff
@@ -14,25 +17,45 @@ const AnimatedBox = Animated.createAnimatedComponent(Box);
...
   animatedStyle,
   title,
   rightElement,
+  animatedBlurOverlayStyle,
 }: AnimatedTitleHeaderProps) {
   const router = useRouter();
+  const { themeDerivedFromThemePreference } = useSettings();
+  const theme = useTheme<Theme>();
+  const gradientTransparentStopColor =
+    themeDerivedFromThemePreference === 'light' ? 'rgba(255,255,255,0)' : 'rgba(27,26,23,0)';
```
pete-watters, line 34: Could we add a `const`  / a token for the gradient?

pete-watters (changes requested):
Overall this looks good but there's a couple of issues to fix:
- [extra settings items](https://github.com/leather-io/mono/pull/845/files#r1941005996)
- [unused function](https://github.com/leather-io/mono/pull/845/files#r1941010416)

pete-watters:
> Overall this looks good but there's a couple of issues to fix:
> 
> * [extra settings items](https://github.com/leather-io/mono/pull/845/files#r1941005996)
> * [unused function](https://github.com/leather-io/mono/pull/845/files#r1941010416)

I see you fixed those but then on re-review I noticed this empty file we should remove - `apps/mobile/src/components/headers/animated-header/animated-header.layout.tsx`

apps/mobile/src/app/(home)/settings/index.tsx
tigranpetrossian: Indeed. Thanks for catching. Probably meant to test the scrolling and committed by mistake.

apps/mobile/src/components/headers/animated-header/animated-title-header.tsx
tigranpetrossian: I've debated doing this but gave up. We don't currently have alpha color tokens in the design system, and Figma doesn't support using color variables where alpha colors would matter most (gradients, shadow filters) making adding them kind of unjustified.
What this code should **really** look like is: `const gradientTransparentStopColor 
 = transparentize(theme.colors['ink.background-primary'])`, except pulling in a color library didn't feel justified either.
 
 Do you think we should make an issue out of this to look into alpha color use-cases?

pete-watters: If you want do but it's not a big deal really 👍

pete-watters (approved):
Nice work! 

Please just remove `apps/mobile/src/components/headers/animated-header/animated-header.layout.tsx` before merging

fbwoolf (approved):

PR #850, opened 2025-02-06
feat: support fiat currency setting, closes LEA-2032

alexp3y:
Implements the fiat currency preference user setting at the services-level.

Market data and balance service calls are now aware of the currently selected fiat currency via the injected settings service, which now contains a reference to fiat currency preference, making it available to services. The fiat currency preference has also been included in market data and balance react query keys in `mobile` to trigger refetching after setting updates.

Currency exchange rate information is retrieved via a stubbed fetch call via the Leather API client. This coupled with the fiat currency preference allows us to adjust market data prices accordingly. We'll need to find a backing service for live exchange rate data and implement the API endpoint separately.

Fields nameed "usd" have been changed to "fiat" in services and related UI components to reflect the changes.

The balance formatting utility is hardcoded to only accept "USD" so we're not formatting decimals on other fiat currencies atm.

<div align="center">

</div>


packages/services/src/balances/btc-balances.service.ts
```diff
@@ -50,24 +49,29 @@ export function createBtcBalancesService(
     const accountBalances = await Promise.all(
       balanceRequests.map(req => getBtcAccountBalance(req, signal))
     );
+
+    const cumulativeBtcBalance =
```
pete-watters, line 53: Maybe this logic could be extracted to a util that accepts `accountBalances`? Just to avoid the repeated ternary and make it cleaner?

pete-watters (approved):
Great work here @alexp3y 🚀 thanks for doing this. It's nice to see the hardcoded `fiat` strings disappear. 

- Do we know what currencies we can support? 
- Could we handle the ones [here](https://linear.app/leather-io/issue/LEA-2032/display-settings-enable-change-of-fiat-currency#comment-6507aeaa) ?



Overall it's good to merge but I would suggest to:
- make the logic for `cumulativeStxBalance =` a bit more DRY as it repeats a lot
- link the task in the branch / commit (LEA-2032)[https://linear.app/leather-io/issue/LEA-2032/display-settings-enable-change-of-fiat-currency]
    - that will help us with the release notes 
    - it also helps linking this to the UI connection we will need to do

tigranpetrossian:
@alexp3y Thanks so much for doing this.

Do you think we could fix the balance formatting within this PR to make sure the UI is presentable, or is it too much out of scope?
I just took a look at the [formatting helper](https://github.com/leather-io/mono/blob/b2084b76ccc0bef363aa11899a4f56a8d52a7f43/packages/utils/src/money/format-money.ts#L80) and it looks like a relatively simple change:

```
export function i18nFormatCurrency(quantity: Money, decimals = 2) {
  const currencyFormatter = new Intl.NumberFormat('en-US', {
    style: 'currency',
    currency: quantity.symbol,
    maximumFractionDigits: decimals,
  });

  if (quantity.amount.isLessThan(1)) {
    return formatDustAmount(currencyFormatter.format(0.01)); // Magic number based on existing implementation.
  }

  return currencyFormatter.format(quantity.amount.shiftedBy(-quantity.decimals).toNumber());
}

export function formatDustAmount(value: string) {
  return '<' + thinSpace + value;
}
```

In [Balance.tsx](https://github.com/leather-io/mono/blob/b2084b76ccc0bef363aa11899a4f56a8d52a7f43/apps/mobile/src/components/balance/balance.tsx#L34), we'll need a proper `isFiat()` checker for the above to work, I couldn't find one, but we at least have [the typing](https://github.com/leather-io/mono/blob/70394f672d47f826fbb32f7087e0680b5426bf3d/packages/models/src/currencies.model.ts)

apps/mobile/src/queries/balance/account-balance.query.ts
```diff
@@ -23,10 +24,10 @@ interface AccountBalance {
   totalBalance: FetchState<Money>;
 }
 
-const zeroMoneyUsd = createMoney(0, 'USD');
-
 export function useAccountBalance(accountId: AccountId): AccountBalance {
   const { fingerprint, accountIndex } = accountId;
+  const { fiatCurrencyPreference } = useSettings();
+  const zeroMoneyFiat = createMoney(0, fiatCurrencyPreference);
```
tigranpetrossian, line 30: nit: outside the `useAccountBalance` this would only instantiate once, thought I doubt this particular call would have any measurable impact.

pete-watters:
> @alexp3y Thanks so much for doing this.
> 
> Do you think we could fix the balance formatting within this PR to make sure the UI is presentable, or is it too much out of scope? I just took a look at the [formatting helper](https://github.com/leather-io/mono/blob/b2084b76ccc0bef363aa11899a4f56a8d52a7f43/packages/utils/src/money/format-money.ts#L80) and it looks like a relatively simple change:
> 
> ```tsx
> export function i18nFormatCurrency(quantity: Money, decimals = 2) {
>   const currencyFormatter = new Intl.NumberFormat('en-US', {
>     style: 'currency',
>     currency: quantity.symbol,
>     maximumFractionDigits: decimals,
>   });
> 
>   if (quantity.amount.isLessThan(1)) {
>     return formatDustAmount(currencyFormatter.format(0.01)); // Magic number based on existing implementation.
>   }
> 
>   return currencyFormatter.format(quantity.amount.shiftedBy(-quantity.decimals).toNumber());
> }
> 
> export function formatDustAmount(value: string) {
>   return '<' + thinSpace + value;
> }
> ```
> 
> This removes the exception from the helper as well. We should gradually remove all exceptions from `utils`, as these are heavily used in rendering logic with arbitrary data we don't control.
> 
> In [Balance.tsx](https://github.com/leather-io/mono/blob/b2084b76ccc0bef363aa11899a4f56a8d52a7f43/apps/mobile/src/components/balance/balance.tsx#L34), we'll need a proper `isFiat()` checker for the above to work, I couldn't find one, but we at least have [the typing](https://github.com/leather-io/mono/blob/70394f672d47f826fbb32f7087e0680b5426bf3d/packages/models/src/currencies.model.ts)

We were planning to add the UI for this later on, this is more of an investigation to see how feasible it is so we scoped it to a few hours. Luckily Alex managed to get the base we need done in that time. Next up for @alexp3y is work on activity services which we need more urgently. 

@tigranpetrossian  - you / I can add the UI for this and connect the setting later. Maybe it's good to do once the dust settles on our discussion about balance formatting and precision? (pun intended)

tigranpetrossian (approved):

tigranpetrossian:
> you / I can add the UI for this and connect the setting later. Maybe it's good to do once the dust settles on our discussion about balance formatting and precision? (pun intended) . This setting is not exposed to users now, don't worry. So even the internal testers won't see the horrible UI now

@pete-watters Absolutely. Worst case scenario we can do a follow up PR with these simple changes, while we get the bigger picture.

alexp3y:
@pete-watters All of the currencies in our `FiatCurrency` model are supported (which matches those in the Settings UI).
```typescript
export type FiatCurrency = LiteralUnion<
  'USD' | 'EUR' | 'GBP' | 'AUD' | 'CAD' | 'CNY' | 'JPY' | 'KRW',
  string
>;
```

packages/services/src/balances/btc-balances.service.ts
alexp3y: Yeah I agree the repeated ternaries are ugly. I wanted to make the instantiation of the zero balances conditional so we don't repeat that computation unnecessarily. I'll take a look to see if we can extract it somehow, maybe by caching zero balances for the different supported fiat currencies so we don't repeatedly create them...

alexp3y: Honestly, this is actually more difficult than it would seem... I will make a note to revisit this and find a better / cleaner way of handling these default zero balances.

apps/mobile/src/queries/balance/account-balance.query.ts
tigranpetrossian: My bad, didn't realize that. Not sure if it's worth the trouble unless there's measurable impact.

packages/services/src/balances/btc-balances.service.ts
pete-watters: OK, don't worry for now as adding this currency support so is already moving us towards a stretch goal. It's not a big deal for now 👍

PR #852, opened 2025-02-06
fix: stx call contract types

fbwoolf:
Some type changes I need to fix errors here:
https://github.com/leather-io/extension/pull/6060


kyranjamie (approved):
Thanks @fbwoolf, v similar to what I'm doing for msgs in other mono PR. Think we should remove the `RpcResponse<>` type, such that the schema helper fn is always used.

PR #853, opened 2025-02-07
feat: activity service, LEA-2150

alexp3y:
## Description: 
This PR adds an `Activity` model and `ActivityService` to support app activity / notification lists.

Activity is modeled on this [Activities](https://www.notion.so/trustmachines/e5ba2c2c5e2a46bbb0c5834cf6b25d21?v=120d74b694c78081bc18000c1bd8c1af&pvs=4) Notion page and related [Slack discussion](https://trustmachines.slack.com/archives/C05LRS7G44R/p1740737026176779).

QA will be better supported once a UI is in place. The current iteration of activity mapping is not _perfect_ but establishes a foundation to support continuous refinement as we adjust how different Txs, events, and other system actions translate to activity.

### Model Updates
- The `Activity` model is a composite of individual [Activities](https://www.notion.so/trustmachines/e5ba2c2c5e2a46bbb0c5834cf6b25d21?v=120d74b694c78081bc18000c1bd8c1af&pvs=4) types grouped under 3 high-level categories ("On-Chain" , "Wallet" , and "General"). The model utilizes a type hierarchy, similar to our `CryptoAssetInfo` model, which can be handled collectively or discriminated by type for more specific use.

- Added an `AccountIdentifier` model containing all info needed to uniquely identify a wallet account, which is used on `Activity` to allow matching items in a multi-account list (i.e. "Total Activity"). This model now replaces the more narrow "UnifiedAccountIdentifier" and "AccountServiceRequest" interfaces in `@leather.io/services` with a globally shared type.

### New Services
- `@leather.io/services` adds both a `Sip9Asset` service to fetch SIP-9 token metadata, and an `ActivityService` to fetch on-chain activity lists for Bitcoin and Stacks.   
- The Activity service exposes 3 calls:
  -  `getTotalActivity(account[]): Activity[]` -> Total on-chain activity list of provided accounts
  -  `getAccountActivity(account): Activity[]` -> Activity list for a single account
  -  `getActivityByAsset(account, asset): Activity[]` -> Activity list for a single asset (BTC, STX, or a SIP10 token)
- The Activity service could realistically be renamed "OnChainActivityService" as we still need to determine how and where off-chain activity data will be sourced.

### Hiro Stacks API Paging
- Added a generic multi-page fetching function for Hiro API requests to efficiently grab multiple pages of Tx and Event data (which map to Activity). 
- Supports fetching all pages by grabbing the first page, calculating total pages, and then fetching the remaining pages in parallel. Also supports fetching a set number of pages in parallel. 
- Determined its more performant (and probably sufficient for now) to just grab a one-size-fits-all handful of pages (currently 2) at once than wait for consecutive requests to get all pages.

## Activity Mapping
The core of the logic work here is mapping Bitcoin and Stacks Tx data to Activity objects.

More info here -> [Summary of On-Chain Mapping Rules](https://www.notion.so/trustmachines/On-Chain-Activity-Mapping-Rules-1afd74b694c7803384bfe830caebd894?pvs=4).
  
## Next Steps
- `LockAssetActivity` is not currently mapped. 
  - STX lock events are relatively uncommon. What seems to be more common are delegate() contract calls, which result in an `ExecuteSmartContractActivity`.
  - More investigation here is needed for how / when to map locking activity.
- `BTC` <-> `sBTC` swaps and other multi-tx activity not yet supported.


packages/services/src/transactions/stacks-transactions.service.ts
```diff
@@ -7,12 +7,36 @@ import {
 } from './stacks-transactions.utils';
 
 export interface StacksTransactionsService {
+  getTotalTransactions(address: string, signal?: AbortSignal): Promise<StacksTx[]>;
```
pete-watters, line 10: Do you think it would be good to accept `token` here too? Then we would be able to get SIP-10 / Stacks only data for the Token view pages also directly from the service? 

We can also search in the cache client side for that too but I'm not sure what would be the most efficient 🤔

pete-watters:
> @pete-watters Still working on this first pass at an activity service + queries but wanted to show my thinking here on how we could potentially unify transaction data under this base discriminable [`Activity`](https://github.com/leather-io/mono/compare/feat/activity-service?expand=1#diff-25fea7cc74179bfc7b803cab434bf7bdf5ce2d3817dc713f07d4170575a31efd) model (per our conversation yesterday).

This is awesome, thanks @alexp3y 👏  exactly what we need so the UI can keep light and just get `Activity` rather than data crunching

packages/models/src/activity.model.ts
```diff
@@ -0,0 +1,42 @@
...
+  readonly status: 'pending' | 'success' | 'failed';
+}
+
+export const TokenTransferDirections = {
+  outbound: 'outbound',
+  inbound: 'inbound',
+};
+export type TokenTransferDirection = keyof typeof TokenTransferDirections;
+
+export interface TokenTransferActivity extends BaseActivity {
```
pete-watters, line 23: 🤙  this is perfect and is all the details we need for the UI

apps/mobile/src/queries/activity/account-activity.query.ts
```diff
@@ -0,0 +1,49 @@
...
+  return toFetchState(useAggregateActivityQuery(accountIdentifiers));
+}
+
+export function useAccountActivity(fingerprint: string, accountIndex: number) {
+  const accountIdentifier = useUnifiedAccountIdentifier(fingerprint, accountIndex);
+  return toFetchState(useAccountActivityQuery(accountIdentifier));
+}
+
+export function useAccountActivityQuery(account: UnifiedAccountIdentifier) {
+  const { fiatCurrencyPreference } = useSettings();
```
pete-watters, line 22: Seeing this here now too and it's actually great to have gotten this done now as it will be pre-built for our next tasks from the start

packages/services/src/transactions/stacks-transactions.service.ts
alexp3y: Yeah, good point. It might make sense to eventually separate out into individual asset activity services (similar to how balances are structured). So, there may end up being a different entry point (e.g. `sip10ActivityService.getAllActivity(), sip10ActivityService.getTokenActivity(token)`), but yeah, I will try to support this kind of functionality via service call.

pete-watters: Thanks @alexp3y 👏  separating into separate calls would be good but i think if we can pass a token it would help with further sub-categorisation of SIP-10's. For example if we are viewing our `BANANA` / `TREMP` SIP-10's

apps/mobile/src/hooks/use-account-identifier.ts
```diff
@@ -0,0 +1,129 @@
...
+  useStacksSigners,
+} from '@/store/keychains/stacks/stacks-keychains.read';
+import { useWallets } from '@/store/wallets/wallets.read';
+
+import { inferPaymentTypeFromPath } from '@leather.io/bitcoin';
+import { extractAccountIndexFromPath } from '@leather.io/crypto';
+import { AccountIdentifier } from '@leather.io/models';
+import { isDefined, uniqueArray } from '@leather.io/utils';
+
+function getDescriptorFromKeychain<T extends { keyOrigin: string; xpub: string }>(
```
pete-watters, line 16: Maybe `getDescriptorFromKeychain` could move into `@leather.io/bitcoin` ?

apps/mobile/src/hooks/use-account-identifier.ts
```diff
@@ -0,0 +1,129 @@
...
+            );
+            return uniqueArray(
+              walletKeychains.map(key => extractAccountIndexFromPath(key.keyOrigin))
+            ).map(accountIndex =>
+              createBitcoinAccountIdentifier(
+                wallet.fingerprint,
+                accountIndex,
+                signers
+                  .fromAccountIndex(wallet.fingerprint, accountIndex)
+                  // TODO: throw error when no matching Stacks address
```
pete-watters, line 65: Is it much extra to add this error handling now?

apps/mobile/src/hooks/use-account-identifier.ts
```diff
@@ -0,0 +1,129 @@
...
+                .filter(isDefined)
+            )
+          );
+  }, [accounts.list, signers, keychains, fingerprint]);
+}
+
+export function useAccountIdentifier(fingerprint: string, accountIndex: number) {
+  const keychains = useBitcoinAccounts().fromAccountIndex(fingerprint, accountIndex);
+  const stxAddress = useStacksSignerAddressFromAccountIndex(fingerprint, accountIndex) ?? '';
+  if (!stxAddress) {
```
pete-watters, line 109: If you default `stxAddress` with `?? ''` I don't think this error gets thrown?

apps/mobile/src/hooks/use-account-identifier.ts
```diff
@@ -0,0 +1,129 @@
...
+    case 'p2tr':
+      return `tr(${accountKeychain.xpub})`;
+    case 'p2wpkh':
+      return `wpkh(${accountKeychain.xpub})`;
+    default:
+      return undefined;
+  }
+}
+
+export function createBitcoinAccountIdentifier(
```
pete-watters, line 29: Can you extract the arg types to an `interface`?

apps/mobile/src/hooks/use-account-identifier.ts
```diff
@@ -0,0 +1,129 @@
...
+  fingerprint: string,
+  accountIndex: number,
+  stxAddress: string,
+  descriptors: string[]
+): AccountIdentifier {
+  return {
+    fingerprint,
+    accountIndex,
+    stxAddress,
+    taprootDescriptor: descriptors.find(desc => desc.startsWith('tr(')) ?? '',
```
pete-watters, line 39: Maybe we should have a better guard here so that one of `taprootDescriptor` or `nativeSegwitDescriptor` is needed?

apps/mobile/src/queries/balance/runes-balance.query.ts
```diff
@@ -1,31 +1,31 @@
...
 
-import { BitcoinAccountIdentifier, getRunesBalancesService } from '@leather.io/services';
+import { AccountIdentifier } from '@leather.io/models';
+import { getRunesBalancesService } from '@leather.io/services';
 
 export function useRunesTotalBalance() {
-  // TODO LEA-1982: check with Alex once we have xpub-based runes endpoint from BIS
-  const totalRequests = useTotalBitcoinAccountServiceRequests();
-  return toFetchState(useRunesAggregateBalanceQuery(totalRequests.map(r => r.account)));
+  const accounts = useTotalAccountIdentifiers();
```
pete-watters, line 14: Thanks for fixing this 👍

packages/services/src/activity/activity.service.ts
```diff
@@ -0,0 +1,335 @@
...
+  getStacksAssetTransfers,
+} from './stacks-asset-transfer.utils';
+import {
+  getEventsByTxId,
+  mapContractCallActivity,
+  mapSmartContractActivity,
+  mapTokenTransferActivity,
+} from './stacks-tx-activity.utils';
+
+export interface ActivityService {
```
pete-watters, line 39: 🤤

packages/services/src/activity/activity.service.ts
```diff
@@ -0,0 +1,335 @@
...
+}
+
+export function createActivityService(
+  hiroStacksApiClient: HiroStacksApiClient,
+  stacksTransactionsService: StacksTransactionsService,
+  bitcoinTransactionsService: BitcoinTransactionsService,
+  marketDataService: MarketDataService,
+  sip10AssetService: Sip10AssetService,
+  sip9AssetService: Sip9AssetService
+): ActivityService {
```
pete-watters, line 64: Maybe you can add some unit tests to cover the activity service?

packages/services/src/activity/activity.service.ts
```diff
@@ -0,0 +1,335 @@
...
+      } catch {
+        // if we're unable to find asset metadata, remove transfer
+        return undefined;
+      }
+    }
+  }
+
+  async function applyMarketData(activity: Activity, signal?: AbortSignal) {
+    if (
+      (activity.type === OnChainActivityTypes.sendAsset ||
```
pete-watters, line 265: It would be more readable to add a few `const`s here. 

`const isFungible = activity.asset.category === CryptoAssetCategories.fungible`

```
if(isFungible)
```

packages/services/src/activity/activity.service.ts
```diff
@@ -0,0 +1,335 @@
...
+          crypto: cryptoValue,
+          fiat: baseCurrencyAmountInQuote(cryptoValue, assetMarketData),
+        },
+      };
+    } else if (
+      activity.type === OnChainActivityTypes.swapAssets &&
+      (activity.fromAsset.category === CryptoAssetCategories.fungible ||
+        activity.toAsset.category === CryptoAssetCategories.fungible)
+    ) {
+      let fromValue,
```
pete-watters, line 287: Maybe you don't need to use `let` here? 

You can return `undefined` at the end instead:
```
else {
      return { 
        ...activity,
        fromValue: undefined ,
        toValue: undefined,
    }
```

pete-watters (commented):
This PR is awesome @alexp3y 👍 

I added some refactoring suggestions but nothing major

packages/services/src/activity/bitcoin-tx-activity.utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  baseCurrencyAmountInQuote,
+  createMoney,
+  isDefined,
+  sumNumbers,
+  uniqueArray,
+} from '@leather.io/utils';
+
+import { LeatherApiBitcoinTransaction } from '../infrastructure/api/leather/leather-api.client';
+
+export function mapBitcoinTxBlockTime(tx: LeatherApiBitcoinTransaction) {
```
pete-watters, line 21: It would be good to have tests for this

packages/services/src/activity/stacks-tx-activity.utils.ts
```diff
@@ -0,0 +1,194 @@
...
+    status: mapStacksTxStatus(tx),
+  };
+
+  const sends = transfers.filter(t => t.sender === account.stxAddress);
+  const receives = transfers.filter(t => t.receiver === account.stxAddress);
+  const sendAssetIds = uniqueArray(sends.map(s => s.assetId));
+  const receiveAssetIds = uniqueArray(receives.map(r => r.assetId));
+
+  // Tx not considered a simple transfer if >1 asset sent, or >1 asset received
+  if (sendAssetIds.length <= 1 && receiveAssetIds.length <= 1) {
```
pete-watters, line 165: How about having a `const` here to make the comment code? 

```
const isSimpleTransfer = sendAssetIds.length >= 1 && receiveAssetIds.length >= 1

if(!isSimpleTransfer)
```

packages/services/src/assets/sip9-asset.service.ts
```diff
@@ -0,0 +1,39 @@
...
+import { HiroStacksApiClient } from '../infrastructure/api/hiro/hiro-stacks-api.client';
+import {
+  createSip9CryptoAssetInfo,
+  getContractPrincipalFromAssetIdentifier,
+  getNonFungibleTokenId,
+} from './stacks-asset.utils';
+
+export interface Sip9AssetService {
+  getAssetInfo(
+    assetIdentifier: string,
```
pete-watters, line 12: You should extract these types to an interface then avoid repeating it

packages/services/src/assets/stacks-asset.utils.ts
```diff
@@ -0,0 +1,92 @@
+import { hexToCV } from '@stacks/transactions';
```
pete-watters, line 1: I wonder if `stacks-asset.utils` should live in the `stacks` package?

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -141,10 +253,43 @@ export function createHiroStacksApiClient(
...
+            ),
+          {
+            priority: hiroApiRequestsPriorityLevels.getNftMetadata,
+            signal,
+            throwOnTimeout: true,
+          }
+        );
+        return res.data;
+      },
+      { ttl: HttpCacheTimeMs.infinity }
```
pete-watters, line 283: We want to cache this infinitely once retrieved?

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -114,7 +226,7 @@ export function createHiroStacksApiClient(
         );
         return res.data;
       },
-      { ttl: HttpCacheTimeMs.fiveMinutes }
+      { ttl: HttpCacheTimeMs.oneSecond }
```
pete-watters, line 229: Lets keep an eye on this to see if it gets us rate limited

packages/services/src/transactions/bitcoin-transactions.service.ts
```diff
@@ -12,12 +13,12 @@ export function createBitcoinTransactionsService(
...
     Gets bitcoin transactions for an account
   */
-  async function getAccountTransactions(account: BitcoinAccountIdentifier, signal?: AbortSignal) {
+  async function getAccountTransactions(account: AccountIdentifier, signal?: AbortSignal) {
     const [nativeSegwitTxs, taprootTxs] = await Promise.all([
       getDescriptorTransactions(account.nativeSegwitDescriptor, signal),
       getDescriptorTransactions(account.taprootDescriptor, signal),
     ]);
-    // TODO: combine duplicate txids
+    // TODO: deduplicate txids
```
pete-watters, line 21: Can you make a task for doing this in Linear? Or do it?

pete-watters (approved):
This PR is awesome, great work @alexp3y 👍  

I did an in-depth review and suggested:
- a couple of refactors
- adding some more tests

I added labels as such, however overall I think it's awesome and LGTM. 

Feel free to push back on suggestions / do them in a follow up.

The main thing would be to add more tests to `activity.service`

apps/mobile/src/hooks/use-account-identifier.ts
kyranjamie: +1 would be great to keep the pure methods located in the domain packages

apps/mobile/src/hooks/use-account-identifier.ts
kyranjamie: We already have an `AccountId` type that identifies an account with fingerprint and account index.  I'm not sure `AccountIdentifier` is the best name, as this includes data beyond just the minimal needed to identify it. This has account metadata too.

apps/mobile/src/hooks/use-account-identifier.ts
```diff
@@ -0,0 +1,129 @@
...
+export function useTotalAccountIdentifiers() {
+  const wallets = useWallets();
+  const keychains = useBitcoinAccounts();
+  const signers = useStacksSigners();
+
+  return useMemo(
+    () =>
+      !wallets || !signers
+        ? []
+        : wallets.list.flatMap(wallet => {
```
kyranjamie, line 53: Can this logic be extracted into a function used outside of the hook? That'd make it easier to test. Also, should this logic live in the mobile directly? Won't we want to do the same for extension?

packages/models/src/account.model.ts
```diff
@@ -0,0 +1,7 @@
+export interface AccountIdentifier {
+  fingerprint: string;
+  accountIndex: number;
+  taprootDescriptor: string;
+  nativeSegwitDescriptor: string;
+  stxAddress: string;
+}
```
kyranjamie, line 7: Is this predominantly needed for the activity service? If so, perhaps a more specific name would suit better. `ActivityRegistration`?

The problem with creating generic interface with specific details is that it requires upfront derivation of the key information. Building out the key management of mobile, with several accounts it was already getting kind of slow, so being able to defer derivation until needed is necessary, and this type encourages pre-computation.

Rather than having this be a core interface, why not have a function with signature below. This gives us the flexibility to defer the calculation, memoize it, read it from a cache etc.

```ts
function createAccountActivityRegistration(account: { fingerprint: string,  accountId: number }): ActivityRegistration;
```

Coupling bitcoin and stacks is also something we can only do on software wallets. Hardware wallet implementations need to handle accounts separately, as a user might have only connected bitcoin and not stacks etc, so the API choice here is also incompatible if we do add hw wallet support.

Also, seems inconsistent to have stacks address and bitcoin descriptors. Why not stx pub key?

packages/models/src/activity/activity.model.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export interface ReceiveAssetActivity extends BaseOnChainActivity {
+  readonly type: 'receiveAsset';
+  readonly asset: CryptoAssetInfo;
+  readonly senders: string[];
+  readonly amount: BigNumber;
+  readonly value?: {
+    crypto: Money;
+    fiat: Money;
+  };
```
kyranjamie, line 77: Wouldn't it be better to only have the actual amount as a single `Money` value. Fiat conversions will quickly go stale?

packages/services/src/activity/stacks-tx-activity.utils.ts
```diff
@@ -0,0 +1,194 @@
...
+
+export function isMempoolTx(
+  tx: HiroStacksTransaction | HiroStacksMempoolTransaction
+): tx is HiroStacksMempoolTransaction {
+  return (
+    tx.tx_status !== 'success' &&
+    tx.tx_status !== 'abort_by_post_condition' &&
+    tx.tx_status !== 'abort_by_response'
+  );
+}
```
kyranjamie, line 61: pretty sure we have a similar method to this in stacks package? If not, could this live there as a pure fn?

packages/services/src/assets/stacks-asset.utils.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export function getContractPrincipalFromAssetIdentifier(assetIdentifier: string) {
+  return assetIdentifier.split('::')[0];
+}
+
+export function getAddressFromAssetIdentifier(assetIdentifier: string) {
+  const principal = getContractPrincipalFromAssetIdentifier(assetIdentifier);
+  return principal.split('.')[0];
+}
```
kyranjamie, line 32: I recently saw this comment in extension

image

Logic being copied because of where its been placed. These fns are discreet pieces of logic useful to consume everywhere, without bringing with it a dependency on our entire services infra. cc @fbwoolf

kyranjamie (commented):
Great work @alexp3y, you've been busy!

Only concern here is `AccountIdentifier`. It might appear close enough like a generic account representation, but I think it sees `ActivityService` requirements slip into `models`. It couples bitcoin and stacks which will present issues with hardware wallet support.

Might it better to axe that type entirely from models, and consider it one specific to the service? Offering a utility fn to derive this information, rather than trying to bake it into our core model?

packages/models/src/account.model.ts
alexp3y: The `AccountIdentifier` interface isn't an Activity-specific type. It's also used by the Balance services, the Utxos service, and now the Activity service.

The point of moving it to `@leather.io/models` instead of a type in `@leather.io/services` (as it was before) is so that it can be used in models directly, e.g. Account-level Activity list items belong to an Account, how do we represent that account?

The xpub / address information is arguably a leakage of service-layer concerns into the model as you suggested, but I added it because key-derivation is tightly linked to app-specific implementations and key information needs to be presented at the service interface.

That said, I don't think `AccountIdentifier` is necessarily a misrepresentation of an "Account". A wallet (fingerprint) and index (accountIndex) is the minimal information to uniquely identify an Account. And the Bitcoin xpubs (tr and ns) and a Stacks address are the minimal information needed to fetch Bitcoin & Stacks data for an account.

It's obviously useful for services, but I thought it could be a step in the right direction for establishing a more globally useful Account model. We seem to have lots of one-off, app-specific account and key data representations.

As an alternative, my suggestions would be we keep at least the fingerprint / index as an `AccountIdentifier` model, which I think is most useful and accurate, and move the derived addresses / xpubs back to a service-specific interface.

In terms of performance, with features like aggregate activity lists and aggregate total balances, upfront derivation of all account key information is something apps will have to do immediately whether its on the type or not. I would imagine that properly memoization functions within the app should mitigate performance issues, or at least they will be trivial compared to data fetching.

packages/models/src/activity/activity.model.ts
alexp3y: Could be. The Activity data will probably be refreshed quite frequently because it needs to look for new Activity, meaning fiat prices would update. But applying market data on individual activity list items could also just be the responsibility of the UI... wdyt @pete-watters ?

pete-watters: I think it’s OK to have it this converted client side. 

We could have a single cached crypto amount and then convert  to fiat client side on render. 

Then we could also only be converting what’s visible to users?

I’m happy to keep this model as is too and also update ‘fiat’ client side. Keeping this format keeps it consistent with the balanced service and balance is just a sum of activity

pete-watters: Maybe it’s best to keep this as is but also add client conversion when rendering just to be certain it’s up to date?

apps/mobile/src/hooks/use-account-identifier.ts
alexp3y: Didn't realize we had `AccountId`, but its in mobile so its not like we can use that on other global types. Feel like that is a system-wide concept and should be moved to models. Then `AccountIdentifier` could be just `Account` and have it extend from `AccountId`

apps/mobile/src/hooks/use-account-identifier.ts
alexp3y: Could definitely extract this to a mobile utility function. I designed this as a mobile-specific implementation of account identifier objects because the extension and mobile key / wallet hooks and data structures involved are not 1:1

packages/models/src/account.model.ts
alexp3y: > Agreed, think that makes more sense, essentially moving this file apps/mobile/src/models/domain.model.ts to models?

Yeah, I think that would be perfect.

> Just think it's a name that understates what it really contains, as is. Ideas for other more descriptive names?

Yeah, maybe something like "AccountAddresses"? It can extend `AccountId` and then have optional entries for each supported chain.

In usage scenarios (e.g. services), we check whether specific chain address info is present and process accordingly. Makes it extensible and able to support hardware wallet single chain scenarios.

e.g.
```ts
// can replace these types w/ zod schemas to work with AccountId
interface BitcoinAddressInfo {
  taprootDescriptor: string;
  nativeSegwitDescriptor: string;
}
interface StacksAddressInfo {
  stxAddress: string;
}
interface AccountAddresses extends AccountId {
  bitcoin?: BitcoinAddressInfo;
  stacks?: StacksAddressInfo;
}

function getAccountActivity(account: AccountAddresses) {
  const activity = [];
  if (hasBitcoinAddress(account)) {
    activity.push(...fetchStacksActivity(account.stacks);
  }
  if (hasStacksAddress(account)) {
   activity.push(...fetchBitcoinActivity(account.bitcoin);
  }
  return activity;
}
```

packages/services/src/assets/stacks-asset.utils.ts
fbwoolf: I think until we decide differently, utils like these should be located and made available from the Stacks pkg so they aren't duplicated. Rn it is tough bc the services pkg isn't available in the extension, correct?

packages/models/src/activity/activity-type.model.ts
```diff
@@ -0,0 +1,24 @@
+export const OnChainActivityTypes = {
```
fbwoolf, line 1: Trying to understand things here, so if a contract has multiple contract calls we will show each one as it's own activity item, like we do now? I know there was some indication by design team that they wanted to bundle certain 'activity' together so the UI is cleaner, but it seems like a really challenging problem. Rn, we will list two items that both link to the same tx in the explorer. For example, both of these are one 'activity' but multiple steps to a contract... 
Screenshot 2025-03-07 at 3 26 20 PM

packages/services/src/assets/stacks-asset.utils.ts
pete-watters: I agree chain specific package utils is the place for chain specific stuff like this. 

I think once things stabilise on mobile that the balances and activity service are prime candidates to integrate into the extension and before we know it we will be sharing loads of code with consistent numbers

packages/models/src/activity/activity-type.model.ts
alexp3y: The activity service bundles items more along the lines of the updated UI design you mentioned, where the focus is more on finding the _actual_ asset transfers than showing all contract calls and intermediate assets where there was no transfer.

The two Txs from your screenshot, for example, will result in two activity items - One `ReceiveAssetActivity` of "ALEX Power Token" and one `ReceiveAssetActivity` of STX.

I tried to summarize the [current mapping rules](https://www.notion.so/trustmachines/Activity-Mapping-Rules-1afd74b694c7803384bfe830caebd894?pvs=4). I definitely expect this to be adjusted as we go. Currently contract function calls can still be displayed, but only when asset transfers aren't easily determined.

apps/mobile/src/hooks/use-account-identifier.ts
alexp3y: Stacks address is now optional in the new `AccountAddresses` model that replaces `AccountIdentifier`

apps/mobile/src/hooks/use-account-identifier.ts
alexp3y: Moved this to `utils` and added unit tests.

apps/mobile/src/hooks/use-account-identifier.ts
alexp3y: @pete-watters The bitcoin address info object on `AccountAddresses` is now optional and if both descriptors aren't available in `createAccountAddresses` it will be undefined.

packages/services/src/activity/stacks-asset-transfer.utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  return uniqueArray(transfers.map(transfer => transfer.sender).filter(isDefined));
+}
+
+export function aggregateTransferReceivers(transfers: StacksAssetTransfer[]): string[] {
+  return uniqueArray(transfers.map(transfer => transfer.receiver).filter(isDefined));
+}
+
+export function mapTransferFromTxEvent(
+  event: HiroTransactionEvent
+): StacksAssetTransfer | undefined {
```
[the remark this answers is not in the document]
alexp3y: It's expected that an event can map to either 0 or 1 transfer, so undefined wouldn't be an error here.

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
alexp3y: Maybe not but token metadata should update very infrequently. I have an RFC proposal to audit or caching at some point in the near future, which should include assessing the accuracy of cache times for different requests across the board.

packages/services/src/transactions/bitcoin-transactions.service.ts
alexp3y: Ah thanks, good catch! Meant to do this, will do it now.

packages/services/src/activity/activity.service.ts
alexp3y: Most of the business logic here is delegated to other `utils` files, which are pretty well-covered already. The main service does more orchestration than anything, so I don't think we need to do much. There are still a couple of more straightforward pure-ish fxs here though, so I'll add some tests for those.

alexp3y:
@pete-watters @kyranjamie @fbwoolf Thank you for the great feedback! I responded or made updates to address all comments.

The one exception is to move any utils to `@leather.io/stacks`. I totally agree with the points made by all - There is currently too much domain functionality here in services and it should be moved out. Many of the utils mentioned actually existed here before this PR, but were just moved to a new file. My original intention was to keep things isolated in `@leater.io/services` to prevent conflicts with existing, similar utils that serve the `@leather.io/query` package. I'd like to take a bit more time and attention to migrate those utils and ensure minimal code duplication. But for the sake of constraining an already massive PR and unblocking further Activity-specific work, I'll tackle that as a follow-on issue next week.

packages/models/src/activity/activity-type.model.ts
fbwoolf: Got it, thx for the link. I had missed that. 👍

fbwoolf (approved):

alexp3y:
@camerow What additional steps / settings are needed here to get this merged?

packages/models/src/activity/activity-type.model.ts
pete-watters: The activity stuff is extra complicated 🤯  well done for implementing this now Alex and also originally Fara! 

We can perhaps refine it in future to group these activities but I think that will get extra complicated and buggy so this is better for now

pete-watters:
> @pete-watters @kyranjamie @fbwoolf Thank you for the great feedback! I responded or made updates to address all comments.
> 
> The one exception is to move any utils to `@leather.io/stacks`. I totally agree with the points made by all - There is currently too much domain functionality here in services and it should be moved out. Many of the utils mentioned actually existed here before this PR, but were just moved to a new file. My original intention was to keep things isolated in `@leater.io/services` to prevent conflicts with existing, similar utils that serve the `@leather.io/query` package. I'd like to take a bit more time and attention to migrate those utils and ensure minimal code duplication. But for the sake of constraining an already massive PR and unblocking further Activity-specific work, I'll tackle that as a follow-on issue next week.

Sounds great 👍 just open a cooldown task to move that stuff and tackle it when you can

PR #858, opened 2025-02-10
Redesign Avatar component, closes LEA-1909, LEA-1758

tigranpetrossian:
This is an (almost) complete rewrite of the Avatar component:

* Replicate the functionality of the web version in the native version
* Incorporate design updates, add support for secondary indicators, add custom icon sizes per Avatar size.

### Global updates

* Converted to [self-contained structure instead of compound components](https://www.notion.so/trustmachines/Component-architecture-196d74b694c780c29c76f9c8d36c8309?pvs=4). The component has a rather minimal footprint, this format results in less verbosity at usage sites.
* Based on current usage, baked in the `defaultFallbackDelay` into the component as a default setting. Behavior is the same as before, the exact amount might need tweaking on the go.

### Mobile version

The React Native version now mimics the web version: added support for images with fallback, custom indicators, as well as possible visual and size variants. All occurrences of Avatar in the app were updated to the new format.

### Web version
* Converted to the `styled` API instead of recipes (per our prior conversations). `styled` offers the same feature set, but tucks away some of the boilerplate required with recipes. 
* Extension - Avatar size is temporarily set to 40 (by design should be updated to 48). This will require applying Cell redesign to the web version first in order to look right, which is way out of scope for this PR and is tracked in [LEA-2111](https://linear.app/leather-io/issue/LEA-2111/update-web-version-of-cell-to-new-designs)

The breaking changes for the extension were addressed in https://github.com/leather-io/extension/pull/6104.

#### Mobile Before & After

image

#### Extension Before & After

image

#### Secondary Indicators
image

By design, the cutouts for indicators are much more refined (see below). This isn't possible out of the box in either platform, I'm working on a solution (along with other "infeasible" things) that will come in a later iteration.

image


apps/mobile/src/features/approver/components/nonce-card.tsx
```diff
@@ -12,9 +12,7 @@ export function NonceCard({ nonce, onPress }: NonceCardProps) {
     <Box mx="-5">
       <Cell.Root pressable onPress={onPress}>
         <Cell.Icon>
-          <Avatar bg="ink.background-secondary">
-            <NumberedListIcon />
-          </Avatar>
+          <Avatar icon={<NumberedListIcon />} />
```
pete-watters, line 15: Nice work cleaning this up

packages/ui/src/components/avatar/avatar.web.tsx
```diff
@@ -1,114 +1,155 @@
...
-export function getAvatarFallback(val: string) {
-  return val.slice(0, 2);
+import { isDefined } from '@leather.io/utils';
+
+import { IconProps } from '../../icons/icon/create-icon.web';
+import { AvatarSize, AvatarVariant, defaultFallbackDelay, iconSizeMap } from './avatar.shared';
+
+type AvatarElement = ElementRef<typeof AvatarRoot>;
+
+export interface AvatarProps extends ComponentPropsWithoutRef<typeof AvatarRoot> {
```
pete-watters, line 19: Can we move this into a shared `types` and share more between web and native?

packages/ui/src/components/avatar/avatar.web.tsx
```diff
@@ -1,114 +1,155 @@
...
+          {indicator}
+        </styled.div>
+      ) : null}
+    </AvatarRoot>
+  );
+});
+
+Avatar.displayName = 'Avatar';
+
+const AvatarRoot = styled(RadixAvatar.Root, {
```
pete-watters, line 77: What do you think about having a separate `styles` file purely to keep the component size down and easier to work with?

packages/ui/src/components/avatar/brc20-avatar-icon.web.tsx
```diff
@@ -1,51 +1,57 @@
+import { styled } from 'leather-styles/jsx';
```
pete-watters, line 1: Maybe we can put these into a folder like `lib` / `instances` or something? Then we have a clean interface to the base component - `avatar/avatar` , `avatar/lib/brc20...` 

I'm not sure on the naming but it has historically been an avatar dumping ground in here. 

Maybe the issue is that they need to be refactored and have the `icon` passed in the same as the other avatars?

packages/ui/src/components/avatar/btc-avatar-icon.native.tsx
```diff
@@ -3,8 +3,6 @@ import { Avatar } from './avatar.native';
 
 export function BtcAvatarIcon() {
   return (
-    <Avatar bg="ink.background-primary" p="0">
-      <BtcIcon />
-    </Avatar>
+    <Avatar outlineColor="ink.border-transparent" icon={<BtcIcon width="100%" height="100%" />} />
```
pete-watters, line 6: Maybe we don't need this file? And can instantiate it on demand?

apps/mobile/src/components/settings/settings-list-item.tsx
```diff
@@ -15,7 +15,7 @@ interface ItemBaseProps {
   type?: ItemType;
   title: string;
   caption?: string;
-  icon?: ReactNode;
+  icon?: ReactElement;
```
pete-watters, line 18: I guess this is enforcing it must be a component?

pete-watters (approved):
Great work here @tigranpetrossian on not an easy task! 

I added some discussion points but no blockers to merge for me. Nice job 🚀

packages/ui/src/components/avatar/avatar.web.tsx
tigranpetrossian: I'm afraid not–this is a web/panda specific type signature.

apps/mobile/src/components/settings/settings-list-item.tsx
tigranpetrossian: Exactly–a narrower type. ReactNode is a broad superset that includes this, + `string | number | boolean | null | undefined` and probably more, none of which is applicable here. We could technically narrow this even further limiting which elements exactly can be passed (i.e. just our icons), but too much strictness is often a footgun.

packages/ui/src/components/avatar/brc20-avatar-icon.web.tsx
tigranpetrossian: I'd frankly even take this further and move those into their respective apps altogether, I don't think this is the right place for these domain-specific icons to live. These also are sort of special cases, we're abusing the `icon` prop to display what in reality is an image that should be passed into Avatar at the usage point.
Had plans to do this initially, but with all the extension changes the PR was snowballing already.
Can we make an issue on improved handling token images? Might be worth including the fallback/image alt improvements too.

packages/ui/src/components/avatar/btc-avatar-icon.native.tsx
tigranpetrossian: 💯 See my [comment here](https://github.com/leather-io/mono/pull/858#discussion_r1950284816)

packages/ui/src/components/avatar/avatar.web.tsx
tigranpetrossian: I find this kind of separation a bit arbitrary to be honest. Extracting things like styles, helpers, types that only have one consumer into separate files does more harm then good: 
1. More diffs across files for simple changes 
2. Unnecessary exports floating around the app/packages slowing down the compiler, messing with editor auto-import suggestions 
3. Potential circular imports, when sharing dependencies between these modules, which causes to split into even more files deepening the issue.

I feel like excessive file organization is a bit like sweeping under the rug–gives us false sense of decoupling for things that are inherently coupled.

pete-watters: OK 👍 I just don't like working with massive files 😄

kyranjamie (approved):
Great work @tigranpetrossian, thanks for writing up details on component conventions 👍🏼

packages/ui/src/components/avatar/brc20-avatar-icon.web.tsx
pete-watters: Sounds great 👍 a future improvement issue is fine

tigranpetrossian: Done: [LEA-2122](https://linear.app/leather-io/issue/LEA-2122/issues-with-token-avatar-usage)

PR #865, opened 2025-02-11
feat: init web app, LEA-2115

alter-eggo:
Set up of new web app using https://developers.cloudflare.com/pages/framework-guides/deploy-a-remix-site/

will fix configs in sep prs


apps/web/app/tailwind.css
```diff
@@ -0,0 +1,12 @@
+@tailwind base;
```
kyranjamie, line 1: Saw your comment on doing this another time, but for removing code, suggest we remove this in this PR, so it never enters the history

kyranjamie (commented):
Thanks @alter-eggo excited to get this started

apps/web/README.md
```diff
@@ -0,0 +1,47 @@
+# Welcome to Remix + Cloudflare!
```
camerow, line 1: Would be nice to commit a basic version of our own readme here - describe the app, present it to the community at large and describe what we are doing here.

apps/web/tsconfig.json
```diff
@@ -0,0 +1,36 @@
+{
+  "include": [
```
camerow, line 2: We have:

  
```
"extends": "../../packages/tsconfig-config/tsconfig.base",
```

do we want to use that like in `mobile`?

apps/web/README.md
alter-eggo: have some description in mind?

kyranjamie:
Catching up on the latest here, and think we should hang fire for now. [By the sounds of this post](https://remix.run/blog/incremental-path-to-react-19) Remix as is scaffolded here is deprecated, and there are [migration guides](https://reactrouter.com/upgrading/remix) for how to move to "Remix 3" otherwise known as React Router v7.

pete-watters:
> Catching up on the latest here, and think we should hang fire for now. [By the sounds of this post](https://remix.run/blog/incremental-path-to-react-19) Remix as is scaffolded here is deprecated, and there are [migration guides](https://reactrouter.com/upgrading/remix) for how to move React Router v7.

Perhaps as it's still so early stages a POC using [Svelte Kit](https://svelte.dev/docs/kit/introduction) is worth a consideration to trial it?

kyranjamie:
I'm sure SvelteKit's great, but it is a big decision to leave the React ecosystem when we have 2 other React-based products. Do you think we should explore this more @pete-watters ?

pete-watters:
> I'm sure SvelteKit's great, but it is a big decision to leave the React ecosystem when we have 2 other React-based products. Do you think we should explore this more @pete-watters ?

I think it's worth creating a demo MVP if we are checking out different setups and unhappy with React Router anyway which has annoyed us in the past. Here is a good [discussion of the differences](https://hygraph.com/blog/svelte-vs-react). 

You can [blend React and Svelte](https://github.com/bfanger/svelte-preprocess-react) and although it's not an ideal solution, it would help us leverage our existing codebase. [More here](https://bfanger.medium.com/combining-react-and-svelte-in-a-single-app-interop-6f78aed96ce2). 

They have a CLI too for scaffolding and we could try out a demo with their version of [routable modals](https://svelte.dev/docs/kit/shallow-routing)? They have an adapter to deploy to [Cloudflare also](https://svelte.dev/docs/kit/adapter-cloudflare).

We are committed to React so perhaps I'm just adding un-welcome noise but if we want to move more away from React in general we could try out a demo of Svelte to see if we like it. Svelte seems better as it's closer to the native browser implementation and builds on native APIs and if the DX is a lot better progress could be fast or we could hate it and appreciate React more.

camerow:
@alter-eggo Thank you for kickstarting this but I think we can close it in favor of the RR7 work @kyranjamie is doing.

PR #874, opened 2025-02-13
feat(sdk): leather sdk

kyranjamie:
This work builds on the [RPC refactor](https://github.com/leather-io/mono/pull/866) and a conversation with @fbwoolf and @markmhendrickson.

Currently, we only publish RPC types and schemas to developers. With a clean object of known methods, it's fairly straightforward to generate a JS client with methods users can call, which is an improvement on having to access the provider directly from the global object. Parameter and result types are inferred from the RPC spec.

How much we want to expand and maintain this it still in question. Perhaps few will use this, in favour of connect or other wallet wrapping libraries. But, at least in its current bare bones state, it'd be minimal effort to maintain, and might be an interesting package to publish, even if only used for internal use, or docs/demos

```ts
import { createLeatherClient } from '@leather.io/sdk';

const client = createLeatherClient();

const addresses = await client.getAddresses();
```

It also supports other providers with this API, though of course it would have to be Leather-compatible.

```ts
createLeatherClient({ 
  getProvider() {
    return window.AnotherProvider;
  } 
})
```

**Update 2025-03-06**: Working on the web app, I could really do with a client. I've made some small amendments to get the types working, but otherwise think this is in a workable state to publish.

I've added the [`ts-expect`](https://github.com/TypeStrong/ts-expect) package to verify that the types we derive as function params, match the originally defined ones. This was fairly tricky when params can be one of 3 states: defined, optional, undefined. Consider `getAddresses`, which can be called `client.getAddresses()` or `client.getAddresses({ network: 'mainnet' })`.

Note that method names are turned into consistent camelcase, so we use `client.stxSignMessage()` not `client.stx_signMessage`.

image

### DIFF
--- a/packages/rpc/src/index.ts
+++ b/packages/rpc/src/index.ts
@@ -64,41 +64,42 @@ export const endpoints = {
   signPsbt,
   signMessage,
   sendTransfer,
-};
+} as const;
 
 type EndpointMap = (typeof endpoints)[keyof typeof endpoints];
 
-export type LeatherRpcMethodMap = {
+/**
+ * Request map keyed by exact method name e.g. `stx_signMessage`
+ */
+export type RpcEndpointMap = {
   [E in EndpointMap as E['method']]: {
     request: z.infer<E['request']>;
     response: z.infer<E['response']>;
   };
 };
 
-export type RpcRequests = ValueOf<LeatherRpcMethodMap>['request'];
+/** @deprecated */
+export type LeatherRpcMethodMap = RpcEndpointMap;
 
-export type RpcResponses = ValueOf<LeatherRpcMethodMap>['response'];
+export type RpcRequests = ValueOf<RpcEndpointMap>['request'];
 
-export type RpcMethodNames = keyof LeatherRpcMethodMap;
+export type RpcResponses = ValueOf<RpcEndpointMap>['response'];
+
+export type RpcMethodNames = keyof RpcEndpointMap;
 
 export interface RequestFn {
   <
     T extends RpcMethodNames,
-    P extends LeatherRpcMethodMap[T]['request'] extends { params: infer P } ? P : never,
+    P extends RpcEndpointMap[T]['request'] extends { params?: infer P } ? P : never,
   >(
     arg: T,
     params: P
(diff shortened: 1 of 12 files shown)


fbwoolf:
Love it 🚀

packages/sdk/src/client.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+describe('Leather SDK', () => {
+  const client = createLeatherClient();
+
+  const clientMethods = Object.keys(client);
+  test('that it has a method for each endpoint', () => {
+    Object.values(endpoints).forEach(endpoint => clientMethods.includes(endpoint.method));
+  });
+
+  test('Types', async () => {
+    // Optional params, compiler should be happy no parameters
```
pete-watters, line 20: Maybe you could move these comments into `test('` declarations instead? Then you'd get more feedback on what failed

e.g. 
```
test('Types: compiler should be happy no parameters', async () => {
...

test('Types: Sign message should cause a type error', async () => {
```

alexp3y (approved):
Nice. This already looks great as-is to access the Leather provider RPC typings. IMO in the future we shouldn't shy away from adding higher-level abstractions to make this as useful / usable as possible.

packages/rpc/src/index.ts
```diff
@@ -64,41 +64,39 @@ export const endpoints = {
...
-export type LeatherRpcMethodMap = {
+export type RpcEndpointMap = {
   [E in EndpointMap as E['method']]: {
     request: z.infer<E['request']>;
     response: z.infer<E['response']>;
   };
 };
 
-export type RpcRequests = ValueOf<LeatherRpcMethodMap>['request'];
+/** @deprecated */
```
pete-watters, line 78: Is this a temporary deprecation? 

Will you refactor the extension to remove `LeatherRpcMethodMap`?

pete-watters (approved):
Nice work 👍

kyranjamie:
More feedback welcome from @edgarkhanzadian @tigranpetrossian @fbwoolf @camerow @alter-eggo — I'm merging for now though to move on with other work

alter-eggo (approved):

PR #879, opened 2025-02-17
feat: move leather provider to a separate package in monorepo [LEA-2142]

edgarkhanzadian:
Tested using both extension and mobile, works well. As soon as this one is merged i'll open PRs for extension and mobile.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -46,6 +46,7 @@
     "@leather.io/constants": "workspace:*",
     "@leather.io/crypto": "workspace:*",
     "@leather.io/models": "workspace:*",
+    "@leather.io/provider": "workspace:*",
     "@leather.io/query": "workspace:*",
     "@leather.io/rpc": "workspace:*",
     "@leather.io/services": "workspace:*",
--- a/packages/provider/src/add-leather-to-providers.ts
+++ b/packages/provider/src/add-leather-to-providers.ts
@@ -1,0 +1,52 @@
+import type { RpcMethodNames } from '@leather.io/rpc';
+
+interface WalletProvider {
+  id: string;
+  name: string;
+  icon: string;
+  webUrl?: string;
+  chromeWebStoreUrl?: string;
+  mozillaAddOnsUrl?: string;
+  googlePlayStoreUrl?: string;
+  iOSAppStoreUrl?: string;
+  methods?: string[];
+}
+
+// https://wbips.netlify.app/wbips/WBIP004
+export function addLeatherToProviders() {
+  const win = window as unknown as Window & { btc_providers?: WalletProvider[] };
+
+  if (!win.btc_providers) win.btc_providers = [];
+
+  win.btc_providers.push({
+    id: 'LeatherProvider',
+    name: 'Leather',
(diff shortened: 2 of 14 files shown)


apps/mobile/package.json
```diff
@@ -45,6 +45,7 @@
     "@leather.io/bitcoin": "workspace:*",
     "@leather.io/constants": "workspace:*",
     "@leather.io/crypto": "workspace:*",
+    "@leather.io/leather-provider": "workspace:*",
```
kyranjamie, line 48: Let's not call leather+leathe

Suggest just `provider` or `injected-provider`?

packages/leather-provider/scripts/prepare-script.cjs
```diff
@@ -0,0 +1,3 @@
+const { updateInjectedProvider } = require('./update-injected-provider.cjs');
```
kyranjamie, line 1: why not esm?

packages/leather-provider/scripts/update-injected-provider.cjs
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+const path = require('path');
+
```
kyranjamie, line 3: Node can all be esm these days

packages/leather-provider/src/index.ts
```diff
@@ -0,0 +1,189 @@
...
+const COMMIT_SHA = 'TEST_COMMIT_SHA';
+const VERSION = 'VERSION';
+
+function isRN(): boolean {
+  return !!(window as any)?.ReactNativeWebView;
+}
+
+enum DomEventName {
+  request = 'request',
+}
```
kyranjamie, line 19: Given this is the only one now, let's avoid the enum

packages/leather-provider/src/index.ts
```diff
@@ -0,0 +1,189 @@
...
+      jsonrpc: '2.0',
+      id,
+      method,
+      params: (params ?? {}) as any,
+    };
+    if (isRN()) {
+      (window as any).ReactNativeWebView.postMessage(JSON.stringify(rpcRequest)); // RN
+    } else {
+      document.dispatchEvent(new CustomEvent(DomEventName.request, { detail: rpcRequest }));
+    }
```
kyranjamie, line 89: Can we pass in this as a callback, rather than having ternary logic? What if we add a third platform, do we keep extending? 

IMO here we should call `dispatchCallback(rpcRequest)` and pass in to a method to dispatch on each respective platform

packages/provider/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+    };
+
+    dispatchCallback(rpcRequest);
+
+    return new Promise((resolve, reject) => {
+      function handleMessage(event: MessageEvent<WalletResponses>) {
+        const response =
+          typeof event.data === 'object' ? event.data : JSON.parse(event.data as any);
+        if (response.id !== id) return;
+        window.removeEventListener('message', handleMessage);
```
kyranjamie, line 99: isn't this web specific? and would this also need to be a handler/eventemitter?

packages/provider/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+function dispatchCallback(rpcRequest: RpcRequests) {
+  // if running in mobile app
+  if (isRN()) {
+    (window as any).ReactNativeWebView.postMessage(JSON.stringify(rpcRequest)); // RN
+    return;
+  }
+  // if running in extension
+  document.dispatchEvent(new CustomEvent('request', { detail: rpcRequest }));
+}
```
kyranjamie, line 25: This isn't quite what I meant. Here, we're handling platform-specific behaviours _inside_ `@leather.io/providers`.

Why not pass in the platform-specific stuff as callbacks? e.g.

```ts
// extension repo
import { initInpageProvider } from '@leather.io/providers';

initInpageProvider({ 
  onDispatch(event) {
    document.dispatchEvent(new CustomEvent('request', { detail: rpcRequest }));
  }
})

// mobile repo
import { initInpageProvider } from '@leather.io/providers';
initInpageProvider({ 
  onDispatch(event) {
    (window as any).ReactNativeWebView.postMessage(JSON.stringify(rpcRequest));
  }
})
// some bundler script that takes output of above and stringifies it
```

now, providers doesn't know about platform concerns

packages/provider/src/rpc-methods.ts
```diff
@@ -0,0 +1,6 @@
+import type { ValueOf } from '@leather.io/models';
+import { type LeatherRpcMethodMap } from '@leather.io/rpc';
+
+export type WalletResponses = ValueOf<LeatherRpcMethodMap>['response'];
+
+export type WalletMethodNames = keyof LeatherRpcMethodMap;
```
kyranjamie, line 6: This is already exported from rpc as RpcMethodNames

packages/provider/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+
+type Platform = 'mobile' | 'extension';
+
+export function initInpageProvider({
+  onDispatch,
+  env,
+}: {
+  onDispatch(rpcRequest: RpcRequests): void;
+  env: { branch: string; commitSha: string; version: string; platform: Platform };
+}) {
```
kyranjamie, line 16:
```suggestion
interface InitInpageProviderArgs {
  onDispatch(rpcRequest: RpcRequests): void;
  env: { branch: string; commitSha: string; version: string; platform: Platform };
}
export function initInpageProvider({ onDispatch, env }: InitInpageProviderArgs) {
```

🧘🏼‍♂️

packages/provider/src/add-leather-to-providers.ts
```diff
@@ -0,0 +1,38 @@
...
+    chromeWebStoreUrl:
+      'https://chromewebstore.google.com/detail/leather/ldinpeekobnhjjdofggfgjlcehhmanlj',
+
+    methods: [
+      'getAddresses',
+      'signMessage',
+      'sendTransfer',
+      'signPsbt',
+      'stx_signMessage',
+      'stx_signTransaction',
```
kyranjamie, line 35: Can we pull this from RPC, as this is outdated

`endpoints.map(endpoint => endpoint.methods`

kyranjamie (approved):

PR #880, opened 2025-02-17
feat: connect leather api client to backend

alexp3y:
This PR connects the existing `LeatherApiClient` (and by extension, the services layer) to the now live Leather Workers API (staging version), greatly reducing and simplifying our API client infrastructure.

Clients Removed:
* Coincap
* Coingecko
* Binance
* ALEX

Remaining clients:
* Hiro
* Best In Slot
* Leather API

Now that we have a single endpoint serving native asset crypto prices, I also simplified the existing `MarketDataService` logic for STX and BTC (and its usage in `mobile`).


apps/mobile/src/queries/market-data/stx-market-data.query.ts
```diff
@@ -1,22 +1,8 @@
...
-    staleTime: oneMinInMs,
-    gcTime: oneMinInMs,
-  } as const;
-}
+import { useMarketDataQuery } from './market-data.query';
 
 // LEA-1761 FIXME: this query returns the wrong precision - price in dollars rather than cents
 export function useStxMarketDataQuery() {
-  return useQuery(createStxMarketDataQueryOptions());
+  return useMarketDataQuery(stxCryptoAsset);
```
pete-watters, line 7: I think we can remove the `// LEA-1761 FIXME: th` now as it was solved elsewhere

pete-watters (approved):
Nice work! 👍

PR #885, opened 2025-02-18
chore: add btc transaction validation and branded type for address

pete-watters:
This PR:
- adds more validation to the `Send BTC` flow on `mobile` 
- does some refactoring in `packages/bitcoin` 
- gives an overview of using a [Branded type](https://egghead.io/blog/using-branded-types-in-typescript) for `BitcoinAddress` - we discussed this on Slack - [here](https://trustmachines.slack.com/archives/C05LAP952E7/p1739537734726239)

~~Opened as a **DRAFT** now as it needs E2E maestro tests to be completed~~

Opening up for full review as the whole maestro suite is failing and needs to be fixed - [LEA-2149](https://linear.app/leather-io/issue/LEA-2149/maestro-tests-failing-and-need-to-be-added-to-ci)

Demo:

https://github.com/user-attachments/assets/d70690ab-2c59-474d-80a1-3f9d2e9ec6bd


pete-watters:
@kyranjamie @tigranpetrossian : Opening this now as a DRAFT to demo what we discussed [here](https://trustmachines.slack.com/archives/C05LAP952E7/p1739537734726239) about using a branded type.

I find it useful but can also reverse the change if needed. 

The PR itself is functional but missing E2E tests which I will add next

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
```diff
@@ -12,56 +13,76 @@ import {
...
+}: SendFormBtcSchema): ParsedSendFormValues {
+  const amount = createMoneyFromDecimal(new BigNumber(inputAmount), 'BTC');
   return {
-    amount: createMoneyFromDecimal(new BigNumber(values.amount), 'BTC'),
+    amount,
     recipients: [
       {
-        address: values.recipient,
-        amount: createMoneyFromDecimal(new BigNumber(values.amount), 'BTC'),
+        address: createBitcoinAddress(recipient),
```
pete-watters, line 53: Here I create a type `BitcoinAddress` using the raw form value but it could be better to add that validation in `SendFormBtcSchema`.

We have more UI changes to make to the forms though so am trying to impact those as little as I can. 

I think it's nicer for users to have the validation triggered on the `Review` button then after entering the value as we show a `Toast`

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
```diff
@@ -12,56 +13,76 @@ import {
...
-  const { account } = route.params;
   const { displayToast } = useToastContext();
+  const {
+    networkPreference: {
+      chain: {
+        bitcoin: { bitcoinNetwork },
+      },
+    },
+  } = useSettings();
+  const payer = createBitcoinAddress(address);
```
pete-watters, line 77: `address` is a route `param` so it would likely make more sense to set the `address` as `BitcoinAddress` there too

packages/bitcoin/src/utils/bitcoin.utils.ts
```diff
@@ -100,7 +100,10 @@ export function decodeBitcoinTx(tx: string): ReturnType<typeof btc.RawTx.decode>
   return btc.RawTx.decode(hexToBytes(tx));
 }
 
-export function getAddressFromOutScript(script: Uint8Array, bitcoinNetwork: BtcSignerNetwork) {
+export function getAddressFromOutScript(
+  script: Uint8Array,
+  bitcoinNetwork: BtcSignerNetwork
+): BitcoinAddress | string {
```
pete-watters, line 106: This function sometimes returns an empty string / unknown so it needs to be `BitcoinAddress | string`. Maybe that's something we need to fix?

```
case "unknown":
      return "unknown";
    case "tr_ms":
    case "tr_ns":
    default:
      return "";
```

packages/utils/src/validation/error.ts
```diff
@@ -0,0 +1 @@
+export type TransactionErrorKey = 'InvalidAddress' | 'InsufficientFunds' | 'InvalidNetworkAddress';
```
pete-watters, line 1: Adding this to share keys with `Stacks`

apps/mobile/src/features/send/send-form/validation/btc.validation.ts
```diff
@@ -0,0 +1,37 @@
+import { t } from '@lingui/macro';
+
+import { BitcoinErrorKey, minSpendAmountInSats } from '@leather.io/bitcoin';
+import { match, satToBtc } from '@leather.io/utils';
+
+const minimumBtcSpend = satToBtc(minSpendAmountInSats);
+
```
pete-watters, line 7: We need to map the error key to a lingui crowdin key. 

I wonder should we investigate if we can move this into `packages/bitcoin` / somewhere shared and then consume it from the extension to have translated errors

tigranpetrossian (approved):
@pete-watters Thanks so much for the validation code and the tests–this isn't something I would be even remotely comfortable tackling. My only issue is with when exactly the assertion for valid address happens, but it isn't a blocker.

packages/utils/src/validation/error.ts
kyranjamie: I'm not sure details like this should go in utils, in favour of utils being a home for js helpers, and more generic, non-domain specific code.

Could we have a ValidationService, or something?

packages/bitcoin/src/validation/amount-validation.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (!spendableBtc) return false;
+  const desiredSpend = new BigNumber(amount);
+  if (desiredSpend.isGreaterThan(spendableBtc)) return false;
+  return true;
+}
+
+interface isBtcMinimumSpendArgs {
+  amount: number;
+}
+export function isBtcMinimumSpend({ amount }: isBtcMinimumSpendArgs) {
```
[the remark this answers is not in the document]
pete-watters: Good catch

fbwoolf (approved):
Nice refactor @pete-watters 👍

packages/utils/src/validation/error.ts
pete-watters: I'll move it to the bitcoin package now as that's the only place using it. I could add it to `models`? I just want to share the type across packages.

I can investigate a validation service. I have a few other functions to migrate but I think thinks like `isValidPrecision` can live in `utils` and we may not need a full service.

pete-watters:
@tigranpetrossian @kyranjamie : I updated to code to add runtime validation. Let me know if it's looking better.

That broke a lot of the unit tests so I think it shows it's an improvement. I'll fix those ASAP

pete-watters:
> @tigranpetrossian @kyranjamie : I updated to code to add runtime validation. Let me know if it's looking better.

Time waits for no man. Please let me know if you have concerns and I can update my code in a subsequent PR

pete-watters:
> Sorry @pete-watters, but yes throwing when initialising the branded type makes sense

Thanks! And no problem, just want to keep momentum but I will always retrospectively make fixes if needed

PR #887, opened 2025-02-19
feat: browser rpc [LEA-2210] [LEA-2209] [LEA-2223]

edgarkhanzadian:
- Implement a basic browser widget for entering the browser
- Implement RPC handling of signPsbt and getAddresses
- Implement dApp redux slice for saving dApp data
- Implement release feature flag for browser work

https://github.com/user-attachments/assets/cc438348-b23f-40c6-ad23-4c92f34b9e54


apps/mobile/src/components/browser/approver-sheet/browser-approver.tsx
```diff
@@ -0,0 +1,41 @@
...
+  message: BrowserMessage;
+  sendResult(result: object): void;
+  origin: string;
+  app: App;
+  closeApprover(): void;
+}
+
+export function BrowserApprover(props: BrowserApproverProps) {
+  switch (props.message?.method) {
+    case 'getAddresses':
```
kyranjamie, line 17: `getAddresses.method`

apps/mobile/src/components/browser/approver-sheet/sign-psbt.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useBitcoinAccounts } from '@/store/keychains/bitcoin/bitcoin-keychains.read';
+import { z } from 'zod';
+
+import { useCurrentNetworkState } from '@leather.io/query';
+import { LeatherRpcMethodMap, signPsbt } from '@leather.io/rpc';
+
+import { infuseBip32 } from './utils';
+
+interface SignPsbtApproverProps {
+  message: LeatherRpcMethodMap['signPsbt']['request'];
```
kyranjamie, line 12:
```suggestion
  message: RpcRequest<typeof signPsbt>
```

apps/mobile/src/components/browser/approver-sheet/sign-psbt.tsx
```diff
@@ -0,0 +1,45 @@
...
+    });
+    return (
+      <PsbtSigner
+        broadcast={props.message.params.broadcast}
+        allowedSighash={props.message.params.allowedSighash}
+        signAtIndex={infusedPsbt.signAtIndex}
+        psbtHex={infusedPsbt.psbtHex}
+        network={props.message.params.network}
+        onBack={props.closeApprover}
+        onSuccess={({ hex, txid }: z.infer<(typeof signPsbt)['result']>) => {
```
kyranjamie, line 38:
```suggestion
        onSuccess={({ hex, txid }: RpcResult<typeof signPsbt>) => {
```

packages/provider/src/mobile.ts
```diff
@@ -11,3 +11,6 @@ initInpageProvider({
     version: 'replace_version',
   },
 });
+// note: this is required, or you'll sometimes get silent failures
+// eslint-disable-next-line @typescript-eslint/no-unused-expressions
+true;
```
kyranjamie, line 16: source?

apps/mobile/src/app/(home)/developer-console/index.tsx
```diff
@@ -86,20 +85,11 @@ export default function DeveloperConsoleScreen() {
...
-                accountIndex={0}
-                message={getAddressesMessage}
-                sendResult={() => setGetAddressesMessage(null)}
-              />
-            )}
-          </WalletLoader>
-        )}
-      </WalletFingerprintLoader>
+      <ApproverSheet
+        origin={'https://leather.io'}
```
pete-watters, line 89: Can you use a constant for this URL?

apps/mobile/src/components/browser/approver-sheet/get-addresses.tsx
```diff
@@ -0,0 +1,169 @@
+import { useRef, useState } from 'react';
```
pete-watters, line 1: Can you split this into a layout file and hooks file?

apps/mobile/src/components/browser/browser-empty-state.tsx
```diff
@@ -122,6 +65,18 @@ export function BrowserEmptyState({
   const { top, bottom } = useSafeAreaInsets();
   const [currentTab, setCurrentTab] = useState<CurrentTab>('suggested');
   const theme = useTheme<Theme>();
+  const { list: connectedApps } = useApps('connected');
+  const { list: recentApps } = useApps('recently_visited');
+
+  const tabMap: Record<CurrentTab, Pick<App, 'name' | 'origin' | 'icon'>[]> = {
+    connected: connectedApps,
+    recent: recentApps,
+    suggested: [
```
pete-watters, line 74: Maybe this can be in a config file?

apps/mobile/src/components/browser/browser-widget.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const wallets = useWallets();
+  const theme = useTheme<Theme>();
+
+  const router = useRouter();
+  const { totalBalance } = useTotalBalance();
+  // TODO LEA-1726: handle balance loading & error states
+  if (totalBalance.state !== 'success') return;
+
+  return (
+    <>
```
pete-watters, line 27: Extra un-needed fragment here `<>`

apps/mobile/src/components/browser/browser-widget.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <Widget.Body>
+          <ScrollView
+            horizontal
+            showsHorizontalScrollIndicator={false}
+            contentContainerStyle={{
+              gap: theme.spacing['2'],
+              paddingHorizontal: theme.spacing['5'],
+            }}
+            style={{
+              // prevent card shadows being cut off
```
pete-watters, line 49: Is this something we need to fix globally?

apps/mobile/src/features/accounts/account-selector/account-selector-sheet.tsx
```diff
@@ -1,18 +1,21 @@
...
 
 import { SheetRef } from '@leather.io/ui/native';
 
 import { AccountSelectorSheetLayout } from './account-selector-sheet.layout';
 
-export function AccountSelectorSheet({ sheetRef }: { sheetRef: RefObject<SheetRef> }) {
+export function AccountSelectorSheet({
+  sheetRef,
+  onAccountPress,
+}: {
```
pete-watters, line 14: Can you move this to an `interface`?

pete-watters (commented):
Nice work here @edgarkhanzadian . It's pretty good overall but the code could be cleaned up, particularly to use separate interfaces and types. 

Is there a demo of it working to see?

packages/provider/src/mobile.ts
camerow: That is so strange - kind of annoying the docs have that comment but no real explanation on what's going on.

apps/mobile/src/store/apps/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  accountId: z.string(),
+});
+const recentlyVisitedAppSchema = z.object({
+  type: z.literal('recently_visited'),
+});
+
+const baseAppSchema = z.object({
+  origin: z.string(),
+  name: z.string(),
+  icon: z.string(),
```
kyranjamie, line 14: Where/how is this used?

apps/mobile/src/store/apps/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  type: z.literal('connected'),
+  accountId: z.string(),
+});
+const recentlyVisitedAppSchema = z.object({
+  type: z.literal('recently_visited'),
+});
+
+const baseAppSchema = z.object({
+  origin: z.string(),
+  name: z.string(),
```
kyranjamie, line 13: what is the name? We show origin only, or should derive it from

apps/mobile/src/store/apps/utils.ts
kyranjamie: > Though I am thinking about changing the name to the homepage (leather.io) title.
So for leather.io it should be "Leather Bitcoin Wallet - Bitcoin for the Rest of Us"

Don't think we should do this, we want to pull as little as possible from the app as possible for security reasons.

markmhendrickson:
@edgarkhanzadian is there a way to test this PR via Expo?

packages/bitcoin/src/utils/bitcoin.utils.ts
```diff
@@ -103,33 +104,37 @@ export function decodeBitcoinTx(tx: string): ReturnType<typeof btc.RawTx.decode>
...
-      return btc.p2pk(outputScript.pubkey, bitcoinNetwork).address ?? '';
+      return createBitcoinAddress(btc.p2pk(outputScript.pubkey, bitcoinNetwork).address ?? '');
     case 'unknown':
-      return 'unknown';
+      return createBitcoinAddress('unknown');
     case 'tr_ms':
     case 'tr_ns':
     default:
-      return '';
+      return createBitcoinAddress('');
```
kyranjamie, line 137: Doesn't this throw an error? This isn't a bitcoin address

apps/mobile/src/store/keychains/bitcoin/bitcoin-keychains.read.ts
```diff
@@ -132,3 +132,10 @@ export function useBitcoinAddresses() {
     [accounts]
   );
 }
+
+export function deriveBitcoinAccountFromAddress(
```
kyranjamie, line 136:
```suggestion
export function deriveBitcoinAddressFromAccount(
```

apps/mobile/src/store/apps/utils.ts
kyranjamie: But on the icon, we derive the icon from the origin via the google api no? So why the record here?

apps/mobile/src/store/apps/apps.read.ts
```diff
@@ -0,0 +1,33 @@
+import { createSelector } from '@reduxjs/toolkit';
+
+import { RootState } from '..';
+import { useAppSelector } from '../utils';
+import { appsAdapter } from './apps.write';
+import { AppType } from './utils';
+
+const selectors = appsAdapter.getSelectors((state: RootState) => state.apps);
+
+export function selectApps(type?: AppType) {
```
kyranjamie, line 10: Semantically its confusing that an `AppType`  is `'recently_visited'`.

apps/mobile/src/features/psbt-signer/psbt-signer.tsx
```diff
@@ -82,11 +91,22 @@ export function PsbtSigner(props: PsbtSignerProps) {
   return <BasePsbtSigner feeRates={feeRates} {...props} />;
 }
 
+function getPsbtNetwork(network: string) {
```
kyranjamie, line 94:
```suggestion
function getPsbtNetwork(network: BitcoinNetworkMods) {
```

apps/mobile/src/features/approver/components/approver-permissions.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { Box, Cell, CheckmarkIcon, Text } from '@leather.io/ui/native';
+
+type Permissions = 'view_balance_activity' | 'request_approval';
+
+interface ApproverPermissionsProps {
+  permissions: Permissions[];
+}
+
+export function ApproverPermissions({ permissions }: ApproverPermissionsProps) {
```
kyranjamie, line 11: Does this component need to be in any way associated to Approver? Why not just PermissionsList or something, outside Approver feature

kyranjamie (changes requested):
Looking good @edgarkhanzadian. Requesting changes only for the important point regarding persisting accountid and fingerprint.

apps/mobile/src/store/keychains/bitcoin/bitcoin-keychains.read.ts
kyranjamie: also can we say index zero address here, because the name infers 1:1 relationship between account and address.

packages/bitcoin/src/utils/bitcoin.utils.ts
kyranjamie: Empty string is an unexpected value though, and not a valid bitcoin address (that BitcoinAddress ought to throw an error if it's given).

apps/mobile/src/components/browser/approver-sheet/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface InfuseBip32Props {
+  psbtHex: string;
+  networkMode: BitcoinNetworkModes;
+  bitcoinAccounts: ReturnType<typeof useBitcoinAccounts>['list'];
+  accountId: string;
+  signAtIndex: number | number[] | undefined;
+}
+
+export function infuseBip32({
```
kyranjamie, line 42: Perhaps `addBip32DerivationToInputs` would be a clearer name here?

apps/mobile/src/components/browser/approver-sheet/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  const signAtIndex = normalizeSignAtIndex(_signAtIndex) ?? [];
+  const tx = getPsbtAsTransaction(psbtHex);
+
+  const inputs = getPsbtTxInputs(tx);
+  inputs.forEach((input, idx) => {
+    if (input.bip32Derivation || input.tapBip32Derivation) return;
+    const bitcoinNetwork = getBtcSignerLibNetworkConfigByMode(networkMode);
+    const addr = getBitcoinInputAddress(input, bitcoinNetwork);
+    const paymentType = inferPaymentTypeFromAddress(addr);
+    const bitcoinAccount = deriveBitcoinAccountFromAddress(bitcoinAccounts, addr);
```
kyranjamie, line 58: I wonder if this step is necessary. Here, we're looking through _all_ the accounts a wallet has, looking to see if the address matches. `deriveBitcoinAccountFromAddress` also calls the derive fn which can be expensive with many accounts.

We know from the connection history of the app, that it only knows the keys of a given fingerprint/accountId. Can we not then assume that we need to use the account it connected with?

apps/mobile/src/components/browser/approver-sheet/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+        signAtIndex.push(idx);
+      } else if (accAccountId !== accountId && signAtIndex.includes(idx)) {
+        throw new Error('PSBT asks to sign an input that is not ours');
+      }
+    }
+
+    const payer = bitcoinAccount?.derivePayer({ addressIndex: 0 });
+    if (payer) {
+      if (paymentType === 'p2tr') {
+        tx.updateInput(idx, {
```
kyranjamie, line 74: Doesn't his try and update all the inputs within a psbt? if a psbt has 3 inputs, and the user passes `signAtIndex: [0,1]` then we only want to adjust 0 and 1, and this would do the same for 2 no?

kyranjamie: I think it's important that we only modify the inputs we're asked to. This is an easy fix, right? We just iterate `signAtInputs` instead?

apps/mobile/src/components/browser/approver-sheet/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  payerToTapBip32Derivation,
+} from '@leather.io/bitcoin';
+import {
+  extractAccountIndexFromDescriptor,
+  extractFingerprintFromDescriptor,
+} from '@leather.io/crypto';
+import { BitcoinNetworkModes } from '@leather.io/models';
+import { RpcRequests } from '@leather.io/rpc';
+
+export function normalizeSignAtIndex(signAtIndex: number | number[] | undefined) {
```
pete-watters, line 24: Can you use interfaces for the types?

apps/mobile/src/components/browser/browser-widget.tsx
```diff
@@ -0,0 +1,67 @@
...
+            // prevent card shadows being cut off
+            overflow: 'visible',
+          }}
+        >
+          <BrowserCard
+            onPress={() => {
+              router.navigate(AppRoutes.Browser);
+            }}
+            title={t({
+              id: 'browser.widget.zest.title',
```
pete-watters, line 55: Don't forget to add these keys in crowdin

packages/bitcoin/src/utils/bitcoin.utils.ts
pete-watters: `createBitcoinAddress('')` should throw an error as there is a check for not empty in `isValidBitcoinAddress` so this seems like a bug

pete-watters: createBitcoinAddress('unknown') should also throw an error here too

apps/mobile/src/store/keychains/bitcoin/bitcoin-keychains.read.ts
```diff
@@ -132,3 +132,11 @@ export function useBitcoinAddresses() {
...
```
kyranjamie, line 136: can we rename this function because there's no derivation. `findAccountFromAddress`? `drive` implies there is some derivation going on, and there isn't.

apps/mobile/src/components/browser/approver-sheet/get-addresses/get-addresses.tsx
```diff
@@ -0,0 +1,114 @@
...
+        publicKey: account.stacksAccount.publicKey,
+      },
+    });
+    const result = { addresses: keysToIncludeInResponse };
+
+    const response = getAddresses.response.parse({
+      result,
+      id: props.message.id,
+      jsonrpc: props.message.jsonrpc,
+    });
```
kyranjamie, line 81: Can use `createRpcSuccessResponse` here

kyranjamie (approved):

pete-watters:
Nice work @edgarkhanzadian 👏 It's looking great!

packages/bitcoin/src/psbt/psbt-outputs.ts
```diff
@@ -36,9 +35,11 @@ export function getParsedOutputs({
         // logger.error('Output has no script');
         return;
       }
-      const outputAddress = createBitcoinAddress(
-        getAddressFromOutScript(output.script, bitcoinNetwork)
-      );
+      const outputAddress = getAddressFromOutScript(output.script, bitcoinNetwork);
+      if (outputAddress === null) {
+        throw new Error('PSBT output has unsupported bitcoin address');
```
fbwoolf, line 40: @edgarkhanzadian I think throwing this error here is going to break PSBTs for existing users. There was an edge case with 'unknown' where we needed to allow it. It also needs to be addressed in the extension bc we look for 'unknown'.

fbwoolf: I suppose if this is just used in mobile, might not be a huge deal? At first I thought this was shared.

fbwoolf: I think I made a commit here that will make it ok: https://github.com/leather-io/extension/pull/6177

PR #890, opened 2025-02-20
feat: implement btc outbound balance, closes LEA-1808, fixes LEA-2129

alexp3y:
This PR uses bitcoin transaction data from our API to determine / map outbound UTXOs, which are then used to calculate outbound BTC balance.

The incorporation of outbound balances (i.e. subtraction from available balance), might cause UX issues [described further here](https://www.notion.so/trustmachines/Mobile-BTC-Balance-UX-Issue-1a0d74b694c780329bc9f73cbeff7448?pvs=4). May need to migrate to using an alternative balance type in mobile UI balance components.

The PR also fixes the issue with BTC balances not appearing on testnet4.

### DIFF
--- a/packages/services/src/infrastructure/api/best-in-slot/best-in-slot-api.client.ts
+++ b/packages/services/src/infrastructure/api/best-in-slot/best-in-slot-api.client.ts
@@ -6,6 +6,7 @@ import { bitcoinNetworkModeToCoreNetworkMode } from '@leather.io/bitcoin';
 import { HttpCacheService } from '../../cache/http-cache.service';
 import { HttpCacheTimeMs } from '../../cache/http-cache.utils';
 import { RateLimiterService, RateLimiterType } from '../../rate-limiter/rate-limiter.service';
+import { selectBitcoinNetworkMode } from '../../settings/settings.selectors';
 import { SettingsService } from '../../settings/settings.service';
 import {
   bisBrc20MarketInfoSchema,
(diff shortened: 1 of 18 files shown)


packages/services/src/transactions/bitcoin-transactions.utils.ts
```diff
@@ -0,0 +1,17 @@
+import { LeatherApiBitcoinTransaction } from '../infrastructure/api/leather/leather-api.client';
```
pete-watters, line 1: Can you add some tests for this?

pete-watters (approved):
Nice work @alexp3y thanks for fixing `Testnet 4` too

packages/services/src/infrastructure/api/best-in-slot/best-in-slot-api.client.ts
```diff
@@ -132,28 +132,32 @@ export function createBestInSlotApiClient(
...
-              { params, signal }
+    return bitcoinNetworkModeToCoreNetworkMode(
+      settingsService.getSettings().network.chain.bitcoin.mode
+    ) !== 'mainnet'
+      ? []
+      : await cache.fetchWithCache(
+          [
+            'best-in-slot-runes-valid-outputs',
+            bitcoinNetworkModeToCoreNetworkMode(
+              settingsService.getSettings().network.chain.bitcoin.mode
```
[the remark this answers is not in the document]
kyranjamie: or user a selector fn?

packages/services/src/utxos/utxos.utils.spec.ts
```diff
@@ -317,3 +320,72 @@ describe('selectUniqueUtxoIds', () => {
     ]);
   });
 });
+
+describe('getOutboundUtxos', () => {
```
kyranjamie, line 324:
```suggestion
describe(getOutboundUtxos.name, () => {
```

minimises changes of divergence in case method ever gets renamed

kyranjamie (approved):

packages/services/src/utxos/utxos.utils.spec.ts
alexp3y: https://github.com/leather-io/mono/pull/890#discussion_r1963287123
image

PR #891, opened 2025-02-20
feat: add test app

edgarkhanzadian:
This is a barebone nextjs app made for testing some basic rpc calls for mobile browser. It's expected to be deleted once we implement web app dev console

At this point it supports:
- getAddresses
- signPsbt
- stx_transferStx

https://github.com/user-attachments/assets/8b41e9cf-2633-4c75-b7c0-ea34a87decc9


kyranjamie:
FYI Jannik from Hiro made a test page storybook that has all the APIs https://connect-git-feat-switch-to-sip-030-hirosystems.vercel.app/?path=/story/connect-connect--default

apps/test-app/README.md
```diff
@@ -0,0 +1,36 @@
+This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
```
pete-watters, line 1: Can you update this to explain what it is (even if its soon to be deleted)

pete-watters (approved):
This is good to me as a temporary app we delete. I'd still get rid of un-needed stuff (svgs etc.)

Maybe we don't need it as per Kyrans comment about the storybook page

kyranjamie:
Inspired by storybook though, I mean we do host our ui storybook at storybook.leather.io 👀  Any reason not to add these methods here?

Creating a whole new nextjs app in mono for temporary purpose seems a little bit overkill for some trigger buttons

kyranjamie:
also, if it can wait, we're like less than a week from having our first deployed build of the web app. You can have a page there to add what you like?

camerow:
@edgarkhanzadian to answer monorepo question, yes we plan to add the web app to the monorepo

PR #893, opened 2025-02-20
feat: stacks address validation LEA-1647 

pete-watters:
This PR adds validation of STX addresses to mobile

https://github.com/user-attachments/assets/11584cc0-c5a8-468c-9081-5b7f4cc142fa


packages/stacks/src/validation/transaction-validation.ts
```diff
@@ -0,0 +1,27 @@
...
+  validatePayerNotRecipient,
+} from './address-validation';
+import { StacksError } from './stacks-error';
+
+export function isValidStacksTransaction(
+  senderAddress: string,
+  recipientAddress: string,
+  chainId: ChainId
+) {
+  if (!isValidStacksAddress(senderAddress) || !isValidStacksAddress(recipientAddress)) {
```
pete-watters, line 15: Right now I'm just checking addresses are correct but there is more work to come validating amounts, compliance etc.

packages/stacks/src/mocks/mocks.ts
```diff
@@ -0,0 +1,8 @@
+// Stacks test addresses
+export const TEST_ACCOUNT_1_STX_ADDRESS = 'SPS8CKF63P16J28AYF7PXW9E5AACH0NZNTEFWSFE';
+export const TEST_ACCOUNT_2_STX_ADDRESS = 'SPXH3HNBPM5YP15VH16ZXZ9AX6CK289K3MCXRKCB';
+export const TEST_TESTNET_ACCOUNT_2_STX_ADDRESS = 'STXH3HNBPM5YP15VH16ZXZ9AX6CK289K3NVR9T1P';
+// TODO ask about these prefixs
```
pete-watters, line 5: @fbwoolf do you know about the different STX address types? 

We have a check - `isValidAddressChain` and I wanted to add tests to make sure we correctly validate `SM` and `SN` addresses. I tried to hack them by changing the prefix of other addresses but then they are invalid addresses.

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -10,49 +15,59 @@ import {
...
-      console.log('tx hex:', txHex);
-      // eslint-disable-next-line no-console
-      console.log('fees:', data.fees);
+      try {
+        const { recipient } = values;
+        isValidStacksTransaction(payer, recipient, chainId);
+        const tx = await generateTx(parseSendFormValues(values));
+
+        if (!tx) {
+          // logger('tx:', 'Attempted to generate raw tx, but no tx exists');
```
pete-watters, line 55: I'll add a `StacksError` here in my next work validating amounts

packages/stacks/src/mocks/mocks.ts
fbwoolf: I believe `SM` and `SN` are mainnet/testnet for multisig?

fbwoolf: Reference here: https://github.com/stacksgov/sips/blob/2c3b36c172c0b71369bc26cffa080e8bdd2b3b84/sips/sip-005/sip-005-blocks-and-transactions.md#transaction-authorization

Screenshot 2025-02-20 at 9 53 20 AM

fbwoolf: We do need to support multisig, and I believe it could be for other things too, this is an example of the one being used to deploy sbtc contracts: https://explorer.hiro.so/address/SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4?chain=mainnet

fbwoolf: I haven't tested, but I wonder if Asigna produces these addresses? 🤔 https://www.asigna.io/

pete-watters: Thanks a lot Fara! That will help me cover it and learn more about it

fbwoolf (approved):

packages/stacks/src/mocks/mocks.ts
pete-watters: I'll add these extended tests in my next PR. Just to be clear, we do still support multisig as ever, I just want us to have better tests

PR #905, opened 2025-02-21
chore: move BitcoinAddress to models to avoid circular dependency

pete-watters:
This PR moves the `BitcoinAddress` branded type to the models package as when installing it in the extension I was hitting issues with interface Inscription and Utxo trying to apply my types.

If you try and install the bitcoin package to models you hit circular dependency issues.

I should have tested my original work better inside of the extension first

### DIFF
--- a/packages/models/src/crypto-assets/bitcoin/bitcoin.model.ts
+++ b/packages/models/src/crypto-assets/bitcoin/bitcoin.model.ts
@@ -1,3 +1,6 @@
+// Branded type for Bitcoin addresses
+export type BitcoinAddress = string & { readonly __brand: unique symbol };
+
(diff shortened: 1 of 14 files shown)


packages/models/src/crypto-assets/bitcoin/bitcoin.model.ts
```diff
@@ -1,3 +1,6 @@
+// Branded type for Bitcoin addresses
```
pete-watters, line 1: Moving this type here so it can be used in other `models` but keeping the type guard in `packages/bitcoin` as it's using an external package for validation

kyranjamie (approved):

alexp3y (approved):

PR #906, opened 2025-02-21
fix: mobile stacks network

fbwoolf:
Mobile fix for: https://github.com/leather-io/extension/pull/6133


pete-watters (approved):

pete-watters:
Thanks for fixing this @fbwoolf 👍 Maybe we should migrate `getStacksNetworkFromChainId` into the `stacks` package?

PR #909, opened 2025-02-22
fix: add rpc types for sip9 transfers

fbwoolf:
Need these for the last issue in the extension for SIP-30 requests.

### DIFF
--- a/packages/rpc/src/methods/stacks/stx-transfer-sip9-nft.ts
+++ b/packages/rpc/src/methods/stacks/stx-transfer-sip9-nft.ts
@@ -1,0 +1,20 @@
+import { z } from 'zod';
+
+import { defineRpcEndpoint } from '../../rpc/schemas';
+import {
+  baseStacksTransactionConfigSchema,
+  stacksTransactionDetailsSchema,
+} from './_stacks-helpers';
+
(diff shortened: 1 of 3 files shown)


fbwoolf:
I am wondering if `functionName` needs to be added to the SIP-30 params to handle BNS name transfers?

kyranjamie (approved):

packages/rpc/src/methods/stacks/stx-transfer-sip9-nft.ts
```diff
@@ -0,0 +1,20 @@
...
+  stacksTransactionDetailsSchema,
+} from './_stacks-helpers';
+
+export const stxTransferSip9Nft = defineRpcEndpoint({
+  method: 'stx_transferSip9Nft',
+  params: z.intersection(
+    z.object({
+      recipient: z.string(),
+      asset: z.string(),
+      assetId: z.string(),
```
janniks, line 15: Yep, looks good, since wallets right now use serialized clarity-values. In the future this `assetId` could also be the Clarity value json structure.

fbwoolf: I see the `id` in the confirmed tx now, so nvm here.

PR #910, opened 2025-02-24
chore: add precision checks to stacks and btc

pete-watters:
This PR adds precision validation to `STX` and `BTC` transactions and fixes a bug with minimum BTC spend amount

https://github.com/user-attachments/assets/5ddd566b-78aa-4094-9739-a7c0d5664fb1

### DIFF
--- a/apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
+++ b/apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
@@ -96,6 +96,7 @@ export function useSendFormBtc() {
       try {
         const { amount: inputAmount, recipient: recipientAddress, feeRate } = values;
         if (!isValidPrecision(+inputAmount, BTC_DECIMALS)) {
+          // eslint-disable-next-line lingui/no-unlocalized-strings
           throw new BitcoinError('InvalidPrecision');
         }
         const recipient = createBitcoinAddress(recipientAddress);
--- a/apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
+++ b/apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
@@ -17,8 +17,9 @@ import {
   isValidBitcoinTransaction,
   payerToBip32Derivation,
 } from '@leather.io/bitcoin';
+import { BTC_DECIMALS } from '@leather.io/constants';
 import { AverageBitcoinFeeRates, Money, bitcoinNetworkToNetworkMode } from '@leather.io/models';
-import { createMoneyFromDecimal } from '@leather.io/utils';
+import { createMoneyFromDecimal, isValidPrecision } from '@leather.io/utils';
 
 import {
   CreateCurrentSendRoute,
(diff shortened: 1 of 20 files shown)


packages/bitcoin/src/validation/transaction-validation.spec.ts
```diff
@@ -45,16 +60,38 @@ describe('isValidBitcoinTransaction', () => {
...
   it('should throw an error if the balance is insufficient', () => {
     const transaction = {
       ...mockTransaction,
-      amount: createMoney(200, 'USD'),
+      amount: createMoney(200, 'BTC'),
     };
     expect(() => isValidBitcoinTransaction(transaction)).toThrow(BitcoinError);
   });
+
+  it('should throw an error if the amount precision is invalid', () => {
```
pete-watters, line 76: @fbwoolf can you please help confirm this is correct for the tests?

I'm trying to test various amounts like `1.123456789 BTC` and `0.123456789 BTC` but I'm not sure if I'm using create money properly to create the amount for them

fbwoolf: `createMoney` won't shift the decimals but `createMoneyFromDecimals` will, so to throw an error you would use `createMoney(0.123456789, 'BTC')` bc that would be 9 decimals instead of 8. `createMoney` expects the fractional unit.

fbwoolf: Same error in the extension...

Screenshot 2025-02-24 at 9 51 06 AM

fbwoolf: If you ever use `createMoneyFromDecimals` it will shift the value 8 decimals (or the provided symbol decimals) expecting the base unit as the input.

fbwoolf (approved):

kyranjamie (approved):
Approving, but isn't clear why this validation doesn't take place within the schema of the form?

pete-watters:
> Approving, but isn't clear why this validation doesn't take place within the schema of the form?

You're right, this check should be in the schema of the form. 

@tigranpetrossian has a UI refactor underway so I am limiting my changes to the form and I forgot to add a note to say to move the validation check to the schema

packages/bitcoin/src/validation/transaction-validation.spec.ts
pete-watters: Thanks @fbwoolf ! I'm trying to tap into your wisdom as I go 🙇

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
```diff
@@ -93,7 +94,11 @@ export function useSendFormBtc() {
 
     onInitSendTransfer({ utxos, feeRates }: SendFormBtcContext, values: SendFormBtcSchema) {
       try {
-        const { recipient: recipientAddress, feeRate } = values;
+        const { amount: inputAmount, recipient: recipientAddress, feeRate } = values;
+        if (!isValidPrecision(+inputAmount, BTC_DECIMALS)) {
```
pete-watters, line 98: @tigranpetrossian I am just adding this check here for now to minimise disruption but as @kyranjamie rightly pointed out, this precision check should be in the form schema checks for `amount`. 

I excluded it from the main transaction validation code as it needs to use the raw input amount

tigranpetrossian: @pete-watters Thanks, noted! Will integrate into the schema.

PR #912, opened 2025-02-24
fix: add wallet sheet image bg color

edgarkhanzadian:
Reopening for a fix from this PR https://github.com/leather-io/mono/pull/567

Before: 
beforeDark
beforeLight

After:

afterLight
afterDark

### DIFF
--- a/apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
+++ b/apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
@@ -9,6 +9,7 @@ import Animated, {
 import { WaitlistIds } from '@/features/waitlist/ids';
 import { AppRoutes } from '@/routes';
 import { TestId } from '@/shared/test-id';
+import { useSettings } from '@/store/settings/settings';
 import { t } from '@lingui/macro';
 import { Image } from 'expo-image';
 import { useRouter } from 'expo-router';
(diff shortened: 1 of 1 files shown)


apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
```diff
@@ -77,7 +79,14 @@ export function AddWalletSheetLayout({
       }}
     >
       <AnimatedBox style={animatedStyle}>
-        <Box width="100%" style={{ height: 200, overflow: 'hidden' }} bg="ink.text-primary">
+        <Box
+          width="100%"
+          style={{ height: 200, overflow: 'hidden' }}
+          bg={whenTheme({
```
camerow, line 85: I would love it for our themes to be much less `theme` aware - eg. the tokens adapt better to the theme, `ink.background-secondary` should be all the `bg` needs and the theme adapts more elegantly.

camerow (approved):

apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
camerow: cc @kyranjamie / @tigranpetrossian

pete-watters (approved):

apps/mobile/src/components/add-wallet/add-wallet-sheet.layout.tsx
kyranjamie: `ink.background-secondary` is a semantic token, is restyle not set up to handle it conditionally?

kyranjamie (approved):

PR #915, opened 2025-02-25
fix: migrate compliance checks and add to mobile

pete-watters:
The PR migrates some compliance check code from the extension to `mobile`. 

We only use `checkEntityAddressIsCompliant` at the moment but I migrated the other functions also `useBreakOnNonCompliantEntity` 

Now if we try and send a transaction to a non-compliant address in `mobile` we see this message
Screenshot 2025-02-26 at 14 24 22


apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
```diff
@@ -53,11 +51,12 @@ function parseSendFormValues({
     ],
   };
 }
-interface GetTxFeesArgs {
-  feeRates: AverageBitcoinFeeRates;
-  recipients: CoinSelectionRecipient[];
-  utxos: CoinSelectionUtxo[];
-}
+// TODO LEA-2028 - revise fees - this was here but unused
```
pete-watters, line 54: I'm commenting out this fee stuff as it wasn't doing anything yet but I guess it's needed for something soon, maybe for LEA-2028? 

`getBitcoinFees` is used in the flow for `psbt-signer` which is the next step after this form so it could be just not needed here at all

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
```diff
@@ -95,63 +94,61 @@ export function useSendFormBtc() {
     onInitSendTransfer({ utxos, feeRates }: SendFormBtcContext, values: SendFormBtcSchema) {
       try {
         const { amount: inputAmount, recipient: recipientAddress, feeRate } = values;
+        const recipient = createBitcoinAddress(recipientAddress);
+        // TODO LEA-1852 - move to form schema validation
+        // 1. validate precision
```
pete-watters, line 99: Annotating the steps for now to assist with refactoring here

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
fbwoolf: It may have been something I did early to get things going, but maybe fees were moved to the approver flow once Edgar took over?

pete-watters: OK, thanks!

I'm just going to delete it to clean up some noise and we can bring it back if we need to

apps/mobile/src/features/send/send-form/validation/btc.validation.ts
```diff
@@ -13,7 +13,6 @@ export function formatBitcoinError(errorMessage: BitcoinErrorKey) {
     }),
     InsufficientAmount: t({
       id: 'bitcoin-error.insufficient-amount',
-      // TODO test this - avoids useLingui
       message: `Minimum is ${minimumBtcSpend}`,
```
pete-watters, line 16: I confirmed this works 
Screenshot 2025-02-26 at 14 23 27

packages/query/src/common/compliance-checker/compliance-checker.hooks.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  refetch: vi.fn(),
+  remove: vi.fn(),
+};
+
+describe('useBreakOnNonCompliantEntity', () => {
+  it('should return isCompliant true when no addresses are on sanctions list', () => {
+    // Mock implementation for compliant addresses
+    vi.mocked(useCheckAddressComplianceQueries).mockReturnValue([
+      // TODO investigate this type error / if its worth adding tests
+      // @ts-expect-error - mockQueryResult is not a UseQueryResult
```
pete-watters, line 64: I tried generating tests for this and they work but I was battling types and gave up 🫨 

Maybe some TS wiz can help here but I think it's OK to suppress in a test for now

fbwoolf (approved):
Nice work @pete-watters, do you know if there is an issue for removing this code from the extension and installing it back? I can look or make one if not.

pete-watters:
> Nice work @pete-watters, do you know if there is an issue for removing this code from the extension and installing it back? I can look or make one if not.

Thanks @fbwoolf 👍 I made an issue for the `extension` refactors and tagged it in some comments [LEA-2189](https://linear.app/leather-io/issue/LEA-2189/refactor-extension-to-use-mono-compliance-check-code).

Ideally I'd be also tackling that as I go but we are a bit short on time now so it would be great if you fancy taking it on (and possibly exposing errors!)

apps/mobile/src/features/send/send-form/hooks/use-send-form-btc.tsx
tigranpetrossian: Thanks so much @pete-watters, made a note to include in the refactored version.

PR #916, opened 2025-02-25
fix: add stacks readme.md

pete-watters:
Adding a stacks README.md to initiate a new release  to hopefully fix issues with https://github.com/leather-io/mono/pull/913


packages/stacks/README.md
```diff
@@ -0,0 +1,3 @@
+# @leather.io/stacks
+
+This package is a copy of [stacks](https://github.com/leather-io/extension/tree/dev/src/shared/crypto/stacks) shared directory in extension repo.
```
kyranjamie, line 3: This isn't really an accurate description of the stacks package. The code came from here, but it isn't intended to just be a copy

```suggestion
`@leather.io/stacks` is intended to be the home of domain logic relating to the Stacks blockchain operations, such as address generation, and transaction signing.
```

Probably could come up with something better than this, but it's more descriptive at least.

pete-watters: Agreed 👍  I just copied it directly from `@leather.io/bitcoin` as I thought about updating that description but now's not the time for writing READMEs. 

I'll commit this improvement and we can consider updating the bitcoin one also. We also need to fix the `stacks` package as it's not publishing the `src` files [LEA-2180](https://linear.app/leather-io/issue/LEA-2180/leatheriostacks-is-not-publishing-src)

PR #918, opened 2025-02-25
feat(web): scaffold react router v7

kyranjamie:
This PR supersedes @alter-eggo's work setting up Remix in https://github.com/leather-io/mono/pull/865, adding React Router 7 in framework mode, following their merge https://remix.run/blog/merging-remix-and-react-router

This serves as the foundation for our Earn experience to be hosted on `leather.io`.

**Build chain issues**
Until now, we didn't have any dependencies on `panda-preset` package, as this was used exclusively in the extension. Now that it's a dependency of `apps/web` I've had to add some `prepare` scripts to get this building, also for downstream deps like `tokens`. I've removed the dependency `tokens` had on `utils`, as only one method was used and this simplifies the amount of pre-building going on. Took ages to figure out how to get this building correctly on CI.

**React 18…19**
I attempted to use React 18 for RR7, but R19 resolves many hydration issues that R18 had been suffering with, so it's best to start fresh with the latest version. Given the inability to use R19 for our mobile app, we now have both versions installed in our monorepo. This creates some React typing issues, I've removed most the `as any`s added when fighting with this, but some remain, e.g. in `ui`. React has been added to `syncpack`s list of version exemptions.


apps/web/app/app.css
```diff
@@ -0,0 +1,44 @@
+@layer reset, base, tokens, recipes, utilities;
```
pete-watters, line 1: Would it be a pain to move this into `packages/ui` eventually? Then we could also load it from there for the `extension`

apps/web/README.md
```diff
@@ -0,0 +1,25 @@
+# Leather.io Web App
+
+Home of Leather's web app, built with [React Router v7](https://reactrouter.com/) in frameework mode.
```
pete-watters, line 3:
```suggestion
Home of Leather's web app, built with [React Router v7](https://reactrouter.com/) in framework mode.
```

pete-watters (approved):

tigranpetrossian (approved):
Not sure this is something we can do before the release, but the DX around importing and using the `ui` package could definitely be simplified quite a bit. I've been slowly looking into this, will post an update once I have something tangible.

apps/mobile/src/features/accounts/components/account-avatar.tsx
```diff
@@ -26,7 +26,7 @@ import {
 } from '@leather.io/ui/native';
 import { isString } from '@leather.io/utils';
 
-export const accountIconMap = {
+export const accountIconMap: Record<string, ForwardRefExoticComponent<any>> = {
```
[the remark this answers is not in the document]
kyranjamie: This is because of a type exports issue (probs introduced by version update)

fbwoolf (approved):
Looks great @kyranjamie 👍

kyranjamie:
I think this is in a good place to merge. Will add Cloudflare deployment set up and some initial nav changes next.

PR #927, opened 2025-02-27
feat: stacks transaction validation, ref LEA-1681

pete-watters:
This PR adds validation checks for Stacks transactions and also analytic events on errors which match the extension behaviour 

https://github.com/user-attachments/assets/ed8e8118-a093-4a92-b278-b7ac2c3a492a

### DIFF
--- a/apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
+++ b/apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
@@ -1,7 +1,6 @@
 import { useGenerateStxTokenTransferUnsignedTransaction } from '@/common/transactions/stacks-transactions.hooks';
 import { useToastContext } from '@/components/toast/toast-context';
 import { useSettings } from '@/store/settings/settings';
-import { analytics } from '@/utils/analytics';
 import { t } from '@lingui/macro';
 import BigNumber from 'bignumber.js';
 
@@ -18,7 +17,7 @@ import {
 } from '../../send-form.utils';
 import { SendFormStxContext } from '../providers/send-form-stx-provider';
 import { SendFormStxSchema } from '../schemas/send-form-stx.schema';
-import { formatStacksError } from '../validation/stx.validation';
+import { formatStacksError, trackAnalyticsForError } from '../validation/stx.validation';
 
 export type CurrentRoute = CreateCurrentSendRoute<'send-form-stx'>;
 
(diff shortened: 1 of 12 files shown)


packages/models/src/transactions/transaction-error.model.ts
```diff
@@ -1,7 +1,12 @@
+export enum TransactionErrors {
```
pete-watters, line 1: Introducing an `enum` for these as otherwise I need to add `// eslint-disable-next-line lingui/no-unlocalized-strings` wherever I am throwing the errors in the mobile app

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -43,27 +42,32 @@ export function useSendFormStx() {
...
       try {
-        const parsedValues = parseSendFormValues(values);
-        const { amount, recipient } = parsedValues;
         // TODO LEA-1852 - move to form schema validation
         // 1. validate precision
         if (!isValidPrecision(+values.amount, STX_DECIMALS)) {
-          // eslint-disable-next-line lingui/no-unlocalized-strings
-          throw new StacksError('InvalidPrecision');
+          // throw new StacksError('InvalidPrecision');
+          throw new Error('Pete');
```
camerow, line 60: Do we really want to throw you as an error @pete-watters ? Seems a bit tragic.

apps/mobile/src/features/stacks-tx-signer/stacks-tx-signer.tsx
```diff
@@ -47,7 +47,7 @@ interface ReviewTxSummaryProps {
 function formReviewTxSummary({ tx, symbol }: ReviewTxSummaryProps) {
   if (symbol !== 'STX') throw new Error('No support for SIP10');
 
-  // WARNING: Danerous type casting
+  // WARNING: Dangerous type casting
   const payload = tx.payload as TokenTransferPayloadWire;
```
camerow, line 51: Quickly investigated this and perhaps would be better to do: 

```
  if (tx.payload.payloadType !== PayloadType.TokenTransfer) {
    throw new Error('Unsupported payload type');
  }
```
rather than typecast which is definitely not type safe in this instance.

apps/mobile/src/features/stacks-tx-signer/stacks-tx-signer.tsx
```diff
@@ -47,7 +47,7 @@ interface ReviewTxSummaryProps {
 function formReviewTxSummary({ tx, symbol }: ReviewTxSummaryProps) {
   if (symbol !== 'STX') throw new Error('No support for SIP10');
 
-  // WARNING: Danerous type casting
+  // WARNING: Dangerous type casting
```
camerow, line 50:
```suggestion
  if (tx.payload.payloadType !== PayloadType.TokenTransfer) {
    throw new Error('Unsupported payload type');
  }
```

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
pete-watters: haha sorry, my bad!

apps/mobile/src/features/stacks-tx-signer/stacks-tx-signer.tsx
pete-watters: Sure, I can check it out but I just fixed a spelling error here in someone else's code

pete-watters:
Thanks for the review @camerow ! 

A few sloppy errors there 😢  - more haste less speed!

camerow (approved):

PR #930, opened 2025-02-27
feat: add use clipboard

alter-eggo:
adds [copy to clipboard hook](https://github.com/leather-io/extension/blob/dev/src/app/common/hooks/use-copy-to-clipboard.ts) from extension, we will need this in web app as well


alter-eggo:
check pls if place is right for this

kyranjamie:
You aren't exporting it though

PR #941, opened 2025-02-28
ci: test eas workflow

camerow:
This PR enables Maestro tests to be run and verified on `push` to `dev`/`main` and there are changes to `apps/mobile`. This leverages an integration with Expo and Github. I would strongly recommend reading up on the documentation [here](https://docs.expo.dev/eas/workflows/get-started/). The short of it is, though, that this is interface is much like GH Workflows - but a smaller subset of syntax - to enable us to run builds / tests / etc on EAS servers and integrate those pipelines with our Github Checks.

@pete-watters I have commented out any failing tests, and, for now, removed `pull_request` as a trigger because tests are slow and flaky. We can create some alerting mechanisms for failures. I think this will be an ongoing thing to work at and improve as we experience some of the issues. 

I fixed many issues locally, but they seem to be now failing in the CI - subsequent runs might yield different results and this is the frustrating experience with E2E Maestro tests, and ultimately the reason I don't think we should run them on PRs.

Screenshot 2025-03-06 at 2 09 38 PM

In this screenshot for example locally those initial tests pass and the passing test fails. In the CI we get different results.


apps/mobile/maestro/shared/add-wallet.yaml
```diff
@@ -18,7 +18,9 @@ appId: io.leather.mobilewallet
     commands:
       - tapOn:
           text: 'Skip for now'
+      - waitForAnimationToEnd:
```
pete-watters, line 21: Nice work finding this `wait`

apps/mobile/eas.json
```diff
@@ -77,22 +77,16 @@
         "image": "latest"
       }
     },
-    "maestro-test-ios": {
+    "maestro": {
       "node": "20.11.0",
       "pnpm": "10.4.1",
-      "withoutCredentials": true,
-      "config": "maestro-test-ios.yml",
+      "environment": "preview",
```
pete-watters, line 83: Can you explain what these do?

apps/mobile/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
pete-watters, line 3: What's the difference here?

pete-watters (approved):
Nice work @camerow 👍 I can tackle trying to get the other tests to work / rewriting them so they are faster and less flaky. 

We still have that double job though, can you delete one of those? 

Screenshot 2025-03-07 at 05 52 43

We don't need `eas-deploy.yml` and `mobile-preview-builds.yml`, I think we can delete `eas-deploy.yml` or do it in a separate PR? 

I also cannot see the test result in this PR, I guess as you removed `pull_request` as a trigger? 

What would you think about adding a label trigger so that we could add a label to PRs and start the tests that way? That would make it easier for debugging / testing them too? 

Based on your investigations it seems like we should:
- have some Maestro tests but less of them and faster / more reliable
- expand our testing to have more integration testing run on each branch

apps/mobile/android/gradle/wrapper/gradle-wrapper.properties
camerow: It's a minor version lower for the build tool gradle, but i have been having a ton of issues with `8-6`. To fix it I deleted `./gradle` and allowed Android Studio to reinstall / rebuild gradle dependencies and it aut-generated this version. Since nobody here really cares / uses gradle other than implicitly through `pnpm run android` and when building I don't think this minor version change matters to us at all.

camerow:
> Nice work @camerow 👍 I can tackle trying to get the other tests to work / rewriting them so they are faster and less flaky.
> 
> We still have that double job though, can you delete one of those?
> 
> Screenshot 2025-03-07 at 05 52 43
> 
> We don't need `eas-deploy.yml` and `mobile-preview-builds.yml`, I think we can delete `eas-deploy.yml` or do it in a separate PR?
> 
> I also cannot see the test result in this PR, I guess as you removed `pull_request` as a trigger?
> 
> What would you think about adding a label trigger so that we could add a label to PRs and start the tests that way? That would make it easier for debugging / testing them too?
> 
> Based on your investigations it seems like we should:
> 
> * have some Maestro tests but less of them and faster / more reliable
> * expand our testing to have more integration testing run on each branch

A label trigger is no problem - I won't use the `needs:` syntax since that is for other things more PR content related. How about I do `e2e-test:android`/`e2e-test:ios`/`e2e-test:mobile`?

I will remove one of those other builds - I think I'll do that in a separate PR.

pete-watters:
> > Nice work @camerow 👍 I can tackle trying to get the other tests to work / rewriting them so they are faster and less flaky.
> > We still have that double job though, can you delete one of those?
> > Screenshot 2025-03-07 at 05 52 43
> > We don't need `eas-deploy.yml` and `mobile-preview-builds.yml`, I think we can delete `eas-deploy.yml` or do it in a separate PR?
> > I also cannot see the test result in this PR, I guess as you removed `pull_request` as a trigger?
> > What would you think about adding a label trigger so that we could add a label to PRs and start the tests that way? That would make it easier for debugging / testing them too?
> > Based on your investigations it seems like we should:
> > 
> > * have some Maestro tests but less of them and faster / more reliable
> > * expand our testing to have more integration testing run on each branch
> 
> A label trigger is no problem - I won't use the `needs:` syntax since that is for other things more PR content related. How about I do `e2e-test:android`/`e2e-test:ios`/`e2e-test:mobile`?
> 
> I will remove one of those other builds - I think I'll do that in a separate PR.

That sounds great 👍 the label itself isn't important, I just think its useful for us if we want to trigger the suite of tests on a PR we can easily enough. 

Removing the other stuff on another PR is fine too, just adding a reminder!

camerow:
> That sounds great 👍 the label itself isn't important, I just think its useful for us if we want to trigger the suite of tests on a PR we can easily enough.
> 
> Removing the other stuff on another PR is fine too, just adding a reminder!

@pete-watters what do you think about running this for PRs when they are in the merge queue? Then we don't waste cycles running the tests but we don't merge broken e2e tests? Flaky test could still make our lives quite frustrating

pete-watters:
> @pete-watters
> 
> I rethought a couple things:
> 
> 1. `needs:e2e-tests` made more sense after thinking about it, since we want to block the PR if it does need them.
> 2. Removed the duplicate job here

Isn't `needs:e2etests` a bit confusing though if you do add them and just want to run them? 

I don't mind about the naming really, just looking for a nice way to run them on demand in CI and a way of protecting them from being broken. 

Let's go with what you've done now (as it's done, great job!) and we can always improve on it as needed

PR #943, opened 2025-03-04
chore: add default to ES and possibility for users to change language

pete-watters:
In this PR I am:
- introducing `react-native-localize` so we can detect a users region and specify language if it's `ES`. 
- adding a new setting to the display settings menu to allow us to toggle easily (not in the designs)
- made the `theme` translation key dynamic so we can localise `System` , `Dark`, `Light` etc.


alexp3y (approved):
Muy bueno!

apps/mobile/src/locales/index.ts
```diff
@@ -41,6 +45,12 @@ export async function initiateI18n() {
   // run load and activate so the I18nProvider doesn't block the render
   i18n.load('en', {});
   i18n.activate(DEFAULT_LOCALE);
+  const locales = getLocales();
+
+  const deviceLocale =
+    locales.find(locale => isDefined(locale.languageTag))?.languageTag ?? DEFAULT_LOCALE;
+  // LEA-2138 Only support es-ES and en
```
markmhendrickson, line 52: @pete-watters would it be easy enough to remove this conditional check for `es-ES` so that any local device language loads upon toggle?

markmhendrickson: Tracking [here](https://linear.app/leather-io/issue/LEA-2307/add-language-selector-view-for-setting-languages)

pete-watters: Yes we can do that easily enough when the time is right. There's no point right now with only `ES` ready

PR #945, opened 2025-03-04
fix: stx token transfer fees condition

alter-eggo:
We should use our specific token fees estimations only in case we got them from config

### DIFF
--- a/packages/query/src/common/remote-config/remote-config.query.ts
+++ b/packages/query/src/common/remote-config/remote-config.query.ts
@@ -129,7 +129,7 @@ export function useConfigTokensEnabledByDefault(): string[] {
 
 export function useConfigTokenTransferFeeEstimations() {
   const config = useRemoteConfig();
-  return get(config, 'tokenTransferFeeEstimations', []);
+  return get(config, 'tokenTransferFeeEstimations', undefined);
 }
 
 export function useConfigSpamFilterWhitelist(): string[] {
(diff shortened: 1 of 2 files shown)


packages/query/src/common/remote-config/remote-config.query.ts
```diff
@@ -129,7 +129,7 @@ export function useConfigTokensEnabledByDefault(): string[] {
 
 export function useConfigTokenTransferFeeEstimations() {
   const config = useRemoteConfig();
-  return get(config, 'tokenTransferFeeEstimations', []);
+  return get(config, 'tokenTransferFeeEstimations', undefined);
```
kyranjamie, line 132: This is the default argument that if undefined, returns undefined

kyranjamie (commented):
Why don't we change the fallback if there are no fees defined in the config, to our hardcoded defaults?

packages/query/src/common/remote-config/remote-config.query.ts
alter-eggo: yeah, I'm working now in extension on fallback, done this pr before we decided to do that in thread

alter-eggo: actually can close this one prob

alter-eggo:
close in favour of https://github.com/leather-io/mono/pull/951

PR #951, opened 2025-03-06
feat: add fallback if config fetch failed, LEA-1806

alter-eggo:
This pr adds config fallback in case of error 
this fixes stx send page fee errors

### DIFF
--- a/packages/query/src/common/remote-config/remote-config.query.ts
+++ b/packages/query/src/common/remote-config/remote-config.query.ts
@@ -33,12 +33,14 @@ export function useRemoteConfig() {
   const { data } = useQuery({
     queryKey: ['walletConfig'],
     queryFn: fetchLeatherMessages(env, leatherGh),
+    initialData: leatherGh.localConfig,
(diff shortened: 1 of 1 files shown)


packages/query/src/common/remote-config/remote-config.query.ts
```diff
@@ -39,6 +39,11 @@ export function useRemoteConfig() {
     staleTime: 1000 * 60 * 10,
     retryDelay: 1000 * 60,
   });
+
+  if (error || failureCount > 0) {
+    return leatherGh.localConfig;
+  }
```
kyranjamie, line 45: afaik `initialData` is for this?

alter-eggo: yeah, exactly.. have forgotten about it

kyranjamie (approved):

kyranjamie:
Approved, but imo we should move the whole config back to extension. doesn't belong her

alter-eggo:
@kyranjamie why move this to extension? won't we need config in web app?

PR #952, opened 2025-03-06
feat: browser UI

edgarkhanzadian:
- Create Browser sheet UI 
- Close it off using a browser feature flag
- Update the webview listeners s.t. those would also work for android (this would require and update to extension)
- Implement screenshots for dApps
- Implement keyboard avoiding view and animations that are built on top of the keyboard open/close animation (like url input)
- Implement a redux state for dApps

Note: not squashing commits until right before merging the PR

Android:

https://github.com/user-attachments/assets/d5317392-87a2-4b30-a216-9c41dff1ec45

iOS:

https://github.com/user-attachments/assets/bce2ba2c-81d1-4ecf-a03d-2885288bb15f


apps/mobile/src/components/browser/browser-sheet/url-input.tsx
```diff
@@ -0,0 +1,285 @@
+import { ReactNode, useMemo, useRef } from 'react';
```
fbwoolf, line 1: This is a massive component file, any way to break this up a bit better?

packages/provider/src/index.ts
```diff
@@ -15,10 +15,17 @@ import { Platform } from './types';
 
 interface initInpageProviderArgs {
   onDispatch(rpcRequest: RpcRequests): void;
+  addEventListener(listener: (this: Window, ev: MessageEvent<any>) => any): void;
+  removeEventListener(listener: (this: Window, ev: MessageEvent<any>) => any): void;
```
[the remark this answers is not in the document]
fbwoolf: If you make an issue to track the change, you can assign it to me.

apps/mobile/src/components/action-bar/action-bar-container.tsx
```diff
@@ -125,94 +137,84 @@ export function useActionBarContext() {
...
   const { sendSheetRef, receiveSheetRef } = useSheetNavigatorContext();
   const wallets = useWallets();
   const addWalletSheetRef = useRef<SheetRef>(null);
+  const browserSheetRef = useRef<SheetRef>(null);
+  const browserRef = useRef<SheetRef>(null);
+  const releaseBrowserFeature = useReleaseBrowserFeatureFlag();
+
+  useEffect(() => {
+    browserRef.current?.present();
+  }, []);
```
fbwoolf, line 175: Should this have any dependencies, or use `useOnMount`?

fbwoolf (approved):
🚀

apps/mobile/src/components/browser/browser-sheet/url-input.tsx
tigranpetrossian, line 1: Looks like this component could use a little refactoring–mostly naming & organization issues.

Few glaring ones:
* There's `searchBarStyle` and `SearchBarButton` but nothing explicitly named search bar? Perhaps `SearchBarButton -> ToolbarButton`, and extract the Toolbar into its own component?
* `UrlInputContainer`doesn't seem to contain the input
* Unlike SearchBarButton, Pressables inside UrlInputContainer aren't named 

A round or two of "moving things around until it feels right" should make this much easier to reason about 👌

packages/provider/src/index.ts
```diff
@@ -60,12 +67,12 @@ export function initInpageProvider({ onDispatch, env }: initInpageProviderArgs)
           const response =
             typeof event.data === 'object' ? event.data : JSON.parse(event.data as any);
           if (response.id !== id) return;
-          window.removeEventListener('message', handleMessage);
+          removeEventListener(handleMessage);
           // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors
           if ('error' in response) return reject(response);
```
tigranpetrossian, line 72: is `response.error` a string? Wouldn't it better to `reject(new Error(error))` instead of working around the eslint rule?

packages/provider/src/mobile.ts
```diff
@@ -10,6 +16,13 @@ initInpageProvider({
     commitSha: 'replace_commit_sha',
     version: 'replace_version',
   },
+  // WebView on Android doesn't support window.addEventListener('message')
+  addEventListener(listener) {
+    (document as any).addEventListener(LEATHER_WALLET_RPC_REQUEST, listener);
```
tigranpetrossian, line 21:
```
declare global {
  interface DocumentEventMap {
    'leather-wallet-rpc-request': MessageEvent; 
  }
}
```
Should communicate clearer than casting entire `document` as any.  You'd also need to alter the type of the listener where it's defined:
```
addEventListener(listener: (this: Window | Document, ev: MessageEvent<any>) => any): void;
removeEventListener(listener: (this: Window | Document, ev: MessageEvent<any>) => any): void;
```

tigranpetrossian (changes requested):
Very cool @edgarkhanzadian.
Left few comments: mostly types & refactoring.
On top of that, I think `browser` as a whole should live in `features`, as it's a major feature.

On a separate note, it would be very nice to see the feature as a whole decoupled from sheet logic (including in naming), I think Browser and its internals don't necessarily need to be aware of the environment it's being rendered in, and the few sheet-specific elements can be injected. This is true for other major features rendered in sheets as well (Send, Receive, Swap). 
I wouldn't expect this change to be done within this PR, but this is something worth keeping on our radars and come up with global sheet management handling all of these.

apps/mobile/src/assets/sbtc-rewards.png
kyranjamie, line 1: have these been imageoptim'd?

apps/mobile/src/components/browser/browser-sheet/tabs/recent.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import { ScreenshotCard } from '../screenshot-card';
+import { URL_SEARCH_HEIGHT } from '../utils';
+
+const { width } = Dimensions.get('window');
+interface RecentProps {
+  searchUrl(url: string): void;
+}
+
+export function Recent({ searchUrl }: RecentProps) {
```
kyranjamie, line 19: Names like this are bad for searchability imo. `BrowserRecentTab`?

apps/mobile/src/components/browser/browser-sheet/tabs/suggested.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { URL_SEARCH_HEIGHT } from '../utils';
+
+interface SuggestedProps {
+  searchUrl(url: string): void;
+}
+
+export function Suggested({ searchUrl }: SuggestedProps) {
+  const { bottom } = useSafeAreaInsets();
+  return (
+    <BottomSheetScrollView contentInset={{ bottom: bottom + URL_SEARCH_HEIGHT }}>
```
kyranjamie, line 19: I would love to see use of Layout components here. Where `BrowserSuggestedTab` adds content & functionality to `BrowserSuggestedTabLayout` that encapsulates all the styles

packages/provider/src/mobile.ts
kyranjamie: Environment specific events concerns could be passed in using EventEmitter API. Provider packages doesn't need to know abut RN or web envs. Then no casting is needed.

packages/provider/src/index.ts
tigranpetrossian: Missed that this isn't part of the PR, but still hopefully an easy fix.

packages/provider/src/index.ts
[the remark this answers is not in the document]
camerow: @edgarkhanzadian you'd probably be best to capture what needs to happen here, could you create this issue for @fbwoolf please?

apps/mobile/src/components/action-bar/action-bar-container.tsx
```diff
@@ -125,94 +130,84 @@ export function useActionBarContext() {
...
+      <ActionBarButton onPress={() => receiveSheetRef.current?.present()} icon={<InboxIcon />} />
+      {isDev() && (
         <ActionBarButton
-          onPress={() => sendSheetRef.current?.present()}
-          icon={<PaperPlaneIcon />}
-          label={t({
-            id: 'action_bar.send_label',
-            message: 'Send',
-          })}
+          onPress={() => {
```
pete-watters, line 188: Lets just get rid of this button if it's not doing anything? Who knows when we will add `Swap`

apps/mobile/src/components/browser/browser-sheet/screenshot-card.tsx
```diff
@@ -0,0 +1,55 @@
...
+  app: App;
+  onPress(): void;
+}
+export function ScreenshotCard({ app, onPress }: ScreenshotCardProps) {
+  const dispatch = useAppDispatch();
+  function onDeleteApp() {
+    dispatch(userRemovesApp({ origin: app.origin }));
+  }
+  return (
+    <Pressable onPress={onPress} pressEffects={legacyTouchablePressEffect} maxWidth={width * 0.4}>
```
pete-watters, line 22: Maybe you can have a const:
`const maxWidth = width * 0.4`

apps/mobile/src/components/browser/browser-sheet/search-bar/browser-navigation-bar.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const hostname = new URL(searchUrl).hostname;
+  return (
+    <BrowserNavigationBarContainer>
+      <Pressable
+        p="3"
+        onPress={onGoBack}
+        disabled={!canGoBack}
+        pressEffects={legacyTouchablePressEffect}
+      >
+        <ChevronLeftIcon style={{ opacity: canGoBack ? 1 : 0.7 }} variant="large" />
```
pete-watters, line 36: Maybe this repeated ternary can be avoided using a function? 

```
function isPossible(value: boolean) {
 return value ?  1 : 0.7
}
```

apps/mobile/src/components/browser/browser-sheet/search-bar/search-bar.tsx
```diff
@@ -0,0 +1,153 @@
...
+          isUrlFocused.value = false;
+        }}
+        keyboardType="web-search"
+        textContentType="URL"
+        autoCapitalize="none"
+        onChangeText={setTextUrl}
+        value={textUrl}
+        autoComplete="url"
+        onSubmitEditing={onSubmit}
+        width={'100%'}
```
pete-watters, line 112: I think you don't need the braces here? `width="100%"`

apps/mobile/src/components/browser/browser-sheet/search-bar/search-bar.tsx
```diff
@@ -0,0 +1,153 @@
...
+  refresh,
+}: SearchBarProps) {
+  const theme = useTheme<Theme>();
+  const { bottom } = useSafeAreaInsets();
+  const textInputRef = useRef<RNTextInput>(null);
+
+  const { keyboardAvoidingStyle, searchBarStyle, isUrlFocused, browserNavigationBarStyle } =
+    useSearchBarAnimatedStyles();
+
+  const baseSearchBar = (
```
pete-watters, line 54: Maybe `baseSearchBar` can be a component in it's own file? Or else use style objects instead of props? 

It seems to use a mixture of a style object and then also lots of props anyway. Some of them seem to be repeated?

```
   style={[
...
        {
          paddingHorizontal: theme.spacing['5'],
          paddingTop: theme.spacing['4'],
          paddingBottom: theme.spacing['4'] + bottom,
        },
      ]}
... 
  pt="4"
      pb="4"
      px="5"
```

apps/mobile/src/components/browser/browser-sheet/search-bar/search-bar.tsx
```diff
@@ -0,0 +1,153 @@
...
+            setTextUrl(copiedString);
+          }}
+          onShare={async () => {
+            if (navState?.url) await Sharing.shareAsync(navState?.url);
+            KeyboardController.dismiss();
+          }}
+        />
+      )}
+
+      <TextInput
```
pete-watters, line 96: Maybe you can wrap this component as `SearchInput` and pass the props there so it's easier to read in this file?

apps/mobile/src/components/browser/browser-sheet/search-bar/toolbar-button.tsx
```diff
@@ -0,0 +1,25 @@
+import { ReactNode } from 'react';
+
+import { Pressable, Text, legacyTouchablePressEffect } from '@leather.io/ui/native';
```
pete-watters, line 3: Is there a `legacyTouchablePressEffect` and a new version? Maybe we can just have a new one. I don't understand the `legacy` part?

apps/mobile/src/components/browser/browser-sheet/tabs/browser-connected-tab.tsx
```diff
@@ -0,0 +1,79 @@
...
+        maxWidth={width * 0.5}
+        style={{
+          paddingBottom: bottom + URL_SEARCH_HEIGHT,
+        }}
+        gap="3"
+      >
+        <Image
+          style={{ height: width * 0.5, width: width * 0.5 }}
+          contentFit="cover"
+          source={require('@/assets/connected_empty.png')}
```
pete-watters, line 40: Can we import this a different way than an inline `require`? 

Maybe this should live in the UI package?

apps/mobile/src/components/browser/browser-sheet/use-browser-search-state.ts
```diff
@@ -0,0 +1,60 @@
+import { useReducer } from 'react';
```
pete-watters, line 1: Nice work on this 👍  It would be good to add some unit tests for the reducer states

apps/mobile/src/components/browser/browser-sheet/use-browser-search-state.ts
```diff
@@ -0,0 +1,60 @@
...
+        ...state,
+        textUrl: formatURL(action.url),
+        searchUrl: formatURL(action.url),
+        browserType: 'active',
+      };
+    }
+    case 'resetSearchBar':
+      return initialState;
+    default:
+      // TODO: handle error here?
```
pete-watters, line 42: Maybe handle the error or get rid of this?

apps/mobile/src/components/browser/browser-sheet/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+
+function getScreenshotPath(hostname: string) {
+  return `${FileSystem.documentDirectory}${hostname}_screenshot.jpg`;
+}
+// Capture screenshot using react-native-view-shot
+export async function captureScreenshot(viewShotRef: RefObject<ViewShot>, hostname: string) {
+  if (!viewShotRef.current) return;
+
+  try {
+    const format = Platform.OS === 'ios' ? 'jpg' : 'png';
```
pete-watters, line 35: Can we use `jpg` for both platforms?

pete-watters (approved):
Great stuff @edgarkhanzadian 🚀

PR #954, opened 2025-03-07
fix: add runes + fallback sip-10 avatars. Hide token symbol, show all…

pete-watters:
### Description

This PR updates the mobile app to

- add an avatar for `runes` and update their text label
- add `sip-10` avatars
- introduce a gradient fallback for missing `sip-10` avatars
- Hide the token symbol in the token widget

### Testing

To test this load the mobile app with a wallet that has `sip-10`s / `runes` 

#### Before
Screenshot 2025-03-07 at 15 55 49
Screenshot 2025-03-07 at 15 55 36
Screenshot 2025-03-07 at 15 55 29

#### After 

https://github.com/user-attachments/assets/fbd67617-034c-4e07-aa7f-33137b947fda

### Notes
- I didn't add tests for this yet as work is on going with `maestro` improvements. I'll follow up
- I changed it to show more tokens on the `home` / `account` view. More work will be done there
- I added a fallback `avatar` to match the extension but will next setup fetching the correct `avatars` from `ALEX`

### DIFF
--- a/apps/mobile/src/components/balance/balance.tsx
+++ b/apps/mobile/src/components/balance/balance.tsx
@@ -6,7 +6,7 @@ import { useLingui } from '@lingui/react';
 import { currencyNameMap } from '@leather.io/constants';
 import { Money } from '@leather.io/models';
 import { BulletSeparator, Text, TextProps } from '@leather.io/ui/native';
-import { formatMoney, i18nFormatCurrency } from '@leather.io/utils';
+import { formatMoneyWithoutSymbol, i18nFormatCurrency } from '@leather.io/utils';
 
 interface BalanceProps extends TextProps {
   balance: Money;
@@ -19,7 +19,7 @@ export function formatBalance(balance: Money, isFiat: boolean) {
     return i18nFormatCurrency(balance, isLargeBalance ? 0 : balance.decimals);
   }
 
-  return formatMoney(balance);
+  return formatMoneyWithoutSymbol(balance);
 }
 
 export function Balance({
--- a/apps/mobile/src/features/balances/stacks/sip10-balance.tsx
+++ b/apps/mobile/src/features/balances/stacks/sip10-balance.tsx
@@ -1,37 +1,51 @@
-import { TokenIcon } from '@/components/widgets/tokens/token-icon';
 import {
   useSip10AccountBalance,
   useSip10TotalBalance,
 } from '@/queries/balance/sip10-balance.query';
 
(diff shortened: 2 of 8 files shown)


apps/mobile/src/components/balance/balance.tsx
```diff
@@ -19,7 +19,7 @@ export function formatBalance(balance: Money, isFiat: boolean) {
     return i18nFormatCurrency(balance, isLargeBalance ? 0 : balance.decimals);
   }
 
-  return formatMoney(balance);
+  return formatMoneyWithoutSymbol(balance);
```
pete-watters, line 22: Changing this as otherwise `Runes` break the whole UI. 

Screenshot 2025-03-07 at 15 55 29

apps/mobile/src/features/balances/stacks/sip10-balance.tsx
```diff
@@ -1,37 +1,51 @@
...
   symbol,
-  ...rest
 }: Sip10TokenBalanceProps) {
   return (
     <TokenBalance
       ticker={symbol}
-      icon={<TokenIcon ticker={symbol} />} // TODO LEA-1909: add images from uri
+      icon={
+        <Avatar
+          image={imageCanonicalUri !== '' ? imageCanonicalUri : getFallbackAvatar(contractId)}
```
pete-watters, line 38: Doing this as sometimes the API returns `''` for the URI

On the `extension` we show them like this with a gradient fallback and a letter
Screenshot 2025-03-07 at 16 03 52

I'll soon update this to use real icons from ALEX which I add to the `sip10` balance service

fbwoolf (approved):

packages/ui/src/components/avatar/runes-avatar-icon.native.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+export function RunesAvatarIcon({ theme }: RunesAvatarIconProps) {
+  const darkFill =
+    theme === 'light'
+      ? colorThemes.base['ink.background-primary']
+      : colorThemes.dark['ink.background-primary'];
+  const lightFill =
+    theme === 'light'
+      ? colorThemes.base['ink.action-primary-default']
+      : colorThemes.dark['ink.action-primary-default'];
```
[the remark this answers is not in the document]
pete-watters: I just saw this now, apologies! I'll try it out in my next PR, thanks

PR #957, opened 2025-03-08
refactor: stacks transactions hooks

fbwoolf:
This PR helps my work in the extension bc I need access to the newer functions to gen Stacks txs. It was pointed out by @pete-watters that the current Stacks txs hooks were unnec and could be simplified to functions instead. I did that refactor here so I can use them in the extension. I tried to rethink/simplify how they were setup originally in the extension code. ~~Now, you can use the more general `generateStacksUnsignedTransaction` with a txType passed in, or you can use individual functions for each tx type on its own. Open to other suggestions here.~~

~~Curious thoughts on the type utils I added in an attempt to handle default/required props better; to pass props now, I separated them into `required` and `optional` instead of using a `Partial<>` type that caused me to have to use defaults for required params.~~

EDIT: I ended up doing a second commit here rethinking the type narrowing. I started thinking just making the one `generateStacksUnsignedTransaction` function was better using the discriminated union type to handle the options. I'm not sure we need to expose anything else.

cc @edgarkhanzadian,the functions here might be helpful for your rpc work?


apps/mobile/ios/Podfile.lock
```diff
@@ -2207,7 +2207,7 @@ SPEC CHECKSUMS:
   segment-analytics-react-native: 53785e35d44a0643beffa40eada68a4cbdf7292e
   SocketRocket: abac6f5de4d4d62d24e11868d7a2f427e0ef940d
   sovran-react-native: 5f02bd2d111ffe226d00c7b0435290eae6f10934
-  Yoga: 348f8b538c3ed4423eb58a8e5730feec50bce372
+  Yoga: b9a182ab00cf25926e7f79657d08c5d23c2d03b0
```
pete-watters, line 2210: 🧘

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -35,16 +39,17 @@ export function useSendFormStx() {
   const {
     params: { account, address: payer, publicKey },
   } = useSendSheetRoute<CurrentRoute>();
+  const { displayToast } = useToastContext();
   const navigation = useSendSheetNavigation<CurrentRoute>();
-  const { networkPreference } = useSettings();
-  const chainId: ChainId = networkPreference.chain.stacks.chainId;
+  const stacksNetwork = useNetworkPreferenceStacksNetwork();
```
pete-watters, line 44: I'd go for a double destructure here:
```
const { networkPreference: { chainId }} = useNetworkPreferenceStacksNetwork();
```

apps/mobile/src/features/send/send-form/hooks/use-send-form-stx.tsx
```diff
@@ -1,13 +1,17 @@
-import { useGenerateStxTokenTransferUnsignedTransaction } from '@/common/transactions/stacks-transactions.hooks';
 import { useToastContext } from '@/components/toast/toast-context';
-import { useSettings } from '@/store/settings/settings';
+import { useNetworkPreferenceStacksNetwork } from '@/store/settings/settings.read';
```
pete-watters, line 2: I didn't know we had this 🫨  I'll refactor some of my other code soon. I was using `useSettings` the whole time

packages/stacks/src/index.ts
```diff
@@ -8,4 +8,4 @@ export * from './transactions/transaction.types';
 
 export * from './validation/stacks-error';
 export * from './validation/address-validation';
-export * from './validation/transaction-validation';
\ No newline at end of file
+export * from './validation/transaction-validation';
```
pete-watters, line 11: Weird it's highlighting this

pete-watters (dismissed):
Great work @fbwoolf thanks for improving this!  🚀 

I'll make an effort to reduce these hooks as we go also. Usually it's to do with us fetching the network in a hook / the current account. 

The more we can decouple functions from hooks in `mono` the better as we will have a much better TS lib then with better tests

packages/stacks/src/transactions/generate-unsigned-transaction.ts
```diff
@@ -24,60 +24,58 @@ export function initNonce(nonce: number | string) {
...
+    nonce: initNonce(options.nonce).toString(),
   };
-
-  return makeUnsignedSTXTokenTransfer({ ...options, ...parsedOptions });
 }
 
-export async function generateStacksUnsignedTransaction(options: StacksUnsignedTransactionOptions) {
-  const { txType } = options;
+export function generateStacksUnsignedTransaction() {
+  return async (options: StacksUnsignedTransactionOptions) => {
```
kyranjamie, line 63: Not sure I fully see the benefit/need of creating a HOF here? There are no outer arguments, so the resulting fn is always the same.

fbwoolf: Good point, I'll change it today.

apps/mobile/ios/Podfile.lock
fbwoolf: I actually don't know what this means, should this be reverted?

kyranjamie (dismissed):

kyranjamie (approved):

apps/mobile/ios/Podfile.lock
pete-watters: I dunno either, I think it's just some package that needed updating. I think it's OK. 

Seems to be `Yoga is an embeddable and performant flexbox layout engine with bindings for multiple languages.` 🤷

PR #961, opened 2025-03-11
ci: remove unused ci jobs

kyranjamie:
Yesterday I added some jobs investigating the best process for secure merges. [Following slack msg](https://trustmachines.slack.com/archives/C05LAP952E7/p1741680965918159?thread_ts=1741623664.745619&cid=C05LAP952E7), and not having signed commits on core branch, they're no longer needed.

Also adds the `all-checks` job used in extension so that we have a single job that checks all others have passed.


pete-watters (approved):
Thanks for removing this. 

Retrospectively I think maybe we should also delete the CrowdIn job? https://github.com/leather-io/mono/pull/956/files 

We have a few un-used actions and it just causes clutter and noise. There is another double job `eas-deploy.yml` that @camerow [promised to remove  ](https://github.com/leather-io/mono/pull/941#pullrequestreview-2666317094)

camerow:
FYI I just removed it as a part of that PR

PR #967, opened 2025-03-11
refactor(analytics)!: wrap event functions and ensure defaults are passed, ref LEA-2288

camerow:


apps/mobile/.env.development
```diff
@@ -1 +1 @@
-export EXPO_PUBLIC_NODE_ENV="development"
```
camerow, line 1: `export` is bash syntax and not valid for env files I don't believe

fbwoolf (approved):

camerow:
> What was the impact of this, we weren't logging any extension event metadata?
> 
> Also, is this a breaking change? Seems like a pretty innocuous fix to do a major bump?

We weren't logging default properties for all events correctly (ie. `platform` was often "Not Set" as a value in Mixpanel. 

I consider any API interface change breaking, we tend to be pretty loose with that but in this case I did change the function definition for one of the methods and had to make code changes to that so technically it is a breaking change.

camerow:
> What was the impact of this, we weren't logging any extension event metadata?
> 
> Also, is this a breaking change? Seems like a pretty innocuous fix to do a major bump?

@kyranjamie I added a couple changes so need re-review. Also changed the commit to `feat`

PR #969, opened 2025-03-12
feat: add ALEX avatars as fallback, ref LEA-2264

pete-watters:


packages/services/src/assets/stacks-asset.utils.ts
```diff
@@ -85,8 +87,39 @@ export function createSip10CryptoAssetInfo(
...
+      'createSip10ImageCanonicalUri metadata',
+      metadata.image_canonical_uri,
+      assetName,
+      tokenInfoFromAlexLabs?.icon
+    );
+    return tokenInfoFromAlexLabs?.icon ?? '';
+  }
+  return metadata.image_canonical_uri;
+}
+// TODO: use this to fetch data from alex lab once determined where to call it from
```
pete-watters, line 117: @alexp3y : I was working on adding this to patch in where we are missing some token URIs. Sometimes the `HiroFtMetadataResponse` contains empty / undefined `metadata.image_canonical_uri`.

Can you recommend a sensible place in the service to:
- call `fetchDataFromAlexLab` and fetch their JSON once 
- cache the result in a `const` 
- use it as a lookup object in `createSip10CryptoAssetInfo` to add in missing avatars

PR #972, opened 2025-03-13
fix: sip10 balance service error handling

alexp3y:
Adds improved error handling to SIP10 and Runes balance services that prevent top-level failures of account and total balance service calls when any individual token balance lookup fails (e.g. failures during asset info + market data fetching).

This fixes a bug preventing the display of sip10 tokens and the incorrect summing of total balance.

PR also adds:
* Filtering of SIP10 token lists to remove zero-balance tokens. 
* Sort function to order SIP10 and Runes token balance lists by available fiat value.


pete-watters (approved):
Great stuff, thanks @alexp3y 🚀

PR #974, opened 2025-03-13
ci: enforce gpg commits for leather-bot

pete-watters:
I made a start on adding GPG commits for leather-bot but I don't have access to add a secret. Can you please help me finish this @camerow ?


kyranjamie:
Made another PR with this so we can close

PR #975, opened 2025-03-13
feat(web): add connection flow to app

kyranjamie:
This PR adds the basic connection flow and the basic ui around it. 

**Screencast**

https://github.com/user-attachments/assets/28675325-83dd-4be3-9f85-d114a4d4e66d

I started setting up Redux Toolkit as it'd be nice to remain consistent with the rest of our stack. As mentioned on Slack though, SSR gets in the way here that raises questions whether it's the best choice. Despite our history with atomic state based approaches in the past, they might make more sense here. I'm using Jotai, as it works much better with SSR, but happy to reevaluate whether that's a good choice as we go through the process of adding the rest of the earn flow.

[Demo at this link →](https://leather-web.wallet-6d1.workers.dev/)


alter-eggo (approved):
tested, worked for me👍
btw I think would be great to open switch account modal automatically on switch account btn click

fbwoolf (approved):

PR #976, opened 2025-03-14
refactor: switch to class-based services, ref LEA-2319

alexp3y:
This PR is part of the `@leather.io/services` clean-up. It migrates the package from factory-functions to class-based services. It also upgrades the InversifyJS dependency to the latest version.

With class-based services, the DI container's auto-bind feature can now be used, removing the reliance on boilerplate config to specify service dependencies. Service interfaces are now also obsolete as the classes themselves become the service type defining the public interface.

The result is a greatly simplified and slimmed-down services layer that takes full advantage of the DI container framework. Using services now simply requires annotating each `class` with `@injectable()` and passing dependent services as constructor arguments.

### DIFF
--- a/packages/services/src/inversify.config.ts
+++ b/packages/services/src/inversify.config.ts
@@ -1,320 +1,90 @@
-import { Container } from 'inversify';
+import { Container, Newable } from 'inversify';
 
-import { ActivityService, createActivityService } from './activity/activity.service';
-import { RuneAssetService, createRuneAssetService } from './assets/rune-asset.service';
-import { Sip9AssetService, createSip9AssetService } from './assets/sip9-asset.service';
-import { Sip10AssetService, createSip10AssetService } from './assets/sip10-asset.service';
-import { BtcBalancesService, createBtcBalancesService } from './balances/btc-balances.service';
-import {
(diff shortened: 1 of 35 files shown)


packages/services/src/inversify.types.ts
```diff
@@ -0,0 +1,4 @@
+export const TYPES = {
```
kyranjamie, line 1:
```suggestion
export const Types = {
```

No all caps naming needed here imo, should effectively use enum convention naming, pascal like the properties 🧘🏼‍♂️

packages/services/src/inversify.config.ts
```diff
@@ -1,320 +1,90 @@
...
 }
 
-export function getContainer(): Container {
+export function getServicesContainer() {
   if (!servicesContainer) {
-    throw new Error('Containers must be initialized before accessing');
+    throw new Error('Container must be initialized before accessing');
   }
   return servicesContainer;
 }
```
kyranjamie, line 48: Huge improvement removing this config. Honestly found this very off putting.

kyranjamie (approved):

pete-watters (approved):
🚀

PR #977, opened 2025-03-14
Feat/lea 2285/activity UI

pete-watters:


pete-watters:
@tigranpetrossian @alexp3y @edgarkhanzadian : adding you as reviewers as I'll be OOO. I need to review it myself but please flag any thing you notice! thanks!

apps/mobile/src/app/(home)/activity/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+        const { type, timestamp } = activity;
+        const activityTitle = formatActivityType(type);
+        const activityDate = dayjs(timestamp * 1000).format('MMM D, YYYY');
+        switch (type) {
+          case 'sendAsset':
+          case 'receiveAsset':
+            const { asset, value, txid } = activity;
+            return (
+              <ActivityCell
+                key={`activity.${index}`}
```
tigranpetrossian, line 34: Is there any piece of unique data we could use as `key` here? If not, perhaps it the activity service should attach uuid-s to those? Index as key will become a footgun if we attempt to virtualize this list, or reshuffle based on filters.

apps/mobile/src/app/(home)/activity/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { data: activity, isLoading } = useTotalActivityQuery(accounts);
+
+  const { mode } = useCurrentNetworkState();
+  // TODO LEA-1726: Handle loading and error states
+  if (isLoading) {
+    return null;
+  }
+  return (
+    <ActivityLayout>
+      {activity?.map((activity, index) => {
```
tigranpetrossian, line 24: I think this list and the Tokens list desperately need virtualization. I could tell from the screencast there's a significant freeze when opening the page, and the same is true with the token list now that we display sip-10 assets there. I've only used FlatList before, but the industry standard seems to be [FlashList](https://shopify.github.io/flash-list/) now.

apps/mobile/src/app/(home)/activity/components/activity-card.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onPress?(): void;
+  onLongPress?(): void;
+  testID?: string;
+  iconTestID?: string;
+}
+
+export function ActivityCard({ children, icon, onPress, ...props }: ActivityCardProps) {
+  const { pressed, onPressIn, onPressOut } = usePressedState(props);
+  const animatedStyle = useAnimatedStyle(() => ({
+    transform: [{ scale: withSpring(pressed ? 0.95 : 1) }],
```
tigranpetrossian, line 18: Our Pressable now accepts declarative presseEffect config. This can be replaced with:
```tsx
<Pressable         
  pressEffects={{
    opacity: { from: 1, to: 0.95 },
  }}
/>

```

apps/mobile/src/app/(home)/account/[account]/activity.tsx
```diff
@@ -0,0 +1,72 @@
...
+                icon={
+                  <ActivityIcon
+                    type={activity.type}
+                    ticker={'symbol' in activity.asset ? activity.asset.symbol : ''}
+                    protocol={activity.asset.protocol}
+                  />
+                }
+                title={formatActivityType(activity.type)}
+                date={dayjs(activity.timestamp * 1000).format('MMM D, YYYY')}
+                amount={isTransferActivity ? activity.value?.crypto : undefined}
```
Comment (tigranpetrossian) [inline apps/mobile/src/app/(home)/account/[account]/activity.tsx:43]: This check is redundant, activity should already be narrowed down to `SendAssetActivity | ReceiveAssetActivity` within this case block, and the fallback for `value?.crypto` and `value?.fiat`is already undefined.

apps/mobile/src/app/(home)/account/[account]/activity.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { fingerprint, accountIndex } = deserializeAccountId(accountId);
+  const accountAddresses = useAccountAddresses(fingerprint, accountIndex);
+  const { data: activity, isLoading } = useAccountActivityQuery(accountAddresses);
+
+  if (isLoading) {
+    return null;
+  }
+  return (
+    <ActivityLayout>
+      {activity?.map((activity, index) => {
```
Comment (tigranpetrossian) [inline apps/mobile/src/app/(home)/account/[account]/activity.tsx:26]: There seem to be two variations of this list, would it make sense to extract it into a single configurable ActivityList component that encapsulates the selection logic, virtualization, etc?

apps/mobile/src/app/(home)/account/[account]/activity.tsx
```diff
@@ -0,0 +1,72 @@
...
+                key={`activity.${index}`}
+                icon={
+                  <ActivityIcon
+                    type={activity.type}
+                    ticker={'symbol' in activity.asset ? activity.asset.symbol : ''}
+                    protocol={activity.asset.protocol}
+                  />
+                }
+                title={formatActivityType(activity.type)}
+                date={dayjs(activity.timestamp * 1000).format('MMM D, YYYY')}
```
Comment (tigranpetrossian) [inline apps/mobile/src/app/(home)/account/[account]/activity.tsx:42]: This seems to be repeated for different kinds of items, I think formatting the date could be the responsibility of the cell itself.

tigranpetrossian (approved):
Looks really good @pete-watters. Left few notes, mostly refactoring. 
I think virtualization is an absolute must, but if there's time constraints it could be done in a follow up. Ping me any time if I can be of help with that.

pete-watters:
> Looks really good @pete-watters. Left few notes, mostly refactoring. I think virtualization is an absolute must, but if there's time constraints it could be done in a follow up. Ping me any time if I can be of help with that.

Thanks @tigranpetrossian 👍 lets team up and work on virtualization / a performance plan. Alex has some work ongoing with workers that should make things faster but we will probably need to still improve it further

apps/mobile/src/app/(home)/activity/index.tsx
alexp3y: I might push back on that one. UUIDs won't persist and would misleading to use as an ID on activity objects. It could lead to bigger problems if we consider them so.

apps/mobile/src/app/(home)/activity/index.tsx
pete-watters: I'll take a look at FlashList, thanks! @alexp3y is working on improving the load times in his worker so I will kick this down the road until he's finished with that and then we should team up and virtualise this.

Right now it takes ~ 5 seconds to get the activity data ready so we can improve on that first and then add virtualisation on top

alexp3y: @tigranpetrossian The latency here and on SIP-10 asset lists is largely due to token metadata fetching. We should address that as the low-hanging fruit before we reach for other solutions. FYI [LEA-2316](https://linear.app/leather-io/issue/LEA-2316/add-consolidated-token-metadata-endpoints)

apps/mobile/src/app/(home)/activity/index.tsx
pete-watters: We don't have something unique unfortunately. I started adding a `uuid` to the service but I double checked with @alexp3y  and he recommended against that as it won't persist and will change each time it refetches so its misleading to consider it an ID. 

Right now, we are loading the data as fast as we can but I believe we can make this faster via Alex's workers API. 

I tried to 'improve' what we have now by making a key with `activity.timestamp` and `activity.type` but still hit non-unique key messages so I will hold off for now and leave it at `index`. 

We can return and optimise then once the workers API updates are included

apps/mobile/src/app/(home)/activity/index.tsx
tigranpetrossian: These two are orthogonal issues. I'm fairly certain the latency when tapping "My Tokens" or "Acivity" (takes a little time before navigating) is due to rendering bottleneck–not fetching data, there are simply too many items. Fetching latency doesn't result in lag between tapping a button and navigating between screens, but rather displays the screen's loading state for a long time.

apps/mobile/src/app/(home)/activity/utils/format-activity.ts
```diff
@@ -0,0 +1,53 @@
+import { t } from '@lingui/macro';
+
+import { ActivityType } from '@leather.io/models';
+import { match } from '@leather.io/utils';
+
+export function formatActivityType(activityType: ActivityType) {
+  return match<ActivityType>()(activityType, {
```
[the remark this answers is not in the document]
pete-watters: I've fixed this, thanks for pointing it out!

apps/mobile/src/app/(home)/activity/index.tsx
pete-watters: Lets fix both issues here - improve load-time and introduce FlashList

apps/mobile/src/app/(home)/account/[account]/activity.tsx
Comment (pete-watters) [reply apps/mobile/src/app/(home)/account/[account]/activity.tsx]: I refactored this whole thing a lot, thanks for helping

pete-watters:
Thanks for the review @tigranpetrossian 👍 I've fixed the things you mentioned but if I missed something let me know and I'll revisit. 

We can work together on the virtualization next week

PR #978, opened 2025-03-15
feat(mobile): add device id to settings

tigranpetrossian:
Adds a copiable device ID to settings screen. This will let testers share their ID to get access to in-progress features.

### DIFF
--- a/apps/mobile/src/hooks/use-copy-to-clipboard.ts
+++ b/apps/mobile/src/hooks/use-copy-to-clipboard.ts
@@ -1,0 +1,34 @@
+import { useCallback, useState } from 'react';
+
+import * as Clipboard from 'expo-clipboard';
+
+const defaultSuccessStateDuration = 2000;
+
+export type UseCopyToClipboardStatus = 'idle' | 'success' | 'error';
+
+interface UseCopyToClipboardOptions {
+  successStateDuration?: number;
+}
+
+export function useCopyToClipboard(value: string, options: UseCopyToClipboardOptions = {}) {
+  const { successStateDuration = defaultSuccessStateDuration } = options;
+  const [status, setStatus] = useState<UseCopyToClipboardStatus>('idle');
+
+  const onCopy = useCallback(async () => {
+    if (value.trim() === '') return;
+
+    try {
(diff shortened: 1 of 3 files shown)


alexp3y (approved):
PR description would benefit from a bit of added context but otherwise it LGTM.

tigranpetrossian:
> PR description would benefit from a bit of added context but otherwise it LGTM.

@alexp3y True–curse of knowledge. Updated.

apps/mobile/src/hooks/use-device-id.ts
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    if (!deviceIdCache) {
+      deviceIdCache = getDeviceId();
+    }
+
+    deviceIdCache.then(id => setValue(id)).catch(() => setValue(null));
+  }, []);
+
+  return value;
+}
```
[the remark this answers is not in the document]
tigranpetrossian: Good catch. This is probably a leftover from random experiments while trying to make the presence of the ID guaranteed. Will remove.

apps/mobile/src/hooks/use-copy-to-clipboard.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    try {
+      await Clipboard.setStringAsync(value);
+      setStatus('success');
+    } catch {
+      setStatus('error');
+    }
+  }, [value]);
+
+  setTimeout(() => setStatus('idle'), successStateDuration);
```
kyranjamie, line 28: Unsafe use of `setTimeout`. If component dismounts before this period ends, react will throw an error.

kyranjamie (approved):

apps/mobile/src/hooks/use-copy-to-clipboard.ts
tigranpetrossian: The warning [was removed in React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#react), this is a negligible memory leak, safe to ignore. 
But thanks for catching this–should be in `finally`, no idea why I placed into the render

PR #979, opened 2025-03-17
feat: add stacking queries, LEA-2322

alter-eggo:
This pr migrates necessary stacking queries from earn app


packages/query/src/stacking/stacking.query.ts
```diff
@@ -0,0 +1,140 @@
...
+          senderAddress,
+          network,
+          client,
+        });
+      } else {
+        return Promise.resolve(noneCV());
+      }
+    },
+    ...defaultQueryOptions,
+  } as const;
```
kyranjamie, line 60: could we use `satisfies `UseQueryOptions` here for stronger typing?

`refetchOnRefreshy` is a valid prop here and won't cause type err

packages/query/src/stacking/stacking.query.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+export function useGetAllowanceContractCallersQuery(
+  params: CreateGetTransactionByIdQueryOptionsArgs & ClientParam
+) {
+  return useQuery(createGetAllowanceContractCallersQueryOptions(params));
+}
```
kyranjamie, line 67: Can we not make this per convention of only exporting option configs?

kyranjamie (approved):
Thanks @alter-eggo

PR #984, opened 2025-03-18
Refactor send flow

tigranpetrossian:
This is a complete refactor of the Send flow, addressing lots of outstanding issues, missing features, and cosmetic tweaks. The feature is under a flag really only for easier reviewing, it's enabled globally and will be swapped either within this PR or in an upcoming one.

**Most notable changes:**
* Add validation (dynamic, data-aware schemas) and display error states
* Incorporate the Amount Field component
* Add Send Max
* Match transitions to designs
* Add inline asset picker
* Change the order of screens to: Select Asset -> Select Account

Code-wise, the flow is now decoupled from the sheet it's currently being displayed in, with the ability to pass a pre-selected asset and account and skip the selection altogether. With future QR support, the plan is to launch the form with more preopulated fields.

There's still some amount of duplication between the two forms, that I'm already working on on removing. Still thinking of finding the best point of abstraction that isn't unnecessarily complex, along with enabling easier SIP-10 support, similar to how it's done in the extension.

### Descoped items:

1. **Address picker.** This wasn't in the scope from the get go, but I think this is a crucial feature we should implement ASAP.
2. **Fiat entry:** I've descoped this for this PR, as it turned out to be more complex than expected. It's in progress, and needs incorporating currency & locale support that was added while this work was in progress.
3. **Amount field transitions:** I have a semi-working version of character animations, however without the [new architecture](https://reactnative.dev/blog/2024/10/23/the-new-architecture-is-here#uselayouteffect) improvements, it seems to be impossible to effectively measure layout on time for animations without constantly being a frame behind and having flickers. At the moment I've just given up on this, hoping to pick it up once we upgrade.

### Basic demo with single account:

https://github.com/user-attachments/assets/35284614-1287-4ee2-86fb-3659bc2c792d

### Multiple accounts:

https://github.com/user-attachments/assets/88e260b2-fda2-477c-a4b7-05d6e2959dd4

Closes LEA-2173, LEA-2168, LEA-2167, LEA-2160, LEA-2159, LEA-2158, LEA-2156, LEA-2153, LEA-2152, LEA-2105, LEA-2077, LEA-1873, LEA-1868, LEA-1867, LEA-1854


apps/mobile/src/components/action-bar/action-bar.tsx
```diff
@@ -13,7 +14,7 @@ import { useTheme } from '@shopify/restyle';
 import { BlurView, Box, Theme } from '@leather.io/ui/native';
 
 export const ACTION_BAR_HEIGHT = 70;
-export const ACTION_BAR_BOTTOM_OFFSET = 40;
+export const ACTION_BAR_BOTTOM_OFFSET = 10;
```
kyranjamie, line 17: Thank you! This had been annoying me for so long haha

apps/mobile/src/components/action-bar/action-bar.tsx
```diff
@@ -84,22 +85,21 @@ export const ActionBar = forwardRef<ActionBarMethods, ActionBarProps>(function (
...
+        overflow="hidden"
+        borderColor="ink.border-transparent"
+        borderWidth={1}
+        style={{ marginHorizontal: theme.spacing[5] - 1 }}
+      >
         <BlurView
+          experimentalBlurMethod="dimezisBlurView"
           themeVariant={themeDerivedFromThemePreference}
-          intensity={90}
+          intensity={Platform.OS === 'ios' ? 95 : 80}
```
kyranjamie, line 98: Is this to get to a fairly similar level of blur? Ought this be abstracted to a value so it can be reused outside this component?

apps/mobile/src/components/amount-field/amount-field-primary-value.tsx
```diff
@@ -12,6 +16,57 @@ const commonTextProps: TextProps = {
...
+  fiatCurrency,
+  cryptoCurrency,
+}: AmountFieldPrimaryValueProps) {
+  const [dynamicFontSize, setDynamicFontSize] = useState(maxFontSize);
+  // Maintain the relative lineHeight to prevent text shifting down as font size decreases
+  const staticLineHeight = maxFontSize * lineHeightRatio;
+  const dynamicLineHeight = dynamicFontSize * lineHeightRatio;
+  const displayValue = formatPrimaryValue({
+    value,
+    currency: currencyMode === 'crypto' ? cryptoCurrency : fiatCurrency,
```
kyranjamie, line 42: Best to avoid ternaries of data sets that could possibly be more than 2 (even though unlikely in this case), and I see this ternary repeated in many places.

I like when helpers `whenCurrency(currencyMode)({ fiat: x, crypto: y })`, but other approaches work well too.

apps/mobile/src/components/amount-field/amount-field.utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    minimumFractionDigits: 0,
+    maximumFractionDigits: currencyDecimalsMap[currency],
+  });
+
+  return formatter.format(Number(value));
+}
+
+interface CalculateSecondaryValueParams {
+  value: string;
+  currency: Currency;
```
kyranjamie, line 55: isn't currency implicit based on base value of market data?

apps/mobile/src/features/send-refactor/components/amount-field.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <RawAmountField {...amountFieldProps} />
+      <FieldConnectorArrow />
+    </AnimatedBox>
+  );
+}
+
+const AnimatedBox = Animated.createAnimatedComponent(Box);
+
+function enteringAnimation(): LayoutAnimation {
+  'worklet';
```
kyranjamie, line 32: 👏🏼 for getting shared element transitions working. I failed on my attempt

apps/mobile/src/features/send-refactor/components/asset-picker/asset-picker.tsx
```diff
@@ -0,0 +1,50 @@
...
+  account: Account | null;
+  onSelectAsset(asset: SendableAsset, assetElementOffsetTop: number | null): void;
+}
+
+export function AssetPicker({ account, onSelectAsset }: AssetListProps) {
+  function handleSelectAsset(asset: SendableAsset) {
+    return function (top: number | null) {
+      onSelectAsset(asset, top);
+    };
+  }
```
kyranjamie, line 23:
```suggestion
export function AssetPicker({ account, onSelectAsset }: AssetListProps) {
  function handleSelectAsset(asset: SendableAsset) {
    return (top: number | null) => onSelectAsset(asset, top);
  }
```
?

apps/mobile/src/features/send-refactor/forms/btc/btc-form-schema.ts
```diff
@@ -0,0 +1,105 @@
...
+  isBtcMinimumSpend,
+  isValidBitcoinAddress,
+  isValidBitcoinNetworkAddress,
+} from '@leather.io/bitcoin';
+import { BTC_DECIMALS } from '@leather.io/constants';
+import { AverageBitcoinFeeRates, BitcoinNetworkModes } from '@leather.io/models';
+import { Utxo } from '@leather.io/query';
+import { createMoneyFromDecimal, isValidPrecision } from '@leather.io/utils';
+
+export function useBtcSchema(
```
kyranjamie, line 22:
```suggestion
export function useBtcSendFormSchema(
```
?

apps/mobile/src/features/send-refactor/utils.ts
```diff
@@ -0,0 +1,124 @@
...
+  return input === '0' || /^0\.0*$/.test(input);
+}
+
+export function getInitialRelevantAccounts(accounts: Account[]) {
+  return accounts.filter(account => account.status !== 'hidden');
+}
+
+export function validateDecimalPlaces(value: string, maxDecimalPlaces: number) {
+  return (value.split('.')[1] ?? '').length <= maxDecimalPlaces;
+}
```
kyranjamie, line 63: unit tests?

apps/mobile/src/features/send-refactor/utils.ts
```diff
@@ -0,0 +1,124 @@
...
+      ],
+      feeRate,
+      isSendingMax,
+      utxos: createCoinSelectionUtxos(utxos),
+      bip32Derivation: [payerToBip32Derivation(nativeSegwitPayer)],
+      network: getBtcSignerLibNetworkConfigByMode(networkMode),
+    });
+    return Promise.resolve(bytesToHex(tx.psbt));
+  } catch (error) {
+    return Promise.reject(error instanceof Error ? error : new Error(String(error)));
```
kyranjamie, line 91: we have an `isError` helper for this type narrowing

packages/stacks/src/validation/memo-validation.ts
```diff
@@ -0,0 +1,9 @@
+import { MEMO_MAX_LENGTH_BYTES } from '@stacks/transactions';
+
+function exceedsMaxLengthBytes(string: string, maxLengthBytes: number): boolean {
+  return string ? Buffer.from(string).length > maxLengthBytes : false;
+}
```
kyranjamie, line 5:
```suggestion
function exceedsMaxLengthBytes(value: string, maxLengthBytes: number): boolean {
  return value ? Buffer.from(value).length > maxLengthBytes : false;
}
```

something doesn't sit right about using a var name of a primitive type.

kyranjamie (commented):
Fantastic work @tigranpetrossian. Loads to review here, I've skimmed mostly.

fbwoolf:
I can't say I totally understand the overall send form refactor here? Why remove the context/provider setup and opt for prop drilling? I find it more difficult to follow these changes and it will be harder to scale if we add more blockchains? Can you help me understand why this is better?

tigranpetrossian:
> I can't say I totally understand the overall send form refactor here? Why remove the context/provider setup and opt for prop drilling? I find it more difficult to follow these changes and it will be harder to scale if we add more blockchains? Can you help me understand why this is better?

Every time I attempt to explain this in text it comes out as ambiguous rambling, I meant to do this in the last eng sync but ran out of time. Would it be helpful if we go through this over a call?

apps/mobile/src/components/action-bar/action-bar.tsx
tigranpetrossian: I thought of this, but there doesn't seem to be a common denominator between these, so it's just an eyeballed value for this component. I could extract it into `getPlatformSpecificActionBarBlurIntensity` to imply the intention better, until we look into Skia for all our filtering/canvas needs.

apps/mobile/src/components/action-bar/action-bar.tsx
tigranpetrossian: Same. The plan is the improve this a bit further in the future: [LEA-2369: Adjust the positioning of the action bar](https://linear.app/leather-io/issue/LEA-2369/adjust-the-positioning-of-the-action-bar)

apps/mobile/src/components/amount-field/amount-field.utils.ts
tigranpetrossian: It is, this is now outdated, thanks for pointing this out. I'm also realizing dynamic fiat currency support **has to be** plugged into the form within this PR (it's currently static) irrespective of Fiat input support that comes later. Working on this.

fbwoolf:
> @fbwoolf Every time I attempt to explain this in text it comes out as ambiguous rambling, I meant to do this in the last eng sync but ran out of time. Would it be helpful if we go through this over a call?

MIght be best in our next eng sync for all to hear, or happy to do a call ...but I always feel calls that are 1:1 just leave people out of the loop on things. cc @camerow for maybe doing during eng sync next week?

apps/mobile/src/features/send-refactor/utils.ts
```diff
@@ -0,0 +1,125 @@
+import { type BtcFormSchema } from '@/features/send-refactor/forms/btc/btc-form-schema';
+import { type StxFormSchema } from '@/features/send-refactor/forms/stx/stx-form-schema';
+import { type Account } from '@/store/accounts/accounts';
+import { bytesToHex } from '@noble/hashes/utils';
+import { type PublicKey } from '@stacks/common';
+import { ChainId, type StacksNetwork } from '@stacks/network';
+import BigNumber from 'bignumber.js';
+import memoize from 'p-memoize';
```
pete-watters, line 8: Is this `p-memoize` library truly necessary? It seems like over-engineering to add a new library to memoize validation of address compliance. 

What happens if an address suddenly becomes non-compliant, how will we know if the address is still the same?

apps/mobile/src/features/send-refactor/utils.ts
```diff
@@ -0,0 +1,125 @@
+import { type BtcFormSchema } from '@/features/send-refactor/forms/btc/btc-form-schema';
```
pete-watters, line 1: This `utils` file seems to be a dumping ground of multi chain functions. 

Chain specific code should live in the respective package - `leather.io/stacks` or `leather.io/bitcoin` .

`validateAddressCompliance` shouldn't mix checks on stacks and bitcoin address

apps/mobile/src/components/amount-field/amount-field.tsx
```diff
@@ -1,45 +1,53 @@
 import { AmountSendMaxButton } from '@/components/amount-field/amount-send-max-button';
 
+import { CryptoCurrency, FiatCurrency, MarketData } from '@leather.io/models';
 import { Box, Theme } from '@leather.io/ui/native';
 
 import { AmountFieldPrimaryValue } from './amount-field-primary-value';
 import { AmountFieldSecondaryValue } from './amount-field-secondary-value';
 
-type CurrencyMode = 'crypto' | 'fiat'; // TODO: Should be moved into containing form types
+type CurrencyMode = 'crypto' | 'fiat';
```
pete-watters, line 9: This type is duplicated in your temporary types. 

I think that this PR is difficult to review as you still have a lot of files in a new folder - `send-refactor` and you're not yet deleting the things that they are replacing. 

This makes it difficult to tell if it's really improving things or if there are some things that are missing

apps/mobile/src/components/amount-field/amount-field.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+import { baseCurrencyAmountInQuote, createMoneyFromDecimal } from '@leather.io/utils';
+
+interface FormatPrimaryValueParams {
+  value: string;
+  currency: Currency;
+  showCurrency: boolean;
+  locale: string;
+}
+
+export function formatPrimaryValue({
```
pete-watters, line 15: It seems like `formatPrimaryValue` and `formatSecondaryValue` could be the same function with a little refactoring?

apps/mobile/src/components/amount-field/amount-field.utils.ts
```diff
@@ -0,0 +1,77 @@
+import { decimalSeparator } from '@/features/send-refactor/temporary-constants';
+import { type CurrencyMode } from '@/features/send-refactor/types';
+
+import { currencyDecimalsMap } from '@leather.io/constants';
+import { type Currency, type MarketData, type Money } from '@leather.io/models';
+import { baseCurrencyAmountInQuote, createMoneyFromDecimal } from '@leather.io/utils';
+
```
pete-watters, line 7: We need to add unit tests for these amount functions

apps/mobile/src/features/send-refactor/components/asset-picker/asset-picker-item.tsx
```diff
@@ -0,0 +1,43 @@
+import { ReactElement, ReactNode, cloneElement, useRef, useState } from 'react';
+import { View } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+
+import { HEADER_HEIGHT } from '@/shared/constants';
+
+const parentTopPadding = 16;
+
+// Wrapper around asset items for measuring and reporting vertical offsets for animation.
```
pete-watters, line 9: I'm trying to understand what's happening here. Is this why you also are adding an `onPress` to `BitcoinBalance` ? 

What's the benefit of doing this vs using `forwardRef` if needed?

apps/mobile/src/features/send-refactor/components/asset-picker/asset-picker.tsx
```diff
@@ -0,0 +1,48 @@
+import {
+  BitcoinBalance,
+  BitcoinBalanceByAccount,
+} from '@/features/balances/bitcoin/bitcoin-balance';
+import { StacksBalance, StacksBalanceByAccount } from '@/features/balances/stacks/stacks-balance';
+import { AssetPickerItem } from '@/features/send-refactor/components/asset-picker/asset-picker-item';
+import { SendableAsset } from '@/features/send-refactor/types';
+import { Account } from '@/store/accounts/accounts';
+
+// Future post-release iteration should include tabs and search:
```
pete-watters, line 10: Can you capture this in a Linear task and get rid of this comment?

apps/mobile/src/features/send-refactor/error-messages.ts
```diff
@@ -0,0 +1,50 @@
...
+export const errorMessages = {
+  notANumber: i18n._({
+    id: 'send.validation.not_a_number',
+    message: 'Not a number',
+  }),
+  notAPositiveNumber: i18n._({
+    id: 'send.validation.greater_than_zero',
+    message: 'Amount must be greater than zero',
+  }),
+  invalidPrecision: i18n._({
```
pete-watters, line 16: Stacks and Bitcoin have different precision so this shouldn't  be tied to `BTC_DECIMALS`. This will show the wrong precision for invalid stacks amounts

apps/mobile/src/features/send-refactor/forms/stx/stx-form.tsx
```diff
@@ -0,0 +1,159 @@
...
+  fiatCurrency,
+  marketData,
+  nonce,
+  onOpenAssetPicker,
+  account,
+}: StxFormProps) {
+  const {
+    state: { currencyMode },
+  } = useSendFlowContext();
+  const currency = currencyMode === 'crypto' ? symbol : fiatCurrency;
```
pete-watters, line 50: This code seems to get repeated a lot
```
const currency = currencyMode === 'crypto' ? symbol : fiatCurrency;
```

Maybe it's worthy of a lookup helper?

apps/mobile/src/features/send-refactor/send-sheet.tsx
```diff
@@ -0,0 +1,24 @@
+import { useSheetNavigatorContext } from '@/common/sheet-navigator/sheet-navigator-provider';
+import { FullHeightSheet } from '@/components/full-height-sheet/full-height-sheet';
+import { Send } from '@/features/send-refactor/send';
+
+import { useHaptics } from '@leather.io/ui/native';
+
+// TODO: Ideally this should live on an upper level. Send as a feature should be agnostic of the environment
```
pete-watters, line 7: If you think we should do this, lets create a Linear task to capture it and ditch the comment. 

Otherwise we are throwing more `TODO`s on the pile

apps/mobile/src/features/send-refactor/send-flow-provider.tsx
```diff
@@ -0,0 +1,104 @@
+import { ReactNode, createContext, useContext, useReducer } from 'react';
```
pete-watters, line 1: Nice work here 👍  I like the `useReducer` pattern. It would be good to add some tests to cover state being set correctly

apps/mobile/src/features/send/send-sheet.tsx
```diff
@@ -1,11 +1,19 @@
...
 import { useSheetNavigatorContext } from '@/common/sheet-navigator/sheet-navigator-provider';
 import { FullHeightSheet } from '@/components/full-height-sheet/full-height-sheet';
+import { SendSheet as RefactoredSendSheet } from '@/features/send-refactor/send-sheet';
+import { useBoolVariation } from '@launchdarkly/react-native-client-sdk';
 
 import { SendSheetNavigator } from './send-sheet-navigator';
 
 export function SendSheet() {
   const { sendSheetRef } = useSheetNavigatorContext();
+  const sendRefactorEnabled = useBoolVariation('send_refactor', false);
```
pete-watters, line 11: I think it's now or never. Lets make this change now and remove old code while we're at it or else not make the change

packages/bitcoin/src/coin-selection/calculate-max-spend.ts
```diff
@@ -1,22 +1,24 @@
...
-} from '../coin-selection/coin-selection.utils';
+import type { AverageBitcoinFeeRates, Money } from '@leather.io/models';
+import { createMoney, satToBtc } from '@leather.io/utils';
+
+import { CoinSelectionUtxo } from './coin-selection';
+import { filterUneconomicalUtxos, getSpendableAmount } from './coin-selection.utils';
 
 interface CalculateMaxSpendArgs {
-  recipient: BitcoinAddress;
+  // recipient is intentionally string instead of BitcoinAddress, as it's being validated
```
pete-watters, line 10: I think this validation is done in the `mobile/src` right? So it doesn't make sense to add this comment to the bitcoin package?

If it needs to be a string that's OK but we should keep this package as an independent entity

pete-watters (changes requested):
Nice work here @tigranpetrossian 👍 It's pretty big and I spend some time initially going over it, I'll spend some more time testing it.

I'd recommend:
- squashing your commits / linking with an issue
- for more independent change commits pick them to smaller PRs
- add some more unit tests

I think you've made some good improvements here but as it's feature flagged it's hard to see the impact. That's why I am requesting changes. 

I think we should get rid of the old if we are adding the new as otherwise when will we do that? It makes it hard to tell what's replacing what and what the final architecture will be. Feel free to push back on this but that's how I feel on first glance

apps/mobile/src/features/send-refactor/utils.ts
pete-watters: I think it's related to the `Chainalysis` issue? If so I guess it is needed to stop that 👍

tigranpetrossian: This does reduce the requests to Chainalysis, but more importantly avoids a significant UI lag during subsequent validations.

apps/mobile/src/features/send-refactor/utils.ts
tigranpetrossian: I agree. some are there mostly due to my lack of knowledge on how to promote them into library-level helpers, e.g. `btcFormValuesToPsbtHex`is form-specific, it was migrated from an inline logic in huge hook from the previous version, but `validateAddressCompliance` is poorly implemented, it should accept a chain-specific address validator as a callback (the way it's done in the extension) instead of baking it in.

apps/mobile/src/components/amount-field/amount-field.utils.ts
tigranpetrossian: There are some significant differences between the two, the primary value is a live input so it should allow trailing zeros, and only shows currency symbol for crypto entry. I'd argue that these should be always separate, but `formatPrimaryValue` is poorly named, hence the confusion. Will revisit.

apps/mobile/src/features/send-refactor/error-messages.ts
tigranpetrossian: Good catch, this is an unfortunate refactoring result.

apps/mobile/src/features/send-refactor/forms/stx/stx-form.tsx
tigranpetrossian: Hate this wholeheartedly. Working on it already.

apps/mobile/src/features/send/send-form/sheets/recipient-sheet.tsx
```diff
@@ -16,6 +16,7 @@ interface RecipientSheetProps {
 export function RecipientSheet<T extends SendFormBaseContext<T>>({
   sheetRef,
 }: RecipientSheetProps) {
+  // XXX: Recipient sheet doesn't need to be aware of the form
```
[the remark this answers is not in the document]
tigranpetrossian: Leftover, will remove.

apps/mobile/src/features/send-refactor/send-sheet.tsx
tigranpetrossian: True. Now that we have several major features in sheets, we need a better way of handling global sheets. Ditch this comment altogether. Tracking here: [LEA-2380: Improve global sheet handling in the mobile app](https://linear.app/leather-io/issue/LEA-2380/improve-global-sheet-handling-in-the-mobile-app)

packages/bitcoin/src/coin-selection/calculate-max-spend.ts
tigranpetrossian: Perhaps the comment is a bit vague, "validated in subroutine" means `getSpendableAmount` validates the address, so it doesn't have to yet be valid at this stage.

apps/mobile/src/components/amount-field/amount-field.tsx
tigranpetrossian: This is intentional, I'd rather not couple the amount field to Send form types, as it will likely be used in at least the Swap form too. I might end up replacing this with a boolean `isFiatEntry` when bringing back the fiat entry.

apps/mobile/src/features/send-refactor/error-messages.ts
tigranpetrossian: ✅

apps/mobile/src/features/send-refactor/send-flow-provider.tsx
tigranpetrossian: I'd love to, but apparently vitest refuses to even test hooks. Did you get a chance at looking into making it work with RN?

apps/mobile/src/features/send/send-form/sheets/recipient-sheet.tsx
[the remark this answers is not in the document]
tigranpetrossian: ✅

apps/mobile/src/components/amount-field/amount-field.utils.ts
tigranpetrossian: Agreed. Kind of related to your Send utils comment. At least the `baseCurrencyAmountInQuote` doesn't belong here. I'll be re-working some of this when adding fiat mode, will add tests then to avoid doing it twice. Added a note in the issue: [LEA-2382: Enable Fiat entry](https://linear.app/leather-io/issue/LEA-2382/enable-fiat-entry)

apps/mobile/src/components/amount-field/amount-field-primary-value.tsx
tigranpetrossian: ✅ Renamed to less vague `InputCurrencyMode`, made both the type and the mapping helper global, as I'm planning to turn this into an implicit user preference.

apps/mobile/src/features/send-refactor/forms/stx/stx-form.tsx
tigranpetrossian: Addressed [here](https://github.com/leather-io/mono/pull/984/files/fd83e6cc86c5f4e8f396de1bd34cca46fe528665#r2009694420)

apps/mobile/src/components/amount-field/amount-field.tsx
tigranpetrossian: In hindsight, made this a global type. Details [here](https://github.com/leather-io/mono/pull/984/files/fd83e6cc86c5f4e8f396de1bd34cca46fe528665#r2009694420)

apps/mobile/src/components/full-height-sheet/full-height-sheet.tsx
```diff
@@ -1,14 +1,13 @@
...
 
-import { Sheet, SheetRef } from '@leather.io/ui/native';
+import { Sheet, SheetProps, SheetRef } from '@leather.io/ui/native';
 
-interface FullHeightSheetProps extends HasChildren {
+interface FullHeightSheetProps extends Omit<SheetProps, 'themeVariant'> {
   sheetRef: RefObject<SheetRef>;
 }
-export function FullHeightSheet({ children, sheetRef }: FullHeightSheetProps) {
+export function FullHeightSheet({ sheetRef, ...sheetProps }: FullHeightSheetProps) {
```
pete-watters, line 10: This is great farewell children

apps/mobile/src/components/sheets/sheet.layout.tsx
```diff
@@ -1,22 +1,21 @@
-import { ReactNode, RefObject } from 'react';
+import { ReactElement, RefObject } from 'react';
```
pete-watters, line 1: I haven't seen `ReactElement` before, nice! Is that something you recommend we use more? 

Like a better `JSX.Element`?

tigranpetrossian: Depends on the use case. ReactElement is a subset of ReactNode–basically just JSX, whereas ReactNode can be a literal string, number, null, etc. Icon doesn't expect literal values, so it made sense to narrow it down here.

apps/mobile/src/features/balances/stacks/stacks-balance.tsx
```diff
@@ -22,18 +26,24 @@ export function StacksTokenBalance({ availableBalance, fiatBalance }: StacksToke
...
       fiatBalance={fiatBalance}
       availableBalance={availableBalance}
+      {...rest}
     />
   );
 }
 
-export function StacksBalance() {
+interface StacksBalanceProps {
+  onPress?(): void;
```
pete-watters, line 35: Same as `BitcoinBalance` do we need this `onPress` ? Apologies if we do and you mentioned it before

apps/mobile/src/features/balances/stacks/stacks-balance.tsx
```diff
@@ -10,7 +10,11 @@ interface StacksTokenBalanceProps extends PressableProps {
   availableBalance: Money;
   fiatBalance: Money;
 }
-export function StacksTokenBalance({ availableBalance, fiatBalance }: StacksTokenBalanceProps) {
+export function StacksTokenBalance({
+  availableBalance,
+  fiatBalance,
+  ...rest
```
pete-watters, line 16: I don't think we need this `...rest` as we don't need `onPress` and if we do maybe just pass that prop instead?

apps/mobile/src/features/balances/stacks/stacks-balance.tsx
pete-watters: Ah I remember it's related to this now right? 

https://github.com/leather-io/mono/pull/984/files#r2004902921

tigranpetrossian: The onPress is definitely used, all those asset items need to be interactive in many flows.

apps/mobile/src/features/send-refactor/forms/btc/btc-form.tsx
```diff
@@ -0,0 +1,153 @@
...
+  type AverageBitcoinFeeRates,
+  type FiatCurrency,
+  type MarketData,
+  type Money,
+} from '@leather.io/models';
+import { type Utxo } from '@leather.io/query';
+import { BtcAvatarIcon, Button } from '@leather.io/ui/native';
+import { isNumber } from '@leather.io/utils';
+
+const protocol = 'nativeBtc';
```
pete-watters, line 29: Maybe we can use `CryptoAssetProtocols` from `models` for this? I guess we should have a more shared `symbol = 'BTC'` also

apps/mobile/src/features/send-refactor/forms/btc/btc-form.tsx
```diff
@@ -0,0 +1,153 @@
...
+            onBlur={onBlur}
+            invalid={invalid}
+            isDirty={isDirty}
+            isTouched={isTouched}
+            error={error}
+          />
+        )}
+      />
+
+      {!form.formState.isValidating ? (
```
pete-watters, line 118: Should this also have a check to only render if `isValidating` and `isError`?

apps/mobile/src/features/balances/stacks/stacks-balance.tsx
tigranpetrossian: Its type signature initially extended PressableProps, so I think it's only fair that it passes down what's defined, and it enabled expected behavior. It's probably worth revisiting these components both for polishing type signatures, but also improving the naming (these aren't exactly Balance components), and maybe trimming down the variations to just 1-2. But that's way out of scope for this PR.

apps/mobile/src/features/send-refactor/components/send-form-layout.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export function SendFormContainer(props: BoxProps) {
+  return <Box px="5" gap="3" flex={1} {...props} />;
+}
+
+export function SendFormFooter(props: BoxProps) {
+  return <Box flex={1} gap="3" justifyContent="flex-end" {...props} />;
+}
+
+export function SendFormLoadingSpinner() {
```
pete-watters, line 13: Maybe this can be more generic as we need it in most places? I can refactor it when revisiting [LEA-1726](https://linear.app/leather-io/issue/LEA-1726/handle-balance-loading-state-elegantly)

apps/mobile/src/features/send-refactor/forms/btc/use-btc-form.ts
```diff
@@ -0,0 +1,82 @@
...
+  const handleSubmit = form.handleSubmit(values =>
+    btcFormValuesToPsbtHex(
+      values,
+      nativeSegwit.derivePayer({ addressIndex: 0 }),
+      utxos,
+      networkPreference.chain.bitcoin.mode
+    )
+      .then((psbtHex: string) => navigate('approval', { hex: psbtHex }))
+      .catch(() =>
+        displayToast({
```
pete-watters, line 66: Maybe we can throw the error here and it can be caught by `formatBitcoinError` / `formatStacksError` ?

apps/mobile/src/features/send-refactor/forms/stx/stx-form-schema.ts
```diff
@@ -0,0 +1,118 @@
...
+      .refine(stxComplianceValidator(chainId), errorMessages.nonCompliantAddress),
+
+    memo: z.string().refine(isValidStacksMemo, errorMessages.memoExceedsLimit),
+    nonce: z.number(),
+    fee: z.number(),
+    isSendingMax: z.boolean(),
+  });
+}
+
+function stxAddressNetworkValidator(chainId: ChainId) {
```
pete-watters, line 58: It could be good to move these validators into `@leather.io/stacks`? We can do it in a future task. 

That would make it easier to share this with the `extension` and replace `yup` with `zod`

apps/mobile/src/features/send-refactor/forms/btc/btc-form.tsx
tigranpetrossian: Other displaying heuristics are baked into the Component, there's no reason why `isValidating` shouldn't be too. Thanks, fixed.

apps/mobile/src/features/send-refactor/components/send-form-layout.tsx
tigranpetrossian: That would very nice. I intentionally kept it scoped to this flow for now because building a decent global page-level loading state will take a little more than this.

apps/mobile/src/features/send-refactor/forms/btc/use-btc-form.ts
tigranpetrossian: Generally not a good idea to throw errors in the render cycle. What would be the benefit of that?

apps/mobile/src/features/send-refactor/hooks/use-calculate-btc-max-spend.ts
```diff
@@ -0,0 +1,27 @@
...
+
+interface CalculateBtcMaxSpendParams {
+  recipient: string;
+  feeRate: number;
+}
+
+export type CalculateBtcMaxSpend = ReturnType<typeof useCalculateBtcMaxSpend>;
+
+export function useCalculateBtcMaxSpend(feeRates: AverageBitcoinFeeRates, utxos: Utxo[]) {
+  return useCallback(
```
pete-watters, line 17: Does `useCallback` give a big benefit here? 

[Kent](https://kentcdodds.com/blog/usememo-and-usecallback) generally says to not bother

apps/mobile/src/features/send-refactor/forms/stx/stx-form.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { locale } from '@/features/send-refactor/temporary-constants';
+import { Account } from '@/store/accounts/accounts';
+import { whenInputCurrencyMode } from '@/utils/when-currency-input-mode';
+import { t } from '@lingui/macro';
+
+import { FiatCurrency, MarketData, Money } from '@leather.io/models';
+import { Button, StxAvatarIcon } from '@leather.io/ui/native';
+import { isNumber } from '@leather.io/utils';
+
+const protocol = 'nativeStx';
```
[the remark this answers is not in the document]
tigranpetrossian: ✅

apps/mobile/src/features/send-refactor/screens/form.tsx
```diff
@@ -0,0 +1,118 @@
...
+              id: 'select_asset.header_subtitle',
+              message: '{subtitle}',
+              values: { subtitle: selectedAccount.name },
+            })}
+            leftElement={canGoBack() ? <HeaderBackButton onPress={goBack} /> : null}
+            rightElement={<NetworkBadge />}
+          />
+        }
+      >
+        {/* TODO: Use pattern matching once we add support for remaining tokens */}
```
pete-watters, line 67: I wouldn't bother with these comments as I'm not sure when we will do this and can figure it out when we do

apps/mobile/src/features/send-refactor/screens/select-asset.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export function SelectAsset() {
+  const { navigate } = useSendNavigation();
+  const route = useSendRoute<'select-asset'>();
+  const canGoBack = route.params?.previousRoute === 'select-account';
+  const {
+    state: { selectedAccount },
+    selectAsset,
+  } = useSendFlowContext();
+  // Preload relevant data to ensure smooth transition animation to form screen
```
pete-watters, line 22: Is this to get around other lags we have? 

I'm not sure if it's good to always preload data from a different chain if we won't always use it?

apps/mobile/src/features/send-refactor/hooks/use-calculate-btc-max-spend.ts
tigranpetrossian: The form is extremely dynamic with state updates propagating on each "key stroke" (whatever the mobile equivalent is). This limits the callback being recreated only when remote data changes. I'd say we probably need more `useCallback`s than fewer.

apps/mobile/src/features/send-refactor/send-flow-provider.tsx
pete-watters: You can do it with `vi.mock` I think. Like how we mocked `useSelector` in `stacks-keychains.read.spec.ts` :
```
vi.mock('react-redux', () => ({
  useSelector: vi.fn(),
}));
```

We can re-visit this in the cooldown, I added a note [here](https://linear.app/leather-io/issue/LEA-2079)

apps/mobile/src/features/send-refactor/utils.ts
pete-watters: Can you please add tests for these functions?

apps/mobile/src/utils/types.ts
```diff
@@ -1,5 +1,7 @@
 import { ReactNode } from 'react';
 
+export type InputCurrencyMode = 'crypto' | 'fiat';
```
pete-watters, line 3: Maybe this should be in `models` ? I'm not sure though . It could be more `CurrencyFormat` or something like that

packages/bitcoin/src/validation/transaction-validation.ts
```diff
@@ -33,12 +33,12 @@ export function isValidBitcoinTransaction({
...
   }
 
-  if (!isBtcMinimumSpend({ amount })) {
+  if (!isBtcMinimumSpend(amount)) {
     throw new BitcoinError('InsufficientAmount');
   }
 
   const { amount: spendable } = calculateMaxSpend({ recipient, utxos, feeRate, feeRates });
-  if (!isBtcBalanceSufficient({ amount, spendable })) {
+  if (!isBtcBalanceSufficient({ desiredSpend: amount, maxSpend: spendable })) {
```
pete-watters, line 41: I'm not sure about this arg renaming as now you need to specify here also.  Maybe it makes it easier to reason with and aligns with the extension?

apps/mobile/src/utils/types.ts
tigranpetrossian: This is an app-specific type that will be bound to user's implicit preference. I don't think it's in any way related to models.

packages/bitcoin/src/validation/transaction-validation.ts
tigranpetrossian: I borrowed desiredSpend & maxSpend from the extension yeah. I found the old naming to be very confusing to be honest. Missed this usage point though, will fix the naming here.

apps/mobile/src/features/send-refactor/types.ts
```diff
@@ -0,0 +1 @@
+export type SendableAsset = 'btc' | 'stx';
```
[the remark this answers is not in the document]
tigranpetrossian: Not sure this is a model. Some types are just local to a flow they're defined in. This will likely be scraped once we add support for more assets/collectibles.

pete-watters (commented):
Great work here @tigranpetrossian . I've finished the code review and it LGTM apart from some small suggestions I made. 

A thing necessary to do is add some unit tests [here](https://github.com/leather-io/mono/pull/984/files#r2000956254). 

I'll do some local testing now

apps/mobile/src/features/send-refactor/forms/btc/use-btc-form.ts
pete-watters: It's just to keep consistency with how we are handling all the other errors and to avoid repeating `displayToast`. 

Maybe we need to do more refactoring to deprecate `formatBitcoinError` instead?

apps/mobile/src/features/send-refactor/types.ts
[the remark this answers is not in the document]
pete-watters: OK. I just think we have enough usages of `'btx' | 'stx'` we should avoid repeating it

packages/bitcoin/src/validation/transaction-validation.ts
pete-watters: Whatever you think is best I'm happy with. If I recall it took me a while to get my head around `desiredSpend ` and `maxSpend ` so I changed it to be `amount` and `spendable`.

I'm happy to go with the general consensus here

apps/mobile/src/features/send-refactor/forms/btc/use-btc-form.ts
tigranpetrossian: I'm not sure I'm following. There just two `displayToast` occurences in the entire flow, both are for very unlikely unhandled exceptions. Most of the validation and error display is done through `zod` + RHF, the form by design doesn't allow submission when errors are present. This is just for an unlikely case of final TX somehow being invalid. I've excluded analytics from this PR to avoid growing it even larger, but we'll be also tracking those events.

pete-watters (approved):
I can't test locally as Alex is doing some work on the API and balances are `0`. 

Based on code quality I am happy to `Approve` but am adding two labels:
- `needs:tests` as we need some tests for [utils](https://github.com/leather-io/mono/pull/984/files#r2000956254).
- `needs:refactor` as I think [we just need to enable this ](https://github.com/leather-io/mono/pull/984/files#r2004917107)now

apps/mobile/src/features/send-refactor/forms/btc/use-btc-form.ts
pete-watters: The pattern we inherited was that - packages throw and error and it gets caught and displayed by `formatBitcoinError` in `use-send-form-btc`:
```
 catch (e) {
        if (e instanceof BitcoinError) {
          displayToast({ title: formatBitcoinError(e.message), type: 'error' });
        } else {
          displayToast({
            title: t({ id: 'something-went-wrong', message: 'Something went wrong' }),
            type: 'error',
          });
        }
```

I agree it's not good to throw errors in the render cycle but am suggesting that we keep consistent and do that until we can refactor and improve. 

It's kind of hard to tell the changes here as you aren't deleting the old code yet like - `use-send-form-btc`

apps/mobile/src/features/send-refactor/utils.ts
tigranpetrossian: Sure: [LEA-2384: Refactor & Test Send flow helpers](https://linear.app/leather-io/issue/LEA-2384/refactor-and-test-send-flow-helpers)

PR #988, opened 2025-03-20
fix: has inscribed utxos during broadcast

fbwoolf:
We had a failing test in the extension from this early/silent return in the broadcast hook. I think we can avoid the utxo check for testing env, and I also added a log so we can see why it is exiting the broadcast.

### DIFF
--- a/packages/query/src/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
+++ b/packages/query/src/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
@@ -42,6 +42,9 @@ export function useBitcoinBroadcastTransaction() {
           const hasInscribedUtxos = await checkIfUtxosListIncludesInscribed(utxos);
 
           if (hasInscribedUtxos) {
+            // Temp log until we relocate query hooks
+            // eslint-disable-next-line no-console
+            console.log('Inscribed utxo detected, aborting broadcast');
             return;
           }
         }
@@ -62,7 +65,7 @@ export function useBitcoinBroadcastTransaction() {
         onFinally?.();
       }
     },
-    [checkIfUtxosListIncludesInscribed, client]
+    [checkIfUtxosListIncludesInscribed, client.transactionsApi]
   );
 
   return { broadcastTx, isBroadcasting };


alter-eggo:
@fbwoolf we have integration test for utxo check, won't it fail bc of it?

camerow:
> @fbwoolf we have integration test for utxo check in `send-btc.spec.ts`, won't it fail bc of it?

Hmm, dang that's a good question. Bit of a whack-a-mole situation if so. IMO this seems better than what I did (just setting `all` regardless of context) but definitely worth checking that test. 

@alter-eggo - @fbwoolf and I spoke today about the challenges between `mono`/`extension` repos as well and here is one of those places that makes developing across them hard. We did discuss maybe Fara exploring that migration in the near future as well. Just FYI

fbwoolf:
Good point, @alter-eggo, I came across that after and realized prob would break things as-is. Let me think. 🤔 Hoping @markmhendrickson responds and says ok to spend all, haha.

markmhendrickson:
@fbwoolf what can I help clarify here exactly?

fbwoolf:
> @fbwoolf what can I help clarify here exactly?

I pinged you in Slack asking if we can spend all utxos for `sendTransfer` rpc requests? We do for psbt requests. Our test account fails bc it must have an inscription on a utxo, so we added a change in the extension to spend all utxos and bypass the check in the hook here. It is currently merged into dev, so I wanted to check if it is ok?

markmhendrickson:
Ah sorry, that ping got lost in my notifications. I think we do need to check for UTXOs for `sendTransfer` whereas we don't for `signPsbt` because the former needs to avoid sending inscriptions invisibly as if it were normal BTC whereas the latter will show the user visibly exactly which inscriptions will be sent before proceeding. 

For PSBTs, this is usually intention since the app has chosen the UTXOs specifically given their inscriptions whereas for simple BTC transfers, the app hasn't done that so any inscriptions included in the UTXOs that have been selected is unintentional at best.

fbwoolf:
Closing bc these hooks are being relocated here: https://github.com/leather-io/mono/pull/991#discussion_r2008836891

PR #989, opened 2025-03-21
feat: add query provider and upgrade alex sdk package

alter-eggo:
This pr 
- adds tanstack query lib/provider
- adds vite-plugin-node-polyfill
- updated wrangler bc of this [err](https://github.com/leather-io/mono/actions/runs/13988601288/job/39167244719)
- upgrades alex-sdk library, otherwise get err with stacks/transaction lib (on the screenshot)
Screenshot 2025-03-21 at 12 54 25

### DIFF
--- a/apps/web/app/root.tsx
+++ b/apps/web/app/root.tsx
@@ -1,22 +1,16 @@
-import {
-  Links,
-  Meta,
-  Outlet,
-  Scripts,
-  ScrollRestoration,
-  isRouteErrorResponse,
-} from 'react-router';
+import { Buffer } from 'safe-buffer';
 
-import { Flex, styled } from 'leather-styles/jsx';
-
-import type { LeatherProvider } from '@leather.io/rpc';
 import leatherUiStyles from '@leather.io/ui/styles?url';
 
 import type { Route } from './+types/root';
 import stylesheet from './app.css?url';
-import { Footer } from './layouts/footer/footer';
-import { GlobalLoader } from './layouts/nav/global-loader';
-import { Nav } from './layouts/nav/nav';
+
+/**
+ * Polyfill global Buffer
+ * (should be called before another code)
+ */
+// @ts-expect-error // safe-buffer typings are too old
+globalThis.Buffer = Buffer;
 
 export function links() {
   return [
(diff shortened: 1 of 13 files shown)


apps/web/app/queries/leather-query-provider.tsx
```diff
@@ -0,0 +1,97 @@
...
+import type { RemoteConfig } from '@leather.io/query';
+
+export interface LeatherEnvironment {
+  env: string;
+  github: {
+    org: string;
+    repo: string;
+    branchName?: string;
+    localConfig?: RemoteConfig;
+  };
```
kyranjamie, line 16: Now that we're using these directly, do we still need the provider at all?

apps/web/app/types/network.ts
```diff
@@ -0,0 +1,21 @@
...
+  wsUrl?: string;
+  isCustomNetwork?: boolean;
+}
+
+interface WhenStacksModeMap<T> {
+  mainnet: T;
+  testnet: T;
+  devnet: T;
+  mocknet: T;
+}
```
kyranjamie, line 17: Stacks only has two "modes" in the way were using mode for bitcoin. These are networks so `WhenStacksNetworkMap` would be better

apps/web/app/types/network.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function whenStacksNetworkMode(mode: StacksNetworkName) {
```
kyranjamie, line 19:
```suggestion
export function whenStacksNetwork(mode: StacksNetworkName) {
```

apps/web/vite.config.ts
```diff
@@ -27,11 +30,39 @@ const knownIcons = [
...
+        global: 'globalThis',
+      },
+      /**
+       * Enables Esbuild (dev) Buffer support (vite-plugin-node-polyfills doesn't work)
+       * @see https://github.com/vitejs/vite/discussions/6180#discussioncomment-7254293
+       */
+      plugins: [
+        NodeGlobalsPolyfillPlugin({
+          buffer: true,
+        }),
```
kyranjamie, line 63: Out of curiosity, did you try a classic polyfil by just doing 


```ts
import { Buffer } from 'buffer'
globalThis.Buffer = Buffer
```

that'd be much nicer than having to change all the build process for this

apps/web/app/queries/leather-query-provider.tsx
alter-eggo: hm, yeah, we can remove it

apps/web/vite.config.ts
alter-eggo: I get an err trying this
`app/root.tsx (1:9): "Buffer" is not exported by "__vite-browser-external", imported by "app/root.tsx".`
seems like under the hood vite changes original buffer to it’s own empty one, quite weird

kyranjamie: and using a different lib like https://www.npmjs.com/package/safe-buffer ?

alter-eggo: yeah, works now

apps/web/app/root.tsx
```diff
@@ -1,86 +1,10 @@
...
-      <p>{details}</p>
-      {stack && (
-        <pre>
-          <code>{stack}</code>
-        </pre>
-      )}
-    </main>
-  );
-}
+export { links, Layout, default, ErrorBoundary } from './app';
```
kyranjamie, line 10: I think it's convention to leave the named link/meta/error boundary exports in the root tsx

So all good to move the layout components, but lets leave the rest here

kyranjamie (approved):

PR #991, opened 2025-03-21
feat: move queries away from @leather.io/query

edgarkhanzadian:
### Why? 
Mobile and web platforms use different versions of react which causes internal packages that use react as a peerDependency not work on one of the platforms (in our case that's mobile as it has an older react version). The only package in question is @leather.io/query. To add to the argument,
1. @leather.io/query is planned to be sunsetted as it is not going to provide much utility when we have @leather.io/services package fully setup.
2. @leather.io/query exports the queries themselves which doesn't leave much room for customizability on the platform's side.

### Notes

This PR is part of moving all of the react related code from @leather.io/query to mobile and extension platforms.

It mostly deletes stuff @leather.io/query package (and adds a bit of stuff to mobile platform)

[This one](https://github.com/leather-io/extension/pull/6190) would add those queries to extension

Tested both mobile and extension with this one, waiting for this one to get merged to fully test the extension with @leather.io/query as an external dependency


apps/mobile/src/features/psbt-signer/psbt-signer.tsx
```diff
@@ -8,7 +8,10 @@ import { BitcoinOutcome } from '@/features/approver/components/bitcoin-outcome';
 import { BitcoinFeeCard } from '@/features/approver/components/fees/bitcoin-fee-card';
 import { InputsAndOutputsCard } from '@/features/approver/components/inputs-outputs-card';
 import { OutcomeAddressesCard } from '@/features/approver/components/outcome-addresses-card';
+import { useAverageBitcoinFeeRates } from '@/queries/fees/fee-estimates.hooks';
```
camerow, line 11: Would we eventually want a package that exports hooks so that each app _can_, if it desires, use that as a dependency?

camerow:
Would we eventually want a package that exports hooks so that each app can, if it desires, use that as a dependency?

apps/mobile/src/queries/stacks/stacks-client.ts
```diff
@@ -0,0 +1,8 @@
+import { type StacksClient, stacksClient } from '@leather.io/query';
+
+import { useLeatherNetwork } from '../leather-query-provider';
+
+export function useStacksClient(): StacksClient {
```
camerow, line 5: We also have `queries/clients/` should this live there?

camerow:
> Nope, we are only going to have hooks associated with the queries in the platform's code, not in internal packages.

Ok, I can see how that creates good separation of concerns, however seems like complex hooks like this are often going to be recreated all over the place:
Screenshot 2025-03-21 at 8 59 08 AM

I do think code duplication like this is often justified. Just thinking out loud more than anything here.

camerow:
> My move was mostly mindless, didn't want to get into specifics of each query when there is a lot of code to move + it needs to be done fast to unblock the dev branch for devs and designers

Makes total sense, and I understand at this point these issues are blocking progress on mobile. Thanks for fixing!

fbwoolf (approved):
Going to approve this bc I know we need to move the hooks back. We def need to get it merged and installed back in the extension quickly, prob a good time to do it bc we haven't merged anything that needs to be released. Once this is done, I agree, we can then look at what we are actually exporting again with the queryFn. Prob best to get Kyran's approval here too bc he had opinions on this work.

camerow:
Sounds good Fara

packages/query/src/bitcoin/transaction/use-check-utxos.ts
```diff
@@ -34,124 +19,3 @@ export function filterOutIntentionalUtxoSpend({
...
-      if (!input.txid) throw new Error('Transaction ID is missing in the input');
-      const idWithIndex = `${bytesToHex(input.txid)}:${input.index}`;
-      return resp.data.satpoint.includes(idWithIndex);
-    });
-  });
-
-  return hasInscribedUtxos;
-}
-
-export function useCheckUnspendableUtxos(blockTxAction?: () => void) {
```
fbwoolf, line 76: How can this be relocated, but `useBitcoinBroadcastTransaction`? Isn't it used there?

apps/mobile/src/queries/transaction/use-bitcoin-broadcast-transaction.ts
```diff
@@ -3,10 +3,11 @@ import { useCallback, useState } from 'react';
 import { TransactionInput } from '@scure/btc-signer/psbt';
 
 import { decodeBitcoinTx } from '@leather.io/bitcoin';
+import { filterOutIntentionalUtxoSpend } from '@leather.io/query';
 import { delay } from '@leather.io/utils';
 
 import { useBitcoinClient } from '../clients/bitcoin-client';
-import { filterOutIntentionalUtxoSpend, useCheckUnspendableUtxos } from './use-check-utxos';
+import { useCheckUnspendableUtxos } from './use-check-utxos';
```
fbwoolf, line 10: You should add a log for the silent return on L45 here. I have a PR up, but will close it bc of this work.
https://github.com/leather-io/mono/pull/988

apps/mobile/src/common/leather-query-provider.tsx
```diff
@@ -1,12 +1,16 @@
...
+
+import { LeatherQueryProvider as LeatherProvider } from '@/queries/leather-query-provider';
 import { queryClient } from '@/queries/query';
 import { GITHUB_ORG, GITHUB_REPO } from '@/shared/constants';
 import { BRANCH_NAME, WALLET_ENVIRONMENT } from '@/shared/environment';
 import { useSettings } from '@/store/settings/settings';
 
-import { LeatherQueryProvider as LeatherProvider } from '@leather.io/query';
-import { HasChildren } from '@leather.io/ui/native';
+interface LeatherQueryProviderProps {
```
pete-watters, line 9: Why change this from `HasChildren` ?

apps/mobile/src/queries/transaction/use-check-utxos.ts
```diff
@@ -0,0 +1,139 @@
...
+import { BitcoinClient, getNumberOfInscriptionOnUtxoUsingOrdinalsCom } from '@leather.io/query';
+import { isUndefined } from '@leather.io/utils';
+
+import { useBitcoinClient } from '../clients/bitcoin-client';
+import { useCurrentNetworkState, useIsLeatherTestingEnv } from '../leather-query-provider';
+
+class PreventTransactionError extends Error {
+  constructor(message: string) {
+    super(message);
+    // eslint-disable-next-line lingui/no-unlocalized-strings
```
pete-watters, line 15: A side note that this `lingui/no-unlocalized-strings` rule is annoying sometimes 

Maybe this can instead use `BitcoinError` which I moved to `packages/bitcoin/`?

apps/mobile/src/queries/transaction/use-check-utxos.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export function useCheckUnspendableUtxos(blockTxAction?: () => void) {
+  const client = useBitcoinClient();
+  const [isLoading, setIsLoading] = useState(false);
+  const { isTestnet } = useCurrentNetworkState();
+  const isTestEnv = useIsLeatherTestingEnv();
+
+  const preventTransaction = useCallback(() => {
+    if (blockTxAction) return blockTxAction();
+    throw new PreventTransactionError(
```
pete-watters, line 65: You will need to add a translation key for this error message so you should just use `BitcoinError` and add a translation to `formatBitcoinError`

apps/mobile/src/queries/transaction/use-check-utxos.ts
```diff
@@ -0,0 +1,139 @@
...
+  const checkIfUtxosListIncludesInscribed = useCallback(
+    async (inputs: TransactionInput[]) => {
+      setIsLoading(true);
+      const txids = inputs.map(input => {
+        if (!input.txid) throw new Error('Transaction ID is missing in the input');
+        return bytesToHex(input.txid);
+      });
+
+      try {
+        // no need to check for inscriptions on testnet
```
pete-watters, line 80: I don't think you need this comment

apps/mobile/src/queries/transaction/use-check-utxos.ts
```diff
@@ -0,0 +1,139 @@
...
+        if (isTestnet && !isTestEnv) {
+          return false;
+        }
+
+        if (txids.length === 0) {
+          throw new Error('Utxos list cannot be empty');
```
pete-watters, line 86: It would be good to use the same approach when throwing all of these errors - `BitcoinError`

pete-watters (approved):
Thanks for doing this @edgarkhanzadian 👍  

The code LGTM but I added some notes about better consistency handling errors.

apps/mobile/package.json
```diff
@@ -79,6 +79,8 @@
     "@stacks/transactions": "7.0.5",
     "@stacks/wallet-sdk": "7.0.2",
     "@tanstack/react-query": "5.59.16",
+    "@types/lodash.get": "4.4.9",
```
pete-watters, line 82: Should we add an explicit `react` version resolution to `mobile` to stop us hitting this problem again? 

It could at least serve as a reminder of the fixed version we need until we manage to upgrade React

alexp3y:
> > Nope, we are only going to have hooks associated with the queries in the platform's code, not in internal packages.
> 
> Ok, I can see how that creates good separation of concerns, however seems like complex hooks like this are often going to be recreated all over the place: Screenshot 2025-03-21 at 8 59 08 AM
> 
> I do think code duplication like this is often justified. Just thinking out loud more than anything here.

@camerow Chiming in late here to respond to this point - This type of complexity should be encapsulated by `@leather.io/services` going forward. There will still be some minor code duplication in app implementations of hooks but most complex operations like this should be pushed down into application / infrastructure services.

PR #994, opened 2025-03-22
fix: contract interface query response type

fbwoolf:
I am thinking we can use this type now looking at `fetchAbi` from `@stacks/transactions`:
https://github.com/hirosystems/stacks.js/blob/654ca25a5993005ddbb6869376ea03cfb95ce131/packages/transactions/src/fetch.ts#L261


fbwoolf:
@kyranjamie thoughts here, any concerns with changing this?

kyranjamie (approved):
No think this is fine 👍🏼

PR #995, opened 2025-03-24
fix: add new route for account balances and fix but with total, ref L…

pete-watters:
This PR fixes a mistake with the original implementation of the `account` `balances` page. 

I made a change to:
- add a new `balances` route off of `account`
- pass the correct account specific total balance to the page for display in the tokens widget

Before this the `account` tokens page was actually just showing all wallet tokens

Demo:

https://github.com/user-attachments/assets/e8d4c9dc-65c5-4690-a291-c268983e118c


alexp3y (approved):
Looks great! 🔥

apps/mobile/src/app/(home)/account/[account]/balances.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { configureAccountParamsSchema } from './index';
+
+export default function BalancesScreen() {
+  const params = useLocalSearchParams();
+  const { accountId } = configureAccountParamsSchema.parse(params);
+  const { fingerprint, accountIndex } = deserializeAccountId(accountId);
+
+  const { totalBalance } = useAccountBalance({ fingerprint, accountIndex });
+  if (totalBalance.state !== 'success') return;
```
Comment (kyranjamie) [inline apps/mobile/src/app/(home)/account/[account]/balances.tsx:20]: Should there be a loading state here?

apps/mobile/src/app/(home)/home/home.tsx
```diff
@@ -23,14 +24,19 @@ export function Home() {
 
   const accounts = useTotalAccountAddresses();
   const { data: activity, isLoading } = useTotalActivityQuery(accounts);
+  const { totalBalance } = useTotalBalance();
 
   useOnDetectNoNotificationPreference(notificationSheetRef.current?.present);
+  if (totalBalance.state !== 'success') return;
```
kyranjamie, line 30: and here?

kyranjamie (approved):

PR #1005, opened 2025-03-25
fix(mobile): collapse the sheet and show toast on fee selection in approver, closes LEA-2022

tigranpetrossian:
Adds auto-closing and a confirmation toast to fee selection in the Approver flow.

### Preview
https://github.com/user-attachments/assets/52994d3e-db9b-4b99-867e-6aefd385a092

### DIFF
--- a/apps/mobile/src/features/approver/components/fees/base-fee-option.tsx
+++ b/apps/mobile/src/features/approver/components/fees/base-fee-option.tsx
@@ -1,6 +1,6 @@
 import { ReactElement } from 'react';
 
-import { Avatar, Box, ItemLayout, Pressable } from '@leather.io/ui/native';
+import { Avatar, ItemLayout, Pressable } from '@leather.io/ui/native';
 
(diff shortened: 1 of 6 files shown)


kyranjamie (approved):
Approved

But, imo this toast isn't a particularly nice ui pattern. All the user interaction takes place in the bottom drawer, and then the confirmation happens at the top in a unrelated location to the action that took place.

A more contextual, inline confirmation prompt/animation would be nicer

tigranpetrossian:
> Approved
> 
> 
> 
> But, imo this toast isn't a particularly nice ui pattern. All the user interaction takes place in the bottom drawer, and then the confirmation happens at the top in a unrelated location to the action that took place.
> 
> 
> 
> A more contextual, inline confirmation prompt/animation would be nicer

@kyranjamie Agreed, I think even a subtle visual feedback on the toggle would be better. I attempted doing this along with polishing few other details, but it would require a major refactoring for at least the Fee handling part.

PR #1006, opened 2025-03-25
feat: integrate collectibles, ref LEA-1426

pete-watters:
This PR connects the `collectibles` service to the mobile UI and adds some basic visual integration 

To test it:
- load a wallet with `collectibles`
    - observe you can see all wallet collectibles on `Home` in the widget
    - observe you can navigate to a `collectibles` page and see all collectibles
- create a second account and select it
    - observe you can see that accounts collectibles the widget
    - observe you can navigate to a `collectibles` page and see only that accounts collectibles

**Demo**

Here is a multi wallet demo using the `dev` wallet with 10 accounts and one of my wallets

https://github.com/user-attachments/assets/b6a17ef3-0863-4a9e-8ce2-56f50c1774bf

**NOTE**
- it's **very slow** to load collectibles so you need to wait
- performance improvements coming 🔜 ™️


pete-watters:
Opening this up for initial review. I have some work to do organising the collectibles into a grid on that page but it's close

apps/mobile/src/components/widgets/collectibles/collectibles.layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+                      title={collectible.name}
+                      protocol={collectible.protocol}
+                      size={thumbnailSize}
+                    />
+                  )
+                }
+              />
+            );
+          default:
+            return null;
```
tigranpetrossian, line 76: Since you're already filtering collectibles and we're only interested in these two protocols:
`collectibles.filter(c => c.protocol === 'inscription' || c.protocol === 'sip9')`
TS should then be able to narrow the type to just these two, and you can use `assertUnreachable` if exhaustiveness is required.

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -20,23 +20,24 @@
...
-        */}
-        {/* {content ? sanitize(content) : ''} */}
-        {content}
+    <CollectibleCardLayout bg="ink.text-primary" p="4" width={size} height={size}>
+      {/* 
+        Remove all HTML tags from the content
+        Tried using RenderHtml, but it was not working as expected
+       */}
+      <Text color="ink.background-secondary" variant="code">
+        {content.replace(/<[^>]*>?/gm, '')}
```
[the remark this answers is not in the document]
tigranpetrossian: This is very cool, except `sanitize-html` seems enormous. DOMPurify was very decent from what I remember.

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -1,14 +1,14 @@
...
 interface CollectibleTextProps {
   src: string;
+  size?: number;
 }
 
-export function CollectibleText({ src }: CollectibleTextProps) {
+export function CollectibleText({ src, size = 200 }: CollectibleTextProps) {
   const [content, setContent] = useState<string | null>(null);
 
   useEffect(() => {
```
tigranpetrossian, line 14: Probably worth turning this into a query, instead of `effect` + `state` + `throw`. `fetch` doesn't throw on HTTP errors, but regardless, we wouldn't want to crash the page/app for this.

packages/ui/src/components/collectibles/native/collectible-card.native.tsx
```diff
@@ -8,27 +10,47 @@ export interface CollectibleCardProps {
...
 
-export function CollectibleCard({ name, type, src, mimeType }: CollectibleCardProps) {
+function CollectibleCardContent({ name, type, src, mimeType, size }: CollectibleCardProps) {
   const isOrdinal = type === 'inscription';
 
   if (isOrdinal) {
     switch (mimeType) {
       case 'audio':
-        return <CollectibleAudio />;
+        return <CollectibleAudio size={size} />;
```
tigranpetrossian, line 23: Nice! Itching to make this playable.

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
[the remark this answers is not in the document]
tigranpetrossian: Ugh, didn't realize DOMPurify depends on DOM, just saw your deleted comment.
I'm out of my depth and trying to understand the risks here, since RN isn't going to be able execute the content, I suppose at least one of the risks users copying this over to elsewhere? Should this be more carefully considered in terms whether we display it or not, and how exactly?

tigranpetrossian (changes requested):
Thanks @pete-watters!
I think better pre-processing of the collectible list outside of render should be an improvement.
Requesting changes because loading `CollectibleText` content should be handled without `fetch` or throw, bad response or even a wifi hiccup along with render will crash the whole thing, which is probably not what we want. I think this warrants a query with loading & error states, even if it requires design additions.

P.S.
Re: sanitization, I don't have a good understanding of risks there, mostly thinking out loud there, but happy to look into it if I can be of help.

apps/mobile/src/components/widgets/collectibles/collectibles.layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+function filterCollectibles(collectible: NonFungibleCryptoAssetInfo) {
+  switch (collectible.protocol) {
+    case 'inscription':
+      return !isEmptyString(collectible.src) && isDefined(collectible.src);
+    case 'sip9':
+      return (
+        !isEmptyString(collectible.cachedImage) &&
+        isDefined(collectible.cachedImage) &&
+        // BNS - Archive is not a valid collectible
```
pete-watters, line 20: I added filtering out collectibles with the name `BNS - Archive` as I'm not sure what that is. I guess BNS names? 

I cross checked the `extension` and BNS names seem to be loaded a different way in `StacksBnsName` in `StacksCryptoAssets` so I need to do more work there

apps/mobile/src/components/widgets/collectibles/collectibles.layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+export function CollectiblesLayout({ collectibles, view = 'gallery' }: CollectiblesLayoutProps) {
+  const displayLimit = 9;
+  const thumbnailSize = view === 'widget' ? 200 : 164;
+  return (
+    <Box flexDirection="row" alignItems="center" paddingHorizontal="5" gap="4" flexWrap="wrap">
+      {collectibles.filter(filterCollectibles).map((collectible, index) => {
+        if (view === 'widget' && index >= displayLimit) return null;
```
pete-watters, line 34: Adding this limit on the `widget` on `Home` as I don't think we should render all of them there. 

I'll check with design
Screenshot 2025-03-26 at 15 25 31

apps/mobile/src/components/widgets/collectibles/collectibles.layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+              <CollectibleCard
+                key={index}
+                name={collectible.title}
+                type={collectible.protocol}
+                src={collectible.src}
+                mimeType={collectible.mimeType}
+                size={thumbnailSize}
+                caption={
+                  view === 'gallery' && (
+                    <CollectibleCaption
```
pete-watters, line 47: I've improvised adding this caption to `gallery` view, copying the extension.  

It's not in the design but it's helpful for now, even just for debugging. Looking at the data and we seem to have some duplicate collectibles so more filtering could be needed.

Displaying the name makes that easier for now

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -20,23 +20,28 @@ export function CollectibleText({ src }: CollectibleTextProps) {
...
-        */}
-        {/* {content ? sanitize(content) : ''} */}
-        {content}
+    <CollectibleCardLayout bg="ink.text-primary" p="4" width={size} height={size}>
+      {/* 
+        Remove all HTML tags from the content
+        Tried using RenderHtml, but it was not working as expected
+       */}
+      <Text color="ink.background-secondary" variant="code">
+        {content
```
pete-watters, line 40: I'm adding this code to strip away HTML tags as a precaution. @tigranpetrossian and I were discussing it and we could use `sanitize-html` but that's 80KB gzipped .

I did some research, asking my pal Claude and it seems that we are safe to render un-sanitised text in React Native 
Screenshot 2025-03-26 at 15 43 47

pete-watters: @kyranjamie what do you think about this? 

- Should we cautiously strip the tags away as a precaution? 
- Maybe it's better to take the hit on `sanitize-html`?
- Are we being overly paranoid and should just let the text render?

pete-watters:
> Thanks @pete-watters! I think better pre-processing of the collectible list outside of render should be an improvement. Requesting changes before loading `CollectibleText` content should be handles without `fetch` or throw, bad response or even a wifi hiccup along with render will crash the whole thing, which is probably not what we want. I think this warrants a query with loading & error states, even if it requires design additions.
> 
> P.S. Re: sanitization, I don't have a good understanding of risks there, mostly thinking out loud there, but happy to look into it if I can be of help.

Thanks @tigranpetrossian 👍 I'll add these improvements ASAP

apps/mobile/src/components/widgets/collectibles/collectibles.layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+      return false;
+  }
+}
+
+export function CollectiblesLayout({ collectibles, view = 'gallery' }: CollectiblesLayoutProps) {
+  const displayLimit = 9;
+  const thumbnailSize = view === 'widget' ? 200 : 164;
+  return (
+    <Box flexDirection="row" alignItems="center" paddingHorizontal="5" gap="4" flexWrap="wrap">
+      {collectibles.filter(filterCollectibles).map((collectible, index) => {
```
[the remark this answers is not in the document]
pete-watters: Thanks @tigranpetrossian 👍 I was a bit lazy not doing this, and unsure if that should be done at the `service` level. I'll add it here to clean things up

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
pete-watters: Thinking about this now and that maybe this should be done in the `collectibles` service. What do you think @alexp3y ?

apps/mobile/src/components/widgets/collectibles/collectibles-serializer.tsx
```diff
@@ -0,0 +1,51 @@
+// TODO LEA-2402: investigate moving this to services / UI to share with the web
```
pete-watters, line 1: @alexp3y maybe we can move this serialization into the service? Then we can return a standard list to web + native?

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
pete-watters: @camerow this is annoying so I will return the raw `content` for now. Let's chat about this and if you think it's a security risk I will introduce `sanitize-html`. I think it's OK not to

apps/mobile/src/components/widgets/collectibles/collectibles-serializer.tsx
alexp3y: I'lll tag this on [an existing ticket](https://linear.app/leather-io/project/cloudflare-worker-api-6bfffabb8f18/issues) for migrating similar kinds of model-associated utils out of `@leather.io/services`.

I think this falls under the same category of model-associated utils that need to find a new home (e.g. `@leather.io/models/utils`).

pete-watters: I'll add a note in that task too for https://github.com/leather-io/mono/pull/1006/files#r2016083714

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
pete-watters: I'm going to leave this as is for now as I think we should fetch inscription text in the services.

tigranpetrossian (approved):

apps/mobile/src/components/widgets/collectibles/collectibles-serializer.tsx
pete-watters: It is, do you want me the rename it? The idea is to do this object formatting at a higher level to avoid repeating it across platforms

PR #1008, opened 2025-03-25
refactor: query package

edgarkhanzadian:
A simple refactor of the query package. Mostly just removing *.hooks.ts files and moving the functions from there to *.utils.ts or *.query.ts


packages/query/src/bitcoin/address/address.utils.ts
```diff
@@ -21,3 +22,39 @@ export function createBitcoinCryptoCurrencyAssetTypeWrapper(balance: Money) {
 export function hasInscriptions(utxos: UtxoResponseItem[]) {
   return !isEmptyArray(utxos);
 }
+
+export function calculateOutboundPendingTxsValue(pendingTxs: BitcoinTx[], address: string) {
+  // sum all inputs
```
pete-watters, line 27: I'd delete these comments as the code is clear enough

pete-watters (approved):

packages/query/src/bitcoin/address/address.utils.ts
```diff
@@ -21,3 +22,39 @@ export function createBitcoinCryptoCurrencyAssetTypeWrapper(balance: Money) {
...
+        `${utxo.txid}:${utxo.vout.toString()}` === `${inscription.txid}:${inscription.output}`
+    );
+  };
+}
+
+export function filterUtxosWithRunes(runes: RunesOutputsByAddress[]) {
+  return <T extends UtxoIdentifier>(utxo: T) => {
+    return !runes.some(rune => rune.output === `${utxo.txid}:${utxo.vout}`);
+  };
+}
```
kyranjamie, line 60: I wonder if these should better live in Bitcoin package? There's no inherent coupling to the API because they're using our Inscription abstraction/model

kyranjamie (approved):

fbwoolf (approved):

PR #1010, opened 2025-03-25
feat: integrate leather api types into client, ref LEA-2389

alexp3y:
Removes the need for manual recreation of Leather API types by using the OpenAPI doc hosted on the staging API (always has latest / up-to-date spec) to generate a fully typed client. Uses generated client in `LeatherApiClient`.

Type generation is performed via `gen-api-types` npm script in `@leather.io/services`. Requires `.env` at package root with API basic auth header ("LEATHER_API_BASIC_AUTH").

Services container initialization now requires a "WALLET_ENV" to provided; Gets registered in the services container as an injectable property. Client uses it to switch between staging and production APIs.


packages/services/src/infrastructure/api/leather/leather-api.client.ts
```diff
@@ -60,65 +51,69 @@ export class LeatherApiClient {
...
-          `${LEATHER_API_URL}/v1/market/crypto-prices`,
-          {
-            signal,
-          }
-        );
-        return leatherApiCryptoPricesSchema.parse(res.data);
+        const { data, error } = await this.client.GET('/v1/market/crypto-prices', { signal });
+        if (error) {
+          throw new Error(error.error);
+        }
```
kyranjamie, line 101: Worth making a wrapper to not have to repeat throwing the error? And, doesn't the generated client have an option to throw rather than safely return?

fbwoolf:
This looks great @alexp3y!

pete-watters:
Great stuff @alexp3y 🚀

PR #1013, opened 2025-03-26
feat(web): side bar v2

kyranjamie:
This PR adds all the new nav items per @fabric-8's latest designs.

**Spacing compared to designs**
image

Renamed `earn` to `stacking`. We still need new button variants added to design system. Need to add new icons.

**Outstanding issues**
[Mentioned in a Figma comment the nature of SSR loading this side panel](https://www.figma.com/design/6yskfJQgHk4e1b8gEcj9fy?node-id=3123-261595&m=dev#1190011262). Issue is that server doesn't know if extension is stilled, thus might flash content on hydration. Presume idea is to remove this component if it is installed.


apps/web/app/layouts/nav/nav.tsx
```diff
@@ -33,70 +32,75 @@ export function Nav() {
...
+          Install
+        </styled.h3>
 
-          <NavItem href="/support" icon={<SupportIcon variant="small" />}>
-            Support
-          </NavItem>
-        </styled.div>
-      )}
+        <NavItem
+          href="https://chromewebstore.google.com/detail/leather/ldinpeekobnhjjdofggfgjlcehhmanlj?hl=en&pli=1"
```
alter-eggo, line 94: let's put this in sep variable?

alter-eggo (approved):
let's disable this codecov warnings for now may be? they are rather annoying

apps/web/app/layouts/nav/nav.tsx
kyranjamie: done

kyranjamie:
@alter-eggo not sure we can disable entirely without disabling for monorepo entirely?

PR #1014, opened 2025-03-26
feat: add stack in a pool page

alter-eggo:
This pr adds
- stacking route
- query hooks
- polyfill required packages

### DIFF
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -17,7 +17,28 @@ export default defineConfig({
   resolve: {
     alias: {
       'leather-styles': path.resolve(__dirname, 'leather-styles'),
+      // polyfill required packages
+      crypto: 'crypto-browserify',
+      stream: 'stream-browserify',
+      vm: 'vm-browserify',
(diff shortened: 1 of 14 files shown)


apps/web/vite.config.ts
```diff
@@ -17,8 +17,29 @@ export default defineConfig({
...
+      'node:fs': 'browserify-fs',
+      'node:path': 'path-browserify',
+      'node:os': 'os-browserify/browser',
+      'node:assert': 'assert',
+      'node:process': 'process/browser',
+      'safe-process': 'process/browser', // vite marks "process" package as external, this is workaround
     },
   },
+  define: {
+    global: 'globalThis', // required for some libs (e.g. pbkdf2)
```
kyranjamie, line 41: do we use `pbkdf2`?

kyranjamie:
Are you sure we need **all** these polyfills?

apps/web/vite.config.ts
alter-eggo: this is subdep of crypto-browserify

PR #1020, opened 2025-03-28
fix: cleanup of small bugs

pete-watters:
This PR groups together a few bug fixes:
- positioning of mnemonic error [LEA-2177](https://linear.app/leather-io/issue/LEA-2177) 
- trim trailing whitespace on paste of mnemonic [LEA-2146](https://linear.app/leather-io/issue/LEA-2146) 
- don't show empty widgets if no data 
- auto focus on key entry [LEA-2361](https://linear.app/leather-io/issue/LEA-2361)
- refactor `recover-wallet` to use `useReducer`


apps/mobile/src/app/(home)/activity/components/activity-widget.tsx
```diff
@@ -17,8 +17,9 @@ interface ActivityWidgetProps {
 export function ActivityWidget({ activity, isLoading, onPressHeader }: ActivityWidgetProps) {
   const theme = useTheme<Theme>();
 
+  const hasActivity = activity && activity.length > 0;
   // TODO LEA-1726: handle  loading & error states
-  if (isLoading || !activity) return null;
+  if (isLoading || !hasActivity) return null;
```
tigranpetrossian, line 22: This will likely need an empty state design, because once loading state is implemented, showing a loading state then nothing would be strange. Worth creating a ticket for the design team to take care of this.

tigranpetrossian (approved):

apps/mobile/src/app/(home)/activity/components/activity-widget.tsx
pete-watters: Thanks 👍  They have some empty states we need to build I linked in LEA-1726

PR #1026, opened 2025-03-28
refactor(mobile): reorganize the global layout and providers

tigranpetrossian:


apps/mobile/src/common/global-sheet-provider.tsx
```diff
@@ -0,0 +1,29 @@
+import { createContext, useContext, useRef } from 'react';
```
pete-watters, line 1: Do you have any thoughts on getting rid of the `common` folder? 

Maybe we can have the `providers` at root level and move the other items elsewhere?

We have `common/` and `shared/` and also another folder `filesystem/`.

apps/mobile/src/common/status-bar.tsx
```diff
@@ -0,0 +1,13 @@
+import { StatusBar as NativeStatusBar } from 'react-native';
```
pete-watters, line 1: Could this live in `components`? We could pass in the `barStyle`

apps/mobile/src/components/headers/developer-console-header.tsx
```diff
@@ -0,0 +1,13 @@
+import { TitleHeader } from '@/components/headers/title-header';
```
pete-watters, line 1: Maybe we can just get rid of this? 
Screenshot 2025-03-28 at 14 52 48

It's only for Developer tools so we don't really need it. Then it could help simplify how we handle the other Headers?

apps/mobile/src/common/global-sheet-provider.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { HasChildren, SheetRef } from '@leather.io/ui/native';
+
+export interface GlobalSheetContextValue {
```
pete-watters, line 5: How about not labelling this as `Global` and just calling it `Sheet`?

pete-watters (approved):
Nice work @tigranpetrossian, thanks for doing this 👍 

I added some discussion points but overall LGTM

apps/mobile/src/common/global-sheet-provider.tsx
tigranpetrossian: I actually moved some stuff out of this one with intention to rename it to something like `core`. What do you think?

apps/mobile/src/components/headers/developer-console-header.tsx
tigranpetrossian: Yeah, honestly I'd love to. Will do.

apps/mobile/src/common/status-bar.tsx
tigranpetrossian: Definitely. No idea why I placed this here.

apps/mobile/src/common/global-sheet-provider.tsx
pete-watters: It seems like this is the central point for sheet through which all sheets need to go? 

Maybe I'm mis-understanding though and there will be more localised `Sheet`s also. 

I'm just suggesting to call it `sheet` instead of `global-sheet` and use it's filesystem placement to infer it's global.

apps/mobile/src/common/global-sheet-provider.tsx
pete-watters: Perfect! I'm happy to have one folder for all the common/shared etc.

apps/mobile/src/common/global-sheet-provider.tsx
tigranpetrossian: That's the nuance, not *all* sheets–just global sheets, hence the name.

This isn't a new implementation, but simply renaming existing implementation which used to be called sheet-navigator and co-located with SheetNavigationContainer–misleading because this has no overlap with navigation.
I propose we keep it as is within this PR, and revisit within [LEA-2380: Improve global sheet handling in the mobile app](https://linear.app/leather-io/issue/LEA-2380/improve-global-sheet-handling-in-the-mobile-app)

fbwoolf (approved):
Nice refactor @tigranpetrossian! 👍

PR #1046, opened 2025-04-02
fix: specify WALLET_ENVIRONMENT for production build

pete-watters:


alexp3y (approved):
🚀

alexp3y:
Added the `needs:refactor` tag because it looks like if we move forward with "WALLET_ENVIRONMENT" we'll need to refactor to use the "EXPO_PUBLIC_" prefix. And it looks like that could be set up in expo directly.

That said, looking more carefully through the codebase I actually think we can (and should) try to refactor or remove "WALLET_ENVIRONMENT" entirely from `mobile` in favor of NODE_ENV.

It seems like this is a very extension-specific concept used in some places to trigger different logic when running tests. And its more of a "WALLET_RUN_MODE" than an "environment", which is misleading. 

I think with relatively minor effort we could refactor reliance on this var out of `mobile` and possibly `mono` entirely.

(Related: https://github.com/leather-io/mono/pull/1047)

pete-watters:
I'll close this off for now

PR #1055, opened 2025-04-03
feat: stx_signTransaction

edgarkhanzadian:
Implement a part of the stx_signTransaction rpc request. This one only features `TransactionTypes.StxTokenTransfer`.

To keep PRs smaller, i'm going to post a separate PR for `TransactionTypes.ContractCall` and `TransactionTypes.ContractDeploy`

https://github.com/user-attachments/assets/eff946c3-2505-4750-ad10-faaa1f376ef0

### DIFF
--- a/apps/mobile/src/features/approver/components/memo-sheet.tsx
+++ b/apps/mobile/src/features/approver/components/memo-sheet.tsx
@@ -2,17 +2,53 @@ import { RefObject, useState } from 'react';
 
 import { SheetLayout } from '@/components/sheets/sheet.layout';
 import { TextInput } from '@/components/text-input';
+import { useToastContext } from '@/components/toast/toast-context';
 import { t } from '@lingui/macro';
+import { PayloadType, deserializeTransaction } from '@stacks/transactions';
 
+import { TransactionTypes, generateStacksUnsignedTransaction } from '@leather.io/stacks';
 import { Button, NoteTextIcon, SheetRef, UIBottomSheetTextInput } from '@leather.io/ui/native';
+import { createMoney } from '@leather.io/utils';
+
+import { TxOptions } from '../utils';
 
 interface MemoSheetProps {
   sheetRef: RefObject<SheetRef>;
   memo: string;
-  onChangeMemo(memo: string): void;
+  txHex: string;
+  setTxHex(txHex: string): void;
+  txOptions: TxOptions;
 }
-export function MemoSheet({ sheetRef, memo: _memo, onChangeMemo }: MemoSheetProps) {
+export function MemoSheet({ sheetRef, memo: _memo, txHex, setTxHex, txOptions }: MemoSheetProps) {
   const [memo, setMemo] = useState(_memo);
+  const tx = deserializeTransaction(txHex);
+  const { displayToast } = useToastContext();
+
+  async function onChangeMemo(memo: string) {
+    try {
+      if (tx.payload.payloadType === PayloadType.TokenTransfer) {
+        const newTx = await generateStacksUnsignedTransaction({
+          txType: TransactionTypes.StxTokenTransfer,
+          amount: createMoney(tx.payload.amount, 'STX'),
+          fee: createMoney(tx.auth.spendingCondition.fee, 'STX'),
+          nonce: Number(tx.auth.spendingCondition.nonce),
+          recipient: tx.payload.recipient,
+          memo,
+          ...txOptions,
+        });
+        const newTxHex = newTx.serialize();
+        setTxHex(newTxHex);
+      }
+    } catch {
+      displayToast({
+        title: t({
+          id: 'approver.send.stx.error.change-memo',
+          message: 'Failed to change memo',
+        }),
+        type: 'error',
+      });
+    }
+  }
 
   return (
     <SheetLayout
@@ -42,7 +78,7 @@ export function MemoSheet({ sheetRef, memo: _memo, onChangeMemo }: MemoSheetProp
         buttonState="default"
         onPress={() => {
           sheetRef.current?.close();
-          onChangeMemo(memo);
+          void onChangeMemo(memo);
         }}
         title={t({
           id: 'approver.add_memo.confirm_button',
--- a/apps/mobile/src/features/approver/stacks-fees.section.tsx
+++ b/apps/mobile/src/features/approver/stacks-fees.section.tsx
@@ -1,0 +1,77 @@
+import { useRef, useState } from 'react';
+
+import { useCalculateStacksTxFees } from '@/queries/stacks/fees/fees.hooks';
+import { deserializeTransaction } from '@stacks/transactions';
+
+import { FeeTypes } from '@leather.io/models';
+import { Approver, Box, SheetRef } from '@leather.io/ui/native';
+import { createMoney } from '@leather.io/utils';
+
+import { StacksFeeCard } from './components/fees/stacks-fee-card';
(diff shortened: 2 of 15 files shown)


fbwoolf (approved):

apps/mobile/src/features/approver/stacks-fees.section.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function StacksFeesSection({ txHex, setTxHex, txOptions }: StacksFeesSectionProps) {
+  const tx = deserializeTransaction(txHex);
+  const { feeMoney } = formReviewTxSummary({
+    tx,
+    symbol: 'STX',
+  });
+  const { data: stxFees } = useCalculateStacksTxFees(tx);
+  const fee = tx.auth.spendingCondition.fee;
+  function getFeeType() {
+    const estimates = stxFees?.estimates;
```
pete-watters, line 29: How about refactoring this a bit to not repeat `estimates?.[X]` so often and to make the intention clearer in the future? 

You could have `getFee(estimates,index) { ` for example

apps/mobile/src/features/approver/stacks-fees.section.tsx
```diff
@@ -0,0 +1,76 @@
...
+      return FeeTypes.High;
+    }
+    return FeeTypes.Custom;
+  }
+
+  const feeSheetRef = useRef<SheetRef>(null);
+
+  const [selectedFeeType, setSelectedFeeType] = useState<FeeTypes>(getFeeType());
+  const fees = {
+    [FeeTypes.Low]: stxFees?.estimates[0]?.fee || createMoney(0, 'STX'),
```
pete-watters, line 46: How about just calling `createMoney(0, 'STX')` once and setting it in a constant?

apps/mobile/src/features/approver/stacks-fees.section.tsx
```diff
@@ -0,0 +1,76 @@
...
+            feeSheetRef.current?.present();
+          }}
+        />
+      </Approver.Section>
+      <StacksFeesSheet
+        sheetRef={feeSheetRef}
+        selectedFeeType={selectedFeeType}
+        setSelectedFeeType={setSelectedFeeType}
+        fees={fees}
+        currentFee={createMoney(tx.auth.spendingCondition.fee, 'STX')}
```
pete-watters, line 69: You already set `tx.auth.spendingCondition.fee` as `fee` on line 27 so you can use that here?

apps/mobile/src/features/approver/utils.tsx
```diff
@@ -103,3 +110,49 @@ export function getStacksFeeData(feeType: FeeTypes) {
     time,
   };
 }
+
+interface ReviewTxSummaryProps {
+  tx: StacksTransactionWire;
+  symbol: CryptoCurrency;
+}
+export function formReviewTxSummary({ tx, symbol }: ReviewTxSummaryProps) {
```
pete-watters, line 118: At first glance I thought this was a component. Maybe change the name to be `getFormReviewTxSummary`?

Also, how likely is it that the user can get here with a `SIP-10`? 🤔 Just wondering if we can prevent it sooner?

apps/mobile/src/features/browser/approver-sheet/stx/sign-transaction/sign-transaction.tsx
```diff
@@ -0,0 +1,88 @@
...
+  closeApprover(): void;
+}
+
+export function SignTransactionApprover(props: SignTransactionApproverProps) {
+  const network = useNetworkPreferenceStacksNetwork();
+  function getNetwork() {
+    if (!props.message.params.network) {
+      return network;
+    }
+    if (StacksNetworks.includes(props.message.params.network as StacksNetworkName)) {
```
pete-watters, line 38: Can we:
- clean `props.message.params.network` up with a destructure? 
- avoid casting `as StacksNetworkName)`
- add copy for the Errors? - will the users see those?

apps/mobile/src/store/settings/settings.read.ts
```diff
@@ -118,31 +123,44 @@ export function useAccountDisplayAddress(fingerprint: string, accountIndex: numb
...
-  );
+    client: {
+      baseUrl: networkConfig.chain.stacks.url,
+    },
+    chainId: networkConfig.chain.stacks.subnetChainId ?? networkConfig.chain.stacks.chainId,
+    bnsLookupUrl: networkConfig.chain.stacks.url || '',
+  };
+}
+
+export function getStacksNetworkFromName(stacksNetworkName: StacksNetworkName) {
```
pete-watters, line 155: Can we avoid this by just using `ChainID` as a prop to `SignTransactionApprover` ?

pete-watters:
Nice work @edgarkhanzadian 🚀  I think it looks really good overall but doing a little more refactoring will give it some extra 🪄

apps/mobile/src/features/approver/components/memo-sheet.tsx
```diff
@@ -2,17 +2,53 @@ import { RefObject, useState } from 'react';
...
 }
-export function MemoSheet({ sheetRef, memo: _memo, onChangeMemo }: MemoSheetProps) {
+export function MemoSheet({ sheetRef, memo: _memo, txHex, setTxHex, txOptions }: MemoSheetProps) {
   const [memo, setMemo] = useState(_memo);
+  const tx = deserializeTransaction(txHex);
+  const { displayToast } = useToastContext();
+
+  async function onChangeMemo(memo: string) {
+    try {
+      if (tx.payload.payloadType === PayloadType.TokenTransfer) {
```
kyranjamie, line 29: What if it is a contact call?

apps/mobile/src/features/approver/stacks-fees.section.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (Number(fee) === estimates?.[1]?.fee.amount.toNumber()) {
+      return FeeTypes.Middle;
+    }
+    if (Number(fee) === estimates?.[0]?.fee.amount.toNumber()) {
+      return FeeTypes.Low;
+    }
+    if (Number(fee) === estimates?.[2]?.fee.amount.toNumber()) {
+      return FeeTypes.High;
+    }
+    return FeeTypes.Custom;
```
kyranjamie, line 39: unusual order

apps/mobile/src/features/browser/approver-sheet/stx/sign-transaction/sign-transaction.tsx
kyranjamie: Agree with all but Pete's [overzealous destructuring](https://www.aleksandrhovhannisyan.com/blog/overzealous-destructuring/)

kyranjamie: If we don't want to either read from deeply nested, or use deep destructuring, I'd suggest selector or lens pattern

kyranjamie (commented):
Great start @edgarkhanzadian. But why all the `if (tx.payload.payloadType === PayloadType.TokenTransfer) {`s? Does this only work with token transfers?

apps/mobile/src/features/browser/approver-sheet/stx/sign-transaction/sign-transaction.tsx
kyranjamie: On type casting https://trustmachines.slack.com/archives/C05LAP952E7/p1743775904159909

PR #1065, opened 2025-04-07
Comprehensive recipient flow (work in progress), part of LEA-1869

tigranpetrossian:
This feature is growing to be quite large, will be integrating it in chunks to keep it manageable.

PR is for just the layout and behavior of the recipient view. This version is under a flag, and adds the following:
* A search input component with clear button
* Layout for displaying own accounts and recents in a SectionList
* Sheet variations based on suggestion existence.

Suggestion generation and more will come in an upcoming PR.

### With existing suggestions
https://github.com/user-attachments/assets/c0fd092e-a502-4fe7-aca2-1318abf7e9a2

### Initial state

While at it, simplified feature flag handling to be managed in one file.


pete-watters (approved):
🚀

PR #1070, opened 2025-04-08
fix: tx request analytics event

fbwoolf:
Minor update needed for the extension to track legacy tx signing events.

### DIFF
--- a/packages/analytics/src/events.ts
+++ b/packages/analytics/src/events.ts
@@ -9,6 +9,7 @@ export interface Events extends HistoricalEvents {
   app_unlocked: undefined;
   app_locked: undefined;
   submit_feature_waitlist: SubmitWaitlist;
+  legacy_request_initiated: { method: string };
 }
 
 // These are historical events that we'll maintain but that do not follow the object-action framework.


fbwoolf:
Actually thinking I should change a few more here to mark them all as `legacy`.

fbwoolf:
@camerow is it ok to change these to track moving forward as `legacy` or will this screw up past events?

packages/analytics/src/events.ts
```diff
@@ -62,8 +62,8 @@ interface HistoricalEvents {
   view_rpc_send_transfer_confirmation: { symbol: string };
   view_rpc_sign_and_broadcast_psbt_confirmation: { symbol: string };
   view_transaction_confirmation: { symbol: string };
-  view_transaction_signing: undefined;
-  view_transaction_signing_error: { reason: string };
+  view_legacy_request_transaction_signing: { type: string };
+  view_legacy_request_transaction_signing_error: { reason: string };
```
camerow, line 66:
```suggestion
  view_transaction_signing_error: { reason: string, isLegacy: boolean };
```

fbwoolf:
I am going to close this bc the event should only be triggered by legacy requests once I refactor the 3 remaining rpc requests using the legacy token. I don't think it is nec to change.

fbwoolf:
> Tracking a user "viewing" the tx signing page is the wrong way to do it imo. `track.page` exists for this purpose. Further, a user could "view" the same request many times (i bet refresh counts them again).
> 
These are the legacy events, they already exists in the tx signing flow, so I'm not sure what you are referring to here?

> We want to know how many times the requests have triggered, and been passed through our wallet. How can we track that?
> 
Pretty confident this event only happens when the `view` popups up after the request is `triggered` bc it is in an `onMount`, maybe it was just poorly named from the beginning?

> Also, what's the rationale for trying to group the old requests together with the ones one? When would we ever fire `legacy: false`? We have separate RPC signing events for the new ones. This event, as it is, _is_ the legacy request.
> 
They aren't grouped, and I did think we would add new event tracking for rpc requests, thought not part of the issue I'm working on. Unfortunately, atm, rpc requests do triggered the legacy request components so this event does get triggered until I get them all refactored away from using that old request token.

> We haven't really thought about how to best track the RPC events, but given their distinct names, it could work well to track each event with its own method name, perhaps with prefix.
>
Yep, totally agree, and should be shared using the new events with mobile.

kyranjamie:
> Pretty confident this event only happens when the view popups up after the request is triggered bc it is in an onMount, maybe it was just poorly named from the beginning?

Yeah, that's how I understand it works too. But it's poor for two reasons, 1) there's a specific tool in analytics systems to collect page visit data `.page()`, and here we're using `.track()` 2) if user refreshes, that'll fire another event for the same request.

As such, I think this request is best left untouched. It'll quickly become legacy once we remove entirely, then we can delete them later.

In order to track the legacy events, I suggest we add new tracking events that more accurately track their usage. I believe this could be done from the extension's background script only, using the [`queueAnalyticsRequest`](https://github.com/leather-io/extension/blob/a05e58935b5357f7486aa32857b29a7379424f40/src/background/background-analytics.ts#L7-L20) fn, where we register the events coming in, [`handleLegacyExternalMethodFormat`](https://github.com/leather-io/extension/blob/a05e58935b5357f7486aa32857b29a7379424f40/src/background/messaging/legacy/legacy-external-message-handler.ts#L51-L55).

Given their temporary nature, I'd prefix with `legacy_`. 

Separately, @camerow perhaps the 3 of us can have a chat how we can best track RPC events best?

kyranjamie (approved):

fbwoolf:
I didn't see the `untypedTrack` until just now in extension, so I guess this change wasn't really nec.

PR #1073, opened 2025-04-09
fix: update maestro e2e tests

pete-watters:
This PR refactors the maestro test suite to fix broken tests and make them more reliable by using IDs rather than text


pete-watters:
@camerow - I'll return to this soon. I need to swap some more text assertions for `testID`s although if we can get on top of `CrowdIn` / `lingui` maybe we can change that. Text assertions act as better documentation. 

The CI  is doing better with almost all `iOS` tests passing. `android` tests fail due to differences in native UI (e.g. `tapOn: 'Select All'` is different in `android`)

pete-watters:
> @pete-watters Is this PR still relevant?

Yes as we need to fix the tests however I will close it and start over

PR #1077, opened 2025-04-10
fix(mobile): refactor feature flags, gate collectibles and waitlist, …

pete-watters:
This PR:
- refactors the `feature-flag` code to centralise to a single file
- gates collectibles behind a feature flag - [ENG-16](https://linear.app/leather-io/issue/ENG-16/feature-flag-collectibles-feature)
- gates waitlist features behind a feature flag - [PRO-12](https://linear.app/leather-io/issue/PRO-12/mpc-would-limit-or-not-sure-unless-truly-planned-as-a-feature-in-the)

https://github.com/user-attachments/assets/779b5399-c8e0-4120-bdb8-59c32635b78a


apps/mobile/src/features/feature-flags/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function useReleaseLocaleFeatureFlag() {
+  return useBoolVariation('release_locale_feature', false);
+}
+
+export function useReleasePushNotificationsFlag() {
+  return useBoolVariation('release_push_notifications', false);
+}
+
+export function useReleaseNewRecipientFlowFeatureFlag() {
```
pete-watters, line 48: @tigranpetrossian I renamed this hook to `useReleaseNewRecipientFlowFeatureFlag` to match the others although it's quite the mouthful. I left the key alone but maybe we should change that?

tigranpetrossian (approved):
Thanks @pete-watters 

Doesn't have to be within PR, but we need to tighten the naming around flags. There's a lot of repetitive & redundant wording that could be removed both in the code and in LD flag identifiers. E.g. `useReleaseCollectiblesFeatureFlag` -> `useCollectiblesFlag`

alexp3y (approved):
🏴‍☠️

PR #1080, opened 2025-04-10
feat(web): add PR preview deploys with SSR disabled for pull-request

kyranjamie:
Draft


kyranjamie:
It isn't as easy as just switching the `ssr` flag to `false`, we'd have to remove the loaders too. Will investigate later a proper CF worker set up

PR #1088, opened 2025-04-11
fix(mobile): fix lingui crowdIn sync

pete-watters:
This PR fixes `lingui` and adds a check to the CI.

It has been silently failing when it had compile issues and this means it doesn't sync with CrowdIn and add newly created keys

### DIFF
--- a/.github/workflows/code-checks.yml
+++ b/.github/workflows/code-checks.yml
@@ -96,3 +96,27 @@ jobs:
         with:
           token: ${{secrets.CODECOV_TOKEN}}
           files: ./app/mobile/coverage/.tmp/coverage-0.json
+
+  lingui-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for changes in app/mobile
+        id: check_changes
+        run: |
+          if git diff --name-only HEAD^ HEAD | grep -q "^app/mobile/"; then
+            echo "changes_detected=true" >> $GITHUB_OUTPUT
+          else
+            echo "changes_detected=false" >> $GITHUB_OUTPUT
+          fi
+
+      - uses: ./.github/actions/provision
+        if: steps.check_changes.outputs.changes_detected == 'true'
+
+      - name: Run lingui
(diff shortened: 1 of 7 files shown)


.github/workflows/code-checks.yml
```diff
@@ -96,3 +96,30 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/provision
+
+      - name: Run lingui
+        working-directory: ./apps/mobile
+        run: pnpm lingui compile --strict
+      # TODO ENG-14 - this should only run on changes to the mobile app
```
pete-watters, line 112: I tried to make it run only on changes to `apps/mobile` but couldn't get it working. 

I can remove this / try improve it by fixing it

.github/workflows/crowdin-action.yml
```diff
@@ -1,3 +1,6 @@
+# FIXME ENG-14 - I can't see this running anywhere
```
pete-watters, line 1: @camerow: I added some notes here as I don't think this is doing anything now. We should probably set `upload_translations: true`? 

We could do better reverting to a checked in `.po` file as now:
- crowdin creates new `id`s added in code with an empty translation
- in the production build it loads the translations from CrowdIn and has no backup
- if a new key is added and CrowdIn not updated it will render the key in the UI e.g. `my.key.id` 

The production app should have a fallback `.po` file that we check in and can also then use for when we are testing. It's unclear what will happen in the production build if it cannot connect to CrowdIn

apps/mobile/src/features/approver/components/approver-buttons.tsx
```diff
@@ -30,7 +30,7 @@ export function ApproverButtons({ approverState, onBack, onApprove }: ApproverBu
             buttonState="default"
             title={t({
               id: 'approver.button.approve',
-              message: 'Approve',
+              message: 'Confirm',
```
pete-watters, line 33: Changing this to `Confirm` to match the other `message:`s for `'approver.button.approve'` and so that `lingui` can compile

tigranpetrossian (approved):

PR #1095, opened 2025-04-14
feat: sip30 transfers

edgarkhanzadian:
Closes LEA-1967 LEA-1965 LEA-1963

This PR implements 3 rpc requests:
1. stx_transferStx
2. stx_transferSip9Nft
3. stx_transferSip10Ft

https://github.com/user-attachments/assets/91adae16-61f3-4b21-913b-6069e064ef1e

### DIFF
--- a/apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
+++ b/apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
@@ -1,0 +1,29 @@
+import { useNextNonce } from '@/queries/stacks/nonce/account-nonces.hooks';
+import { useStacksSigners } from '@/store/keychains/stacks/stacks-keychains.read';
+import { assertStacksSigner } from '@/store/keychains/stacks/utils';
+
+export function NonceLoader({
+  accountId,
+  children,
+}: {
+  accountId: string;
+  children(nonce: number): React.ReactNode;
+}) {
+  const { fromAccountId } = useStacksSigners();
+  const signer = fromAccountId(accountId)[0];
+  assertStacksSigner(signer);
+  const currentStacksAddress = signer.address;
+
+  const {
+    data: nonceResponse,
+    isLoading: isNonceLoading,
+    isError,
+  } = useNextNonce(currentStacksAddress);
+  if (isNonceLoading) return null;
+  if (isError || !nonceResponse?.nonce) {
+    // TODO: track this
+    throw new Error('Nonce request failed');
+  }
+
+  return children(nonceResponse?.nonce);
+}
--- a/apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/hooks.ts
+++ b/apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/hooks.ts
@@ -1,0 +1,84 @@
+import { useCallback } from 'react';
+
+import { useStacksSigners } from '@/store/keychains/stacks/stacks-keychains.read';
+import { assertStacksSigner } from '@/store/keychains/stacks/utils';
+import { bytesToHex } from '@stacks/common';
+import { Pc, serializeCV } from '@stacks/transactions';
+
+import { RpcRequest, stxTransferSip10Ft } from '@leather.io/rpc';
+import {
+  TransactionTypes,
+  createSip10FnArgs,
+  formatContractIdString,
+  generateStacksUnsignedTransaction,
+  getStacksAssetStringParts,
+} from '@leather.io/stacks';
+import { useOnMount } from '@leather.io/ui/native';
+
+import { getDefaultFee } from '../utils';
+
+interface UseTransferSip10FtTxHex {
+  request: RpcRequest<typeof stxTransferSip10Ft>;
+  accountId: string;
+  setTxHex(txHex: string): void;
+  nonce: number;
+}
+
+export function useTransferSip10FtTxHex({
+  request,
+  accountId,
+  setTxHex,
+  nonce,
+}: UseTransferSip10FtTxHex) {
+  const { fromAccountId } = useStacksSigners();
+
+  const getTxHex = useCallback(
+    async function getTxHex() {
+      const signer = fromAccountId(accountId)[0];
+      const fee = getDefaultFee();
+      assertStacksSigner(signer);
+
+      const { contractAddress, contractAssetName, contractName } = getStacksAssetStringParts(
+        request.params.asset
+      );
+
+      const currentStacksAddress = signer.address;
+
+      const fnArgs = createSip10FnArgs({
+        amount: request.params.amount,
+        senderStacksAddress: currentStacksAddress,
+        recipientStacksAddress: request.params.recipient,
+      });
+
+      const tx = await generateStacksUnsignedTransaction({
+        txType: TransactionTypes.ContractCall,
+        publicKey: bytesToHex(signer.publicKey),
+        contractAddress,
+        contractName,
+        functionArgs: fnArgs.map(arg => serializeCV(arg)),
+        functionName: 'transfer',
+        nonce,
+        fee,
+        postConditions: [
+          Pc.principal(currentStacksAddress)
+            .willSendEq(request.params.amount)
+            .ft(formatContractIdString({ contractAddress, contractName }), contractAssetName),
+        ],
+      });
+
+      const txHex = tx.serialize();
+      return txHex;
+    },
+    [
+      fromAccountId,
+      request.params.asset,
+      request.params.recipient,
+      request.params.amount,
+      accountId,
+      nonce,
+    ]
+  );
+  useOnMount(() => {
+    void getTxHex().then(txHex => setTxHex(txHex));
+  });
+}
--- a/apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/transfer-sip10-ft.tsx
+++ b/apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/transfer-sip10-ft.tsx
@@ -1,7 +1,9 @@
 import { useState } from 'react';
 
+import { useToastContext } from '@/components/toast/toast-context';
 import { BaseStxTxApproverLayout } from '@/features/approver/layouts/base-stx-tx-approver.layout';
 import { getTxOptions } from '@/features/approver/utils';
+import { useBroadcastStxTransaction } from '@/queries/stacks/use-broadcast-stx-transaction';
 import { useAccounts } from '@/store/accounts/accounts.read';
 import { useStacksSigners } from '@/store/keychains/stacks/stacks-keychains.read';
 import { assertStacksSigner } from '@/store/keychains/stacks/utils';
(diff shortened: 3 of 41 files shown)


apps/mobile/src/store/keychains/stacks/utils.ts
```diff
@@ -12,3 +14,7 @@ const stacksKeychainSchema = z.object({
 export type StacksKeychain = z.infer<typeof stacksKeychainSchema>;
 export const stacksKeychainStoreSchema = entitySchema(stacksKeychainSchema);
 export type StacksKeychainStore = z.infer<typeof stacksKeychainStoreSchema>;
+
+export function assertSigner(signer: StacksSigner | undefined): asserts signer is StacksSigner {
```
kyranjamie, line 18:
```suggestion
export function assertStacksSigner(signer: StacksSigner | undefined): asserts signer is StacksSigner {
```
?

packages/rpc/src/methods/stacks/stx-transfer-stx.ts
```diff
@@ -1,20 +1,14 @@
...
-      amount: z.coerce.number(),
-      memo: z.string().optional(),
-    }),
-    baseStacksTransactionConfigSchema
-  ),
+  params: z.object({
+    recipient: z.string(),
+    amount: z.coerce.number(),
+    memo: z.string().optional(),
+  }),
```
kyranjamie, line 12: let's use 

```suggestion
  params: z.object({
    recipient: z.string(),
    amount: z.coerce.number(),
    memo: z.string().optional(),
  }).passthrough(),
```

which allows other props, without this, we'd be introducing a worse breaking change

apps/mobile/src/store/keychains/stacks/stacks-keychains.read.ts
```diff
@@ -56,3 +57,9 @@ export function useStacksSignerAddressFromAccountIndex(fingerprint: string, acco
   const signers = useStacksSigners().fromAccountIndex(fingerprint, accountIndex);
   return signers.map(signer => signer.address)[0];
 }
+
+export function stacksSignerFromAddress(address: string) {
+  return function (signer: StacksSigner) {
+    return signer.address === address;
+  };
```
kyranjamie, line 64: is this used?

kyranjamie (commented):
Missing from this I think is the handling of closeApprover. A user closing the approver should tell the app that the user cancelled the flow, returning an error response

apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/hooks.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      const publicKey = createStacksPublicKey(extractKeyFromDescriptor(signer?.descriptor)).data;
+      const currentStacksAddress = publicKeyToAddressSingleSig(publicKey, 'mainnet');
+
+      const fnArgs: ClarityValue[] = [
+        uintCV(request.params.amount),
+        standardPrincipalCVFromAddress(createAddress(currentStacksAddress)),
+        standardPrincipalCVFromAddress(createAddress(request.params.recipient)),
+        noneCV(), // Add memo to SIP-30?
+      ];
```
kyranjamie, line 62: this is useful logic that'd be great to see in the stacks package

apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/hooks.ts
```diff
@@ -0,0 +1,95 @@
...
+        contractAddress,
+        contractName,
+        functionArgs: fnArgs.map(arg => serializeCV(arg)),
+        functionName: 'transfer',
+        nonce,
+        fee,
+        postConditions: [
+          Pc.principal(currentStacksAddress)
+            .willSendEq(request.params.amount)
+            .ft(`${contractAddress}.${contractName}`, contractAssetName),
```
kyranjamie, line 76: there's a helper for this string manip. in stacks package.

apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/transfer-sip10-ft.tsx
```diff
@@ -0,0 +1,88 @@
...
+          sendResult(response);
+        },
+      }
+    );
+  }
+
+  return (
+    <BaseStxTxApproverLayout
+      txHex={txHex}
+      setTxHex={setTxHex}
```
kyranjamie, line 80: I see here we pass around the `setTxHex` method, such that the child component can set the fee. This is quite side effecty behaviour on a flow (creating a tx hex), that should ideally be pretty deterministic. This is kinda sketchy, because what if some bug manipulates the tx in an unexpected way? It'll both be very hard to pin down, and potentially dangerous, causing unexpected transfer of funds etc.

How might we structure this such that `BaseStxTxApproverLayout` doesn't set the fee? Perhaps a `onUserSelectFee` handler we pass to up?

apps/mobile/src/features/browser/approver-sheet/stx/nonce-provider.tsx
```diff
@@ -0,0 +1,29 @@
...
+  assertStacksSigner(signer);
+  const publicKey = createStacksPublicKey(extractKeyFromDescriptor(signer?.descriptor)).data;
+  const currentStacksAddress = publicKeyToAddressSingleSig(publicKey, 'mainnet');
+
+  const { data: nonceResponse, isLoading: isNonceLoading } = useNextNonce(currentStacksAddress);
+  if (isNonceLoading) return null;
+
+  return cloneElement(children, {
+    nonce: nonceResponse?.nonce ?? 1,
+  });
```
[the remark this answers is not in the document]
kyranjamie: > I'm not sure what we should default it to here 🤔 i think if nonce request fails we shouldn't block the user from using the flow, wdyt?

I'd err on side of failing completely, and allowing us to track the frequency of this happening.

On `assertCorrectMethod`, can't we utilise the schemas to narrow the type, rather than using this new assert method? If using `parse` on a new var rather than the prop, you can consume the value in any fn scope, without need to reassert?

apps/mobile/src/features/approver/hooks.ts
```diff
@@ -0,0 +1,78 @@
+import { useToastContext } from '@/components/toast/toast-context';
+import { t } from '@lingui/macro';
+
+export function useStxTransactionUpdatesHandler() {
+  const { displayToast } = useToastContext();
+  function changeMemoToastHandler(onChangeMemo: (memo: string) => void) {
+    return function (memo: string) {
```
kyranjamie, line 7: Why the currying here?

apps/mobile/src/features/approver/hooks.ts
```diff
@@ -0,0 +1,78 @@
+import { useToastContext } from '@/components/toast/toast-context';
```
kyranjamie, line 1: On this file, these hooks are really specific only to stacks transaction operations, that happen to use the approver ux pattern.

How might we structure the files to make that relationship clear? If stacks message signing hooks go here too, we could end up with a generic do-all Approver codebase that goes against the design of it being a self-contained use-agnostic pattern that takes care of approving things only.

kyranjamie (commented):
Great work @edgarkhanzadian, and thanks for being patient with these reviews, lots going on here and trying to be thorough.

Last main point from me, will approve after discussed, but recommend getting at least one others' eyes on it before merging.

Even with fewer side effects in this version, I feel that setting the tx hex on to state, then reading that, is a dangerous pattern. We had issues with this approach in extension several years ago that we tried to refactor out.

<details>

<summary>Made these diagrams previously, which are loosely relevant</summary>

image

</details>

Wouldn't it be better to have a fn that's called with all the parameters needed to create a tx, every time, than it is it mutate one and then take the most recent? Given that tx generation is an async operation, there are risks of race conditions when building it this way. Even if the fn to create the tx is memoized for perf, this is safer than "taking the most recent copy" when its ready.

apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  children,
+}: {
+  accountId: string;
+  children(nonce: number): React.ReactNode;
+}) {
+  const { fromAccountId } = useStacksSigners();
+  const signer = fromAccountId(accountId)[0];
+  assertStacksSigner(signer);
+  const publicKey = createStacksPublicKey(extractKeyFromDescriptor(signer?.descriptor)).data;
+  const currentStacksAddress = publicKeyToAddressSingleSig(publicKey, 'mainnet');
```
pete-watters, line 19: Do we need to account for `testnet` in `publicKeyToAddressSingleSig`?

apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip10-ft/transfer-sip10-ft.tsx
```diff
@@ -0,0 +1,88 @@
...
+  sendResult,
+  nonce,
+  accountId,
+}: TransferSip10FtApproverProps) {
+  const { displayToast } = useToastContext();
+  const network = useNetworkPreferenceStacksNetwork();
+
+  const { list: accounts } = useAccounts();
+  const [txHex, setTxHex] = useState<null | string>(null);
+  useTransferSip10FtTxHex({ request, accountId, setTxHex, nonce });
```
pete-watters, line 42: I got a bit confused here as we don't use any values from `useTransferSip10FtTxHex`. 

I'm curious if it needs to be a hook? 

Maybe we could instead pass in `fromAccountId` also? 

**  const { fromAccountId } = useStacksSigners();**

apps/mobile/src/features/browser/approver-sheet/stx/transfer-sip9-nft/transfer-sip9-nft.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const network = useNetworkPreferenceStacksNetwork();
+
+  const { list: accounts } = useAccounts();
+  const [txHex, setTxHex] = useState<null | string>(null);
+  useTransferSip9NftTxHex({ request, accountId, setTxHex, nonce });
+  const signer = useStacksSigners().fromAccountId(accountId)[0];
+  const { mutateAsync: broadcastTransaction } = useBroadcastStxTransaction();
+
+  assertStacksSigner(signer);
+  if (!txHex) return null;
```
pete-watters, line 47: Should we do something else to handle `!txHex`?

apps/mobile/src/queries/stacks/use-broadcast-stx-transaction.ts
```diff
@@ -19,6 +20,11 @@ export function useBroadcastStxTransaction() {
         return Promise.reject(new Error(getErrorMessage(response.reason)));
       }
 
+      // Sometimes we might get "API rate limit exceeded" with no txid
```
pete-watters, line 23: How about making this comment into code?

```
const isRateLimited = !response.txid;

if(isRateLimited){

```

packages/stacks/src/transactions/get-contract-fn-args.ts
```diff
@@ -0,0 +1,46 @@
...
+export function createSip10FnArgs({
+  amount,
+  senderStacksAddress,
+  recipientStacksAddress,
+}: GetSip10FnArgs) {
+  const fnArgs: ClarityValue[] = [
+    uintCV(amount),
+    standardPrincipalCVFromAddress(createAddress(senderStacksAddress)),
+    standardPrincipalCVFromAddress(createAddress(recipientStacksAddress)),
+    noneCV(), // Add memo to SIP-30?
```
pete-watters, line 25: Do we need to add the memo to `SIP-30`? Can we open a task for that?

pete-watters:
Great work here @edgarkhanzadian , pretty complex stuff 🤯  I flagged a couple of things to improve but overall LGTM. 

I didn't approve yet as I think Kyran makes a good point [here](https://github.com/leather-io/mono/pull/1095#pullrequestreview-2781631158) that's important to solve. If doing things this way in the past caused problems then it could be better to avoid this pattern and come up with a better one. I

 don't fully understand the complications but feel like it's safer not to store this in state or even use memoization. I'd veer more towards re-creation even if it's slow to avoid potential issues with incorrect transactions

As this PR is quite large and we are not in production yet I think we could approve and merge the bulk of this and then immediately change that. I think it would be good to get Fara to review also when she can as she could have valuable insights here.

pete-watters (approved):
Thanks for the great work here @edgarkhanzadian 🎉 

[As discussed](https://trustmachines.slack.com/archives/C05LAP952E7/p1745405960162119)  lets revisit this to improve soon and once @kyranjamie  and @fbwoolf have time to go over it in more detail. I think it's good to get it merged now so you can progress elsewhere but we should re-visit it urgently and implement the best solution we can.

PR #1096, opened 2025-04-14
feat: sip30 message signing

edgarkhanzadian:
Closes LEA-1961 LEA-2216

Sign text and structured messages

https://github.com/user-attachments/assets/5492c90f-c9ef-4535-bfd3-43850500a4d3

### DIFF
--- a/apps/mobile/src/features/approver/clarity-value.section.tsx
+++ b/apps/mobile/src/features/approver/clarity-value.section.tsx
@@ -1,0 +1,71 @@
+import { useTheme } from '@shopify/restyle';
+import { ClarityType, ClarityValue, cvToString } from '@stacks/transactions';
+
+import { Box, Text, Theme } from '@leather.io/ui/native';
+import { assertUnreachable } from '@leather.io/utils';
+
+function wrapText(val: string) {
+  return <Text>{val}</Text>;
+}
+
+export function ClarityValueListDisplayer({
+  val,
+  level = 0,
+}: {
+  val: ClarityValue;
+  level?: number;
+}) {
+  const theme = useTheme<Theme>();
+  const encoding = 'tryAscii';
+
+  switch (val.type) {
+    case ClarityType.BoolTrue:
+      return wrapText('true');
+    case ClarityType.BoolFalse:
+      return wrapText('false');
+    case ClarityType.Int:
+      return wrapText(val.value.toString());
+    case ClarityType.UInt:
+      return wrapText(`u${val.value.toString()}`);
+    case ClarityType.Buffer:
+      if (/[ -~]/.test(val.value)) return wrapText(JSON.stringify(val.value));
+      return wrapText(`0x${val.value}`);
+    case ClarityType.OptionalNone:
+      return wrapText('none');
+    case ClarityType.OptionalSome:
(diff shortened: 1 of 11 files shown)


apps/mobile/src/features/approver/clarity-value.section.tsx
```diff
@@ -0,0 +1,70 @@
...
+}: {
+  val: ClarityValue;
+  level?: number;
+}) {
+  const theme = useTheme<Theme>();
+  const encoding = 'tryAscii';
+
+  switch (val.type) {
+    case ClarityType.BoolTrue:
+      return wrapText('true');
```
pete-watters, line 22: Could we just use `<Text>` here? I'm not sure if `wrapText` adds much benefit

apps/mobile/src/features/approver/layouts/base-stx-message-approver.layout.tsx
```diff
@@ -0,0 +1,155 @@
...
+  return capitalize(stacksChainIdToCoreNetworkMode(chainId));
+}
+
+export type StructuredMessageDataDomain = TupleCV<{
+  name: StringAsciiCV;
+  version: StringAsciiCV;
+  'chain-id': UIntCV;
+}>;
+
+function Structured({
```
pete-watters, line 55: Maybe we can:
- name this **StructuredMessage** or something with more info?
- give it an interface type 
- move it to it's own file?

apps/mobile/src/features/browser/approver-sheet/stx/sign-structured-message/sign-structured-message.tsx
```diff
@@ -0,0 +1,54 @@
...
+  sendResult,
+  accountId,
+}: StxSignStructuredMessageApproverProps) {
+  const { list: accounts } = useAccounts();
+  const signer = useStacksSigners().fromAccountId(accountId)[0];
+
+  if (!signer) throw new Error('No signer found');
+
+  async function onApprove() {
+    if (!signer) throw new Error('No signer found');
```
pete-watters, line 32: Maybe we can use `StacksError` for these errors instead? Then we can have a key for the message too

pete-watters (approved):
Nice work @edgarkhanzadian. I flagged a couple of possible improvements but LGTM overall

apps/mobile/src/features/approver/clarity-value.section.tsx
fbwoolf: Yeah, have no idea where `wrapText` came from ...even predates me, haha.

fbwoolf (approved):
Nice work @edgarkhanzadian ...really too bad we can't share more UI components b/w web/mobile.

apps/mobile/src/features/approver/clarity-value.section.tsx
pete-watters: Thinking on what Fara said and maybe we could share this with the extension, using `wrapText` to our advantage. 

We could pass that in as a callback that uses either `Box` / `Text` depending on platform. Then we could share all of it if we also pass in a component for `ClarityType.Tuple`

PR #1098, opened 2025-04-14
feat: sip30 get addresses

edgarkhanzadian:
Closes LEA-1959

Implement stx_getAddresses

https://github.com/user-attachments/assets/de84a847-8f38-4ec7-ac99-daec4ed42479


apps/mobile/src/features/browser/approver-sheet/stx/get-addresses/get-addresses.tsx
```diff
@@ -0,0 +1,86 @@
...
+      publicKey: bytesToHex(account.publicKey),
+      derivationPath: account.derivationPath,
+    };
+
+    const response = createRpcSuccessResponse('stx_getAddresses', {
+      result: [stacksAddressResponse],
+      id: props.message.id,
+    });
+
+    if (props.origin) {
```
kyranjamie, line 55: When would we ever run this without an origin? Think we should abort sooner if that is the case

apps/mobile/src/features/browser/approver-sheet/stx/get-addresses/get-addresses.tsx
```diff
@@ -0,0 +1,86 @@
...
+  RpcRequest,
+  RpcResponse,
+  createRpcSuccessResponse,
+  stxGetAddresses,
+} from '@leather.io/rpc';
+import { SheetRef } from '@leather.io/ui/native';
+
+interface GetAddressesApproverProps {
+  app: App;
+  message: RpcRequest<typeof stxGetAddresses>;
```
kyranjamie, line 22:
```suggestion
  request: RpcRequest<typeof stxGetAddresses>;
```

We used `message` in extension a lot in past, but this was really a relic from pre-following RPC correctly. This was updated recently.

Suggest only using `request` to keep consistent with RPC terminology.

apps/mobile/src/features/browser/approver-sheet/get-addresses/get-addresses.tsx
```diff
@@ -57,14 +57,12 @@ export function GetAddressesApprover(props: GetAddressesApproverProps) {
...
 
-    if (props.origin) {
-      dispatch(
-        userConnectsApp({
-          origin: props.origin,
-          accountId: makeAccountIdentifer(account.fingerprint, account.accountIndex),
-        })
-      );
-    }
+    dispatch(
```
pete-watters, line 60: Is it safe to remove `if (props.origin) {` ?

pete-watters (approved):
Nice work! 🚀

PR #1104, opened 2025-04-17
fix(mobile): bring back action bar, tweak the design

tigranpetrossian:
@edgarkhanzadian Looks like Action Bar got knipped by accident. This brings it back along with some cosmetic tweaks to match the designs.


tigranpetrossian:
No worries, no one liked it anyway.

PR #1106, opened 2025-04-17
refactor: unused code (part 2)

edgarkhanzadian:
Continuing my unused code refactor.

Had to bring back some queries from the [previous deletion PR](https://github.com/leather-io/mono/pull/1103) that are supposed to be an example of services usage.
We can review their usage later after the release

Deleting unused dependencies and some unused exports


apps/mobile/src/features/send/navigation.tsx
```diff
@@ -82,7 +82,7 @@ function getInitialRouteName({
   return 'form';
 }
 
-export function cardStyleInterpolator({
+function cardStyleInterpolator({
```
pete-watters, line 85: Maybe we can refactor this? Checking the usage and it's only used once so it could be easier to pass inline?

```
 <SendStack.Navigator
      initialRouteName={initialRouteName}
      screenOptions={{
        headerShown: false,
        cardStyle: { backgroundColor: theme.colors['ink.background-primary'] },
        cardStyle: {
      opacity: current.progress,
    },,
      }}
    >
```

apps/mobile/src/queries/activity/account-activity.query.ts
```diff
@@ -1,9 +1,39 @@
+import { useAccountAddresses, useTotalAccountAddresses } from '@/hooks/use-account-addresses';
+import { toFetchState } from '@/shared/fetch-state';
 import { useSettings } from '@/store/settings/settings';
 import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
 
-import { AccountAddresses } from '@leather.io/models';
+import { AccountAddresses, CryptoAssetInfo } from '@leather.io/models';
 import { getActivityService } from '@leather.io/services';
 
+/**
```
pete-watters, line 9: I can fix this but this is actually a mistake I made when integrating activity. I should have used these hooks - `useTotalActivity` / `useAccountActivity` rather than the queries directly. 

I will fix it when adding the loading states in [LEA-1726](https://linear.app/leather-io/issue/LEA-1726/handle-loading-states-elegantly)

apps/mobile/src/queries/balance/sip10-balance.query.ts
```diff
@@ -13,6 +13,9 @@ export function useSip10TotalBalance() {
   return toFetchState(useSip10AggregateBalanceQuery(addresses));
 }
 
+/**
```
pete-watters, line 16: Lets just get rid of this if not needed. We only need `useSip10TotalBalance` and `useSip10AccountBalance`

apps/mobile/src/queries/balance/runes-balance.query.ts
```diff
@@ -11,6 +15,14 @@ export function useRunesTotalBalance() {
   return toFetchState(useRunesAggregateBalanceQuery(accounts));
 }
 
+/**
```
pete-watters, line 18: Lets just ditch this if unused

apps/mobile/src/queries/common/remote-config/remote-config.query.ts
```diff
@@ -31,7 +29,7 @@ function fetchLeatherMessages(env: string, leatherGh: LeatherEnvironment['github
   };
 }
 
-export function useRemoteConfig() {
+function useRemoteConfig() {
```
pete-watters, line 32: I wonder if we even need this at all?

apps/mobile/src/queries/market-data/market-data.query.ts
```diff
@@ -5,6 +5,10 @@ import { FiatCurrency, FungibleCryptoAssetInfo } from '@leather.io/models';
 import { getMarketDataService } from '@leather.io/services';
 import { oneMinInMs } from '@leather.io/utils';
 
+/**
```
pete-watters, line 8: Do we need this comment? Does `knip` fail without it?

apps/mobile/src/queries/market-data/market-data.query.ts
```diff
@@ -26,7 +30,9 @@ export function useMarketDataQuery(asset: FungibleCryptoAssetInfo) {
   const { fiatCurrencyPreference } = useSettings();
   return useQuery(createMarketDataQueryOptions(asset, fiatCurrencyPreference));
 }
-
+/**
+ * Keeping this as an example of services usage
```
pete-watters, line 34: I'd ditch this if not used, we can bring it back if we need to

apps/mobile/src/store/keychains/keychains.ts
```diff
@@ -38,7 +38,7 @@ export function filterKeychainsToRemove<T extends { descriptor: string }>(
 interface WithDescriptor {
   descriptor: string;
 }
-
+/** @knipignore */
```
pete-watters, line 41: Does `knip` explode here? 🤔

pete-watters (approved):
Nice work 👍 I added some suggestions but LGTM overall

camerow:
IMO overall looks good, nice to get deps cleaned up specifically. Perhaps worth a bit of documentation in our README what this tool is for, how we should use it on the team?

camerow (approved):

PR #1115, opened 2025-04-23
feat(mobile): add search empty state to the recipient flow

tigranpetrossian:
Adds empty state view when searching for address, and better handling of query states.
Sneaking in a small case-sensitivity bugfix too, along with moving some helpers around.

image

### DIFF
--- a/apps/mobile/src/features/send/components/recipient/v2/recipient-selector/recipient-selector.tsx
+++ b/apps/mobile/src/features/send/components/recipient/v2/recipient-selector/recipient-selector.tsx
@@ -1,16 +1,24 @@
 import { useState } from 'react';
 
+import {
+  isBnsLookupCandidate,
+  normalizeSearchTerm,
+} from '@/features/send/components/recipient/v2/build-recipient-suggestions';
 import { RecipientInput } from '@/features/send/components/recipient/v2/recipient-input';
+import { RecipientSelectorHeader } from '@/features/send/components/recipient/v2/recipient-selector/recipient-selector-header';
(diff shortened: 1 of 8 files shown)


apps/mobile/src/features/send/components/recipient/v2/recipient-selector/recipient-selector.tsx
```diff
@@ -38,27 +46,41 @@ export function RecipientSelector({
...
-          keyExtractor={item => item.id}
-          stickySectionHeadersEnabled={false}
-          renderSectionHeader={({ section }) => <RecipientSelectorSectionHeader id={section.id} />}
-          renderItem={({ item }) => (
-            <RecipientSelectorItem entry={item} onSelect={onSelectAddress} />
-          )}
-        />
-      )}
+      </RecipientSelectorHeader>
+      {matchSuggestionsResult({
```
pete-watters, line 57: This is a cool syntax but any reason not to just just a component here? `<SuggestionsResult` 

I guess it's a nice pattern passing in the object props?

pete-watters (approved):
Great work! It seems unusual that `matchSuggestionsResult` isn't just a standard component but I guess its for good reason?

apps/mobile/src/features/send/components/recipient/v2/recipient-selector/recipient-selector.tsx
tigranpetrossian: Almost no strong reason really--experimenting with different variations of this for now to pinpoint main use cases and make a single reusable matcher we can use in most places. And it will likely be a component vs. a helper function like this one. There's a lot to cover with this, I'll share some ideas on the call today.

PR #1120, opened 2025-04-25
Add timestamp to recipient recents, + few fixes

tigranpetrossian:
**Adds relative date to recent suggestions:**

image

### Also few fixes for the flow: 
1. Correct recents order.
2. Pass correct validation helper for searching
3. Remove extra checks from address validation primitives to avoid false positives.

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -25,10 +25,14 @@ import { LDProvider } from '@launchdarkly/react-native-client-sdk';
 import { i18n } from '@lingui/core';
 import { I18nProvider } from '@lingui/react';
 import { QueryClientProvider } from '@tanstack/react-query';
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
 import * as SplashScreen from 'expo-splash-screen';
 import { PersistGate } from 'redux-persist/integration/react';
 
 import { Box, SheetProvider } from '@leather.io/ui/native';
+
+dayjs.extend(relativeTime);
 
 // Catch any errors thrown by the Layout component
 export { ErrorBoundary } from 'expo-router';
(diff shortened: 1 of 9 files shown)


packages/stacks/src/validation/address-validation.ts
```diff
@@ -33,11 +29,5 @@ export function isValidAddressChain(address: string, chainId: ChainId) {
 }
 
 export function validatePayerNotRecipient(senderAddress: string, recipientAddress: string) {
-  if (!isValidStacksAddress(senderAddress) || !isValidStacksAddress(recipientAddress)) {
```
pete-watters, line 36: Nice refactor 👍  are the `stacks` and `btc` address validation steps just not needed now or were they bothering you?

apps/mobile/src/app/_layout.tsx
```diff
@@ -25,11 +25,15 @@ import { LDProvider } from '@launchdarkly/react-native-client-sdk';
...
 import { QueryClientProvider } from '@tanstack/react-query';
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
 import * as SplashScreen from 'expo-splash-screen';
 import { PersistGate } from 'redux-persist/integration/react';
 
 import { Box, SheetProvider } from '@leather.io/ui/native';
 
+dayjs.extend(relativeTime);
+
```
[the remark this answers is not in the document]
pete-watters: For re-enabling language change? As yet there is no task for that. 

At the moment we have gated the feature but we are still calling `initiateI18n` and we call `getLocales` from `react-native-localize` . 

For efficiency maybe we should have stopped doing that when gating it but I think it's already loading the locales

pete-watters (approved):
🚀  Code LGTM but I'm not sure why you were removing the address format assertions?

If needed it's fine by me but you will need to update the tests. 

There's also an issue with `lingui` failing, maybe you need to rebase / fix that

packages/stacks/src/validation/address-validation.ts
tigranpetrossian: Those are redundant, the outcome will be the same without those checks. Helpers like `validatePayerNotRecipient` or `isValidAddressChain` can already assume the input is a valid address. 
Forgot the botched tests, will fix.

PR #1124, opened 2025-04-28
Recipient selection guard, part of LEA-1869

tigranpetrossian:
Adds a guard sheet (a custom alert dialog) to the recipient flow for proactive validation.

https://github.com/user-attachments/assets/fdcf8c1d-6e4b-4d4b-8fa1-4a0859f384b2


apps/mobile/src/features/send/components/recipient/v2/recipient-evaluator.ts
```diff
@@ -0,0 +1,130 @@
...
+  canSelfSend,
+  isNewAddress,
+  schema,
+}: CreateRecipientEvaluatorParams) {
+  return async function evaluateRecipient(address: string): Promise<GuardResult> {
+    const validationResult = await schema.safeParseAsync(address);
+    const newRecipient = isNewAddress(address);
+
+    if (recipientSchemaResultContainsError(validationResult, 'recipientIsPayer')) {
+      if (canSelfSend) {
```
pete-watters, line 43: Nice work here 👍 I was going to suggest doing this another way but on review of the other code this is the best way. 

We could call it something like `canConsolidate` or something but `canSelfSend` is fine I think

pete-watters (approved):
Nice work here @tigranpetrossian 👏  

When will `v2` become the only version?

tigranpetrossian:
Thanks @pete-watters. This is now enough of an improvement over the existing version. Going to submit a separate PR for removing the flag, didn't want it to ruin the diffs and make the reviewing harder in this one.

apps/mobile/src/features/send/components/recipient/v2/recipient-evaluator.ts
tigranpetrossian: The best way would probably be not using `zod` altogether. The way this flow turned out made both `zod` and `react-hook-form` mostly redundant--those mostly just add friction now with almost no value. Something to look into way down the line.

> canConsolidate
I thought about this as well but consolidation is a single a special case, where as self send is just a likely mistake we block.

PR #1130, opened 2025-04-29
fix(mobile): add loading states and error handling

pete-watters:
This PR:
- adds loading states and some error handling to mobile data queries
- ensures we never show false `$0` balances
- adds empty widget fallback for activity
- performs an organisational refactor to move files to `features/` and remove excess files in `app/` as they create un-necessary `expo` routes

**Activity widget**

I added a loading state for the activity widget, but it was then decided not  to use that. I decided to leave the code as it could be useful:

Demo of activity loader:

https://github.com/user-attachments/assets/c0fe578e-f119-4dde-afb8-1e0364fd05df

Demo without activity loader = only render widget when available: 

https://github.com/user-attachments/assets/c2c69a1c-2b28-4d40-bc26-e559c6f8614b


pete-watters:
@tigranpetrossian @edgarkhanzadian : I'm opening this up for review now as it's almost done and was getting bigger than planned. 

I need to do some final work on handling empty balances on the `Account Card` + `Account page` 
Screenshot 2025-05-06 at 12 55 29

I'll do that soon but in the meantime please let me know what you think of this so far and if you can recommend any improvements

pete-watters:
@tigranpetrossian @edgarkhanzadian : I added the remaining loading + fetch errors for the `Account Page`, pushing one new commit

https://github.com/user-attachments/assets/7dcf4279-6193-4328-ab50-d3e89df02cfa

Here is a demo of the `Home` loading states:

https://github.com/user-attachments/assets/450c2d57-7ad8-42b0-a69a-f9877161b110

apps/mobile/src/app/account/[account]/activity.tsx
```diff
@@ -1,27 +1,31 @@
...
-  }
   return (
     <ActivityLayout accountName={accountName}>
-      {activity?.map((activity, index) => (
-        <ActivityCell key={`activity.${index}`} activity={activity} />
-      ))}
+      <FetchWrapper data={activity}>
+        {hasActivity(activity) ? (
+          activity.state === 'success' &&
+          activity.value?.map((activity, index) => {
```
Comment (camerow) [inline apps/mobile/src/app/account/[account]/activity.tsx:19]: Perhaps better as a `slice(0, 5).map(...)` but since it's getting changed later anyway maybe doesn't matter.

apps/mobile/src/app/balances/index.tsx
```diff
@@ -1,24 +1,34 @@
...
 export default function BalancesScreen() {
   const { totalBalance } = useTotalBalance();
-  if (totalBalance.state !== 'success') return;
+
   const pageTitle = t({
     id: 'balances.header_title',
     message: 'My tokens',
   });
+
+  const balance = totalBalance.state === 'success' ? totalBalance.value : EmptyBalance;
```
camerow, line 17: I've seen this in a few places and feel like the view is the wrong spot for this check and change to balance, wouldn't it be better for the `value` to be returned as either the value or `EmptyBalance`? If that's a later improvement that is ok, but this is standing out to me, without this line of logic this is a perfectly dumb View which renders stuff and has no logic - this check has a codesmell of a leaky abstraction.

apps/mobile/src/components/loading/fetch-wrapper.tsx
```diff
@@ -0,0 +1,28 @@
+import { Error, FetchState, Loading } from '@/components/loading';
+
+import { assertUnreachable } from '@leather.io/utils';
+
+interface FetchWrapperProps {
```
camerow, line 5: Better as `unknown` to express we don't know or really care what this is?

camerow (approved):
A few comments, but please address in separate smaller PRs. Sorry this one had to sit for so long!

apps/mobile/src/app/balances/index.tsx
fbwoolf: It seems like a reusable `TotalBalance` component could be made rather than repeating the use of `useTotalBalance` with the same state check in multiple places?

fbwoolf (approved):

pete-watters:
@fbwoolf & @camerow thanks for the review and feedback. I have another branch in flight based off this one so I will include your recommendations there and tag you in the next one.

apps/mobile/src/app/account/[account]/activity.tsx
Comment (pete-watters) [reply apps/mobile/src/app/account/[account]/activity.tsx:19]: Thanks! 

I've actually updated this in my next PR already as @tigranpetrossian recommended using `slice` also:
```
export function ActivityList({ activity, mode = 'full' }: ActivityListProps) {
  const Component = mode === 'widget' ? ActivityCard : ActivityListItem;
  return (
    <Box>
      {activity.slice(0, mode === 'widget' ? 5 : undefined).map((activity, index) => (
        <Component key={`activity.${index}`} activity={activity} />
      ))}
    </Box>
  );
}

```

apps/mobile/src/app/balances/index.tsx
pete-watters: Thanks, I'll refactor this to just have a `TotalBalance` component instead, I didn't think of that. Then I can get rid of the repeated logic and centralise things.

We also have another over-complicated component - `components/balance/balance.tsx` so I will split that into two:
- a dumb rendering one
- the new `TotalBalance` 

Then it should also help make that easier to reason with and easier to replicate in the `extension`

pete-watters: I'm going to do this now then fixing `Locked Balances` - [LEA-2627](https://linear.app/leather-io/issue/LEA-2627/refactor-balance-component-and-show-locked-balances). 

I'll add you both on the PR just mentioning here while I visited

apps/mobile/src/components/loading/fetch-wrapper.tsx
pete-watters: Updating this in [LEA-2627](https://linear.app/leather-io/issue/LEA-2627/refactor-balance-component-and-show-locked-balances)

apps/mobile/src/app/balances/index.tsx
pete-watters: @camerow I started implementing this but it got messy fast trying to move the `EmptyBalance` into `FetchState`. I spoke with @alexp3y about it and ran my ideas past him and we respectfully disagree and think it's OK to repeat this pattern for now.

I will create some more wrapper functions, I made one to `TotalBalance` already like @fbwoolf suggested.

We are using the `fetchState` pattern in a few places to help us know the data state. We mostly use it for `balance` but we also use it for `activity`, `collectibles`. For `balance` it's valid for `value: EmptyBalance` if `error` but not for the other cases.

At the moment I can't think of a better less SMELLy way to do it at source. Options are:
- create a balance specific `toBalanceFetchState` 
- pass an `arg` in - `toFetchState(..., isBalance)`

The `balances` code is often very similar but with slight nuances. For example in `select-account` we need to check if the balance is `0` also. 

```
if (balance.state === 'success' && balance.value.fiat.availableBalance.amount.isZero()) {
    return null;
  }
  const availableBalance =
    balance.state === 'success' ? balance.value.fiat.availableBalance : EmptyBalance;

```

For now, I will create a helper to avoid the repetition of `balance.state === 'success' ? ` and try come up with a clean way to minimise repetition while working on fixing other bugs

alexp3y: Chiming in on this. IMO the `EmptyBalance` isn't a value that should be produced by the hook or `FetchState`, which correctly does not return a value when not in the "success" state.

The UI component should make the decision to show the "-.--" string. Pushing that inside the hook or `FetchState` as a default would be burying a display formatting concern into data fetching logic IMO.

Agree that the display string could be baked directly into the `Balance` component, perhaps by making the "balance" param an optional `Money` and having "-.--" be the default display under the hood? That would help to clarify that this isn't an actual numerical balance value but rather a formatting string.

Also would consider renaming `EmptyBalance` to `EmptyBalanceDisplay` or `EmptyBalanceString` to further clarify.

pete-watters: Opening this DRAFT -> https://github.com/leather-io/mono/pull/1190 still working on it but we can take the discussion there

PR #1133, opened 2025-04-30
feat: integrate consolidated token api endpoints

alexp3y:
Integrates new token-data API endpoints, replacing existing direct Hiro and BIS API calls to implement efficiency and performance gains in the services layer.

Introduces a few other performance improvement adjustments: 

- Excludes token data / metadata endpoints from cache invalidation on pull to refresh
- Adds hiro API rate limit header
- Optimizes hiro rate limiter p-queue configuration


apps/mobile/src/components/page/page.layout.tsx
```diff
@@ -22,7 +22,12 @@ export function PageLayout({ children }: HasChildren) {
 
   const onRefresh = useCallback(() => {
     setRefreshing(true);
-    void queryClient.invalidateQueries();
+    void queryClient.invalidateQueries({
+      predicate: query =>
+        query.queryKey[0] !== 'hiro-stacks-get-nft-token-metadata' &&
```
tigranpetrossian, line 27: Worth extracting those keys into a constant array with a descriptive name to communicate the reason for skipping.

It's probably a about time to extract this out of the layout file into a hook as well:

```
export function useRefreshGlobalQueries() {
  const [refreshing, setRefreshing] = useState(false);

  const onRefresh = useCallback(() => {
    ...
  }, []);

  return { refreshing, onRefresh };
}
```

Can do this one myself later if anything.

tigranpetrossian (approved):

apps/mobile/src/components/page/page.layout.tsx
camerow: Would it also be worth extracting keys as `const`s so that they are exportable/sharable?

camerow (approved):

apps/mobile/src/components/page/page.layout.tsx
pete-watters: Adding the `metadata` here is a good idea but if we add `runes` + `sip10`'s here, doesn't that mean we won't get up to date data on refresh? (I could be mistaken there but checkin)

So if I am waiting for funds to show up for example and I hit refresh, how can I get the latest data?

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -52,11 +52,20 @@ export type HiroNftHolding = NonFungibleTokenHolding;
...
 @injectable()
 export class HiroStacksApiClient {
+  private readonly _axios: AxiosInstance;
+
   constructor(
     @inject(Types.CacheService) private readonly cache: HttpCacheService,
     @inject(Types.SettingsService) private readonly settings: SettingsService,
     private readonly limiter: RateLimiterService
-  ) {}
+  ) {
```
pete-watters, line 61: Nice! Thanks for adding this 💪  maybe this negates the need for increased limits elsewhere?

pete-watters (approved):
Great work @alexp3y ! 🚀  I have one [Q here](https://github.com/leather-io/mono/pull/1133/files#r2070010669) but otherwise LGTM

apps/mobile/src/components/page/page.layout.tsx
alexp3y: @tigranpetrossian We should debounce this as well.

PR #1148, opened 2025-05-06
fix: stacks transaction types

fbwoolf:
This PR fixes some type issues with our Stacks transaction types. We had previously made them required, which caused us to have to use `?? 0` etc in places when generating the unsigned tx. This could lead to errors if we accidentally broadcast a tx with a `0` nonce. Our loaders should handle this, and surface the error when necessary.

Also, I was getting type errors with the `.passthrough` added to a few of the schemas rather than using the `baseStacksTransactionConfigSchema`.

cc @edgarkhanzadian for the changes here in case it breaks anything in mobile.

### DIFF
--- a/packages/stacks/src/transactions/generate-unsigned-transaction.ts
+++ b/packages/stacks/src/transactions/generate-unsigned-transaction.ts
@@ -20,16 +20,16 @@ import {
   isTransactionTypeSupported,
 } from './transaction.types';
 
-export function initNonce(nonce: number | string) {
-  return new BigNumber(nonce, 10);
+export function initNonce(nonce?: number | string) {
+  return nonce !== undefined ? new BigNumber(nonce, 10) : undefined;
 }
 
 export function getUnsignedContractCallParsedOptions(options: StacksUnsignedContractCallOptions) {
   return {
     ...options,
-    fee: options.fee.amount.toString(),
+    fee: options.fee?.amount.toString(),
     functionArgs: options.functionArgs.map(arg => deserializeCV(hexToBytes(cleanHex(arg)))),
-    nonce: initNonce(options.nonce).toString(),
+    nonce: initNonce(options.nonce)?.toString(),
     postConditions: options.postConditions,
   };
 }
@@ -39,8 +39,8 @@ export function getUnsignedContractDeployParsedOptions(
 ) {
   return {
     ...options,
-    fee: options.fee.amount.toString(),
-    nonce: initNonce(options.nonce).toString(),
+    fee: options.fee?.amount.toString(),
+    nonce: initNonce(options.nonce)?.toString(),
     postConditions: getPostConditions(
       options.postConditions?.map(pc => ensurePostConditionWireFormat(pc))
     ),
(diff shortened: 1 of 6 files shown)


packages/stacks/src/transactions/transaction.types.ts
```diff
@@ -23,26 +23,26 @@ export function isTransactionTypeSupported(txType: TransactionTypes) {
 export type StacksUnsignedContractCallOptions = ReplaceTypes<
   UnsignedContractCallOptions,
   {
-    fee: Money;
+    fee?: Money;
     functionArgs: string[];
-    nonce: number | string;
+    nonce?: number | string;
```
kyranjamie, line 28: I know this relates to conversation on other PR, and maybe I'm missing some of the context, so good to discuss. 

Afaict, the purpose of this replaced type is that it makes `nonce` and `fee` required, where they're optional as part of stacks lib. So making them optional once more doesn't make much sense, for that matter we could just use the core stacks lib type.

If we don't pass a nonce or fee to the stacks lib, it does a fee estimation and nonce look up itself, both values are needed to form a legal transaction.

As I understand, all this complexity stems from needing a tx for fee estimation prior to having a fee or nonce value. If so, then it'd be nice if this problematic use-case didn't make an impact on our core tx construction logic. 

Here, I think the values should be required, (unless we want to rely on the internal look up of stacks lib, which I think we would rather not?)

In the ext I'd think it'd make sense to wrap tx construction such that we have... pseudo code hierarchy

- `generateStacksContractCall()`
  - Use case: to make a legal tx used for fee estimation
    - `generateStacksContractCallForFeeEstimation({ ...knownTxValues, fee: dummyFee, nonce: dummyNonce })`
    - Here we don't care that we use dummy values, all good
  - Use case: to make actual transaction to broadcast to network
    - `generateStacksContractCall({ ...txValuesIncludingDefinedNonceAndFee })`
    - The core fn requires the fee and nonce, compiler error if undefined, no mixing of logic with fee estimation concerns

Structuring it like this, we scope the fallbacks only the place they're needed. The changes here worry me that we're adjusting the core logic for the fee estimation use case only. 

Not sure if I've explained well here so would be helpful to chat live later when you're online

fbwoolf: The replace types here aren't just to make them required, it is also to narrow the type and convert fee to `Money`.

But, happy to go the pseudo code route you suggested. I think maybe I just misinterpreted your previous feedback.

packages/stacks/src/transactions/generate-unsigned-transaction.ts
```diff
@@ -20,16 +20,16 @@ import {
   isTransactionTypeSupported,
 } from './transaction.types';
 
-export function initNonce(nonce: number | string) {
-  return new BigNumber(nonce, 10);
+export function initNonce(nonce?: number | string) {
+  return nonce !== undefined ? new BigNumber(nonce, 10) : undefined;
 }
```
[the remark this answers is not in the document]
fbwoolf: I copied this from the extension, but that doesn't mean a whole lot.

PR #1150, opened 2025-05-07
fix(mobile): fix mobile dev

pete-watters:
The mobile build was failing due to updates to `packages/tokens`. Again exposing our lack of E2E tests in CI 😰 

Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-05-07 at 06 33 02

PR #1151, opened 2025-05-07
fix: mobile builds

edgarkhanzadian:
Ref: https://github.com/leather-io/mono/pull/1150

Kudos to @pete-watters for finding and fixing the issue! This PR just makes it in a way that we still enable that thing for web app

NOTE: EAS builds are failing due to an unrelated issue, looking into it now


tigranpetrossian (approved):

pete-watters:
Thanks @edgarkhanzadian 👏

PR #1152, opened 2025-05-07
fix: eas builds

edgarkhanzadian:
This PR fixes EAS build as our app was not supported by the latest xcode. 

It also updates cocoapods for consitency

### DIFF
--- a/apps/mobile/eas.json
+++ b/apps/mobile/eas.json
@@ -12,8 +12,8 @@
       },
       "ios": {
         "simulator": true,
-        "cocoapods": "1.15.2",
-        "image": "latest"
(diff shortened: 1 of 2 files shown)


tigranpetrossian (approved):

apps/mobile/eas.json
```diff
@@ -12,8 +12,8 @@
       },
       "ios": {
         "simulator": true,
-        "cocoapods": "1.15.2",
-        "image": "latest"
+        "cocoapods": "1.16.2",
+        "image": "sdk-51"
```
pete-watters, line 16: I guess we need to lock this `sdk` as a newer one is not working?

PR #1154, opened 2025-05-07
feat: add real data on active pooled stacking page

alter-eggo:
WIP
This pr adds real data on active pooled stacking page


alter-eggo:
@edgarkhanzadian nope, closed

PR #1155, opened 2025-05-08
fix: tsc mobile build

edgarkhanzadian:
Building both mobile and web using pnpm build throws this type of error while building @leather.io/ui:

```
Type 'import(".../mono/node_modules/.pnpm/@types+react@19.0.12/node_modules/@types/react/index").ReactNode' is not assignable to type 'React.ReactNode'.
```
So for some reason @types/react package gets mixed up when trying to build both at the same time.

But building these repos separately works perfectly fine. Introducing 2 scripts here, `pnpm build:mobile` and `pnpm build:web` for separate platform building.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
   "type": "module",
   "scripts": {
     "build": "turbo run build",
+    "build:mobile": "turbo run build --filter=!./apps/web",
     "build:watch": "turbo run build:watch",
+    "build:web": "turbo run build --filter=!./apps/mobile",
     "check:all": "turbo run check:all",
     "clean": "git clean -dfX",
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
Also noticing some annoying errors (on web, api changes in how ref is used) issues, can't wait to get on latest RN architecture and R19 😇

kyranjamie:
> we would need to use expo sdk 54 for that

@edgarkhanzadian [You can use Expo53 for R19](https://expo.dev/changelog/sdk-53#highlights:~:text=0.79%C2%A0with-,React%2019,-and%20React%20Native)

PR #1157, opened 2025-05-08
Loading state for page loading in the browser [LEA-2488]

edgarkhanzadian:
https://github.com/user-attachments/assets/72543d2d-9360-4134-8fd7-5ccdf798a36d

### DIFF
--- a/apps/mobile/src/features/accounts/components/browser-loading.tsx
+++ b/apps/mobile/src/features/accounts/components/browser-loading.tsx
@@ -1,0 +1,60 @@
+import { forwardRef, useImperativeHandle, useState } from 'react';
+import Animated, {
+  Easing,
+  useAnimatedStyle,
+  useSharedValue,
+  withRepeat,
+  withTiming,
+} from 'react-native-reanimated';
+
+import { useTheme } from '@shopify/restyle';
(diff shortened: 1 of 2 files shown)


apps/mobile/src/features/accounts/components/browser-loading.tsx
```diff
@@ -0,0 +1,60 @@
+import { forwardRef, useImperativeHandle, useState } from 'react';
```
pete-watters, line 1: Could we put this in `packages/ui`? I'm using `SkeletonLoader` from there

apps/mobile/src/features/accounts/components/browser-loading.tsx
```diff
@@ -0,0 +1,60 @@
...
+      },
+    };
+  }, [gradientWidth]);
+
+  if (!isActive) return null;
+
+  return (
+    <Box position="absolute" width="100%" alignItems="center" height={5} top={0} zIndex="100">
+      <AnimatedLinearGradient
+        colors={['#FFFFFF00', colors['blue.action-primary-default'], '#FFFFFF00']}
```
pete-watters, line 52: Should we use all tokens instead of `'#FFFFFF00'`

pete-watters (approved):

PR #1158, opened 2025-05-08
Exiting mobile browser to return 'home' [PRO-3]

edgarkhanzadian:
pressing the back button when:
- nowhere else to go back -> go to empty state
- long pressed -> go to empty state

https://github.com/user-attachments/assets/3bbef822-bf5e-4d46-bda9-947cb7fe336d


pete-watters (approved):
Nice 👍

PR #1159, opened 2025-05-08
Browser history (latest to earliest) [PRO-8]

edgarkhanzadian:
Quick fix for reversing latest searches.

Need to implement a better fix here with a FlatList, but can be handled later

https://github.com/user-attachments/assets/0a61391f-fe51-419f-95bc-3a2b1003c5ac


pete-watters (approved):
🚀

PR #1160, opened 2025-05-08
feat: clean search text with X button [LEA-2596]

edgarkhanzadian:
Delete search text with X button

https://github.com/user-attachments/assets/0310b36d-1a86-4ace-b03d-c561106a1ed1

### DIFF
--- a/apps/mobile/src/features/browser/browser/search-bar/search-input.tsx
+++ b/apps/mobile/src/features/browser/browser/search-bar/search-input.tsx
@@ -6,7 +6,14 @@ import { SharedValue } from 'react-native-reanimated';
 import { t } from '@lingui/macro';
 import { useTheme } from '@shopify/restyle';
 
-import { TextInput, Theme } from '@leather.io/ui/native';
+import {
+  Box,
+  CloseIcon,
+  Pressable,
+  TextInput,
+  Theme,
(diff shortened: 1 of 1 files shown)


apps/mobile/src/features/browser/browser/search-bar/search-input.tsx
```diff
@@ -25,34 +32,51 @@ export function SearchInput({
...
+        borderColor="ink.border-default"
+        borderWidth={1}
+        placeholderTextColor={theme.colors['ink.text-subdued']}
+        color="ink.text-primary"
+        placeholder={t({
+          id: 'browser.input_placeholder',
+          message: 'Type URL or search',
+        })}
+      />
+      <Pressable
```
tigranpetrossian, line 65: Added an IconButton component few days ago that you can use here, see similar example usage in [here](https://github.com/leather-io/mono/blob/fb4c703af9c8394e3fae45a5e04ef5716d240b1f/apps/mobile/src/features/send/components/recipient/recipient-input.tsx)

tigranpetrossian (approved):
👍
We now have at least two similar looking search bars. [Created an issue](https://linear.app/leather-io/issue/LEA-2597/reusable-search-input-component) to build a reusable one:

PR #1163, opened 2025-05-08
feat: add platform to getInfo response

edgarkhanzadian:
This is more of a proposal. I believe platform should be part of the info that we supply to dApps regarding the wallet.


pete-watters (commented):
This LGTM but am not sure if its a good idea

pete-watters (approved):
I'll approve this as it seems like a good idea. We can revert if others later disagree

PR #1166, opened 2025-05-09
fix: build ios app on xcode 16

edgarkhanzadian:
Build ios app with xcode 16 on EAS servers with ios 18.

should fix this:
```
ITMS-90725: SDK version issue - This app was built with the iOS 17.5 SDK. All iOS and iPadOS apps must be built with the iOS 18 SDK or later, included in Xcode 16 or later, in order to be uploaded to App Store Connect or submitted for distribution.
```

### DIFF
--- a/apps/mobile/eas.json
+++ b/apps/mobile/eas.json
@@ -13,7 +13,7 @@
       "ios": {
         "simulator": true,
         "cocoapods": "1.16.2",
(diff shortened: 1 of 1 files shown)


pete-watters (approved):
Thanks @edgarkhanzadian 🚀

PR #1167, opened 2025-05-09
chore: miscellaneous small fixes

pete-watters:
This PR batches together some small and easy mobile fixes:
- feature flag `FeeBadge` until its calculated dynamically [PRO-77](https://linear.app/leather-io/issue/PRO-77/enable-or-remove-congestion-label-in-send-flow-of-mobile-app)
- make some copy changes [LEA-2595]([LEA-2595](https://linear.app/leather-io/issue/LEA-2595))
- add wallet name to account page [LEA-2497]([LEA-2497](https://linear.app/leather-io/issue/LEA-2497))
- adjust padding of network label and settings cog [LEA-2471](https://linear.app/leather-io/issue/LEA-2471/fix-padding-with-the-right-icon-and-testnet-label-in-the-header)

### Wallet name on account page
Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-05-09 at 11 08 14

### Padding for settings / network label

https://github.com/user-attachments/assets/3692d247-80e3-4ddd-82a2-1600caf87c91

### DIFF
--- a/apps/mobile/src/app/account/[account]/index.tsx
+++ b/apps/mobile/src/app/account/[account]/index.tsx
@@ -2,6 +2,7 @@ import { Account } from '@/features/account/account';
 import { EmptyBalance } from '@/features/balances/token-balance';
 import { useAccountBalance } from '@/queries/balance/account-balance.query';
 import { AccountLoader, deserializeAccountId } from '@/store/accounts/accounts';
+import { WalletLoader } from '@/store/wallets/wallets.read';
 import { useLocalSearchParams } from 'expo-router';
 import { z } from 'zod';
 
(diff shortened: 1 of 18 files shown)


apps/mobile/src/app/account/[account]/index.tsx
```diff
@@ -19,15 +20,20 @@ export default function AccountScreen() {
...
   return (
-    <AccountLoader fingerprint={fingerprint} accountIndex={accountIndex}>
-      {account => (
-        <Account
-          account={account}
-          balance={totalBalance}
-          isLoading={isLoadingTotalBalance}
-          isError={isErrorTotalBalance}
-        />
+    <WalletLoader fingerprint={fingerprint}>
```
Comment (pete-watters) [inline apps/mobile/src/app/account/[account]/index.tsx:23]: Maybe there's a better way than this double loader?

Comment (fbwoolf) [reply apps/mobile/src/app/account/[account]/index.tsx:23]: I have gotten used to the loader pattern and have quite a few nested in the extension sometimes for the rpc requests. I really haven't found a better solution.

Comment (pete-watters) [reply apps/mobile/src/app/account/[account]/index.tsx:23]: OK great, just checking in case there is a better way I don't know yet. I actually like this pattern. 

Thanks a lot! 🙇

PR #1132, opened 2025-04-30
Activity improvements

tigranpetrossian:

PR #1171, opened 2025-05-09
chore: ENG-37: Activity UI updates

pete-watters:
This PR adds some changes to `mobile` activity:
- takes in some improvements made by @tigranpetrossian  in [this PR](https://github.com/leather-io/mono/pull/1132) 
- moving text, updating labels, fixing colours and icons - [ENG-37](https://linear.app/leather-io/issue/ENG-37/adjust-and-update-activity-flag-itens-state-according-to-designs)
- adds our own `sBTC` svg to make sure it always renders - [LEA-2589](https://linear.app/leather-io/issue/LEA-2589)
- adds `FlashList` to `activity` to improve rendering performance

Here's a demo of the new activity:

https://github.com/user-attachments/assets/f98d302b-3681-4c51-89bf-e64cfa65f2a7


apps/mobile/src/features/activity/utils/format-activity.ts
```diff
@@ -1,52 +1,134 @@
...
-  }[activityType];
+import { BaseOnChainActivity, OnChainActivity } from '@leather.io/models';
+
+interface FormatActivityStatusProps {
+  type: OnChainActivity['type'];
+  status: BaseOnChainActivity['status'];
+}
+
+export function formatActivityStatus({ type, status }: FormatActivityStatusProps) {
+  switch (true) {
```
pete-watters, line 12: I got some pushback about using `switch(true)` in the past but I think it's a good fit here to map out all the status's and text. 

I'm open to refactoring if there is a better way

apps/mobile/src/features/activity/utils/format-activity.ts
```diff
@@ -1,52 +1,134 @@
...
+  const statusText = formatActivityStatus({ type, status });
+  return statusText ? `${statusText} ${timestampText}` : timestampText;
+}
+
+export function getActivityTitle(activity: OnChainActivity) {
+  switch (activity.type) {
+    case 'sendAsset':
+    case 'receiveAsset':
+      if (!activity.value?.crypto?.symbol) {
+        // TODO ENG-37 - find out about sBTC rewards - Asked design
```
pete-watters, line 99: I'll do more work on this soon as we got some [new designs](https://linear.app/leather-io/issue/ENG-37/adjust-and-update-activity-flag-itens-state-according-to-designs#comment-c068893d) for this now

apps/mobile/src/features/activity/activity-list.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react';
+
+import { ViewMode } from '@/shared/types';
+// NOTE: '/src' is necessary to fix an import issue with FlashList
```
pete-watters, line 4: If I don't point to `@shopify/flash-list/src` `FlashList` doesn't work and crashes the APP. 

If I do point there it works but gives a ton of type errors. I need to spend some extra time investigating. Maybe someone else has ideas?

apps/mobile/src/features/activity/utils/format-activity.ts
pete-watters: Great idea to use the `map`, I hadn't considered that approach. I'll change it and revise the name

fbwoolf (approved):

apps/mobile/src/features/activity/utils/format-activity.ts
pete-watters: I've updated this. FYI there is no such thing as `pending` for `receive` but I have added a placeholder. 

I also added some more activities I found using the map but we don't have a design for yet.

apps/mobile/src/components/balance/balance.tsx
```diff
@@ -19,12 +19,12 @@ interface FormatBalanceProps {
 export function formatBalance({ balance, isFiat, operator }: FormatBalanceProps) {
   if (isFiat) {
     const isLargeBalance = balance.amount.isGreaterThanOrEqualTo(100_000);
-    return i18nFormatCurrency(balance, isLargeBalance ? 0 : balance.decimals);
+    return operator
+      ? `${operator} ${i18nFormatCurrency(balance, isLargeBalance ? 0 : balance.decimals)}`
+      : i18nFormatCurrency(balance, isLargeBalance ? 0 : balance.decimals);
```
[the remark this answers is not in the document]
pete-watters: I'm going to refactor this whole `balance` component so I will do this separately

apps/mobile/src/features/activity/activity-list.tsx
pete-watters: I spent a few hours on this and couldn't solve it yet. [I opened a bug on their repo](https://github.com/Shopify/flash-list/issues/1661 )

I decided to just replace `FlashList` with native `FlatList` instead for now as it seems to give most of the improvements we want for now. 

I opened [a draft PR](https://github.com/leather-io/mono/pull/1184) which shows the problem, I'll keep a more isolated demo of FlashList there in case we revisit

apps/mobile/src/components/headers/animated-header/animated-header-screen.layout.tsx
```diff
@@ -23,6 +23,7 @@ interface AnimatedHeaderScreenLayoutProps extends HasChildren {
   subtitle?: string | React.ReactNode;
   contentContainerStyles?: ScrollViewStylesProps;
   isHeaderReversible?: boolean;
+  scrollable?: boolean;
```
pete-watters, line 26: @tigranpetrossian I tried to get this `AnimatedHeader` working with `FlatList` and it does work however it gives a warning about nested virtualized lists (rightly). 

https://github.com/user-attachments/assets/22b93972-cebf-4dae-ac23-02ff6ef6b639

I refactored this component a bit to enable `scrollable` or not but maybe I should just throw that away and revert to a fixed header for `activity` ?

apps/mobile/src/features/activity/activity-list.tsx
pete-watters: I got this solved with help from @tigranpetrossian . Thanks a lot!

PR #1176, opened 2025-05-13
feat: sip 30 sendTransfer

edgarkhanzadian:
https://github.com/user-attachments/assets/8bf15d3b-d8f1-410a-8853-10782d8d158e

### DIFF
--- a/apps/mobile/src/features/browser/approver-sheet/send-transfer.tsx
+++ b/apps/mobile/src/features/browser/approver-sheet/send-transfer.tsx
@@ -1,0 +1,132 @@
+import { useMemo } from 'react';
+
+import { PsbtSigner } from '@/features/psbt-signer/psbt-signer';
+import { createCoinSelectionUtxos } from '@/features/send/utils';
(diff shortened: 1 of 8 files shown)


apps/mobile/src/features/browser/approver-sheet/send-transfer.tsx
```diff
@@ -0,0 +1,118 @@
...
+  closeApprover(): void;
+  accountId: string;
+}
+
+export function SendTransferApprover(props: SendTransferApproverProps) {
+  const { data: feeRates } = useAverageBitcoinFeeRates();
+  const { fingerprint, accountIndex } = destructAccountIdentifier(props.accountId);
+  const utxos = useAccountUtxos(fingerprint, accountIndex);
+
+  if (!utxos.value) return null;
```
pete-watters, line 37: If `!utxos.value` or `!feeRates` then we return `null` so do we show an empty section / let them retry? 🤔

pete-watters (approved):

PR #1180, opened 2025-05-13
feat(mobile): add send form error view, closes LEA-2362

tigranpetrossian:
Adds error view to the send form, with ability to reload.

https://github.com/user-attachments/assets/168dfcf0-3343-4be5-88fb-c924b79cfa67

### DIFF
--- a/apps/mobile/src/features/send/components/error-state.tsx
+++ b/apps/mobile/src/features/send/components/error-state.tsx
@@ -1,0 +1,27 @@
+import { t } from '@lingui/macro';
+import { Image } from 'expo-image';
+
+import { ArrowRotateClockwiseIcon, Box, Button, Text } from '@leather.io/ui/native';
+
+interface ErrorStateProps {
+  onRetry: () => void;
+}
+
+export function ErrorState({ onRetry }: ErrorStateProps) {
+  return (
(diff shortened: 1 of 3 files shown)


apps/mobile/src/features/send/components/error-state.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ArrowRotateClockwiseIcon, Box, Button, Text } from '@leather.io/ui/native';
+
+interface ErrorStateProps {
+  onRetry: () => void;
+}
+
+export function ErrorState({ onRetry }: ErrorStateProps) {
+  return (
+    <Box alignItems="center" justifyContent="center" gap="4" style={{ paddingTop: 64 }}>
+      <Image
```
pete-watters, line 13: Maybe this can be a `prop` and we can re-use it? 

I can do that in [LEA-2562](https://linear.app/leather-io/issue/LEA-2562/error-boundary)

pete-watters (approved):
Great work 👏

apps/mobile/src/features/send/components/error-state.tsx
pete-watters: Just the `<Image` so we can pass in different ones

PR #1181, opened 2025-05-13
feat: sdk 53

edgarkhanzadian:
Update expo sdk to v53 and update lots of packages to the newer version to support new arch. This also introduces react 19.

Moved web react to v19.0 to keep a consistent version across apps.

The app is building fine on both android and ios. For some reason android is failing on CI, but that is going to be revamped later anyway to include CNG in CI.

This app now uses CNG, so we are going to run prebuild every time we run `pnpm ios` or `pnpm android` which i think is kinda neat.

Benefits of this PR:
1. Upgrade to the newest version of Expo 53, [changelog v52](https://expo.dev/changelog/2024-11-12-sdk-52) [changelog v53](https://expo.dev/changelog/sdk-53)
2. Upgrade to almost latest react-native, that gives us perks like [stable new architecture](https://reactnative.dev/architecture/landing-page) and [react 19](https://react.dev/blog/2024/12/05/react-19) inside of the app
3. This will also give us support for [CNG (Continuous Native Generation)](https://docs.expo.dev/workflow/continuous-native-generation/). Meaning we are not going to keep native code inside of the app anymore and we are going to prebuild it using the app.config.js file. That will help us alleviate ton of headache during next version updates and give us a one file to update most of the ios and android app settings. It also enables better CI (i think it should be faster and easier to use), but it's not yet enabled. Will be done in a follow up PR
4. This PR also gives us a possibility to finally update XCode to the latest version, as expo 51 no longer works on XCode 16.3 or newer
5. Update node to v22

Anyway, dear reviewer, welcome to hell with 190+ file changes

image


alexp3y:
Looks like a ton of great work @edgarkhanzadian. I understand this upgrades the Expo SDK but could you add a bit of additional context or link to an existing discussion around what the upgrade does for us? Just some high-level what/why reasoning on what I'm looking at would be helpful.

apps/mobile/app.config.js
```diff
@@ -44,26 +38,57 @@ export default () => {
   return {
     expo: {
       name: 'Leather',
+      owner: 'leather-wallet',
       slug: 'leather-wallet-mobile',
       version: '2.2.0',
+      runtimeVersion: {
+        policy: 'fingerprint',
+      },
+      newArchEnabled: true,
```
camerow, line 47: I want to say in Expo 53 this is enabled by default? Fine to keep it explicit though.

apps/mobile/package.json
```diff
@@ -29,7 +30,9 @@
     "lingui:extract": "lingui extract --clean",
     "lingui:extract:watch": "lingui extract --watch",
     "lingui:watch": "concurrently  \"pnpm lingui:extract:watch\"  \"pnpm lingui:compile:watch\"  ",
-    "preios": "pnpm run check:cocoapods && pnpm run firebase:ios",
+    "postinstall": "if [ \"$CI\" = \"1\" ]; then pnpm -w build:mobile; fi",
+    "preandroid": "pnpm run firebase && npx expo prebuild",
```
camerow, line 34: Shouldn't need the `npx expo prebuild` step here, if it needs to be run the `expo run` commands should run it i believe? 

A bit of redundancy is fine with me, and maybe running it frequently kind of helps us ensure it's working as intended.

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
```diff
@@ -93,7 +93,7 @@ function ConfigureAccount({ fingerprint, accountIndex, account }: ConfigureAccou
                 onPress={() => {
                   router.navigate({
                     pathname: AppRoutes.Account,
-                    params: { accountId: account.id },
+                    params: { account: account.id, accountId: account.id },
```
Comment (camerow) [inline apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:96]: At some point we should use the typed value of `account` rather than the `accountId`

camerow (approved):

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
Comment (pete-watters) [reply apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:96]: I don't understand this change but I guess it's needed? 

Don't we pass the same thing twice here? 
```
account: account.id, accountId: account.id
```

apps/mobile/src/features/account/account.tsx
```diff
@@ -44,6 +44,7 @@ export function Account({ account, walletName, balance, isLoading, isError }: Ac
                   pathname: AppRoutes.SettingsWalletConfigureAccount,
                   params: {
                     fingerprint: account.fingerprint,
+                    wallet: account.fingerprint,
```
pete-watters, line 47: @edgarkhanzadian I hit some rebase issues here which I can fix but I wanted to ask about this. Is there a reason we are doubling up on this stuff?

Why do we pass `wallet: account.fingerprint` ? 

Do we need to pass `account: account.id, accountId: account.id` twice?

pete-watters: OK that's great, thanks for explaining 👍

PR #1182, opened 2025-05-13
fix(mobile): add custom retry logic for queries

tigranpetrossian:
Fine tunes retrying errored queries, limiting to `5**`, `408 (Request Timeout)` and `429 (Too Many Requests)`, and potential network errors. Limits retry count (rather arbitrarily) to 5.

From what I understand errored requests (particularly `429`) don't count against the rate limit, it makes sense to retry those with the hope of next request happening after the cooldown.

### DIFF
--- a/apps/mobile/src/queries/query.ts
+++ b/apps/mobile/src/queries/query.ts
@@ -1,9 +1,25 @@
 import { QueryClient } from '@tanstack/react-query';
+import { HttpStatusCode, isAxiosError } from 'axios';
+
+const RETRY_LIMIT = 5;
 
 export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
+      retry: (failureCount, error) => isRetryableError(error) && failureCount <= RETRY_LIMIT,
       gcTime: 1,
     },
   },
 });
+
+function isRetryableError(error: Error): boolean {
+  if (!isAxiosError(error)) return false;
+  if (!error.response) return true;
+  const status = error.response.status;
+
+  return (
+    status >= 500 ||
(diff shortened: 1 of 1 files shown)


apps/mobile/src/queries/query.ts
```diff
@@ -1,9 +1,25 @@
 import { QueryClient } from '@tanstack/react-query';
+import { HttpStatusCode, isAxiosError } from 'axios';
+
+const RETRY_LIMIT = 5;
```
pete-watters, line 4: If we stop re-trying, we should show a callout to the users to let them know, like this one for `Balances` 

Screenshot 2025-05-14 at 05 49 52

Otherwise they won't know we aren't fetching anymore. We also need to add more error states to let users initiate re-fetch more easily. 

I think we should bundle this change with [LEA-2562](https://linear.app/leather-io/issue/LEA-2562/error-boundary)

apps/mobile/src/queries/query.ts
```diff
@@ -1,9 +1,25 @@
...
+
+function isRetryableError(error: Error): boolean {
+  if (!isAxiosError(error)) return false;
+  if (!error.response) return true;
+  const status = error.response.status;
+
+  return (
+    status >= 500 ||
+    status === HttpStatusCode.RequestTimeout ||
+    status === HttpStatusCode.TooManyRequests
```
camerow, line 23: Too many requests? I'll show you too many requests. Try it again!

camerow (approved):

alexp3y (approved):
LGTM

pete-watters (approved):

PR #1185, opened 2025-05-14
feat: prevent missing metadata refetch

alexp3y:
This PR ensures that we don't continue attempting to refetch SIP10 & SIP09 token metadata that is malformed (422) or missing (404) by returning null in these cases and caching the response.

The cache time for the metadata requests will still invalidate eventually (set interval is ~1 month) giving missing token metadata a chance to update if the status corrects in the future.


pete-watters (approved):
Thanks @alexp3y 👍 

I tagged you in this PR too - https://github.com/leather-io/mono/pull/1182/files can you check if we need that?

PR #1189, opened 2025-05-15
feat: consolidate http cache config, ref LEA-2101

alexp3y:
This PR addresses both HTTP caching and the protection of HTTP cache keys in the `mobile` pull-to-refresh feature.

HTTP cache configuration has been extracted to a single file (similar to the `workers` repo), simplifying & clarifying the surface area of HTTP calls + TTLs in `@leather.io/services`. 

The `HttpCacheService` has also been refactored into an abstract class to support shared logic adding an "http-" prefix to all cache keys and auto-selecting the TTL for any cache key passed to `fetchWithCache`.

The "http-" prefix is then used in the pull-to-refresh feature in `mobile` to protect all HTTP calls from invalidation.

The result is that only expired cache keys will refetch when a user "refreshes" the mobile app. This helps drives down the number of calls made on page load / refresh and will help prevent rate limit errors.

The full motivation for this change is described this [Linear comment](https://linear.app/leather-io/issue/LEA-2101/protect-http-cache-keys-from-pull-to-refresh-query-invalidation#comment-7a3db710).

### DIFF
--- a/apps/mobile/src/components/page/page.layout.tsx
+++ b/apps/mobile/src/components/page/page.layout.tsx
@@ -9,6 +9,7 @@ import {
 } from '@/components/action-bar/action-bar-container';
 import { queryClient } from '@/queries/query';
 
+import { HttpCacheService } from '@leather.io/services';
 import { Box, HasChildren } from '@leather.io/ui/native';
 
 import { ACTION_BAR_TOTAL_HEIGHT, ActionBarMethods } from '../action-bar/action-bar';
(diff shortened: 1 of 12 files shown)


alexp3y:
> Not familiar with the set up here, but code LGTM
> 
> It isn't super clear to me why this isn't a more mobile sided fix. The code in the linear issue shows that pull to refresh calls `queryClient.invalidateQueries();`. Isn't this the source of the issue? Refreshing shouldn't clear absolutely everything, instead rather selectively refetch balances/transactions, I would have thought.

By default `queryClient.invalidateQueries()` will clear everything - You need to explicitly exclude queries from invalidation. This update makes it easier to exclude all cached HTTP calls.

PR #1192, opened 2025-05-15
feat: authenticate tx broadcast requests to hiro, closes LEA-2631

alexp3y:
Adds the partner header onto Stacks Tx broadcast requests made via `@stacks/transaction`.

Addresses the issue with STX sends failing due to hitting Hiro's unauthenticated rate limit.


pete-watters (approved):
Thanks @alexp3y 👏

PR #1195, opened 2025-05-15
Will/eas update e2e

camerow:


apps/mobile/src/components/headers/home-header.tsx
```diff
@@ -0,0 +1,21 @@
+import { NetworkBadge } from '@/features/settings/network-badge';
+
+import { Box } from '@leather.io/ui/native';
+
+import { HeaderLeatherLogo } from './components/header-leather-logo';
+import { HeaderOptions } from './components/header-options';
+import { HeaderLayout } from './header.layout';
+
+export function HomeHeader() {
```
pete-watters: @camerow what I tried to do [here](https://github.com/leather-io/mono/pull/1410#discussion_r2166486715) was to replace the Developer console by adding a new `Button` gated by `ENV`.  ( I added an ENV var - `EXPO_PUBLIC_MAESTRO_CI`, I was missing `EXPO_PUBLIC` before. )

What we were doing in tests was finding and clicking this button in the `Dev tools` to clear the state

Something like this should work: 
```
  {
    process.env.EXPO_PUBLIC_EAS_CI === 'true' && (
      <Button
        title="Test"
        onPress={() => clearAllPersistedStorage(wallets.list.map(wallet => wallet.fingerprint))}
        data-testid="clear-all-persisted-storage-button"
      />
    );
  }
```

We could be better calling it the `ENV` var `EXPO_PUBLIC_CI` or something as we have another [CI  problem](https://linear.app/leather-io/issue/LEA-2716/remove-postinstall-from-mobile-build) with us always running `postinstall` on any PR to `mono` as we are checking against a CI flag. `postinstall` is only needed in EAS for mobile

PR #1200, opened 2025-05-16
fix: LEA-2623: exclude hidden accounts from activity and balances

pete-watters:
This PR fixes a bug whereby previously we were still including **hidden** accounts in our balance and activity calculations. 

I made these changes:
- filter `hidden` accounts out of `stacksSigners` and `derivedAddresses`
- default `useAccounts()` to use `active`, opt in for `hidden`
- **WHEN** users follow the path `Home` -> `Account` -> `Settings` -> `Hide Account` 
- **THEN** on back send them to `Home` and not the hidden account overview

### **Demo**

https://github.com/user-attachments/assets/c7b11f9e-792b-48d7-a335-eab2fb4100bf

**Home with all accounts hidden**

Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-05-16 at 16 25 23

### **Previous behaviour**

https://github.com/user-attachments/assets/d2d6b2d3-c40e-4c2f-a12f-851fc176ba82

### DIFF
--- a/apps/mobile/src/features/account/account.tsx
+++ b/apps/mobile/src/features/account/account.tsx
@@ -32,6 +32,7 @@ export function Account({ account, walletName }: AccountProps) {
     fingerprint,
     accountIndex,
   });
+
   const activity = useAccountActivity(fingerprint, accountIndex);
   const collectibles = useAccountCollectibles(fingerprint, accountIndex);
   const releaseCollectibles = useCollectiblesFlag();
@@ -42,15 +43,6 @@ export function Account({ account, walletName }: AccountProps) {
     <PageLayout>
       <NakedHeader
         error={isErrorTotalBalance && <FetchError />}
-        onGoBack={
-          account.status === 'hidden'
-            ? () => {
-                router.navigate({
-                  pathname: AppRoutes.Home,
-                });
-              }
-            : undefined
-        }
         rightElement={
           <Box alignItems="center" flexDirection="row" justifyContent="center" mr="2">
             <NetworkBadge />
--- a/apps/mobile/src/features/account/account.tsx
+++ b/apps/mobile/src/features/account/account.tsx
@@ -42,6 +42,15 @@ export function Account({ account, walletName }: AccountProps) {
     <PageLayout>
       <NakedHeader
         error={isErrorTotalBalance && <FetchError />}
+        onGoBack={
+          account.status === 'hidden'
+            ? () => {
+                router.navigate({
+                  pathname: AppRoutes.Home,
+                });
+              }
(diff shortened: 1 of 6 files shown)


apps/mobile/src/features/receive/screens/select-account.tsx
```diff
@@ -12,7 +12,7 @@ type CurrentRoute = CreateCurrentReceiveRoute<'select-account'>;
 
 export function SelectAccount() {
   const navigation = useReceiveSheetNavigation<CurrentRoute>();
-  const accounts = useAccounts('active');
+  const accounts = useAccounts();
```
pete-watters, line 15: @tigranpetrossian is this OK with you? I changed it so we infer `useAccounts` as `active`. 

Maybe it's better to:
- introduce `useActiveAccounts`
- instead of changing this update everywhere else to be `useAccounts('active'``

@edgarkhanzadian we use `useAccounts` a lot in `Approver`. I guess you only really want active accounts there also?

apps/mobile/src/hooks/use-account-addresses.ts
```diff
@@ -98,31 +99,53 @@ function deriveAccountAddresses(
...
+  return derivedAccountAddresses.filter(account =>
+    activeAccounts.some(
+      active =>
+        active.fingerprint === account.id.fingerprint &&
+        active.accountIndex === account.id.accountIndex
+    )
+  );
+}
+
 export function useTotalAccountAddresses() {
```
alexp3y, line 115: May want to consider adding the same `(status: AccountStatus = 'active')` param on these hooks so they can optionally return all vs active `AccountAddresses`

alexp3y (approved):
🦾

apps/mobile/src/hooks/use-account-addresses.ts
pete-watters: Good idea 👍

apps/mobile/src/features/account/account.tsx
```diff
@@ -42,6 +42,15 @@ export function Account({ account, walletName }: AccountProps) {
...
         error={isErrorTotalBalance && <FetchError />}
+        onGoBack={
+          account.status === 'hidden'
+            ? () => {
+                router.navigate({
+                  pathname: AppRoutes.Home,
+                });
+              }
+            : undefined
+        }
```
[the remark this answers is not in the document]
pete-watters: I changed this as suggested but then on testing it seemed quite buggy and also in the `Configure account` screen I need to make more changes. So I decided to just omit this and I can improve on it in the next one instead!

PR #1203, opened 2025-05-16
CI: reduce number of builds

camerow:
We are getting charged a lot for running these builds. I am exploring using `eas update` so that we don't have to run builds so frequently - but for now I think we should disable all these `dev` builds. This will save us a fair amount of resources used.


tigranpetrossian (approved):

pete-watters:
🚀

PR #1205, opened 2025-05-18
feat(web): add leather service package

alter-eggo:
This pr adds leather service package and market data

### DIFF
--- a/apps/web/app/app.tsx
+++ b/apps/web/app/app.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import { Links, Meta, Outlet, Scripts, ScrollRestoration, useNavigate } from 'react-router';
 
 import { QueryClientProvider } from '@tanstack/react-query';
@@ -51,6 +52,17 @@ export default function App() {
   useOnRouteChange(
     location => location.pathname === '/' && navigate('/stacking', { replace: true })
   );
+
+  useEffect(() => {
+    import('~/services/init-app-services')
+      .then(({ initAppServices }) => {
+        initAppServices();
+      })
+      .catch(error => {
+        // eslint-disable-next-line no-console
+        console.error(error);
+      });
+  }, []);
 
   return (
     <QueryClientProvider client={queryClient}>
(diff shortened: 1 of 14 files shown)


apps/web/app/app.tsx
```diff
@@ -52,6 +53,19 @@ export default function App() {
...
+    import('~/services/init-app-services')
+      .then(({ initAppServices }) => {
+        initAppServices();
+      })
+      .catch(error => {
+        // eslint-disable-next-line no-console
+        console.error(error);
+      });
+  }, []);
+
```
alter-eggo, line 68: otherwise get error
Screenshot 2025-05-18 at 11 34 47

apps/web/app/services/web-http-cache.service.ts
```diff
@@ -0,0 +1,20 @@
+import { QueryKey } from '@tanstack/react-query';
+import { queryClient } from '~/constants/query-client';
+
+import { HttpCacheKey, HttpCacheOptions, HttpCacheService } from '@leather.io/services';
+
+export class WebHttpCacheService implements HttpCacheService {
```
alexp3y, line 6: I reworked the `HttpCacheService` recently to be an abstract base class that the app-specific implementations should now extend from:

``` ts
export class WebHttpCacheService extends HttpCacheService {
  fetchWithCacheInternal<T>(
    key: unknown[],
    fetchFn: () => Promise<T>,
    options: HttpCacheOptions = {}
  ): Promise<T> {
    return queryClient.fetchQuery({
      queryKey: key,
      queryFn: fetchFn,
      staleTime: options.ttl ?? 0,
      gcTime: options.ttl ?? 0,
      retry: false,
    });
  }
}
```

alexp3y (approved):
Very nice! Couple of comments but otherwise everything LGTM. Glad to see `@leather.io/services` go multi-app 🚀

apps/web/app/app.tsx
kyranjamie: this is strange? why do you know? because buffer has already been polyfilled

PR #1211, opened 2025-05-19
feat: add leather api requests to queue

alexp3y:
Implements a priority queue to enforce client-side concurrency limits on Leather API requests.

This PR is necessary to prevent an overload of simultaneous requests as users add a high number of wallet accounts to their device and ensures consistent resource usage during high-volume actions like initial load / refresh.

Max concurrency is currently set at 20 RPS, which favors performance and responsiveness while still maintaining a controlled request flow. This design enables us to support an unbounded number of additional accounts per device w/o hitting rate limits.


pete-watters (approved):
🚀

PR #1212, opened 2025-05-19
fix: browser issues

edgarkhanzadian:
1. Patched react-native-webview s.t. it doesn't fail when we open a URL (https://github.com/react-native-webview/react-native-webview/issues/3697#issuecomment-2697390751). Only used the iOS part as we didn't have android build issues
2. Fixed animations connected to browser


patches/react-native-webview@13.13.5.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/src/RNCWebViewNativeComponent.ts b/src/RNCWebViewNativeComponent.ts
```
pete-watters, line 1: Just double checking this is right as I don't do much patching. This looks like a git diff?

pete-watters (approved):

patches/react-native-webview@13.13.5.patch
camerow: @edgarkhanzadian looks like this caused an issue though:

PR #1228, opened 2025-05-20
feat(mobile): support btc in conversion unit, ref ENG-34

alexp3y:
This PR adds both "BTC" and "sats" (Satoshis) as selectable currencies in the mobile "Conversion unit" selection screen.

This was done by adding them as entries on the fiat currency model. I think this is the most suitable and painless way to accomplish as our system doesn't distinguish broadly between fiat currencies and display currency settings.

I added two new `FiatCurrency` symbols: `XBT` and `sats` 

The idea there is to distinguish the Bitcoin fiat currency symbol (XBT) from the existing crypto version (BTC), keeping them totally separate and protecting the assumption that `FiatCurrency` symbols are unique. This allows components to continue determining whether a balance object `isFiat` using only the symbol w/ no further context.

My hunch is that using this `XBT` symbol _may_ be preferable than constantly adding extrinsic `isFiat` context to components, but the tradeoff is that we'll need some kind of fiat currency display transformations for "XBT -> BTC" (cc @tigranpetrossian @pete-watters).

Another UI follow-on that I couldn't figure out is how to make "SATS" lowercase when displayed in the totals.

---

The PR also fixes a market data service bug that was both increasing JPY and KRW currency balances and decreasing Rune token balances by a factor of 100.

---

<div flex flex-wrap='row'>

</div>


packages/models/src/currencies.model.ts
```diff
@@ -3,7 +3,7 @@ import { LiteralUnion } from './types.utils';
 export type CryptoCurrency = LiteralUnion<'BTC' | 'STX', string>;
 
 export type FiatCurrency = LiteralUnion<
-  'USD' | 'EUR' | 'GBP' | 'AUD' | 'CAD' | 'CNY' | 'JPY' | 'KRW',
+  'USD' | 'EUR' | 'GBP' | 'AUD' | 'CAD' | 'CNY' | 'JPY' | 'KRW' | 'XBT' | 'sats',
```
kyranjamie, line 6: But, Bitcoin isn't a fiat currency? Rather than adding here, can't we just extend/make a new string literal type type that contains all the currencies a user can set?

`ViewCurrency` or something

apps/mobile/src/components/balance/balance.tsx
```diff
@@ -14,6 +14,15 @@ interface FormatBalanceProps {
   operator?: string;
 }
 export function formatBalance({ balance, isFiat, operator }: FormatBalanceProps) {
+  if ((isFiat && balance.symbol === 'XBT') || balance.symbol === 'sats') {
```
kyranjamie, line 17: My understanding was that XBT was a long since forgotten ticker Kraken introduced before going back to BTC

kyranjamie: https://support.kraken.com/hc/en-us/articles/360001206766-Bitcoin-currency-code-XBT-vs-BTC

kyranjamie (commented):
Nice Alex, I reckon sats should also be the default in mobile.

apps/mobile/src/components/balance/balance.tsx
alexp3y: XBT is the ISO currency code for Bitcoin.

https://bitcoinmagazine.com/glossary/xbt

packages/models/src/currencies.model.ts
alexp3y: Technically not, but I steered away from trying to add another type because of how embedded `FiatCurrency` already is system-wide as a de-facto "view currency".  We make a lot of assumptions when dealing with something "is fiat". Seems natural to try and leverage existing architecture w/ a unique symbol representing BTC / sats when it is being used in all respects "as a `FiatCurrency`". How important is maintaining the strict semantic distinction of "Fiat Currency" versus how we seem to use `FiatCurrency` in our system? 🤷

kyranjamie: I would say it is pretty important to use accurate terminology, especially when dealing currency. Reading this code I'd assume it was a mistake, because I know that bitcoin isn't a fiat currency.

Would the context of us making a lot of assumptions that _fiat means view currencies_ be clear to others in future? And, what do we do later when we need a type of actual fiat currencies?

apps/mobile/src/components/balance/balance.tsx
pete-watters: Is `formatBalance` the main culprit in why we are including `XBT` / `sats` in `fiat`? 

Honestly, this is some tech debt and SMELLy code I added just to get things working. I can refactor it so we can keep the code more clean and intuitive sticking to `BTC` and `SATS` and adding them to  `CryptoCurrency` instead?

In this code, the reason I use this lookup for `isFiat` is:
- some tokens with an extra long name break the UI
- in that case I was hiding the token name (when `!isFiat`) via `formatMoneyWithoutSymbol` 

We were waiting a while to decide on using `ellipsis` to truncate long token names - [ENG-33](https://linear.app/leather-io/issue/ENG-33/define-and-implement-a-clear-strategy-for-handling-long-token-names). I'll implement that today which means we could / probably should re-add token names to the UI as that was the original design. 

I made a quick change locally to:
- get rid of the `isFiat` check in `balance` 
- add a  `catch` to `i18nFormatCurrency` for `SIP-10's` etc. as they were throwing `Warning: RangeError: Currency is invalid` 

The UI seems to mostly work but we need to make some slight tweaks to symbol position:
Screenshot 2025-05-21 at 05 51 14

@alexp3y how about refactoring this slightly to keep our crypto crypto and our fiat fiat? Then we can handle the rendering in [ENG-33](https://linear.app/leather-io/issue/ENG-33/define-and-implement-a-clear-strategy-for-handling-long-token-names)?

packages/models/src/currencies.model.ts
pete-watters: I added a suggestion to help with the to fiat / not to fiat conundrum  [here](https://github.com/leather-io/mono/pull/1228/files#r2099330932)

pete-watters (commented):
Nice work on this @alexp3y 👍  This looks good to me once we solve this issue of fiat / not. Take a look at [this suggestion ](https://github.com/leather-io/mono/pull/1228/files#r2099330932) when you can

packages/models/src/currencies.model.ts
tigranpetrossian: > Reading this code I'd assume it was a mistake, because I know that bitcoin isn't a fiat currency.
+1, this would be first reaction too--I think semantics are critical here.

apps/mobile/src/components/balance/balance.tsx
pete-watters: I realised I was wrong about `isFiat` here and there's more to it. We use it to format using `i18nFormatCurrency` vs `formatMoney` also

@alexp3y and I had a quick chat and he explained his reasoning to me and he's 💯 correct. 

He'll chime in here and give more details

alexp3y: A good example is that `STX 1` and `$0.56` in @pete-watters's screenshot both use the same `Balance` component and renders them completely differently based on the "type" of the currency, which it understands intrinsically from its symbol. 

If the currency's symbol is in the `FiatCurrency` type, it's assumed to be a view currency (e.g. "$0.56"), otherwise, it's treated like a crypto asset (e.g. "STX 1").

That assumption isn't unique to this component, but system-wide. Adding `XBT` and `sats` directly to `FiatCurrency` protects that assumption.

My point is that, if we want to be accurate, `FiatCurrency` _already is_ our "View Currency" type based on its usage. 

Or even more accurately its our "Base Currency" type, that is "a currency in which any other currency can be denominated / priced via MarketData (and that users can preferentially select as their view currency)".

Further, once we introduce "BTC" and "sats", it begs the question of whether the standalone concept of "fiat currency" is even useful at all?

packages/models/src/currencies.model.ts
alexp3y: Responded to the accuracy point here: https://github.com/leather-io/mono/pull/1228#discussion_r2102216334

apps/mobile/src/components/balance/balance.tsx
alexp3y: > If you were starting this from scratch @alexp3y how would you model the data in the way that allows the view to present as it wishes?

With 20/20 hindsight, I don't think I would have a single model - There are two concerns that need to be separated:
1. Currencies in which we support pricing crypto currencies via MarketData
2. User currency (conversion unit) preference options

The BTC - sat equivalence you mentioned highlights how these are independent.

We might have something in `models` like:
```ts 
export type QuoteCurrency = LiteralUnion<
  'USD' | 'EUR' | 'GBP' | 'AUD' | 'CAD' | 'CNY' | 'JPY' | 'KRW' | 'BTC',
  string
>;
```

Then we could define a separate `DisplayCurrency` to be used as the user currency setting. This one may be best left to apps to define independently, rather than using a globally shared type.

The services layer is only aware of `QuoteCurrency` so we would need to:
* Map `DisplayCurrency` to a `QuoteCurrency` and inject it into services to use during pricing / creating `MarketData`.
* Map `Money` objects (balances) coming out of the services layer to the correct `DisplayCurrency` and correctly format.

Finally we would ensure there are no assumptions that just because a currency symbol matches a `QuoteCurrency` that it can be assumed to be a display currency (e.g. BTC is both a QuoteCurrency AND a CryptoCurrency). All views would need added context to know when a balance should be formatted as a crypto currency or a user display currency.

packages/models/src/currencies.model.ts
pete-watters: How about have a type `BaseCurrency` that's a `LiteralUnion< FiatCurrency | 'XBT' | 'sats',`

Could we even just have `BTC` as a base? And then we format it as BTC / SATS based on another setting we need to implement [LEA-2033](https://linear.app/leather-io/issue/LEA-2033/display-settings-bitcoin-unit-setting)

So we add `BTC` as a base currency and then whether the user sees BTC / SATS depends on the setting in LEA-2033

fbwoolf: Trying to wrap my head around this one, it does feel off to me. I do think an additional type should be made, or the current `FiatCurrency` should be renamed.

pete-watters: The more I think of it `fiatCurrency` is kind of just a localisation / display thing. Maybe we should model it on [trading pairs](https://www.cmcmarkets.com/en-ie/trading-guides/forex-currency-pairs#:~:text=The%20base%20currency%20is%20the,sell%20rate%2Fbuy%20rate).) where we would have `BaseCurrency` + `QuoteCurrency` ?  

We deal with BTC and STX as our `BaseCurrencies` then based on locale some users choose to display their `QuoteCurrency` currency in different ways. Then we don't really have to classify things as `fiat` or `crypto` everywhere

apps/mobile/src/components/balance/balance.tsx
kyranjamie: Makes a lot of sense, let's do this instead? Would it be such a big refactor?

apps/web/app/store/fiat-currency.ts
```diff
@@ -1,9 +1,9 @@
 import { useAtom } from 'jotai/index';
 import { atomWithStorage } from 'jotai/utils';
 
-import { FiatCurrency } from '@leather.io/models';
+import { QuoteCurrency } from '@leather.io/models';
 
-export const fiatCurrencyAtom = atomWithStorage<FiatCurrency>('fiatCurrency', 'USD');
+export const fiatCurrencyAtom = atomWithStorage<QuoteCurrency>('fiatCurrency', 'USD');
```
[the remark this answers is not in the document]
kyranjamie: this isn't important state so we can rename without worrying about the key change

alexp3y:
> I'm fine with this approach, but I think we should remove all the instances where the renamed `QuoteCurrency` type is used to mean `FiatCurrency`. Otherwise developers encountering this code will still understandably be very confused

Addressed the outstanding comments in the latest push. Agree that it helps clarify. There are still instances of vars / pararms with in `mobile` with `fiat` but for now I believe we covered everything directly typed to `QuoteCurrency`. LMK if you think anything else should to be migrated.

apps/mobile/src/components/balance/balance.tsx
pete-watters: Sorry for the delayed reply here @alexp3y , happy you merged it and we can continue to improve on it if needed

PR #1231, opened 2025-05-20
feat(web): add liquid stacking protocol overview

alter-eggo:
This pr adds:
- real data in liquid stacking protocol overview
- real data on active liquid stacking page
- real market data in all the flows 
- fix part of contract calls queries (use stacks client methods to avoid 429 errs)

### DIFF
--- a/apps/web/app/queries/protocols/use-protocol-balance.ts
+++ b/apps/web/app/queries/protocols/use-protocol-balance.ts
@@ -1,0 +1,46 @@
+import { useMemo } from 'react';
+
+import { useQueries } from '@tanstack/react-query';
+import { ProtocolSlug } from '~/features/stacking/start-liquid-stacking/utils/types-preset-protocols';
+import { getNetworkInstanceByName } from '~/features/stacking/start-pooled-stacking/utils/utils-stacking-pools';
+import { createGetDaoBalanceQueryOptions } from '~/queries/protocols/dao/balance';
+import { createGetLisaBalanceQueryOptions } from '~/queries/protocols/lisa/balance';
+import { CreateProtocolBalanceQueryOptionsParams } from '~/queries/protocols/protocol-types';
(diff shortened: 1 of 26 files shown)


apps/web/app/queries/protocols/use-protocol-balance.ts
```diff
@@ -0,0 +1,41 @@
...
+      networkUrl: network.networkPreference.chain.stacks.url,
+    };
+
+    switch (slug) {
+      case 'stacking-dao':
+        return createGetDAOBalanceQueryOptions(options);
+      case 'lisa':
+        return createGetLisaBalanceQueryOptions(options);
+      default:
+        return createEmptyQueryOptions();
```
kyranjamie, line 36: haven't seen this technique before lol, if we need conditional queries, isn't it better to populate an array of useQueries? Which can be empty, rather than mocked?

apps/web/app/queries/protocols/use-protocol-fee.ts
```diff
@@ -0,0 +1,37 @@
...
+      networkMode,
+    };
+
+    switch (slug) {
+      case 'stacking-dao':
+        return createGetDAOFeeQueryOptions(options);
+      default:
+        return createEmptyQueryOptions();
+    }
+  }, [address, client, networkMode, slug]);
```
kyranjamie, line 34: I do wonder if a better approach is to just have each their own route and coming separately rather than having conditional logic everywhere in the code?

alter-eggo: sep routes? I use it in table to display fees
I will need to add conditionals all the same in the table, with different queries

kyranjamie: all good for now, but later I think we should separate in routes more, even if this specifically is used in table

kyranjamie (approved):

PR #1234, opened 2025-05-21
fix: update browser suggested tab [LEA-2554]

edgarkhanzadian:
https://github.com/user-attachments/assets/b31154a5-8331-4150-81e1-63092ba0335e


apps/mobile/src/features/browser/browser/dapp-card.tsx
```diff
@@ -1,28 +1,66 @@
...
+import { BlurView, Box, Pressable, Text, Theme } from '@leather.io/ui/native';
+
+const dAppImageAspectRatio = 342 / 400;
+const { width: screenWidth } = Dimensions.get('window');
+
+interface CaptionWrapperProps {
+  children: ReactNode;
+}
+
+// TODO: this is used until we figure out the BlurView for android
```
pete-watters, line 17: Do we have a ticket we can link to this?

apps/mobile/src/features/browser/browser/tabs/browser-suggested-tab.tsx
```diff
@@ -9,76 +15,127 @@ import { BrowserTabSheetLayout } from './browser-tab-sheet.layout';
 interface SuggestedProps {
   goToUrl(url: string): void;
 }
+const alexLink = 'https://alexgo.io/';
```
pete-watters, line 18: Maybe we should add these in the `constants` package?

PR #1238, opened 2025-05-21
fix: linking to browser instead of safari [LEA-2490]

edgarkhanzadian:
https://github.com/user-attachments/assets/f072129f-bb49-462e-83bd-edcce241d48a


pete-watters (approved):
Thanks @edgarkhanzadian ! It was easier than I thought using `linkingRef.` 

Thanks for refactoring the activity stuff also

tigranpetrossian (approved):

PR #1239, opened 2025-05-21
fix(mobile): ENG-41 stop the app crashing when deleting wallets

pete-watters:
This PR fixes a bug whereby if:
- AS a user I login and load a wallet
- WHEN I delete the wallet 
- AND I have no wallets
- THEN the app crashes with: `It is always expected an account has both Taproot and Native Segwit`


apps/mobile/src/components/headers/animated-header/animated-header-screen.layout.tsx
```diff
@@ -2,7 +2,7 @@ import Animated from 'react-native-reanimated';
 
 import { AnimatedTitleHeader } from '@/components/headers/animated-header/animated-title-header';
 
-import { Box, HasChildren, TextProps } from '@leather.io/ui/native';
```
camerow, line 5: Our linter wasn't calling this out?

PR #1241, opened 2025-05-21
fix: android navigation and blurview

edgarkhanzadian:
BlurView doesn't work properly on Android so we need to just render Box instead of it. Also set the warning on BlurView usage in android

### DIFF
--- a/apps/mobile/src/components/action-bar/action-bar.tsx
+++ b/apps/mobile/src/components/action-bar/action-bar.tsx
@@ -1,5 +1,5 @@
 import { ReactNode, useMemo, useRef } from 'react';
-import { ViewStyle } from 'react-native';
+import { Platform, ViewStyle } from 'react-native';
 import Animated, { FadeInDown, FadeOutDown } from 'react-native-reanimated';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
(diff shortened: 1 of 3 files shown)


apps/mobile/src/components/action-bar/action-bar.tsx
```diff
@@ -92,37 +92,60 @@ interface ActionBarContainerProps {
...
+    marginHorizontal: theme.spacing[5],
+    paddingHorizontal: theme.spacing[6],
+    paddingVertical: theme.spacing[2],
+    overflow: 'hidden',
+    borderRadius: theme.borderRadii['sm'],
+    borderColor: theme.colors['ink.border-transparent'],
+    borderWidth: 1,
+  } as const;
+
+  switch (Platform.OS) {
```
tigranpetrossian, line 117: `Platform.select` you used in [create wallet](https://github.com/leather-io/mono/blob/42eb396a3ba78634465adfeec8c2c33db8e6f115/apps/mobile/src/app/create-new-wallet.tsx#L30) seemed much less verbose. Can we restort to that instead adding a component?

tigranpetrossian (approved):
Thanks Edgar!

apps/mobile/src/components/action-bar/action-bar.tsx
tigranpetrossian: No you don't 😛

PR #1254, opened 2025-05-22
fix: android sluggish sheet [LEA-2648]

edgarkhanzadian:
Multiple things happening here:
1. Moving AddWalletSheet and AddAccountSheet to global sheets as those sheets are called pretty much from every part of the app
2. delete forwardRef on Pressable, Card and Cell components
3. And, most importantly, refactored the `usePressedState` function and the animated styles that depended on it. We should always use reanimated's SharedValue instead of regular state whenever we are dealing with worklets (i.e. useAnimatedStyles, useAnimatedProps). That drastically improves performance of animations

https://github.com/user-attachments/assets/a8b626c9-d82d-4fd7-aa92-979354d25d87

### DIFF
--- a/apps/mobile/src/features/wallet-manager/add-wallet/add-wallet-sheet.tsx
+++ b/apps/mobile/src/features/wallet-manager/add-wallet/add-wallet-sheet.tsx
@@ -1,9 +1,10 @@
-import { RefObject, useCallback, useRef, useState } from 'react';
+import { useCallback, useRef, useState } from 'react';
 
 import {
   NotifyUserSheetData,
   NotifyUserSheetLayout,
 } from '@/components/sheets/notify-user-sheet.layout';
+import { useGlobalSheets } from '@/core/global-sheet-provider';
 import { AppRoutes } from '@/routes';
 import { useSettings } from '@/store/settings/settings';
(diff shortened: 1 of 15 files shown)


apps/mobile/src/features/wallet-manager/add-wallet/add-wallet-sheet.tsx
```diff
@@ -49,7 +46,8 @@ export function AddWalletSheet({ addWalletSheetRef, opensFully }: AddWalletSheet
         restoreWallet={restoreWallet}
         addWalletSheetRef={addWalletSheetRef}
         themeVariant={themeDerivedFromThemePreference}
-        opensFully={opensFully}
+        // TODO: this should be set when we call ref.current.present
```
pete-watters, line 49: Can we do this now?

pete-watters (approved):

apps/mobile/src/features/wallet-manager/add-wallet/add-wallet-sheet.tsx
pete-watters: No worries, I didn't realise it was for waitlist

PR #1257, opened 2025-05-22
Fix/eng 33/long token names

pete-watters:


packages/utils/src/money/format-money.ts
```diff
@@ -51,8 +51,25 @@ export function createMoney(value: NumType, symbol: Currency, resolution?: numbe
 
 const thinSpace = ' ';
 
+function exponentialToNumerical(amount: BigNumber) {
```
pete-watters, line 54: I added this code so that we can handle `exponential` values properly. @camerow we've gone full circle here as I actually[ grabbed this snippet](https://github.com/pete-watters/crypto-view/blob/develop/src/tools/serializer.js) from an interview assessment I did way back when we first met

It could perhaps be improved but it works and allows us to render scientific notation properly

fbwoolf: Are you sure there isn't an issue where you are trying to format a value that was incorrectly converted to the Money type in the first place? I had a similar bug reported in swaps at one point, but if I remember correctly it was just that createMoney was being used instead of createMoneyFromDecimal (I think). I'm not sure if that is the issue here?

pete-watters: Thanks @fbwoolf , I will check that out shortly and refactor as needed. Maybe we need to make that change in the `balance` / `activity` service so it does it correctly at source and then not add this formatting here. 

It only seems to have trouble with some of the rarer / custom SIP-10's

PR #1259, opened 2025-05-22
chore(mobile): gate runes total balances behind feature flag

alexp3y:
Rune balances are still being incorporated into total and account total balance. 

This PR puts the inclusion of runes balances into totals behind the runes feature flag.


pete-watters (approved):
Maybe we should add a hard flag to the `Runes` balances service too? Similar to what we did for collectibles

We don't need to fetch anything for Runes.

PR #1261, opened 2025-05-22
feat: filter non-primary utxos from service, ref LEA-2467

alexp3y:
Filters out non-primary account address UTXOs from `UtxosService` (and therefore all balance services), i.e. no change address or non-zero index UTXOs, i.e. anything not on a derivation path matching `m/84 or 86/0'/< account index >/0/0`.

Shorter-term fix for BTC send flow issue - Allows us to continue using and testing sends.

Ties app BTC balances and send behavior more closely to that currently in extension.

### DIFF
--- a/packages/crypto/src/derivation-path-utils.ts
+++ b/packages/crypto/src/derivation-path-utils.ts
@@ -9,7 +9,7 @@ export enum DerivationPathDepth {
   AddressIndex = 5,
 }
 
-function extractSectionFromDerivationPath(depth: DerivationPathDepth) {
+export function extractSectionFromDerivationPath(depth: DerivationPathDepth) {
   return (path: string) => {
     const segments = path.split('/');
     const accountNum = parseInt(segments[depth].replaceAll("'", ''), 10);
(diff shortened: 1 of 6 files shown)


pete-watters (approved):
Thanks @alexp3y

alexp3y:
> Can merge if you like, but will probably want to revert when tackling change address spends

@kyranjamie Sounds good. I didn't know whether you'd have availability before the mobile release so I just wanted to have something prepared as a fallback. But we can hold off on merging unless / until absolutely needed.

pete-watters:
> > Can merge if you like, but will probably want to revert when tackling change address spends
> 
> @kyranjamie Sounds good. I didn't know whether you'd have availability before the mobile release so I just wanted to have something prepared as a fallback. But we can hold off on merging unless / until absolutely needed.

At first I thought let’s hold off merging this to save us reverting but as the release path for mobile is a bit longer and this is so critical I think we should get this in and then remove.

We will need to merge it here, then release dev, then get it into the TestFlight build. It doesn’t take that long but I think having this workaround in place for now to relieve the pressure is worth it

PR #1264, opened 2025-05-23
fix: revamp splash screen [LEA-2654]

edgarkhanzadian:
Revamp splash screen:
1. Fixes useless authentication checks that can be cancelled when app is opened
2. Hide the screen if user comes back after a minute of the app being in the background

### DIFF
--- a/apps/mobile/src/components/splash-screen-guard/use-auth-state.tsx
+++ b/apps/mobile/src/components/splash-screen-guard/use-auth-state.tsx
@@ -1,99 +1,174 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useReducer } from 'react';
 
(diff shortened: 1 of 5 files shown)


apps/mobile/src/components/splash-screen-guard/use-auth-state.tsx
```diff
@@ -1,99 +1,174 @@
...
+  | { type: 'authFailed' }
+  | { type: 'onFinishAnimation' };
+
+// Initial state
+const initialState: AuthState = {
+  status: 'cold-start',
+  isUnlocked: false,
+};
+
+function reducer(state: AuthState, action: AuthAction): AuthState {
```
pete-watters, line 34: Nice 😎

pete-watters (approved):

PR #1269, opened 2025-05-23
fix(web): page header to follow grid

kyranjamie:
General style and refactors for web


markmhendrickson:
@kyranjamie nudge on the above fix 🙏

PR #1271, opened 2025-05-23
feat: add stacking tracking

camerow:
We would like telemetry on the volume of use for our stacking providers. This PR adds some telemetry about which pools are used and how much people are stacking. There is very little PII here so I think this is safe to do and this is all public blockchain data.

### DIFF
--- a/apps/web/app/features/stacking/increase-liquid-stacking/utils/utils-increase-liquid-stacking.ts
+++ b/apps/web/app/features/stacking/increase-liquid-stacking/utils/utils-increase-liquid-stacking.ts
@@ -7,7 +7,7 @@ import { stxToMicroStxBigint } from '~/utils/unit-convert';
 
 import { LeatherSdk } from '@leather.io/sdk';
 import { formatContractId } from '@leather.io/stacks';
-import { stxToMicroStx } from '@leather.io/utils';
+import { scaleValue, stxToMicroStx } from '@leather.io/utils';
 
 import { StackIncreaseInfo } from '../../direct-stacking-info/get-has-pending-stack-increase';
 import { SignerDetailsFormValues } from './types';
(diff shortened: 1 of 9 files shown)


apps/web/app/features/stacking/start-pooled-stacking/utils/utils-delegate-stx.ts
```diff
@@ -111,6 +112,14 @@ export function createDelegateStxMutationOptions({
 
       const delegateStxOptions = getDelegateStxOptions(values, poxInfo, network);
 
+      await analytics.track('pooled_stacking_started', {
```
markmhendrickson, line 115: Let's add this sort of contract-parameter tracking on the extension side, too, for thoroughness (and without tracking the executor's address per se, for PII-protection purposes). That way, we can see the range of contracts, functions and parameters being executed in general via the extension in addition to seeing them triggered here on the web side.

apps/web/app/queries/balance/account-balance.hooks.ts
```diff
@@ -37,7 +38,13 @@ export function useStxCryptoAssetBalance(address: string) {
     }),
     select: resp => {
       const initialBalance = createMoney(new BigNumber(resp.balance), 'STX');
-      return createStxCryptoAssetBalance(initialBalance, inboundBalance, outboundBalance);
+      const stxBalance = createStxCryptoAssetBalance(
+        initialBalance,
+        inboundBalance,
+        outboundBalance
+      );
+      void analytics.track('stx_balance_updated', stxBalance);
```
markmhendrickson, line 46: When is this triggered exactly?

markmhendrickson (changes requested):

apps/web/app/features/stacking/increase-liquid-stacking/utils/utils-increase-liquid-stacking.ts
```diff
@@ -47,7 +47,9 @@ export function createIncreaseLiquidMutationOptions({
         authId: authId ? parseInt(authId, 10) : undefined,
       });
 
-      void analytics.untypedTrack('liquid_stacking_increased');
+      void analytics.track('liquid_stacking_increased', {
+        amount: increaseBy,
```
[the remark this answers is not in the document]
camerow: I know we did discuss this, but is it really more protective since these are all public blockchain txs?

markmhendrickson (approved):

PR #1272, opened 2025-05-23
DRAFT: feat: add ability to track services data and actions

camerow:
Adds the ability to dependency inject analytics tracking into the services package. Doesn't use it at all in this PR, I wanted to vet the base inclusion of this before I went ahead and started tracking things.


pete-watters (approved):

tigranpetrossian (approved):

alexp3y (commented):
Great usage of the services package here. Will be cool to see this used. 

Just need to adjust how we're defining the service to ensure the client is available in the DI container as intended.

PR #1273, opened 2025-05-23
feat: add min max fns to calc money

fbwoolf:
Possible min/max fns for calculating Money.


pete-watters:
@fbwoolf I tagged you on this PR https://github.com/leather-io/mono/pull/1257 as in `mobile` I noticed we don't handle exponential values in `formatMoney` so I added some code to do that there

packages/utils/src/money/calculate-money.ts
```diff
@@ -60,3 +60,23 @@ export function sumMoney(moneysArr: Money[]) {
...
+
+export function maxMoney(xAmount: Money, yAmount: Money) {
+  if (xAmount.symbol !== yAmount.symbol)
+    throw new Error('Cannot calculate max of different currencies');
+  return createMoney(
+    BigNumber.max(xAmount.amount, yAmount.amount),
+    xAmount.symbol,
+    xAmount.decimals
+  );
+}
```
kyranjamie, line 82: I suggest we write these utilities to support n amount of money objects

kyranjamie (approved):
These could definitely do with some unit tests 🙏🏼 

Other than this, LGTM

PR #1295, opened 2025-05-27
feat: non-zero index/change utxo spends

kyranjamie:
Mobile bitcoin tx signing logic is written with the assumption we only ever spend from the zero change and address index of an account e.g. `m/84'/0'/0'/0/0`, illustrated by the many instances of `nativeSegwitAccount.derivePayer({ addressIndex: 0 })`. Currently, Leather always send change to the zero index address, however users importing from other wallets may have funds on a change address e.g. `m/84'/0'/0'/1/0`.

This task is to refactor the PSBT signing logic to handle transactions constructed with inputs from many paths.

To start, I've refactored the `derivePayer` function to require a `change` value. Previously it was named `receive` and defaulted to `0`, but this is more correct terminology and it's clearer to be explicit.

The issue preventing the spend of the change address UTXO was caused by an incorrect `bip32Derivation` path  being added to the input at _tx creation_ time. The signer logic was reading this, and thus trying to sign from the wrong key. I've removed some code that referred to "output derivation paths", pretty sure this was a mistake, as bitcoin tx outputs don't correspond to keys, in the same way inputs we spend do, so think it's safe to have been deleted.

After fixing this, I was still unable to get the transaction to broadcast, with a slightly different error

```
{"code":-26,"message":"mandatory-script-verify-flag-failed (Witness program hash mismatch)"}
```

Turns out we also had spendable Taproot inputs in our set of UTXOs, which were being passed to the Native Segwit assuming logic, again resulting in us trying to sign the wrong key. I've refactored the tx generation function to support taproot inputs as well, fixing this issue.

I've ran into an issue where in some cases I have some UTXOs but not the information to know which fingerprint they belong to. Issue here: https://linear.app/leather-io/issue/LEA-2691/including-keyorigin-in-utxo-response

In meantime, I've added `AccountId` to `PsbtSigner` so we have some relevant info available.

I'm still fighting some utxo typing issue, as there are conflicts between value being string/number in different cases, don't pay attention to these, chatting with @alexp3y how we can use more consistent utxo types and will fix asap

### DIFF
--- a/apps/mobile/src/features/browser/approver-sheet/utils.ts
+++ b/apps/mobile/src/features/browser/approver-sheet/utils.ts
@@ -22,21 +22,20 @@ import {
 import { BitcoinNetworkModes } from '@leather.io/models';
 import { RpcRequests } from '@leather.io/rpc';
 
-interface InfuseBip32Props {
+interface AddBip32DerivationFieldToInputsProps {
   psbtHex: string;
   networkMode: BitcoinNetworkModes;
   bitcoinAccounts: ReturnType<typeof useBitcoinAccounts>['list'];
   accountId: string;
   signAtIndex: number | number[] | undefined;
 }
-
 export function addBip32DerivationFieldToInputs({
   psbtHex,
   networkMode,
   bitcoinAccounts,
   accountId,
   signAtIndex: _signAtIndex,
-}: InfuseBip32Props) {
+}: AddBip32DerivationFieldToInputsProps) {
   const signAtIndex = normalizeSignAtIndex(_signAtIndex);
   const tx = getPsbtAsTransaction(psbtHex);
 
@@ -56,9 +55,7 @@ export function addBip32DerivationFieldToInputs({
     });
     const bitcoinAccount = findAccountByAddress(bitcoinAccountsById, addr);
 
-    if (!bitcoinAccount) {
-      return;
-    }
+    if (!bitcoinAccount) return;
 
     const accountIdx = extractAccountIndexFromDescriptor(bitcoinAccount?.descriptor);
     const accountFingerprint = extractFingerprintFromDescriptor(bitcoinAccount?.descriptor ?? '');
@@ -66,7 +63,8 @@ export function addBip32DerivationFieldToInputs({
     const bitcoinAccountId = makeAccountIdentifer(accountFingerprint, accountIdx);
 
     if (bitcoinAccountId === accountId) {
-      const payer = bitcoinAccount.derivePayer({ addressIndex: 0 });
+      // TODO in #1295: use dynamic payer info
+      const payer = bitcoinAccount.derivePayer({ change: 0, addressIndex: 0 });
       if (paymentType === 'p2tr') {
         tx.updateInput(idx, {
           tapBip32Derivation: [payerToTapBip32Derivation(payer)],
--- a/apps/mobile/src/features/browser/approver-sheet/utils.ts
+++ b/apps/mobile/src/features/browser/approver-sheet/utils.ts
@@ -44,9 +44,9 @@ export function addBip32DerivationFieldToInputs({
     if (input.bip32Derivation || input.tapBip32Derivation) return;
     const bitcoinNetwork = getBtcSignerLibNetworkConfigByMode(networkMode);
     const addr = getBitcoinInputAddress(input, bitcoinNetwork);
-    if (!addr) {
-      throw new Error('Unsupported bitcoin address');
-    }
+
+    if (!addr) throw new Error('Unsupported bitcoin address');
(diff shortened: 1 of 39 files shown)


tigranpetrossian (approved):
Thanks so much for the help with this @kyranjamie 
The code looks good, I'll need to take another look post-merge to grasp all the details myself.

apps/mobile/src/features/send/screens/form.tsx
```diff
@@ -69,39 +69,35 @@ export function Form() {
...
+                    fiatCurrency={fiatCurrencyPreference}
+                    nonce={nonce}
+                    onOpenAssetPicker={handleOpenAssetPicker}
+                    assetItemAnimationOffsetTop={assetItemElementInitialOffset}
+                  />
+                )}
               </StxDataLoader>
             ),
             btc: (
               <BtcDataLoader account={selectedAccount}>
```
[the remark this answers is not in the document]
kyranjamie: yeah also curious how we can start utilising `use`

apps/mobile/src/utils/special-char.ts
```diff
@@ -0,0 +1,3 @@
+// A minus sign is a wider character than a hyphen, and should be used in
```
pete-watters, line 1: Thanks for fixing this for me! I remember you flagged it previously but I forgot to do it

apps/mobile/src/features/browser/approver-sheet/utils.ts
```diff
@@ -22,21 +22,20 @@ import {
 import { BitcoinNetworkModes } from '@leather.io/models';
 import { RpcRequests } from '@leather.io/rpc';
 
-interface InfuseBip32Props {
+interface AddBip32DerivationFieldToInputsProps {
```
pete-watters, line 25: Great renaming 👏 this will help us a lot when re-visiting

apps/mobile/src/utils/special-char.ts
kyranjamie: Once you see you cannot unsee haha

apps/mobile/src/store/keychains/bitcoin/bitcoin-keychains.read.ts
```diff
@@ -126,16 +136,20 @@ export function useBitcoinPayerAddressFromAccountIndex(fingerprint: string, acco
     accountIndex
   );
 
-  const taprootPayerAddress = taproot?.derivePayer({ addressIndex: 0 }).address ?? '';
-  const nativeSegwitPayerAddress = nativeSegwit?.derivePayer({ addressIndex: 0 }).address ?? '';
+  const taprootPayerAddress = taproot?.derivePayer({ change: 0, addressIndex: 0 }).address ?? '';
+  const nativeSegwitPayerAddress =
+    nativeSegwit?.derivePayer({ change: 0, addressIndex: 0 }).address ?? '';
```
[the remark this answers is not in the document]
pete-watters: I added a lot of horrible code in [this PR](https://github.com/leather-io/mono/pull/1277/files) to stop the APP from crashing on deletion.

I added a lot of optional chains and defaults to `''` in a hurry to get a critical bug fixed. 

It would be good to revisit the whole area and improve on it. Maybe there's a better way of making signers optional

packages/bitcoin/src/coin-selection/coin-selection.ts
```diff
@@ -13,29 +13,21 @@ export interface CoinSelectionOutput {
   address?: string;
 }
 
-export interface CoinSelectionUtxo {
```
pete-watters, line 16: Are you removing this as it's just annoying and easier to use more generic types?

packages/bitcoin/src/fees/bitcoin-fees.ts
```diff
@@ -2,13 +2,12 @@ import { AverageBitcoinFeeRates, Money } from '@leather.io/models';
...
 import {
   CoinSelectionRecipient,
-  CoinSelectionUtxo,
   DetermineUtxosForSpendArgs,
   determineUtxosForSpend,
   determineUtxosForSpendAll,
 } from '../coin-selection/coin-selection';
 
-type GetBitcoinTransactionFeeArgs = DetermineUtxosForSpendArgs & {
+type GetBitcoinTransactionFeeArgs = DetermineUtxosForSpendArgs<{ value: string; txid: string }> & {
```
pete-watters: Should we just add a type for `{ value: string; txid: string }`?

pete-watters: Don't worry, it's not a big deal. It's actually probably easier to read if repeated than if it was abstracted

PR #1299, opened 2025-05-28
feat: sentry

edgarkhanzadian:
After digging through xcode logs it seems like sentry has an implicit dependency for @sentry/cli (and implicit dependency are a big no-no in pnpm). Anyway, adding that package fixed the EAS builds!

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -32,7 +32,7 @@ import { trackFirstAppOpen } from '@/utils/analytics';
 import { LDProvider } from '@launchdarkly/react-native-client-sdk';
 import { i18n } from '@lingui/core';
 import { I18nProvider } from '@lingui/react';
-// import * as Sentry from '@sentry/react-native';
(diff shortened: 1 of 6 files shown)


apps/mobile/src/app/_layout.tsx
```diff
@@ -43,17 +43,12 @@ import { Box, SheetProvider } from '@leather.io/ui/native';
...
-//   sendDefaultPii: true,
-//   integrations: [Sentry.feedbackIntegration()],
-//
-//   // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-//   // spotlight: __DEV__,
-// });
+Sentry.init({
+  dsn: process.env.EXPO_PUBLIC_SENTRY_DSN,
+  debug: __DEV__,
+  tracesSampleRate: 1.0,
```
tigranpetrossian, line 49: I think 1 is way too high for prod, if I remember correctly Sentry charged based on the rate. We can easily go with 0.1 and adjust from there.

PR #1310, opened 2025-05-30
fix: ensure we set proper border radius and use container on browser

camerow:
Closes LEA-2492

@pete-watters / @edgarkhanzadian biggest change here is to use the bottom sheet container for the browser sheet, which required adding `flex: 1`. I did notice that no other sheets set this to `true` (see image) so it's a safe change but we may want to rething how we handle the container, perhaps just an exported component that can be composed in? 
Screenshot 2025-05-30 at 8 39 12 AM

### DIFF
--- a/apps/mobile/src/features/browser/browser/browser-sheet.tsx
+++ b/apps/mobile/src/features/browser/browser/browser-sheet.tsx
@@ -66,7 +66,8 @@ export function BrowserSheet() {
     <Sheet
       ref={browserSheetRef}
       snapPointVariant="fullHeightWithoutNotch"
-      shouldHaveContainer={false}
+      shouldHaveContainer={true}
+      isScrollView={false}
       themeVariant={themeDerivedFromThemePreference}
       onDismiss={resetSearchBar}
     >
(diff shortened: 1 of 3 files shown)


tigranpetrossian:
> we may want to rething how we handle the container, perhaps just an exported component that can be composed in? 

@camerow 
[LEA-2086: Refactor native Sheet component](https://linear.app/leather-io/issue/LEA-2086/refactor-native-sheet-component) 
I've started working on this, but it took a backseat due to the release. Should solve most of tight coupling/overconfiguration issues.

apps/mobile/src/features/browser/browser/browser-sheet.tsx
```diff
@@ -66,7 +66,7 @@ export function BrowserSheet() {
     <Sheet
       ref={browserSheetRef}
       snapPointVariant="fullHeightWithoutNotch"
-      shouldHaveContainer={false}
+      shouldHaveContainer={true}
```
pete-watters, line 69: Not a big deal but I think you can just specify it as `shouldHaveContainer` when its positive (no `={true}`

pete-watters:
@camerow I approved this a little too hastily and it has had some consequences: 
Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-06-03 at 09 16 25

I will try and fix it or else roll back. We need to do an overhaul of sheets.

PR #1311, opened 2025-05-30
fix: spreading was indexing on 0

camerow:
Event properties on mobile were broken because of the spread operator. They were coming through formatted as:

```json
{"0": {"network": "mainnet"}, "platform": "mobile"}
```

but should have been

```json
{"network": "mainnet", "platform": "mobile"}
```

### DIFF
--- a/packages/analytics/src/client.ts
+++ b/packages/analytics/src/client.ts
@@ -8,7 +8,7 @@ export function AnalyticsClient<T extends AnalyticsClientInterface>(
   return {
     async track<K extends keyof Events>(
       event: K,
-      properties?: undefined extends Events[K] ? [] : [param: Events[K]]
+      ...[properties]: Events[K] extends undefined ? [] : [Events[K]]
     ) {
       return analyticsClient.track(event, { ...properties, ...defaultProperties });
     },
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

pete-watters (approved):
I’m good with it if it works but it seems to have a lot of type errors? Maybe it needs to be refactored a bit more?

packages/analytics/src/client.ts
```diff
@@ -8,7 +8,7 @@ export function AnalyticsClient<T extends AnalyticsClientInterface>(
   return {
     async track<K extends keyof Events>(
       event: K,
-      ...properties: undefined extends Events[K] ? [] : [param: Events[K]]
+      properties?: undefined extends Events[K] ? [] : [param: Events[K]]
```
pete-watters, line 11: I find it a bit confusing what this is doing

pete-watters: Thanks @tigranpetrossian 👍 maybe with your update it's a bit clearer but sometimes I favour clarity over brevity. 

I haven't seen `undefined extends ` before so something new for me to learn.

tigranpetrossian: @pete-watters 
Sure. `extends` at type level just means "assignable" in TS, `undefined extends X`  is basically asking "is `undefined` a subset of type X`.

PR #1316, opened 2025-06-02
Add analytics to Send & Receive + setup improvements

tigranpetrossian:
PR Adds few small improvements to the telemetry Setup and event tracking for Send & Receive

### Setup improvements

**No `void`**
Disabled `no-floating-promises` just for this method, as **all** usages of it are fire & forget. I've debated adding a separate `trackAsync` method, but since we basically never wait for the promise resolution, silencing the rule is safe and the least noisy solution.

**No optional chaining**
`analytics` client now uses a proxy to ignore calls and can be safely called as `analytics.track`. I intentionally didn't update usage sites to keep the PR small.

**Disable Sentry in development mode**
Noticed few exceptions in Sentry from local setups. Sneaking in a commit to fix that.

### Tracking

* Fixed `SheetNavigator` tracks not being called due to a missing ref. 
* Added a `base` prefix to `SheetNavigator` screen calls: `/select-asset` -> `send/select-asset`
* Added tracking to **Action Bar**
* Added tracking calls to **Send** and **Receive** flows

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -47,6 +47,7 @@ Sentry.init({
   dsn: process.env.EXPO_PUBLIC_SENTRY_DSN,
   debug: __DEV__,
   tracesSampleRate: 1.0,
+  enabled: !__DEV__,
 });
 
 // Catch any errors thrown by the Layout component
--- a/apps/mobile/src/components/action-bar/use-action-bar.ts
+++ b/apps/mobile/src/components/action-bar/use-action-bar.ts
@@ -1,6 +1,7 @@
 import { useGlobalSheets } from '@/core/global-sheet-provider';
 import { useBrowserFlag } from '@/features/feature-flags';
 import { useWallets } from '@/store/wallets/wallets.read';
+import { analytics } from '@/utils/analytics';
 import { useSegments } from 'expo-router';
 
 const allowedSegmentConditions: ((segments: ReturnType<typeof useSegments>) => boolean)[] = [
@@ -17,18 +18,22 @@ export function useActionBar() {
   const isActionBarVisible = allowedSegmentConditions.some(condition => condition(segments));
 
   function onAddWallet() {
+    analytics.track('add_wallet_sheet_opened', { source: 'action_bar' });
     addWalletSheetRef.current?.present();
   }
 
   function onOpenSend() {
+    analytics.track('send_sheet_opened', { source: 'action_bar' });
     sendSheetRef.current?.present();
   }
 
   function onOpenReceive() {
+    analytics.track('receive_sheet_opened', { source: 'action_bar' });
     receiveSheetRef.current?.present();
   }
 
   function onOpenBrowser() {
+    analytics.track('browser_sheet_opened', { source: 'action_bar' });
     browserSheetRef.current?.present();
   }
 
--- a/apps/mobile/src/utils/analytics.ts
+++ b/apps/mobile/src/utils/analytics.ts
@@ -11,7 +11,7 @@ const FIRST_OPEN_KEY = 'first_open_tracked';
 const segmentClient = createClient({
   writeKey: process.env.EXPO_PUBLIC_SEGMENT_WRITE_KEY || '',
   trackAppLifecycleEvents: true,
-  debug: true,
+  debug: false,
 });
 
 segmentClient.add({ plugin: new AppLifecycleEventPlugin() });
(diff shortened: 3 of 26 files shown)


apps/mobile/src/app/_layout.tsx
```diff
@@ -47,6 +47,7 @@ Sentry.init({
   dsn: process.env.EXPO_PUBLIC_SENTRY_DSN,
   debug: __DEV__,
   tracesSampleRate: 1.0,
+  enabled: !__DEV__,
```
pete-watters, line 50: Thanks for doing this 👍

apps/mobile/src/utils/analytics.ts
```diff
@@ -16,15 +16,43 @@ const segmentClient = createClient({
...
-export let analytics: ReturnType<typeof configureAnalyticsClient<SegmentClient>> | undefined =
-  leatherAnalyticsClient;
+let analyticsEnabled = selectAnalyticsPreference(store.getState()) === 'consent-given';
+
+function setupPreferenceSubscription() {
+  return store.subscribe(() => {
+    const analyticsPreference = selectAnalyticsPreference(store.getState());
+
+    if (analyticsPreference === 'rejects-tracking') {
+      void analytics.track('user_setting_updated', { analytics: 'rejects-tracking' });
```
pete-watters, line 33: If they reject tracking I don’t think we should trigger an event to signify that

apps/mobile/src/components/action-bar/use-action-bar.ts
```diff
@@ -17,18 +18,22 @@ export function useActionBar() {
   const isActionBarVisible = allowedSegmentConditions.some(condition => condition(segments));
 
   function onAddWallet() {
+    analytics.track('add_wallet_sheet_opened', { source: 'action_bar' });
```
pete-watters, line 21: Is this correct? I don’t think we have ‘Add Wallet’ in the action bar?

apps/mobile/src/components/action-bar/use-action-bar.ts
```diff
@@ -17,18 +18,22 @@ export function useActionBar() {
...
     addWalletSheetRef.current?.present();
   }
 
   function onOpenSend() {
+    analytics.track('send_sheet_opened', { source: 'action_bar' });
```
pete-watters, line 26: Do we need to specify source here?

pete-watters (approved):
LGTM, I added a few Q’s but good to merge

apps/mobile/src/components/action-bar/use-action-bar.ts
tigranpetrossian: We'll likely to add more toggles for global sheets elsewhere in the future.

apps/mobile/src/utils/analytics.ts
tigranpetrossian: This isn't an added behavior, it was already there before the refactoring. It's pretty common nad a useful metric to have.

apps/mobile/src/utils/analytics.ts
```diff
@@ -16,15 +16,43 @@ const segmentClient = createClient({
...
+    if (typeof value === 'function') {
+      return (...args: unknown[]) => {
+        if (!analyticsEnabled) return;
+        return value.apply(target, args);
+      };
+    }
+
+    return value;
+  },
+});
```
kyranjamie, line 55: So the proxy just returns undefined to any fn call if analytics isn't enabled? 

I wonder if this is necessary, as the client already works in such a way that you can call all the methods but nothing happens if disabled (as far as I understand).

kyranjamie (approved):

apps/mobile/src/utils/analytics.ts
tigranpetrossian: Correct, just noop-ing all possible (unknown) methods.
The assumption here is that the client is [our own client adapter](https://github.com/leather-io/mono/blob/69566995267061d4661e6094c04d21b9a732deaf/packages/analytics/src/client.ts) not Segment, especially now that we might use something else.

In hindsight though, I'd love our `analytics` module to have `enabled` or any sort off a an on/off switch for this, so other apps could use it too. Something to look into.

apps/mobile/src/utils/analytics.ts
pete-watters: I know it was there before, just seems non consensual.

Tracking an event to signify unwillingness to be tracked

PR #1326, opened 2025-06-03
fix: egg sheet

edgarkhanzadian:
fixEggSheet

### DIFF
--- a/packages/ui/src/components/sheet/sheet.native.tsx
+++ b/packages/ui/src/components/sheet/sheet.native.tsx
@@ -132,8 +132,8 @@ export function Sheet({
             backgroundColor: theme.colors['ink.background-primary'],
             paddingBottom: bottom,
             flex: 1,
-            borderTopLeftRadius: theme.borderRadii.round,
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
LGTM, but shouldn't this be a larger border radius? cc/ @fabric-8 @mica000

pete-watters (approved):
Thanks @edgarkhanzadian . I had started fixing this but got distracted.

we should add this fix but @camerow changed the radius on purpose in his PR so we should follow up with that later to make sure it looks OK in the browser sheet

pete-watters:
@edgarkhanzadian - @camerow added this change here to fix a different issue [LEA-2492](https://linear.app/leather-io/issue/LEA-2492) - [PR](https://github.com/leather-io/mono/pull/1310)

I suggest we merge your change to fix the `Egg Sheet` (😂 ) and we can fix all the sheets properly in another task we have to refactor them - [LEA-2086](https://linear.app/leather-io/issue/LEA-2086/refactor-native-sheet-component)

PR #1337, opened 2025-06-05
fix(mobile): replace image stickers, ref ENG-110

pete-watters:
This PR:
- updates some sticker images
- attempts to update some text 

You can see the new stickers here:

https://github.com/user-attachments/assets/0b0e568a-9d2a-494d-8e08-5083ad8150bb

https://github.com/user-attachments/assets/7ab5693a-1e4f-4b95-b353-dd438578f1e1


apps/mobile/src/features/activity/activity-empty.tsx
```diff
@@ -19,7 +19,7 @@ export function ActivityEmpty() {
         <Text textAlign="center" variant="label01">
           {t({
             id: 'activity-empty.title',
-            message: 'Make your first transaction to get started',
```
pete-watters, line 22: I had to change this text but I can't see it being taken by `lingui compile` so it wasn't updating properly

pete-watters:
I had another commit that was updating text but it wasn't working and I think it relies on first fixing the lingui setup so I removed it:
Screenshot 2025-06-05 at 09 10 57

tigranpetrossian (approved):

PR #1338, opened 2025-06-05
fix(mobile): banish fiat from the app, ref ENG-109

pete-watters:
This PR:
- updates our code to banish instances of `fiat` , replacing them with `quote`
- simplifies the `Balance` component so it accepts a prop for `isQuoteCurrency` rather than a lookup
- tweaks the text size of `activity` to be smaller for when quote currency is BTC

https://github.com/user-attachments/assets/3207fd8e-222b-4ad3-a694-13f769f0908f


alexp3y (approved):
Adios fiat. 🦾

packages/analytics/src/events.ts
```diff
@@ -152,6 +152,7 @@ type UserSettingValue =
   | { bitcoin_unit: 'bitcoin' | 'satoshi' }
   | { email_address: string }
   | { fiat_currency: string }
+  | { quote_currency: string }
```
pete-watters, line 155: @camerow what should we do with this event? I added a new `event` as I'd like to rename it as `quote_currency` but I didn't want to remove `fiat_currency` either

apps/mobile/src/store/settings/settings.ts
```diff
@@ -58,7 +58,7 @@ export const initialState: SettingsState = {
   bitcoinUnitPreference: 'bitcoin',
   createdOn: new Date().toISOString(),
   emailAddressPreference: '',
-  fiatCurrencyPreference: 'USD', // TODO: migrate to quoteCurrencyPreference
+  quoteCurrencyPreference: 'USD', // TODO: migrate to quoteCurrencyPreference
```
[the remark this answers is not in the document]
pete-watters: I'll finish this settings migration in a smaller task - https://linear.app/leather-io/issue/LEA-2723/refactor-mobile-settings-to-migrate-fiatcurrencypreference-to

packages/analytics/src/events.ts
pete-watters: I'll keep both for now and we can retire `fiat_currency` / merge them in future

PR #1339, opened 2025-06-05
fix(mobile): use availableUnlockedBalance for STX in account selection screen, ref LEA-2644

pete-watters:


apps/mobile/src/features/send/screens/select-account.tsx
```diff
@@ -80,18 +80,20 @@ interface AccountItemProps {
...
-  if (balance.state !== 'success') {
+  });
+
+  if (btc.state !== 'success' || stx.state !== 'success') {
     return null;
   }
 
-  if (balance.value.quote.availableBalance.amount.isZero()) {
+  const availableBalance =
+    asset === 'stx' ? stx.value.quote.availableUnlockedBalance : btc.value.quote.availableBalance;
```
pete-watters, line 93: Changing this to use `availableUnlockedBalance` if the asset is STX. Before when we were taking it from the `balance` object using a lookup `[asset]` it was always showing the `quote.availableBalance`. 

I'm not sure if this is the best way to do it as I am making some assertions for `btc.state` / `stx.state`. This may not scale well but I am unsure if we will ever support **Send*** for other currencies so I think it's sufficient here

alexp3y (approved):

PR #1345, opened 2025-06-06
fix: update translations

edgarkhanzadian:


apps/mobile/src/locales/en/messages.po
```diff
@@ -28,7 +28,7 @@ msgstr "{currency}"
 
 #: src/features/approver/components/fees/bitcoin-fee-option.tsx:40
 msgid "fees-sheet.fee-rate-caption"
-msgstr "{feeRate} sats/vB · {quoteFee}"
+msgstr "{feeRate} sats/vB · {fiatFee}"
```
pete-watters, line 31: I refactored most instances of ‘fiat’ to be ‘quite’ so I think we could need to change this to ‘quoteFee’?

pete-watters (approved):
LGTM I just added one Q about ‘fiatFee’

PR #1349, opened 2025-06-06
fix: every commit dev builds

edgarkhanzadian:
Build the app only when the leather bot authored the commit (release PRs)

### DIFF
--- a/.github/workflows/mobile-preview-builds-dev.yml
+++ b/.github/workflows/mobile-preview-builds-dev.yml
@@ -11,6 +11,7 @@ permissions:
 jobs:
   deploy-eas-ios-dev:
     runs-on: ubuntu-latest
+    if: github.event.head_commit.author.email == '163018147+leather-bot@users.noreply.github.com'
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/mobile-preview-builds-dev.yml
```diff
@@ -11,6 +11,7 @@ permissions:
 jobs:
   deploy-eas-ios-dev:
     runs-on: ubuntu-latest
+    if: github.event.head_commit.author.email == '163018147+leather-bot@users.noreply.github.com'
```
pete-watters, line 14: Does the `163018147` mean something? Is it just auto - generated?

pete-watters (approved):
Thanks for updating this 👍  

I noticed that during your testing, all of the Linear issues linked in the commit lists keep re-opening. 

@camerow maybe we should turn off auto moving of issues from Linear? It's handy they update automatically when PRs get merged but then for these release PRs it re-opens lots of old tickets. I guess we can see if they get closed again after merging this

PR #1353, opened 2025-06-09
fix(mobile): update transaction time estimates to match extension, re…

pete-watters:
This PR updates our transaction time estimates to match the new times added in [this PR](https://github.com/leather-io/extension/pull/6287/files#diff-e9b0e3306c22ba600261c464098eef7e157551debead72d6b999d5424c01c3f8)

We could share these better however on `mobile` we are translating them so that makes it tricky to do that.


apps/mobile/src/locales/en/messages.po
```diff
@@ -99,34 +99,34 @@ msgstr "{title}"
...
-#: src/features/approver/utils.tsx:68
-msgid "approver.bitcoin.fee.speed.low"
-msgstr "~40 mins"
-
-#: src/features/approver/utils.tsx:96
+#: src/features/approver/utils.tsx:97
 msgid "approver.stacks.fee.speed.low"
 msgstr "~40 mins"
 
+#: src/features/approver/utils.tsx:105
```
pete-watters, line 118: `lingui` is compiling but not taking the new values

PR #1365, opened 2025-06-10
fix(web): responsive titles

kyranjamie:
Closes LEA-2733


apps/web/app/features/page/page.tsx
```diff
@@ -24,18 +24,20 @@ interface PageHeadingProps {
...
           {title}
         </Page.Title>
       </Box>
-      <Box width="50%">
-        <Page.Subtitle mt={['space.03', 'space.03', 0]} style={{ whiteSpace: 'pre-line' }}>
-          {subtitle}
-        </Page.Subtitle>
+      <Box flex={1}>
+        {subtitle && (
+          <Page.Subtitle mt={['space.03', 'space.03', 0]} whiteSpace="pre-line">
```
pete-watters, line 37: Not related to your work but in the past we moved the `extension` away from using array syntax for breakpoints towards more specific named ones like: ` mt={{ base: 'space.04', sm: 'space.07' }}`. 

Although more verbose I think that can be more easy to read and decipher than array sytax

pete-watters (approved):

apps/web/app/features/page/page.tsx
kyranjamie: I love this syntax, it's super simple to understand and read imo!?

PR #1371, opened 2025-06-11
refactor: drop info suffix from crypto asset model, closes LEA-2734

alexp3y:
Renames the asset type hierarchy (base type + all sub-types) from `CryptoAssetInfo` -> `CryptoAsset`

Removes use of "info" in param / variable names.


apps/mobile/src/features/activity/utils/make-activity-link.ts
```diff
@@ -2,7 +2,7 @@
 import { getStacksNetworkMode } from '@/queries/leather-query-provider';
 
 import { HIRO_EXPLORER_URL } from '@leather.io/constants';
-import { CryptoAssetInfo, NetworkConfiguration } from '@leather.io/models';
+import { CryptoAsset, NetworkConfiguration } from '@leather.io/models';
```
pete-watters, line 5: I see your `CryptoAssetInfo` -> `CryptoAsset` and I raise you to -> `Asset` 🤯

pete-watters (approved):
Nice work @alexp3y 👍  Maybe `CryptoAsset`'s are just `Asset`'s? 🤔

kyranjamie (approved):

alexp3y:
> Nice work @alexp3y 👍 Maybe `CryptoAsset`'s are just `Asset`'s? 🤔

@pete-watters  "Assets" alone might be a bit too generic. I think for base or unified types that have no other context, `CryptoAsset` anchors the domain and still provides some good semantic value / helps clarify "crypto assets" vs "web assets" vs "image assets" etc...

But in our specific type implementations "crypto asset" _**is**_ very redundant and adds a lot of noise. E.g `InscriptionCryptoAsset` or `StxCryptoAssetBalance`. These we already _**know**_ are "crypto assets" and could easily drop to `InscriptionAsset` or `StxBalance`.

In other words, the general convention could be that in the absence of a specific asset / balance implementation, "crypto asset" is the generic stand-in name (used for base types + unified types)... otherwise, we can safely drop to the specific asset type name.

In all other places (directories + filenames) we can also just refer to the simplified top-level entities "balances" and "assets", which we already do a pretty good job of (e.g. `services/assets`, `queries/balances`, `utils/assets`, etc..).

PR #1374, opened 2025-06-11
chore: check fingerprint

edgarkhanzadian:
Compare fingerprints of PRs to check if merging the PR would change the app's fingerprint (and hence require us to submit a new build to store)

### DIFF
--- a/.github/workflows/check-fingerprint.yml
+++ b/.github/workflows/check-fingerprint.yml
@@ -1,0 +1,91 @@
+name: Check Mobile Fingerprint
+on: pull_request_target
+permissions:
+  contents: read
+  pull-requests: write
+jobs:
+  check-mobile-fingerprint:
+    name: Check Mobile Fingerprint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Prepare the app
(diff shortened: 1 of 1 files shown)


.github/workflows/check-fingerprint.yml
```diff
@@ -0,0 +1,72 @@
...
+
+            // Find existing comment by the bot
+            const botComment = comments.find(comment =>
+              comment.user.type === 'Bot' &&
+              comment.body.includes('<!-- LEATHER_FINGERPRINT_TESTER -->')
+            );
+            const basicConfig = true ? {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: tag + "This PR has a different fingerprint. Merging this PR would render"
```
camerow, line 56: > "This PR has a different fingerprint. Merging this PR would render"

Incomplete sentence, probably you just want the inverse of the following one?

.github/workflows/check-fingerprint.yml
```diff
@@ -0,0 +1,72 @@
+name: Check Fingerprint
```
kyranjamie, line 1: Can we name this something so that it's clear it's for mobile?

pete-watters (approved):

PR #1384, opened 2025-06-12
fix(mobile): activity list performance LEA-2660

pete-watters:
This PR:
- refactors the `Error` component to not show 
- adds some fixes and improvements to the activity list. 

**NOTE** it seems a bit sluggish going 'Back' if you load a lot of activity. I will keep checking to see if there is a faster way to un-mount it

**DEMO**

https://github.com/user-attachments/assets/469d6e16-a865-4fdc-b054-03fbb6f2be70

I spent a lot of time on this trialling different approaches:
- upgrading `FlashList` - [PR here](https://github.com/leather-io/mono/pull/1362)
- swapping to `FlatList`
- tweaking parameters passed to `FlashList`
- simplifying the `renderItem` we pass in - [PR here](https://github.com/leather-io/mono/pull/1315/files)

In the end it seems that the bug with content disappearing was more related to the activity list not being inside a `ScrollView`. I had refactored this previously based on warning messages I saw about not nesting it inside a scrollview. 

After figuring that I spent time adjusting the settings of `FlashList` to optimise it for activity list:
- setting the `initialRenderLimit` to `10` makes it load quickly at first
- setting `onLoad` to increase render limit by `20` then quickly adds more data before scrolling starts
- decreasing the `onEndReachedThreshold` to `0.01` helps us load more data sooner
- `onEndReached` then adds `20` more items at a time 
- I am explicitly setting `estimatedItemSize` as it's recommended for performance
- I tried setting `estimatedListSize` but it made the width jumpy

I tried a few different variants of this
**Loading more data initially made it very laggy**

https://github.com/user-attachments/assets/b1645c62-17a3-4a49-bb34-891ca73456ae

**Batch loading more on demand**

`50` introduced visual gaps when scrolling

https://github.com/user-attachments/assets/e1a55886-ad8c-48c0-a39e-8b8c9e7a8953


apps/mobile/src/components/page/page.layout.tsx
```diff
@@ -1,63 +1,32 @@
...
-  scrollable?: boolean;
-}
-export function PageLayout({ children, scrollable = true }: PageLayoutProps) {
   return (
-    <>
-      <Box flex={1} bg="ink.background-primary">
-        {scrollable ? <ScrollableContent>{children}</ScrollableContent> : children}
-      </Box>
-    </>
+    <Box flex={1} bg="ink.background-primary">
```
pete-watters, line 17: Previously, I made `PageLayout` so we could have a wrapper for places we needed to be able to drag and refresh to invalidate queries. 

I added a flag of `scrollable` to wrap it in `ScrollView` or not as during early stages to `FlashList` implementation it gave warnings about using it inside of a scroll area.

It turns out that not having it in a scroll area was the main cause of this bug with disappearing content - LEA-2660. 

@tigranpetrossian is [working on a refactor](https://trustmachines.slack.com/archives/C05LAP952E7/p1749800036550979) of this setup to instead use a `Screen` component. 

For now I've:
- added `FlashList` for `Activity` inside of `ScrollView`
- refactored out the hook for `useRefreshHandler`

apps/mobile/src/features/activity/activity-list.tsx
```diff
@@ -1,48 +1,130 @@
...
+  const { linkingRef } = useBrowser();
   const { refreshing, onRefresh } = useRefreshHandler();
-  const [renderLimit, setRenderLimit] = useState(10);
-  const filteredActivities = activity
-    .slice(0, renderLimit)
-    .filter(activity => activity.type in OnChainActivityTypes) as OnChainActivity[];
+  const initialRenderLimit = 10;
+  const [renderLimit, setRenderLimit] = useState(initialRenderLimit);
+
+  const activityItems = useMemo(
```
pete-watters, line 37: I added `useMemo` here for now but [I am looking into a better performance improvement](https://github.com/leather-io/mono/pull/1392) using React Compiler so will likely revert this.

I am using `formatActivityItem` to preformat the data as much as I can to try and simplify the `renderItem` component used by `FlashList` - (`ActivityListItem`). 

In their docs they recommend this being as simple a component as possible. I refactored it down as much as I could but the data we are showing there still needs some other subcomponents:
- `ActivityAvatarIcon`
- `Balance` 

We can look into further improving those but I think this performance improvement is good enough for now

pete-watters:
I'll close this off for now as @tigranpetrossian is working on a separate bigger PR that also resolves this issue. We an re-open if needed

PR #1393, opened 2025-06-17
fix: code signing on dev

edgarkhanzadian:
Only use expo updates on prod. Fixes this:

Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-06-16 at 16 53 20


pete-watters (approved):
Thanks @edgarkhanzadian 👏  I confirmed this is working after a clean install

PR #1394, opened 2025-06-17
fix: crowdin action

edgarkhanzadian:
This would open PR every time someone pushes a commit to dev branch. It would:
1. upload sources with default translations to crowdin
2. download new translations for existing sources from crowdin.

Currently it is not going to open a PR if someone only updates translations on crowdin (without pushing a commit to dev), but there is a possibility for it with crowdin github integration

So to ensure that we have latest translations during the release, we need to merge crowdin pr and then right after that merge the release pr

Example PR: https://github.com/leather-io/mono/pull/1397


pete-watters:
This change is OK with me as it improves the current process, but it still seems very complicated and I question how important multi language support is at all in the short term. 

Even with this change, we still have issues with basic things like changing a translation in our code. I mentioned that in this [other PR](https://github.com/leather-io/mono/pull/1353) where I am trying to update some translation text in the APP. 

You can see in this video that I am trying to update the text of an existing key but it is not correctly adding it to the `.po` file:

https://github.com/user-attachments/assets/f8a46087-dce3-4ec1-938c-a01027a5dafe

I checked our planning centre and I cannot see anything about having multiple translations as priority there. @camerow what do you think about simplifying this and:
- removing `CrowdIn` for now
- keeping `lingui` but just as a way to have the app translation ready without all of this complication

kyranjamie:
> This would open PR every time someone pushes a commit to dev branch

So for every PR merged, there's a new Crowdin one created?

PR #1399, opened 2025-06-19
fix(web): awaited tracking call, fixes #1391

### LINKED ISSUE #1391: User can't confirm and start pooling on app.leather.io
Steps to reproduce

1. Go to [app.leather.io](http://app.leather.io)
2. select any pool for example Fast Pool v2
3. enter an amount
4. check the first two boxes
5. then try to confirm and start pooling
6. See error: no pop-up to comes up to sign the transaction.

The confirm button is dark black, as soon as you click it it turns grey and nothing happens, there is no pop-up.

Image

This is what is displayed in the console

Image

kyranjamie:
@camerow I understand this was caused by the tracking added, I haven't dug into why it failed yet, but try/catching it solves the issue. We shouldn't ever `await analytics.track` anything, as there's no need for users to be slowed down while an analytic call is made, can be ran in parallel instead. It's a bit of a hack but `void analytics.track` does have the fire and forget behaviour we want

Ah yeah actually it's a hanging promise issue, not that it throws, so the code path never makes it past the analytics call to open the contract call.

### DIFF
--- a/apps/web/app/features/stacking/start-liquid-stacking/utils/utils-liquid-stacking-stx.ts
+++ b/apps/web/app/features/stacking/start-liquid-stacking/utils/utils-liquid-stacking-stx.ts
@@ -109,7 +109,7 @@ export function createDepositStxMutationOptions({ leather, network }: CreateHand
     mutationKey: ['deposit-stx', leather, network],
     mutationFn: async (values: LiquidStackingFormValues) => {
       const liquidStackStxOptions = getOptions(values, network);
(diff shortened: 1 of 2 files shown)


kyranjamie:
This needs to be merged to main after being reviewed. OOO tomorrow if someone else can take care of that 🤝

fbwoolf (approved):

fbwoolf:
@camerow do you know when mentioned above that it needs to be merged into main, is there another step to release web?

PR #1401, opened 2025-06-21
fix: use of flag needs width 100 percent

fbwoolf:
This is needs or the receive modal is broken in the extension...

Screenshot 2025-06-21 at 10 44 31 AM


kyranjamie (approved):

kyranjamie:
thanks @fbwoolf

PR #968, opened 2025-03-12
feat: token view in leather mobile, ref LEA-2017

pete-watters:
### Description

This PR:
 - adds the `tokens` view so you can view all tokens at wallet and account level
 - adds chain `indicator` avatars to tokens
 - adds an animated header to flip balance and text on scroll - `ReversibleHeader` 
 - refactors `TokensWidget` to be named `BalancesWidget`
     - possibly soon to he renamed to `Holdings`

### Demo

https://github.com/user-attachments/assets/d7017dd1-3750-4579-a567-fdc229f95025

### Testing
To test this PR:
- open the mobile APP and load a wallet
- create multiple accounts
- observe all tokens listed on the home view
- click into an account and observe only tokens for that account 

**note** I have added no E2E tests here yet as some infrastructure work is needed there

PR #1404, opened 2025-06-23
Standardize mobile layouts + related bug fixes, closes LEA-2461, LEA-2752, LEA-2660

tigranpetrossian:
This PR introduces a better system for managing screen-level layouts and adds several related outstanding features, as well as related bugfixes.

Touches many files replacing old layouts with new. **Best reviewed commit by commit.**

### Screen component

All screens are now built using a new multi-part Screen component, combinations of which cover all of our current and future use cases. This replaces several variations of headers and layouts we’ve accumulated over time.

#### Basic structure
```tsx
<Screen>
  <Screen.Header />
  <Screen.ScrollView />
</Screen>
```
The Screen component comes with built-in scroll tracking. It’s currently only used for animating header titles (shown below), but can be easily hooked into for controlling all kinds of scroll-aware behavior.

The component is also action bar aware, and will adjust the bottom inset automatically to accommodate when the action bar is shown.

**A detailed README for `Screen` is in progres–will add while this is being reviewed.**

### Headers
Different variations of headers were merged into a single `Screen.Header` component with following sensible defaults based on existing usage:
* Default vaiant is what used to be NakedHeader: showing a back button and a network badge 
* The back button will now automatically handle going to the previous route, with the ability to override as necessary.

### Removed [Stack.Screen definitions](https://github.com/leather-io/mono/blob/dev/apps/mobile/src/core/app-navigation-stack.tsx)

The sole purpose of this file was to disable native headers on most screens, and define a “naked header” in couple of places. This is no longer necessary.

### Lists

Most occurrences of FlashList are replaces with `Screen.List`. This also inadvertently fixes the Activity List rendering issues and simplifies the usage quite a bit. `Screen.List` is provides the same scroll-aware logic, and is internally based on FlatList. We might consider migrating to FlashList again in future, but given all the issues @PETE uncovered while working with it, the FlatList should be fine for now.

There’s still a single `FlashList` usage for tokens, replacing it is a refactor its own – way out of scope for this PR. I’ll log a separate issue for that.

#### Animated title and header blur 

https://github.com/user-attachments/assets/537756af-a43d-4cfb-814b-511e6caec047

#### Animated tokens view header

https://github.com/user-attachments/assets/cdb03d6b-2cc0-48d2-b341-2389e97d9ecc

#### Animated account header 

https://github.com/user-attachments/assets/88c3e6d6-1cd7-43e3-bf3f-64580e8c87a4

#### Correct refresh control position

https://github.com/user-attachments/assets/fecf1d09-90d0-4f87-bbb8-1f38835c2c06

#### Settings/Notifications routing issues

https://github.com/user-attachments/assets/a585715e-824d-4ad6-9a92-219d02ce7060

There's also many spacing/visual adjustments on different screens to bring implementations closer to designs, omitting previews for those–would be too time consuming to document.


apps/mobile/src/features/activity/activity-widget.tsx
```diff
@@ -1,30 +1,39 @@
...
   activity: FetchState<Activity[]>;
   onPressHeader?: () => void;
   title: string;
 }
 export function ActivityWidget({ activity, onPressHeader, title }: ActivityWidgetProps) {
   const theme = useTheme<Theme>();
-  // Don't render anything if loading or no activity
-  if (activity.state === 'loading' || !hasActivity(activity)) {
+
+  if (activity.state !== 'success' || !hasActivity(activity)) {
```
pete-watters, line 33: I don't think this will work for the empty case. It needs to be a success to know how many entries there are

apps/mobile/src/app/account/[accountId]/balances.tsx
```diff
@@ -23,26 +23,26 @@ export default function BalancesScreen() {
...
-        )
-      }
-      isHeaderReversible={true}
-    >
-      <AccountBalances mode="full" fingerprint={fingerprint} accountIndex={accountIndex} />
-    </AnimatedHeaderScreenLayout>
+    <Screen>
+      <Screen.Header />
+      <Screen.ScrollView refreshControl={<RefreshControl />}>
+        {/* TODO: This was previously called "ReverisbleHeader. The behavior wasn't clear. Clarify out and replicate. */}
```
Comment (pete-watters) [inline apps/mobile/src/app/account/[accountId]/balances.tsx:31]: This was me being unclear, apologies! 

Sometimes in the design the order of the header is supposed to get flipped on scroll. So:
1. Full screen - title on top, subtitle below
2. Scrolled down - subtitle on top, subtitle below 

So the `reversible` means that the position of those gets reversed in some cases but not others

apps/mobile/src/app/account/[accountId]/activity.tsx
```diff
@@ -1,25 +1,57 @@
...
+            ListHeaderComponent={
+              <Screen.Title>
+                {i18n._({
+                  id: 'activity.account.header_content_title',
+                  message: '{accountName} Activity',
+                  values: { accountName: accountName },
+                })}
+              </Screen.Title>
+            }
+            data={activity.value as OnChainActivity[]} // TODO: Unclear why was this cast. Needs clearing up.
```
Comment (pete-watters) [inline apps/mobile/src/app/account/[accountId]/activity.tsx:46]: Without casting this you get other type errors like this:
```
Type 'Activity' is not assignable to type 'OnChainActivity'.
  Type 'ConnectAppActivity' is not assignable to type 'OnChainActivity'.
    Type 'ConnectAppActivity' is missing the following properties from type 'SwapAssetsActivity': fromAsset, fromAmount, toAsset, toAmount, and 2 more.
```

Re-visiting this and I think it was a mistake on my part not always specifying this type of activity data as `OnChainActivity` / I struggled to properly narrow it in all cases. 

It would be great to clean this up as the activity list only shows `OnChainActivity`

apps/mobile/src/app/activity/index.tsx
```diff
@@ -1,24 +1,36 @@
...
-          activity.state === 'success' && <ActivityList activity={activity.value} />
-        ) : (
-          <Box width="100%" justifyContent="center" alignItems="center">
-            <ActivityEmpty />
-          </Box>
+        {activity.state === 'success' && (
+          <Screen.List
+            refreshControl={<RefreshControl />}
+            ListHeaderComponent={<Screen.Title>{pageTitle}</Screen.Title>}
+            data={activity.value as OnChainActivity[]} // TODO: Unclear why was this cast. Needs clearing up.
```
pete-watters, line 25: Sloppy implementation is the reason ✌️  [See thread here](https://github.com/leather-io/mono/pull/1404/commits/02bc117694fb87f84a80d03efd61e97f992da897#r2161804675)

pete-watters (commented):
Great work @tigranpetrossian 👏  the code LGTM and is a great improvement. 

I will do a full QA early tomorrow and approve fully then all going well

apps/mobile/src/features/receive/screens/asset-details.tsx
```diff
@@ -57,7 +56,6 @@ export function AssetDetails() {
             values: { subtitle: accountName },
           })}
           leftElement={<HeaderBackButton onPress={navigation.goBack} testID={TestId.backButton} />}
-          rightElement={<NetworkBadge />}
```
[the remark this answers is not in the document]
tigranpetrossian: No worries. The badge is now the [default right element](https://github.com/leather-io/mono/pull/1404/files#diff-44264c3e416359e3d1a2cdcc5b0a1a3168c883155bf0a0444cb870233b27d8d6), because it was the case with every header usage.

apps/mobile/src/app/activity/index.tsx
tigranpetrossian: Thanks for clarifying. Added those before I took time off. We can log and refactor.

apps/mobile/src/app/account/[accountId]/balances.tsx
Comment (tigranpetrossian) [reply apps/mobile/src/app/account/[accountId]/balances.tsx:31]: Ah cool, I couldn't reproduce that behavior. Is there a design for this anywhere I can look at? We can easily add this back.

apps/mobile/src/features/receive/screens/asset-details.tsx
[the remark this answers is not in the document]
pete-watters: OK great, I knew you would have handled it but thought it worth double checking

apps/mobile/src/app/activity/index.tsx
pete-watters: Yeah we can leave it as is for now and then fix it easily after

packages/ui/src/assets/icons/pulse-16-16.svg
camerow, line 1: @tigranpetrossian Wrong icon added? Looks empty.

apps/mobile/src/features/activity/activity-widget.tsx
tigranpetrossian: `hasActivity` was already checking for `success` so this should still work. This simply excludes the error state, since it was rendering a full page broken error boundary which I don't think was intended. 
On the long term I think those widgets need better visual handling of errors/loading/empty states. Logged an [issue](https://linear.app/leather-io/issue/LEA-2970/home-widgets-non-success-states).

pete-watters: OK great 👍

apps/mobile/src/app/account/[accountId]/balances.tsx
Comment (pete-watters) [reply apps/mobile/src/app/account/[accountId]/balances.tsx]: @tigranpetrossian I introduced the `ReversibleHeader` in [this PR](https://github.com/leather-io/mono/pull/968) when working on the Tokens / Balance list view [LEA-2017](https://linear.app/leather-io/issue/LEA-2017/create-tokens-view)

There is a design for it [here](https://www.figma.com/design/BHLQoweGQdc1aONIy6YEIx/%F0%9F%96%BC%EF%B8%8F-Token---Collectibles-widget?node-id=5-1231&t=sln9SnWsVUYHfPnk-1) 

The basic idea is to flip / reverse (insert more apt word) the title and balance on scroll down on some screens. 

Screenshot 2025-06-24 at 06 38 19

Looking at the code and it seems to be only applied on the `Balances` / `My Tokens` pages. It hasn't been working so I don't think there's an urgent need to include it in this PR. It could be worth double checking with design if it's still needed and where and doing it in a follow up.

Comment (pete-watters) [reply apps/mobile/src/app/account/[accountId]/balances.tsx]: Right now on this PR it's not adding anything on scroll (it wasn't before either) so we can add this later / now

https://github.com/user-attachments/assets/a1e3b6b6-ad5e-47b8-b272-9eedde4e6f1e

pete-watters (approved):
Great work here @tigranpetrossian 👏  

I've done a manual QA of the app and confirmed it's working well. You are correct in that these changes also solve [LEA-2660](https://linear.app/leather-io/issue/LEA-2660/activity-goes-blank-after-scrolling-up-and-down), perhaps as `FlashList` is inside a `ScrollView`? I found that was the cause of the disappearing content.

The only slight issue of note here is the animated header for `Balances` view / `Account activity` but they weren't working at the moment anyway.

- Balance view should update the header on scroll [as discussed here](https://github.com/leather-io/mono/pull/1404#discussion_r2161791550)
- Account activity should have the account name in the header on scroll ([Figma](https://www.figma.com/design/Nbq5UKIEHEGYWQDw0IPOxr/%E2%9C%88%EF%B8%8F-Project--Send-BTC-and-STX?node-id=2025-85544&m=dev))
- I'm not sure if there should be more info than just the icon in the `Account page` header

**Account page**
Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-06-24 at 06 56 22

**None** of these should block merge as they weren't working before anyway so happy to move ahead with this and do them later. 

Thanks for doing this, it makes all the screens a lot easier to deal with

apps/mobile/src/app/account/[accountId]/balances.tsx
Comment (tigranpetrossian) [reply apps/mobile/src/app/account/[accountId]/balances.tsx]: Thanks @pete-watters. This is trivial to add with the new layout. Let's get it in.

Comment (tigranpetrossian) [reply apps/mobile/src/app/account/[accountId]/balances.tsx]: ✅ Added a preview in the description. 

Uploading Simulator Screen Recording - iPhone 16 Pro - 2025-06-24 at 10.10.37.mp4…

tigranpetrossian:
Thanks for testing this @pete-watters, really appreciate it. Activity and Balances are now fixed, and the account icon is by design.<br>Re:Flashlist - most likely that was the reason, without coupling removed most of these issues just disappeared by themselves. I've opted into Flatlist for now because I saw a really fast implementation out there that made me change my mind. Ours still needs some revisions to get there, but it should be a followup in the next cooldown.<br><br>I'll add a README for Screen and merge this shortly.

pete-watters:
> Thanks for testing this @pete-watters, really appreciate it. Activity and Balances are now fixed, and the account icon is by design.Re:Flashlist - most likely that was the reason, without coupling removed most of these issues just disappeared by themselves. I've opted into Flatlist for now because I saw a really fast implementation out there that made me change my mind. Ours still needs some revisions to get there, but it should be a followup in the next cooldown.I'll add a README for Screen and merge this shortly.

OK great 🎉  I like `FlashList` and I think it is better, but I saw the same fast implementation with `FlatList` and as good an effect in our APP so happy to use either. I like that `FlatList` is already built in too and I'm not convinced by the `FlashList` / `Shopify` OS products in general

pete:
> We might consider migrating to FlashList again in future, but given all the issues @pete uncovered while working with it, the FlatList should be fine for now.

Thank you, but I cannot take credit for working on FlashList nor uncovering any issues.

pete-watters:
> > We might consider migrating to FlashList again in future, but given all the issues @pete uncovered while working with it, the FlatList should be fine for now.
> 
> Thank you, but I cannot take credit for working on FlashList nor uncovering any issues.

Welcome original @pete 👏  I think you must have guided us somehow although unknowingly

PR #1405, opened 2025-06-23
feat(web): complete zealy quest on wallet connect

alexp3y:
Integrates internal quests API to complete Zealy "connect earn" task when users connect to web app.

This was originally added only to the legacy `earn.leather.io`.


camerow (approved):
Looks good, Kyran has a comment about using an enum, no other feedback on my side.

PR #1410, opened 2025-06-25
chore(mobile): e2e tests

pete-watters:
This PR revisits the mobile e2e tests with **maestro**. Previous tests relied heavily on re-setting the wallet using the Developer console which [we have since removed](https://github.com/leather-io/mono/commit/587cd6a54321487011a34c736b00cbe1b1a3fec1).

I am deprecating all previous tests and will use the code to write a smaller subset of faster tests to help protect the main functionality of the app without slowing down development too much.

For now on this PR I just have one `smoke-test` enabled as I am trying to determine the best way of:
- running tests on all changes to `mono` that effect mobile
- how to effectively but safely clear the app state in tests in lieu of the developer console


apps/mobile/eas.json
```diff
@@ -20,9 +20,11 @@
     "maestro": {
       "extends": "base",
       "environment": "production",
-      "withoutCredentials": true,
```
pete-watters, line 23: I needed to remove `"withoutCredentials": true,` to get around this problem in the Expo CI

Screenshot 2025-06-25 at 12 04 46

apps/mobile/eas.json
```diff
@@ -20,9 +20,11 @@
...
       "channel": "maestro",
-      "distribution": "internal"
+      "distribution": "internal",
+      "ios": {
```
pete-watters, line 25: I need to add `simulator: true` here or else the simulator won't run in EAS 

Screenshot 2025-06-25 at 12 05 58

apps/mobile/.eas/workflows/android-maestro-test.yml
```diff
@@ -26,4 +26,4 @@ jobs:
     params:
       build_id: ${{ needs.build_android.outputs.build_id }}
       flow_path:
-        - maestro/flows/
+        - maestro/flows/smoke-tests.yaml
```
pete-watters, line 29: For now I have set this up to only run `smoke-tests.yaml` on the `needs:e2e` tests label. 

We need to change this so that it runs these tests on any updates to `mono`, similarly to how Kyran setup [playwright.yml](https://github.com/leather-io/mono/blob/dev/.github/workflows/playwright.yml)

@camerow, maybe you have an idea and can help here? 

In Kyrans playwright test workflow, it runs the web app playwright tests on any changes to:
- web app
- packages - excluding `ui/native` files

I tried to setup this EAS workflow in a similar way but couldn't get it to work, maybe as these workflows are inside mobile already? 

I tried to change it to be this:
```
on:
  pull_request:
    paths:
      - '../../../../apps/mobile/**'
      - '../../../../packages/**'
      - '!../../../../packages/ui/**/*.web.ts'
      - '!../../../../packages/ui/**/*.web.tsx'
```

Maybe instead I need a `.github/workflow` that runs on changes to these files and then triggers mobile E2E tests? 

These tests are only really effective if they can run on every PR to `mono` that makes changes that could effect `mobile`

apps/mobile/maestro/flows/create-wallet.yaml
```diff
@@ -0,0 +1,21 @@
+# Leather Mobile Wallet Smoke Tests
+appId: io.leather.mobilewallet
+name: Create Wallet
+---
+# Scenario: User can create a new wallet
+- launchApp
+
+# Given the user is on the home screen
```
pete-watters: This test and subsequent tests are a WIP as I am unsure of the best way to do things. 

Previously our tests relied on re-setting the state via the Developer Console. We have since deprecated that but all it was doing was clearing persisted storage `clearAllPersistedStorage(wallets.list.map(wallet => wallet.fingerprint))}`. 

**Maestro** supposedly lets you to run scripts via `runScript` which I tried out here with a simple `hello-world` script and also a more advanced re-creation of  `clearAllPersistedStorage` 

An alternative solution is to add a button to the UI based on context to do this. Without doing this we cannot clear the state between tests and reset wallets

apps/mobile/maestro/flows/smoke-tests.yaml
```diff
@@ -0,0 +1,14 @@
+# Leather Mobile Wallet Smoke Tests
```
pete-watters, line 1: This is a very basic smoke test but is effective in that it ensures the app has been built successfully and the user reaches the home screen. 

I trialled using `Gherkin` BDD style and after doing some research it seems it can be just as effective to use Gherkin comments to explain the Maestro steps

apps/mobile/src/app/index.tsx
```diff
@@ -39,6 +41,19 @@ export default function HomeScreen() {
           <Box flexDirection="row" alignItems="center" p="2" gap="2">
             <LeatherLogomarkIcon />
             <NetworkBadge />
+
+            {/* FOR E2E TESTS ONLY - allows us to clear the secure store */}
```
pete-watters: I tested this as an alternative to `Developer tools` for clearing the state in tests. 

I'm making some mistake when trying to use a new `.env` variable - `EAS_CI` as I can't get it to work. I added it in my `.env` as:
```
export EAS_CI=true
```

I can get the button to show if I use `EXPO_PUBLIC_ENVIRONMENT` but I can't use that for this purpose. 

Here's a demo of the button:

https://github.com/user-attachments/assets/dfac6bd9-85c6-43c0-a344-311ab80f7b25




I want to add a new env variable for `EAS_CI` so I can also use it for [this fix](https://linear.app/leather-io/issue/LEA-2716/remove-postinstall-from-mobile-build) for `postinstall`

Maybe you have some advice here @camerow  / @edgarkhanzadian ?

kyranjamie (approved):

camerow (approved):

pete-watters:
Thanks for the review @camerow & @kyranjamie 👍  

I have simplified this PR to only add `smoke-tests` that are still run on demand via the label `needs:e2e-tests` 

I have made progress setting up an EAS env variable but will work on that separately and add it in a follow up PR. (That will enable me to safely reset the state and add more test coverage)

I would like to run these tests on each PR to mono but [I'm not quite sure how to do that yet](https://github.com/leather-io/mono/pull/1410#discussion_r2166457277). I can discuss it as a topic in the engineering sync

PR #1416, opened 2025-06-30
feat: add remaining token details data to services, closes LEA-2696

alexp3y:
This PR adds a new service (`FungibleAssetInfoService`) that provides the additional data needed for the token details view.

Providing this data as independent service calls (`getAssetDescription()`, `getAssetPriceChange()`, `getAssetPriceHistory()`) instead of bundling onto the existing `FungibleCryptoAsset` model (and existing asset services) allows callers to fetch ancillary data on-demand instead of requiring all asset API calls be placed together when an asset is referenced.

### DIFF
--- a/packages/services/src/assets/fungible-asset-info.service.ts
+++ b/packages/services/src/assets/fungible-asset-info.service.ts
@@ -1,0 +1,123 @@
+import { inject, injectable } from 'inversify';
+
+import { CryptoAssetProtocols, FungibleCryptoAsset } from '@leather.io/models';
+import { quoteCurrencyAmountToBase } from '@leather.io/utils';
+
+import {
+  LeatherApiClient,
+  LeatherApiTokenPriceHistory,
+} from '../infrastructure/api/leather/leather-api.client';
+import type { SettingsService } from '../infrastructure/settings/settings.service';
+import { Types } from '../inversify.types';
+import { MarketDataService } from '../market-data/market-data.service';
+import {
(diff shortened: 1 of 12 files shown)


packages/services/src/assets/fungible-asset-info.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    switch (asset.protocol) {
+      case CryptoAssetProtocols.nativeBtc:
+      case CryptoAssetProtocols.nativeStx:
+        return this.leatherApiClient.fetchNativeTokenHistory(asset.symbol, signal);
+      case CryptoAssetProtocols.sip10:
+        return this.leatherApiClient.fetchSip10TokenHistory(asset.contractId, signal);
+      case CryptoAssetProtocols.rune:
+        return this.leatherApiClient.fetchRuneHistory(asset.runeName, signal);
+      default:
+        throw Error('Price history not supported for asset type: ' + asset.protocol);
```
tigranpetrossian, line 103: We have [assertUnreachable](https://github.com/leather-io/mono/blob/d79c87b533e52ad559e75e50ee5a384a8312c121/packages/utils/src/index.ts#L208) that can enforce exhaustiveness too.

tigranpetrossian (approved):
Can't wait to have price changes displayed in the app 🔥

pete-watters (approved):
Nice work! 🚀

PR #1418, opened 2025-06-30
chore: simplify translations LEA-2758

pete-watters:
In this PR I simplify how we translate text in the mobile APP by removing the link to `CrowdIn` and simply managing text using only `lingui`. This means we have the application setup for internationalisation without the headache of dealing with the previous `CrowdIn` integration. 

I found that there was a disconnect between the latest text on CrowdIn and in the application so I have also taken the time to fix any discrepancies there. 

I have posted a [more detailed explanation of the changes in Linear](https://linear.app/leather-io/issue/LEA-2758/assess-and-improve-crowdin-lingui-setup#comment-dd7dea0f)

The result of this PR is that:
- local application text is up to date with what is desired in production
- `lingui` builds a `.po` file that is checked into the repo
- developers must manually run 2 `lingui` commands when adding/updating text

The steps I took in this PR are:
- gets rid of **crowdin** integration
- changes the `lingui` setup to force override of `.po` files
- adds `eslint` enforcement of translations - recovers lost overrides by extracting and compiling in CI
- manually updates the APP src text to match CrowdIn as of `2025-06-30 14:12`


.github/workflows/lingui-checks.yml
```diff
@@ -0,0 +1,20 @@
+name: Lingui Checks
```
pete-watters, line 1: I added this workflow at the root level of the `mono` repo (copying how Kyran setup `playwright.yml`).

We could probably put it in an EAS workflow inside of `apps/mobile` but I am unsure if thats better. Here I check `lingui` compiles and has no missing translations when files on the path  `'apps/mobile/**'` change

I have another PR open for E2E tests [here](https://github.com/leather-io/mono/pull/1410#discussion_r2166457277) with an open Q about `.eas/workflows`. They seem great but in that case I am not sure how to trigger the EAS workflow at root `mono` level so that **mobile** E2E tests will run on any change to its mono repo dependancy packages

apps/mobile/src/locales/index.ts
```diff
@@ -1,86 +1,14 @@
-// import { getLocales } from 'react-native-localize';
-// import { isProduction } from '@/shared/environment';
-// import OtaClient from '@crowdin/ota-client';
 import { i18n } from '@lingui/core';
 
-// import { formatter } from '@lingui/format-po';
```
pete-watters, line 6: @camerow If we are paying a fee for `CrowdIn` we should probably cancel the plan for the next while until we use it

apps/mobile/src/locales/en/messages.po
```diff
@@ -7,42 +7,45 @@ msgstr ""
...
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
-"X-Crowdin-Project: leather-mobile\n"
-"X-Crowdin-Project-ID: 701559\n"
-"X-Crowdin-Language: en\n"
-"X-Crowdin-File: /src/locales/en/messages.po\n"
-"X-Crowdin-File-ID: 10\n"
 "Project-Id-Version: leather-mobile\n"
 "Language-Team: English\n"
-"PO-Revision-Date: 2025-06-06 10:34\n"
+"PO-Revision-Date: 2025-06-30 14:12\n"
```
pete-watters, line 12: Working my way through this file I found we have a lot of duplicated text with different IDs. This bloats the amount of translations we have. 

I added the up to date text from CrowdIn by:
- downloading the `.po` file 
- diffing against what we generated from our code
- working my way through discrepancies

eslint.config.js
```diff
@@ -67,6 +67,24 @@ export default tseslint.config(
     name: 'mobile',
     files: ['apps/mobile/src/**/*.{ts,tsx}'],
     extends: [reactConfig, pluginLingui.configs['flat/recommended']],
+    rules: {
```
pete-watters, line 70: I am re-adding this check to flag non-translated strings in the mobile app to enforce adding text via `lingui` to keep the app translation ready

apps/mobile/README.md
```diff
@@ -57,6 +57,46 @@ If you encounter difficulties with `pnpm run ios` you can instead run:
...
+
+`lingui-checks.yml` is used to verify that any new / updated translations have been added to the `messages.po` file. The job is triggered by pull requests to `apps/mobile`.
+
+The workflow runs:
+
+- `pnpm lingui:extract` to generate `messages.po` based on translation
+- `pnpm lingui:compile` to test compilation errors (e.g. same id with different text)
+- `git diff` to see if `messages.po` has been updated by the extract
+
+#### CrowdIn workflow
```
pete-watters: I added this section about how we could use `CrowdIn` in the future but maybe I should delete it. 

If we decide to use CrowdIn in the future I don't think we should auto sync it. We would get the same benefit from simply:
- uploading this `.po` file
- having translators translate it into whatever languages inside of `CrowdIn` 
- download those `.po` files and add them to the codebase

A drawback of doing things this new way is that:
- text changes require developers to make a code change
- developers need to run two commands to generate translations for new text

This is not perfect but with the previous workflow developers took no responsibility for adding translations. They could:
- add new ids and text to the code
- they auto synced to CrowdIn but with empty message strings
- that mean the UI showed id's instead of text if anything

apps/mobile/src/locales/index.ts
camerow: Suspended our account there. We should still have the functionality until Mid August.

apps/mobile/README.md
camerow: We can probably forget about Crowdin for now. @tigranpetrossian has some ideas that he'll be drafting up for how we can better approach all this in the future. I think this is a nice simplification for now. Great work Pete!

camerow (approved):

apps/mobile/README.md
pete-watters: OK great 👏  I will remove this from the `README` and get it merged. I have the same info in Linear so we have a record of it if needed

apps/mobile/src/locales/index.ts
pete-watters: OK great 👍  I have synced all our english text to the app now already. 

I also took backups of our translations for other languages in case we need them / it can help us

Screenshot 2025-07-01 at 16 00 59

PR #1421, opened 2025-07-01
fix(mobile): update transaction time estimates to match extension

pete-watters:
This PR makes a simple text change to the mobile app. 

I previously tried to do this in [this PR](https://github.com/leather-io/mono/pull/1353) but I was unable to as I would need to:
- update the text in crowd in
- publish an OTA update
- have crowdin open a PR to the repo
- approve that PR and merge
- deal with a `dev` PR with that change

Now we have [gotten rid of CrowdIn](https://github.com/leather-io/mono/pull/1418) I can instead:
- make this change in the codebase
- run `pnpm lingui:extract` to add it to the `.po` file
- run `pnpm lingui:compile` to generate the `.js` file needed


camerow (approved):

camerow:
So curious why this triggers a fingerprint mismatch - this is like a textbook PR that should be able to use EAS Update right? cc @edgarkhanzadian

PR #1424, opened 2025-07-02
fix(web): add bg color to pages

fabric-8:


kyranjamie:
if this is a global change, don't we want to add it directly inside Page?

fabric-8:
@kyranjamie I was a bit hesitant / scared, but yeah this might make the most sense 👍

kyranjamie (approved):

PR #1427, opened 2025-07-03
chore(mobile): make sure any errors thrown get reported to Sentry

pete-watters:
During investigation of issues with Granite, I was wondering why none of the crashes seen there were reported to Sentry. 

I realised that this is as for React Native, Sentry doesn't auto catch errors thrown via `throw new Error(`. 

Our app throws a lot of these errors (something we can consider improving perhaps) but we don't have any stats on them. 

I investigated using [react-native-exception-handler](`https://www.npmjs.com/package/react-native-exception-handler`) in [this PR](https://github.com/leather-io/mono/pull/1426) but found it un-reliable. 

This PR makes a slight change to use `ErrorUtils.setGlobalHandler` to make sure all of our thrown errors get reported

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -60,6 +60,9 @@ initAppServices();
 void SplashScreen.preventAutoHideAsync();
 void initiateI18n();
 void setupFeatureFlags();
+ErrorUtils.setGlobalHandler(error => {
+  Sentry.captureException(error);
+});
 
 function App() {
   useWatchNotificationAddresses();


tigranpetrossian (commented):
> Our app throws a lot of these errors (something we can consider improving perhaps) 

Agreed. All exceptions happening directly or indirectly in the render cycle should be removed.

Few notes and questions on this:

1. ErrorUtils doesn't seem to be documented, this might be an internal implementation detail with no guarantees that it won't be changed or removed in the future.
2. Exceptions are either being propagated to the error boundary which does report, or being caught somewhere manually, in which case that would be the place to trace. What are the errors in question that weren't being reported?

pete-watters:
> > Our app throws a lot of these errors (something we can consider improving perhaps)
> 
> Agreed. All exceptions happening directly or indirectly in the render cycle should be removed.
> 
> Few notes and questions on this:
> 
> 1. ErrorUtils doesn't seem to be documented, this might be an internal implementation detail with no guarantees that it won't be changed or removed in the future.
> 2. Exceptions are either being propagated to the error boundary which does report, or being caught somewhere manually, in which case that would be the place to trace. What are the errors in question that weren't being reported?

Thanks @tigranpetrossian 👍  Removing the cases of `throw new Error` is a non trivial task as we have 60 of them in the mobile app. A few related to `useContext` but most handle business logic

1. `ErrorUtils` is a global error handler in React Native, similar to `window.onerror` in Web applications. There are no guarantees it won't change in future but for now we aren't catching these errors in Sentry and this avoids having to add a `try { } catch{}` everywhere we have these instances of `throw new Error` 
2. What led me down this path was an application crash in Granite with the error `'Nonce request failed'`. You can see this ticket - [LEA-2705](https://linear.app/leather-io/issue/LEA-2705/leather-mobile-issues-with-granite) That comes from our code in `<NonceLoader`:
```
export function NonceLoader({
...
  if (isError || !nonceResponse?.nonce) {
    // TODO: track this
    throw new Error('Nonce request failed');
  }

  return children(nonceResponse?.nonce);
}
```

This error is not caught by the error boundary and thus we don't log it in sentry. An option we have here is to add a try catch:
```
try { 
  if (isError || !nonceResponse?.nonce) {
    // TODO: track this
    throw new Error('Nonce request failed');
  }

  return children(nonceResponse?.nonce);
} } catch(e) {
   Sentry.captureException(error); 
}
```

Refactoring away these errors is something we need to do but there are so many of them that I think it will take some time and much thought. 

Adding this catching and reporting of errors in the meantime will at least give us a better idea of the performance of the APP in the real world as none of these are being reported. 

I got this issue myself with the in-app browser when trying to use Granite:
IMG_5044


It comes from:
```
export function assertAppIsConnected(app: App): asserts app is ConnectedApp {
  if (app.status !== 'connected') throw new Error('App is not connected');
}
```

tigranpetrossian:
Sure, I wasn't suggesting to remove those within this PR, although I wish these were handled differently.   Regardless, let's get to the bottom of this. That screenshot is the custom expo error boundary that should never show up in production, can we understand why isn't that error propagating to our error boundary, because it's clearly not being caught anywhere?

pete-watters:
> Sure, I wasn't suggesting to remove those within this PR, although I wish these were handled differently. Regardless, let's get to the bottom of this. That screenshot is the custom expo error boundary that should never show up in production, can we understand why isn't that error propagating to our error boundary, because it's clearly not being caught anywhere?

OK. I can dig into it a bit more. According to cursor the error is not caught as:
```
The error isn't caught by the error boundary because React's error boundaries only catch errors that occur during rendering, lifecycle methods, and constructors. When you throw an error directly in a component function like this, it's not considered a "render error" - it's just a regular JavaScript error that happens to occur during the render phase.
```

I think it could be happening as the connected APP sends an RPC call to initiate the `Approver` which fails and then throws this un-recoverable error. 

I wanted to add tracking of these errors until we can fix it so we understand how many of these problems are happening without us realising

tigranpetrossian:
Sorry I missed this. Those errors in particular are manually thrown in the render cycle. It might be that those aren't propagated to the top boundary because sheets are rendered in portals.<br>Let's merge this as is and deal with it afterwards, aside from logging to Sentry, these will need to be handled properly, at the very least by wrapping the sheets in their own boundaries, or better yet, designing and implementing those failure cases.

tigranpetrossian (approved):

pete-watters:
> Sorry I missed this. Those errors in particular are manually thrown in the render cycle. It might be that those aren't propagated to the top boundary because sheets are rendered in portals.Let's merge this as is and deal with it afterwards, aside from logging to Sentry, these will need to be handled properly, at the very least by wrapping the sheets in their own boundaries, or better yet, designing and implementing those failure cases.

Ah ok, that makes sense now why they aren't caught by the boundary. Thanks for helping me understand. 

I will come back to this soon and add error boundaries around the sheets so we catch this better [LEA-3014](https://linear.app/leather-io/issue/LEA-3014/fix-issues-with-uncaught-errors).

PR #1434, opened 2025-07-09
fix: use text search [LEA-2731]

edgarkhanzadian:
Use google search if the string doesn't look like a URL. Later on we can try adding a feature for selecting a default search engine

https://github.com/user-attachments/assets/ebfdcd3d-f8f3-4173-9112-5671f3470208

### DIFF
--- a/apps/mobile/src/features/browser/browser/utils.ts
+++ b/apps/mobile/src/features/browser/browser/utils.ts
@@ -14,6 +14,8 @@ import {
   supportedMethods,
 } from '@leather.io/rpc';
 
+const bareUrlRegex = /^[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
+
 export type BrowserSheetTab = 'suggested' | 'connected' | 'recent';
 
 export const URL_SEARCH_HEIGHT = 70;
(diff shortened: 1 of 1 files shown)


apps/mobile/src/features/browser/browser/utils.ts
```diff
@@ -22,7 +24,12 @@ export function formatURL(url: string) {
   if (url.startsWith('http')) {
     return url;
   }
-  return 'https://' + url;
+  if (bareUrlRegex.test(url)) {
+    return 'https://' + url;
+  }
+  // This is going to be handled as search text in google.com
+  // TODO: add search selection in settings
```
tigranpetrossian: Let's reserve `TODOs` for indicating that the code in question needs improvement, and track features in Linear instead.

tigranpetrossian (approved):

apps/mobile/src/features/browser/browser/utils.ts
```diff
@@ -22,7 +24,12 @@ export function formatURL(url: string) {
...
+  return 'https://google.com/search?q=' + url;
```
pete-watters: Maybe we can use DuckDuckGo instead at first?  `https://duckduckgo.com/?q=`

Then we already take privacy seriously from the start. For URLs I think the search will be as consistent but without tracking users

pete-watters (approved):

apps/mobile/src/features/browser/browser/utils.ts
edgarkhanzadian: https://linear.app/leather-io/issue/LEA-3040/create-default-browser-search-selection

PR #1437, opened 2025-07-10
feat: adds locale to token description service call

alexp3y:
Adds support for token description translations, which are set to be enabled via the API. Defaults to `en` for now, even though several other languages are supported via the description endpoints. We'll need to do some future coordination around language preference and a supported `Locales` structure to ensure FE/BE alignment.

### DIFF
--- a/apps/mobile/src/queries/assets/fungible-asset-info.query.ts
+++ b/apps/mobile/src/queries/assets/fungible-asset-info.query.ts
@@ -4,11 +4,14 @@ import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
 import { FungibleCryptoAsset } from '@leather.io/models';
 import { getFungibleAssetInfoService } from '@leather.io/services';
 
+const DEFAULT_LOCALE = 'en';
+const DEFAULT_PRICE_HISTORY_PERIOD = '24h';
+
 export function useAssetDescriptionQuery(asset: FungibleCryptoAsset) {
   return useQuery({
     queryKey: ['fungible-asset-info-service-get-asset-description', asset],
     queryFn: ({ signal }: QueryFunctionContext) =>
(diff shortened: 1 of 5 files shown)


packages/services/src/infrastructure/api/leather/leather-api.types.ts
```diff
@@ -1430,7 +1430,9 @@ export interface paths {
     /** @description Get SIP-10 Token Description */
     get: {
       parameters: {
-        query?: never;
+        query?: {
+          locale?: 'en' | 'es' | 'ru' | 'zh';
```
pete-watters, line 1434: Maybe this could be a single shared type?

apps/mobile/src/queries/assets/fungible-asset-info.query.ts
```diff
@@ -8,7 +8,7 @@ export function useAssetDescriptionQuery(asset: FungibleCryptoAsset) {
   return useQuery({
     queryKey: ['fungible-asset-info-service-get-asset-description', asset],
     queryFn: ({ signal }: QueryFunctionContext) =>
-      getFungibleAssetInfoService().getAssetDescription(asset, signal),
+      getFungibleAssetInfoService().getAssetDescription(asset, 'en', signal),
```
pete-watters: We could use a `const` for `en` but its not much difference

pete-watters (approved):
Great work @alexp3y 👍  I added two suggestions but both are nits

packages/services/src/infrastructure/api/leather/leather-api.types.ts
alexp3y: The types in this file are generated from the backend OpenAPI spec, but in `leather-api.client.ts` I do export these together as a single type ->`LeatherApiLocale`, which is used in the service interface.

PR #1446, opened 2025-07-13
fix: stackingdao pool address

### LINKED ISSUE #1445: Stacking with StackingDao pool, passes wrong poolAddress argument.
>In the second step (where the user delegates the STX), you pass the wrong parameter for delegate-to (see screenshot). This should be just our address, and not the whole pool's address (so just SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG).

314159265359879:
should be the address principle not the contract.

I think that would fix what Philip reported, see https://github.com/leather-io/mono/issues/1445 or:
image

### DIFF
--- a/apps/web/app/data/data.ts
+++ b/apps/web/app/data/data.ts
@@ -214,9 +214,9 @@ export const stackingPoolData = {
       "Enter the STX address of the pool with which you'd like to Stack without your STX leaving your wallet.",
     duration: -1,
     poolAddress: {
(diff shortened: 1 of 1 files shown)


314159265359879:
Philips says this isn't fixing the problem yet because they require the allow-contract call to go to the complete pool (contract principle). And the delegate stx call should go to just the address. 

image

kyranjamie:
@314159265359879 there's now a deployed build in one of the comments. Can you QA this there?

314159265359879:
Retested step 1 contract principle argument ✅ 
Retested step 2 address principle argument ✅

kyranjamie (approved):

PR #1447, opened 2025-07-14
refactor: delete unused routes

edgarkhanzadian:
We have unused routes for adding mpc wallets and hardware wallets to the wishlist, this PR deletes those. Also deletes an unused swap screen.

Also delete unused waitlist buttons in add wallet sheet


tigranpetrossian (approved):

apps/mobile/src/app/mpc-wallets.tsx
```diff
@@ -1,148 +0,0 @@
...
-import { SettingsList } from '@/components/settings/settings-list';
-import { SettingsListItem } from '@/components/settings/settings-list-item';
-import {
-  NotifyUserSheetData,
-  NotifyUserSheetLayout,
-} from '@/components/sheets/notify-user-sheet.layout';
-import { WaitlistIds } from '@/features/waitlist/ids';
-import { t } from '@lingui/macro';
-
-import {
```
pete-watters, line 13: We should delete all these logos too as we're never going to use them

pete-watters:
Thanks @edgarkhanzadian . I had planned to do this in the next cooldown but also:
- remove all the logos
- get rid of the `releaseWaitlistFeatures` and get rid of all of those features

PR #1449, opened 2025-07-14
feat: add bottom tab nav

edgarkhanzadian:
This PR is setting up bottom navigation tabs and pretty much changing the navigation flow of the app.

We would have a root stack navigation for tabs / settings / wallet generation

Then the home screen would be a bottom tab navigation, that would contain a home screen / browser / activity tabs.

Browser and activity tabs are only going to be screens without navigation, while the home screen is going to be a stack navigator for account screens / balances / collectibles.

The base of this PR is going to be `feat/updated-navigation`. I propose we make more PRs into that brunch before merging `feat/updated-navigation` into dev (s.t. we don't merge half-baked features). Making this type of change with feature flags would be quite a headache

https://github.com/user-attachments/assets/2bea371d-3a57-4023-b589-09c848c62c24


package.json
```diff
@@ -152,7 +152,8 @@
           "@react-navigation/core": "*",
           "@expo/metro-runtime": "*",
           "expo-constants": "*",
-          "escape-string-regexp": "*"
+          "escape-string-regexp": "*",
+          "query-string": "7.1.3"
```
edgarkhanzadian, line 156: Had to manually override query-string to 7.1.3 in pnpm s.t. expo-router doesn't use the 9.x version that comes with bip21.

https://github.com/expo/expo/issues/35173

apps/mobile/src/app/(tabs)/activity.tsx
```diff
@@ -16,7 +15,7 @@ export default function ActivityScreen() {
 
   return (
     <Screen>
-      <Screen.Header centerElement={<Text variant="label01">{pageTitle}</Text>} />
+      <Screen.Header leftElement={null} />
```
pete-watters, line 18: So you can't go back here, only change tab back to `Home`? 

Maybe we can get rid of the `ActivityWidget` from the `Home` page then?

On the `Account` page accessing `Activity` will be a bit different than on `Home` as it will use the original way but that's OK

pete-watters (approved):
Nice work 👍  

Lets make sure we consider removing the `ActivityWidget` from `Home` if it's in a `Tab` anyway. 

We should also think about if a change is needed for `Account` - `Activity`

apps/mobile/src/app/(tabs)/activity.tsx
edgarkhanzadian: Great idea, removed!

PR #1442, opened 2025-07-11
chore(mobile): add node_modules/.pnpm to fingerprint.ignore

pete-watters:
This PR updates our EAS `.fingerprintignore` file so that it excludes `node_modules/.pnpm`. 

@camerow and I had a call with the Expo team and they helped us pinpoint this as a reason why we keep seeing new fingerprints.

PR #1450, opened 2025-07-14
fix(web): panda borders

pete-watters:
This PR reverts [this commit](https://github.com/leather-io/mono/commit/077f7a281ab8a97d2872fd9f9fd66287a698ede1#diff-0d0efc282d3c6cbfd65434f2af1c2058653036f28927d2c5aa67f8e3e0c77211). It is needed to enable us to [do this release ](https://github.com/leather-io/extension/pull/6316)

It was adding a new border to `panda-preset` that was over-writing the other borders. `panda-preset` gets its borders from the **tokens** package so new borders should instead be added there.

### DIFF
--- a/packages/panda-preset/src/tokens.ts
+++ b/packages/panda-preset/src/tokens.ts
@@ -2,7 +2,6 @@ import { defineTokens } from '@pandacss/dev';
 
 import { tokens as leatherTokens, zIndices } from '@leather.io/tokens';
 
-import { borders } from './borders';
 import { colors } from './colors';
 
 function transformZIndices(zIndices: Record<string, number>) {
@@ -13,5 +12,4 @@ export const tokens = defineTokens({
   ...leatherTokens,
   colors,
   zIndex: transformZIndices(zIndices),
-  borders,
 });
--- a/packages/tokens/src/tokens.ts
+++ b/packages/tokens/src/tokens.ts
@@ -13,6 +13,7 @@ export const tokens = {
     error: { value: '1px solid {colors.red.border}' },
     focus: { value: '2px solid {colors.ink.action-primary-default}' },
     invert: { value: '1px solid {colors.invert}' },
+    none: { value: 'none' },
     subdued: { value: '1px solid {colors.ink.text-subdued}' },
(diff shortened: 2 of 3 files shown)


packages/panda-preset/src/tokens.ts
```diff
@@ -2,7 +2,6 @@ import { defineTokens } from '@pandacss/dev';
 
 import { tokens as leatherTokens, zIndices } from '@leather.io/tokens';
```
pete-watters, line 3: `leatherTokens` here already includes the `borders` from `@leather.io/tokens`. 

Adding these new borders was then resetting them all to be only:
```
{
  none: { value: 'none' },
  default: { value: '1px solid {colors.ink.border-default}' }
```

This meant that all the other borders on the `extension` no longer worked.

packages/tokens/src/tokens.ts
```diff
@@ -13,6 +13,7 @@ export const tokens = {
     error: { value: '1px solid {colors.red.border}' },
     focus: { value: '2px solid {colors.ink.action-primary-default}' },
     invert: { value: '1px solid {colors.invert}' },
+    none: { value: 'none' },
```
pete-watters, line 16: I added this `non` border from `packages/panda-preset/src/borders.ts` here instead. 

This means it will be picked up in the web-app from here via `panda-preset`

pete-watters:
> This PR's build fingerprint differs from the latest production build's fingerprint. Merging this PR would require us to submit the build to the stores. (Note: Make sure your PR is up to date)

I am still seeing this fingerprint warning so [this](https://github.com/leather-io/mono/pull/1442) 'fix' didn't work. The fingerprint shouldn't be different for a JS change to token files

fbwoolf (approved):
Thx @pete-watters!

pete-watters:
Thanks @fbwoolf 👍  lets wait for @kyranjamie's test build to make sure Web is looking OK

pete-watters:
I confirmed this border change also keeps the Web App working properly: 

Screenshot 2025-07-14 at 15 54 02

PR #1458, opened 2025-07-15
fix: browser linking

edgarkhanzadian:
Fix linking in browser

https://github.com/user-attachments/assets/31b73dfa-83e9-49bb-b3c0-1f315e2f8839


apps/mobile/src/features/browser/browser/use-browser-search-state.ts
```diff
@@ -38,13 +38,15 @@ function reducer(state: SearchBarState, action: SearchBarAction): SearchBarState
 }
 export function useBrowserSearchState() {
   const [state, dispatch] = useReducer(reducer, initialState);
+  const goToUrl = useCallback((url: string) => {
```
pete-watters, line 41: I'm not sure if `useCallback` is necessary here if it's wrapping `dispatch`?

pete-watters (approved):

camerow:
This is looking so good @edgarkhanzadian and @mica000!

apps/mobile/src/features/browser/browser/use-browser-search-state.ts
edgarkhanzadian: yeap it is, as i use goToUrl in useEffect dependencies

PR #1461, opened 2025-07-16
feat: add dapp screen

edgarkhanzadian:
Connected and recent apps

https://github.com/user-attachments/assets/616482aa-cb8a-4081-b4e1-3b32d73733d3


apps/mobile/src/features/browser/browser/general-apps-screen.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Box, Text } from '@leather.io/ui/native';
+
+interface GeneralAppsScreenProps {
+  appStatus: AppStatus;
+  title: string;
+  noAppsTitle: string;
+  imageSrc: string;
+}
+
+export function GeneralAppsScreen({
```
tigranpetrossian: GeneralAppsScreen -> AppListing(Screen)?

apps/mobile/src/features/browser/browser/general-apps-screen.tsx
```diff
@@ -0,0 +1,63 @@
...
+  title,
+  noAppsTitle,
+  imageSrc,
+}: GeneralAppsScreenProps) {
+  const { list: appList } = useApps(appStatus);
+  const { openURL } = useOpenURL();
+
+  const noApps = (
+    <>
+      <Box position="absolute" style={{ width: '100%' }}>
```
tigranpetrossian: What's the behavior here? Asking in case it's a common enough use case for me to add as an option into Screen.Header.

apps/mobile/src/features/browser/browser/general-apps-screen.tsx
```diff
@@ -0,0 +1,63 @@
...
+export function GeneralAppsScreen({
+  appStatus,
+  title,
+  noAppsTitle,
+  imageSrc,
+}: GeneralAppsScreenProps) {
+  const { list: appList } = useApps(appStatus);
+  const { openURL } = useOpenURL();
+
+  const noApps = (
```
tigranpetrossian: `noApps` -> `emptyState`

tigranpetrossian (approved):
Thanks @edgarkhanzadian 
Left some naming nits and a question about the layout.

apps/mobile/src/features/browser/browser/general-apps-screen.tsx
edgarkhanzadian: When trying to center the screen body, it centers quite a bit lower because of the header. I'm not sure this is the best way to fix this, but didn't come up with a better solution.

Centered: 

centered

Not centered:

notCentered

apps/mobile/src/features/browser/browser/general-apps-screen.tsx
edgarkhanzadian: Done!

apps/mobile/src/features/browser/browser/general-apps-screen.tsx
edgarkhanzadian: Done!

apps/mobile/src/features/browser/browser/general-apps-screen.tsx
tigranpetrossian: Ah, I see. If we had `calc` in RN it would be as simple as `height: calc(100% - HEADER_HEIGHT)`, but what you have is fine for now. <br>I'll take a note to look into baking this into an empty state component when I get the time.

edgarkhanzadian: Thank you!

PR #1463, opened 2025-07-16
feat: send receive buttons

edgarkhanzadian:
This PR adds send/receive buttons:

https://github.com/user-attachments/assets/65a224c0-71fb-4cb1-b637-c71757c8037b

NOTE: For some reason triggering back button on send navigation flow also trigger back functionality on the main stack navigator:

https://github.com/user-attachments/assets/ec25e1fb-9f8e-477c-a1af-283b2c95458f


packages/ui/src/components/button/button-v2.native.tsx
```diff
@@ -0,0 +1,166 @@
...
+      ghost: 'ink.text-primary',
+    }
+  );
+}
+
+type ButtonSize = 'xs' | 'sm' | 'md';
+
+export const ButtonV2 = forwardRef(
+  (
+    {
```
edgarkhanzadian, line 77: As far as I know we are planning to update a Button to be rounded everywhere (+ plus adding some sizing props to it). So instead of making the change on all of the buttons right now i thought it might be better to create a duplicate ButtonV2 with some added functionality for now until we transition from the old button to the new one everywhere.

apps/mobile/src/features/account/account.tsx
```diff
@@ -88,6 +90,32 @@ export function Account({ account, walletName }: AccountProps) {
           accountName={name}
           walletName={walletName}
         />
+        <Box mt="3" mb="5" px="5" flexDirection="row" justifyContent="center" gap="2">
```
pete-watters, line 93: Can you put this into a shared component? It's almost the same as whats in `acounts-widget`

packages/ui/src/components/button/button-v2.native.tsx
pete-watters: That's OK with me but we will need to make the buttons the same before this goes to `dev` so we should probably just do that.

Are there many changes? I think it's just making them more round? I'd be happy to YOLO change that if it's a small enough change. We have 45 `<Button`s which is a lot but not a crazy amount. I guess most are the same

pete-watters (approved):
Nice work!

apps/mobile/src/features/account/account.tsx
edgarkhanzadian: I mean yea but these are conceptually quite different. E.g. if we decide to run analytics on where the user entered the flow, how should I update the onPress behavior for the buttons? I am kind of hesitant to do it right now but we can get to it later in the "update navigation" PR workflow

packages/ui/src/components/button/button-v2.native.tsx
edgarkhanzadian: I'd hope we could do it after merging the updated navigation PR as it's pretty big by itself

PR #1465, opened 2025-07-17
feat(mobile): implement token details, ref LEA-3015

pete-watters:
This PR adds the first draft of token details. It's close but some more work is needed regarding:
- BTC balances split by address type
- improved transitions between sheets, going back etc. 
- UI polish + final refactor to improve overall code 

I have feature flagged the whole thing so it could be merged safely or not. 

Here's a demo:

https://github.com/user-attachments/assets/4d4ae2a2-0482-412a-beb9-b19cbfd86c53


apps/mobile/src/app/account/[accountId]/activity.tsx
```diff
@@ -43,7 +42,7 @@ export default function AccountActivityScreen() {
                 })}
               </Screen.Title>
             }
-            data={activity.value as OnChainActivity[]} // TODO: Unclear why was this cast. Needs clearing up.
+            data={activity.value.filter(activity => activity && 'asset' in activity)}
```
Comment (pete-watters) [inline apps/mobile/src/app/account/[accountId]/activity.tsx]: Filtering for only `OnChainActivity` to avoid casting. I could probably do this in the Query itself

apps/mobile/src/features/balances/balances.tsx
```diff
@@ -1,42 +1,101 @@
...
 export function AllAccountBalances({ mode }: BalanceViewProps) {
+  const [sheetData, setSheetData] = useState<TokenSheetData | null>(null);
   const runesFlag = useRunesFlag();
+  const tokenDetailsFlag = useTokenDetailsFlag();
+
+  const tokenSheetRef = useRef<SheetRef>(null);
+
+  function onOpenToken({ asset, availableBalance, quoteBalance }: OnOpenTokenProps) {
+    setSheetData({ asset, availableBalance, quoteBalance });
+    // analytics.track('token_sheet_opened', { source: 'action_bar' });
```
pete-watters, line 37: Keeping these as a placeholder. I'll add the analytics keys soon

apps/mobile/src/features/balances/balances.tsx
```diff
@@ -1,42 +1,101 @@
...
+  const tokenSheetRef = useRef<SheetRef>(null);
+  const tokenDetailsFlag = useTokenDetailsFlag();
+
+  function onOpenToken(data: TokenSheetData) {
+    setSheetData(data);
+    // analytics.track('token_sheet_opened', { source: 'action_bar' });
+    tokenSheetRef.current?.present();
+  }
+
+  const onPressToken = tokenDetailsFlag ? onOpenToken : undefined;
```
pete-watters, line 69: I have enabled `tokenDetailsFlag` in `dev` but off in production in launch darkly. 

There is no way to trigger the `TokenDetails` sheet without entering here first

apps/mobile/src/features/balances/bitcoin/bitcoin-balance.tsx
```diff
@@ -37,29 +39,30 @@ export function BitcoinTokenBalance({
...
+  onPress?: ({ asset, availableBalance, quoteBalance }: OnOpenTokenProps) => void;
 }
 
 export function BitcoinBalance({ onPress }: BitcoinBalanceProps) {
   const { state, value } = useBtcTotalBalance();
-
   const availableBalance = value?.btc.availableBalance;
   const quoteBalance = value?.quote.availableBalance;
-
+  if (!availableBalance || !quoteBalance) {
```
pete-watters, line 49: I need to clean this up

apps/mobile/src/features/balances/stacks/sip10-balance.tsx
```diff
@@ -83,6 +87,14 @@ function Sip10BalanceWrapper({ data, mode = 'full' }: Sip10BalanceWrapperProps)
                 contractId={item.asset.contractId}
                 quoteBalance={item.quote.totalBalance}
                 imageCanonicalUri={item.asset.imageCanonicalUri}
+                onPress={() => {
+                  // pass balance and quote balance to the sheet from here
```
pete-watters, line 91: I am passing the balances to the sheet via a callback as we already have that information at this point. 

Without doing this we would need to call the balance hooks again in the sheet and then have 6 different setups for each scenario (excluding Runes) - `BTC`, `STX` and `SIP-10` both all balances and account balance.

apps/mobile/src/features/token/account-list-item.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const availableBalance = tokenBalance?.availableBalance;
+  const quoteBalance = tokenBalance?.quoteBalance;
+
+  return (
+    <AccountListItemComponent
+      px="0"
+      accountName={account.name}
+      walletName={
+        <Text variant="caption01" lineHeight={16}>
+          {/* Should perhaps refactor account to have a wallet name?  Avoids using the wallet store here */}
```
pete-watters, line 62: We need to use `WalletLoader` to get the `wallet` name. We could probably add that to the account object

apps/mobile/src/features/token/account-list-item.tsx
```diff
@@ -0,0 +1,85 @@
...
+          {wallet.name}
+        </Text>
+      }
+      // FIXME LEA-3015: refactor address prop to not be address - leftPrimary or something like  that
+      address={
```
pete-watters, line 67: I'm re-using `AccountListItemComponent` but it needs to be refactored to be more generic

apps/mobile/src/features/token/account-list-item.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Balance
+          balance={quoteBalance}
+          variant="caption01"
+          color="ink.text-subdued"
+          isQuoteCurrency={tokenId === 'BTC'}
+        />
+      }
+      balance={
+        // FIXME LEA-3015: isQuoteCurrency is crashing for non BTC tokens
```
pete-watters, line 76: I need to fix this to handle SIP-10 balances in this context 
Screenshot 2025-07-17 at 19 23 07

apps/mobile/src/features/token/address-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const account = useAccountByIndex(fingerprint, accountIndex);
+  if (!account) {
+    return null;
+  }
+
+  return (
+    <TokenDetailsCard
+      title={
+        <Box flexDirection="row" alignItems="center" gap="2">
+          {/* FIXME - this icon needs to be 16x16 but require refactors to make it work */}
```
pete-watters, line 40: I need to refactor the icons so they can accept a smaller size of `16x16` to fit this design 

Screenshot 2025-07-17 at 19 22 59

apps/mobile/src/features/token/hooks/use-get-token-balance.ts
```diff
@@ -0,0 +1,77 @@
+import { AccountBalance, useAccountBalance } from '@/queries/balance/account-balance.query';
```
pete-watters, line 1: I'll be removing this file completely. I replaced this with passing balances as props as I think that's cleaner than these custom hooks

apps/mobile/src/features/token/address-list.tsx
```diff
@@ -0,0 +1,192 @@
...
+    nativeSegwitPayerAddress,
+    taprootPayerAddress,
+    stxAddress: stxAddress ?? '',
+  });
+
+  const tokenBalance = useGetAccountTokenBalance({ tokenId, account });
+  const availableBalance = tokenBalance?.availableBalance;
+  const quoteBalance = tokenBalance?.quoteBalance;
+
+  if (tokenId === 'BTC') {
```
pete-watters, line 71: This balance section of this code will be re-written once we have balances split by BTC address type. 

For now it's using `useGetAccountTokenBalance` which will be removed and has both BTC balance listed as the same aggregate balance:

Screenshot 2025-07-17 at 19 31 10

apps/mobile/src/features/token/components/token-activity.tsx
```diff
@@ -0,0 +1,79 @@
...
+}: TokenActivityProps) {
+  const scrollViewAdjustmentOffset = 56;
+  const { refreshing, onRefresh } = useRefreshHandler();
+  const activity = useTotalActivity();
+
+  return (
+    <Screen>
+      <Screen.Header
+        blurOverlay={false}
+        // topElement={isErrorTotalBalance && <FetchErrorCallout />}
```
pete-watters: I need to add error handling in case of incorrect balances / timeout etc.

apps/mobile/src/features/token/components/token-activity.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <FetchWrapper data={activity}>
+        {activity.state === 'success' && (
+          <Screen.List
+            refreshControl={<RefreshControl progressViewOffset={scrollViewAdjustmentOffset} />}
+            style={{ marginTop: -scrollViewAdjustmentOffset }}
+            data={activity.value
+              .filter(activity => activity && 'asset' in activity)
+              .filter(activity => filterByTicker(activity, ticker))
+              .filter(activity => {
+                if (isDefined(accountIndex) && isDefined(fingerprint)) {
```
pete-watters, line 61: When an account is selected, this re-renders and filters activity using that account ID. 

We have another query for account specific activity by asset but this data is already here

apps/mobile/src/features/token/components/token-details.tsx
```diff
@@ -0,0 +1,74 @@
...
+            </Text>
+          </Box>
+        }
+        quoteBalance={<Balance balance={quoteBalance} variant="label01" isQuoteCurrency />}
+      />
+      {assetDescription && <TokenDescription>{assetDescription}</TokenDescription>}
+
+      <TokenDetailsTable
+        name={`${capitalize(asset.chain)} (${asset.symbol})`}
+        // for Layer design has 'Layer 1 (Bitcoin)' but no other examples. Only showing 'Layer 1' for now based on previous use of getChainLayerFromAssetProtocol
```
pete-watters, line 57: I need to check with design and get examples of different tokens. The design only has `BTC` and looks like this:

Screenshot 2025-07-17 at 19 35 54

I will check the copy with design but decided to use a function we already had for now `getChainLayerFromAssetProtocol`
Screenshot 2025-07-17 at 19 36 21

apps/mobile/src/features/token/token.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const price = marketData.data?.price;
+  const { data: assetDescription } = useAssetDescriptionQuery(asset);
+  const { data: assetPriceChange } = useAssetPriceChangeQuery(asset);
+
+  return (
+    <TokenActivity
+      ticker={asset.symbol}
+      accountIndex={accountIndex}
+      fingerprint={fingerprint}
+      // TokenDetails is Activity ListHeader to avoid nested scrolling errors
```
pete-watters, line 92: This page is structured to be centered around the `ActivityList` `FlatList`. 

I pass all the non-activity content as the `ListHeader` so that scrolling works smoothly on both and it doesn't give errors.

apps/mobile/src/features/token/components/token-activity.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export function TokenActivity({
+  ListHeader,
+  ticker,
+  accountIndex,
+  fingerprint,
+}: TokenActivityProps) {
+  const scrollViewAdjustmentOffset = 56;
+  const { refreshing, onRefresh } = useRefreshHandler();
+  const activity = useTotalActivity();
```
edgarkhanzadian, line 40: I think this is going to be a big issue for performance on bigger wallets. To my knowledge we have a separate service for BTC/STX/SIP10 activities, so we'd rather use that

apps/mobile/src/features/token/components/token-activity.tsx
```diff
@@ -0,0 +1,79 @@
...
+      />
+      <FetchWrapper data={activity}>
+        {activity.state === 'success' && (
+          <Screen.List
+            refreshControl={<RefreshControl progressViewOffset={scrollViewAdjustmentOffset} />}
+            style={{ marginTop: -scrollViewAdjustmentOffset }}
+            data={activity.value
+              .filter(activity => activity && 'asset' in activity)
+              .filter(activity => filterByTicker(activity, ticker))
+              .filter(activity => {
```
edgarkhanzadian, line 60: i'd throw this filters into a useMemo as it might be a performance issue to filter these all out

edgarkhanzadian (approved):
Thank you @pete-watters ! I think this is good to merge but not to release for now (feature flag is helping tremendously here!). Let's merge this and then open smaller PRs to fix some issues

apps/mobile/src/features/token/components/token-activity.tsx
pete-watters: For this data, we are already loading it on initial load of the home page for total activity. So here it should just come from the cache.

I'm open to changing this if there are improvements to be made

apps/mobile/src/features/token/components/token-activity.tsx
pete-watters: I can do this but I think it's better to start using React Compiler when we can. I was working on a draft of adding that [here](https://github.com/leather-io/mono/pull/1392)

PR #1469, opened 2025-07-17
fix: update design for connections tab

edgarkhanzadian:
https://github.com/user-attachments/assets/a3503165-0ece-458f-9654-6492adc3c30d


tigranpetrossian (approved):
Thanks @edgarkhanzadian,
There's an emerging pattern of empty states that's probably worth standardizing. I'm adding those (along with the centering issue) to my notes. Will log a proper issue once I'm less swamped, so we can standardize.

PR #1485, opened 2025-07-21
fix: activity title [LEA-3051]

edgarkhanzadian:
https://github.com/user-attachments/assets/1cf83bdb-fa87-46d3-8459-a68bf5db878c


apps/mobile/src/app/(tabs)/activity.tsx
```diff
@@ -15,7 +16,20 @@ export default function ActivityScreen() {
 
   return (
     <Screen>
-      <Screen.Header leftElement={null} />
+      <Screen.Header
+        leftElement={null}
+        centerElement={
+          <Screen.HeaderAnimationTarget>
```
tigranpetrossian: `Screen.HeaderAnimationTarget` is not necessary here, the Text alone should be enough. `HeaderAnimationTarget` is wrapping the title internally, so when it leaves the view `centerElement` appears automatically.

tigranpetrossian (changes requested):

apps/mobile/src/app/(tabs)/activity.tsx
edgarkhanzadian: ohh, neat ! Thank you for catching this!

tigranpetrossian (approved):

PR #1486, opened 2025-07-21
fix: approver disabled confirm

edgarkhanzadian:
https://github.com/user-attachments/assets/38c3e654-a34c-426e-a335-5f235057c621


camerow (approved):
Nice little improvement here to the UX @edgarkhanzadian!

PR #1492, opened 2025-07-24
feat: minimum app version enforcement

camerow:
Screenshot 2025-07-24 at 7 30 32 AM


apps/mobile/src/components/version-guard/version-guard.tsx
```diff
@@ -0,0 +1,59 @@
...
+    Sentry.captureException(versionCheck.error, {
+      tags: {
+        version_guard: 'error',
+      },
+      level: 'warning',
+    });
+    return <>{children}</>;
+  }
+
+  // Show loading state while checking versions
```
tigranpetrossian: I'd remove all the redundant comments; the code is pretty self-explanatory.

apps/mobile/src/components/version-guard/version-guard.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <AppUpdateRequired
+        currentVersion={versionCheck.currentVersion}
+        minimumVersion={versionCheck.minimumVersion}
+        isLoading={false}
+      />
+    );
+  }
+
+  // Version is valid, render the normal app
+  return <>{children}</>;
```
tigranpetrossian:
```suggestion
  return children;
```

apps/mobile/src/utils/version-utils.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (parts.length < 2 || parts.length > 3) {
+    throw new Error('Version must be in format x.y or x.y.z');
+  }
+
+  const major = parseInt(parts[0] || '0', 10);
+  const minor = parseInt(parts[1] || '0', 10);
+  const patch = parts[2] ? parseInt(parts[2], 10) : 0;
+
+  if (isNaN(major) || isNaN(minor) || isNaN(patch)) {
+    throw new Error('Version components must be numeric');
```
tigranpetrossian, line 32: Can this be made resilient without throwing errors? I'd love to see a more idiomatic control flow that doesn't depend on try/catch, either by relying on fallback, or a union type returning {valid: false} | {valid: true: result: ...}

apps/mobile/src/components/app-update-required/app-update-required.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { VersionDisplay } from './version-display';
+
+interface AppUpdateRequiredProps {
+  currentVersion: string;
+  minimumVersion: string | null;
+  isLoading?: boolean;
+  onUpdatePress?: () => void;
+}
+
+export function AppUpdateRequired({
```
tigranpetrossian: nitpicking on naming: `UpdateGate`, `UpdateEnforcer`

apps/mobile/src/components/app-update-required/update-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    setIsLoading(true);
+
+    try {
+      await handleStoreRedirect();
+    } catch (error) {
+      // Show user-friendly error message
+      Alert.alert(
+        t({ id: 'version_guard.store_error_title', message: 'Unable to Open Store' }),
+        error instanceof Error
```
tigranpetrossian: We have [isError](https://github.com/leather-io/mono/blob/9c1837d596269dfae743372bca8c174f04d8c230/packages/utils/src/index.ts#L57) in `utils`

apps/mobile/src/components/app-update-required/app-update-required.tsx
tigranpetrossian: Scratch that--didn't scroll down enough to see this is wrapped in VersionGuard.

apps/mobile/src/components/app-update-required/update-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function UpdateButton({ onPress }: UpdateButtonProps) {
+  const [isLoading, setIsLoading] = useState(false);
+
+  async function handlePress() {
+    if (onPress) {
+      onPress();
+      return;
+    }
```
tigranpetrossian:
```suggestion
    onPress?.()
```

apps/mobile/src/components/app-update-required/update-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    setIsLoading(true);
+
+    try {
```
tigranpetrossian, line 22: Would love to see a basic state machine here too instead of `try/catch` for control flow.

apps/mobile/src/components/version-guard/version-guard.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  // If there's an error, allow the app to continue (graceful degradation)
+  if (versionCheck.error) {
+    Sentry.captureException(versionCheck.error, {
+      tags: {
+        version_guard: 'error',
+      },
+      level: 'warning',
+    });
+    return <>{children}</>;
```
tigranpetrossian:
```suggestion
    return children;
```

tigranpetrossian (approved):
Thanks @camerow,
Happy to have this in the app, this is a critical feature.

Couple of notes:
1. I don't think using LD as a back-end for this is optimal, this should ideally be powered by our back end/a runtime config.
2. Small design feedback: I think current and future versions are too prominent, but don't carry much value for the user, as they don't have to deal with versions manually.
I'd make it less prominent and perhaps lay it out as :

```
           Version
       2.56.0 -> 40.0.0 
```

apps/mobile/src/components/app-update-required/update-button.tsx
camerow: I've consolidated this a ton and there is really very little state now.

camerow:
> Thanks @camerow, Happy to have this in the app, this is a critical feature.
> 
> Couple of notes:
> 
> 1. I don't think using LD as a back-end for this is optimal, this should ideally be powered by our back end/a runtime config.
> 2. Small design feedback: I think current and future versions are too prominent, but don't carry much value for the user, as they don't have to deal with versions manually.
>    I'd make it less prominent and perhaps lay it out as :
> 
> ```
>            Version
>        2.56.0 -> 40.0.0 
> ```

FWIW I know that Kraken uses LaunchDarkly for this particular feature - are we ok with it in the mid-term? 

UI is updated to use a bottom sheet - I think you've seen the updated design.

tigranpetrossian:
@camerow Did some light refactoring, commits worth squashing before merge.
Re: LD - I think it's a misplaced responsibility, and those backfire in many ways that are hard to even anticipate.

On the long term, the idiomatic way of doing this kind of thing is our own API or at least runtime config, with decent session negotiation/bootstrapping logic.

PR #1493, opened 2025-07-25
Currency formatter, closes LEA-2198

tigranpetrossian:
This PR adds a library-level currency formatter to be used across our apps for formatting balances, prices, and other amounts.

It's intentionally not tied to our domain model, and has baked-in sensible defaults that follow our [Balance formatting rules](https://www.notion.so/trustmachines/Balance-formatting-rules-19ed74b694c78004956cfc0dafbad1b3?source=copy_link)

**[Detailed README](https://github.com/leather-io/mono/pull/1493/files#diff-6776dc78b52039d93e6af2a466b72341f5184bb306beba4daf0462c7da1c69b9)**

### Example one time factory setup and adapting to `Money` model

_Edit: updated naming below based on [Alex's suggestions](https://github.com/leather-io/mono/pull/1493#discussion_r2230553176)_
```ts
const currencyFormatter = createCurrencyFormatter({ locale: 'en-US' })

function formatAmount(money: Money, options?: formatCurrencyOptions) {
  return currencyFormatter.formatAmount(
    {
      amount: money.amount.shiftedBy(-money.decimals).toNumber(),
      currencyCode: money.symbol,
      decimals: money.decimals,
    },
    options
  );
}
```
At the app-level, this will adjust the currency format (in our case `Money`), to the input signature of the formatter.

### Examples

```ts
// Fiat currencies show currency symbols
formatAmount(usdBalance) // "$1,234.56"
formatAmount(jpyBalance) // "¥1,235" (no decimal places)
formatAmount(eurBalance) // "1.234,56 €" (locale 'de')

// Currency code for crypto is always appended regardless of locale
formatAmount(btcBalance) // "12.3456789 BTC"
formatAmount(ethBalance) // "456.123456 ETH"

// Large amounts use compact notation (≥1M by default)
formatAmount(millionDollarBalance) // "$1.23M"
formatAmount(billionDollarBalance) // "$1.23B"

// Small fiat amounts show approximation
formatAmount(dustAmount) // "< $0.01"
```

### Presets
Presets are meant to help to standardize the way we display money across our apps:
```ts
// prices
formatAmount(smallPrice, { preset: 'token-price' }) // $0.000001 
formatAmount(largePrice, { preset: 'token-price' }) // $3,587.02 
formatAmount(btcPrice, { preset: 'token-price' }) // $122,838.46 

// shorthand balance
formatAmount(btcBalance, { preset: 'shorthand-balance-crypto' }) // // "12.3457 BTC"
formatAmount(usdBalance, { preset: 'shorthand-balance-fiat' }) // // "$12.35"
```

### Percentage

Added a separate `formatPercentage` helper is available to be used once we start showing deltas:

```ts
formatPercentage(1) // 100.00%
formatPercentage(0.123456, 4) // 12.3456%
```


packages/utils/src/formatter/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Setup
+
+The formatter isn’t tied to our domain model and expects a numeric decimal input; therefore, it
+needs a one-time setup at the app-level to adapt the app’s currency to its input:
+
+```ts
+import { createFormatter, formatCurrencyOptions } from '@leather.io/utils';
+
+const formatter = createFormatter({ locale: 'en-US' });
```
alexp3y: Curious about the app-level implementation and how we could possibly set it up to be responsive to changes in locale, for if/when we implement a locale user preference setting. 

Perhaps using a hook w/ something like `useMemo` as the example here can help ensure we don't have to rework the usage pattern in the future.

packages/utils/src/formatter/README.md
```diff
@@ -0,0 +1,133 @@
+# Formatter
```
alexp3y, line 1: Question on the naming convention. "Formatter" feels pretty generic. What if we want to implement a "token name formatter", for example? Couldn't this more specifically be named "currency formatter"?

alexp3y: Question on the naming convention. "Formatter" feels pretty generic. What if we want to implement a "token name formatter", for example? Couldn't this more specifically be named "currency formatter"?

alexp3y (approved):
This is seriously awesome. Will be great to have this logic centralized and be able to apply it globally.

packages/utils/src/formatter/README.md
tigranpetrossian: This is a great question. Ultimately, whether we make this part of a hook at app-level depends on whether we want to reload the app when user wants to change the language or to a live update. There's also a complexity of language !== locale.
I'm leaning towards the former for few reasons:
* Users don't typically change languages very often, usually maybe once. If we get the locale detection just right they might never have to do it manually, which makes fully reloading the app upon language change a bearable experience, but avoids tying this to the state tree
* There a lots of moving parts when it comes to determining the locale that goes into the formatter, because it's not necessarily always the same as the language, some scenarios to consider:
  1. User has a locale for which we don't have translations (therefore a language setting), we fallback to english, but still respect user's regional formatting settings
  2. User has formatting preferences that differ from their locale, I've received feedback in the past on other apps where users locale formatting wasn't the `en` (period for decimal separator, commas for grouping, currency in the beginning), but that's what they were used to because most other apps they used didn't respect their locale.
  
[Here's a past discussion](https://trustmachines.slack.com/archives/C05LRS7G44R/p1742385352972509?thread_ts=1742382625.551349&cid=C05LRS7G44R) we had about localization vs i18n decisions.

Long story short: I would try to avoid trying this to React for as long as we can help it, then make a switch if it turns out not to be viable.

packages/utils/src/formatter/README.md
tigranpetrossian: I agree. I did multiple iterations on the name, and at some point landed on a bold assumption that in the future this might contain non-currency helpers, but I don't think that was a good idea. Will rename to "currency formatter". Do you think it's worth renaming the main helper `formatAmount` to `formatCurrency` as well?

packages/utils/src/currency-formatter/README.md
```diff
@@ -0,0 +1,133 @@
...
+formatAmount(largePrice, { preset: 'token-price' }); // $3,587.02
+formatAmount(btcPrice, { preset: 'token-price' }); // $122,838.46
+```
+
+#### Shorthand balances
+
+Meant to be used in listing items and other areas with scarce real estate.
+
+```ts
+formatAmount(btcBalance, { preset: 'shorthand-balance-crypto' }); // // "12.3457 BTC"
```
tigranpetrossian: The premise of `formatAmount` is to be currency-agnostic, so this isn't ideal.
I added ability to specify presets as callbacks to be input-aware after I added these presets, so these should be merged into a single `shorthand-balance` preset.

packages/utils/src/formatter/README.md
alexp3y: I think `currencyFormatter.formatAmount()` makes sense... It's an "amount" as it relates to currency (i.e. not potatoes). And that keeps it extensible - You can have other utilities alongside it like the `formatPercentage()`. App-level implementations could be then more specific - like `formatCurrencyAmount()`

edgarkhanzadian (approved):
Awesome work!

packages/utils/src/formatter/README.md
tigranpetrossian: Perfect, all set then. Thanks for the input @alexp3y

kyranjamie (approved):
Great work @tigranpetrossian 

I don't see an option to pad decimals? If I'm working with tabular data and I want a consistent number of decimal figures, can I do that with this library?

tigranpetrossian:
> Great work @tigranpetrossian 
> 
> I don't see an option to pad decimals? If I'm working with tabular data and I want a consistent number of decimal figures, can I do that with this library?

Thanks @kyranjamie, 
You should already be able to do that using minFractionDigits. If it's a common enough use case I can also add it to the custom options.

PR #1495, opened 2025-07-26
Use the new currency formatter in the mobile app, closes LEA-3079, LEA-2976

tigranpetrossian:
This replaces existing formatting implementations with the new currency formatter.

* Added necessary `Intl` polyfills to bring in [many of the unavailable features.](https://github.com/facebook/hermes/blob/main/doc/IntlAPIs.md#supported-on-android-only)
* Added formatting options to the Balance component
* Added ability to force display Balance bypassing the privacy setting, to address [LEA-2976](https://linear.app/leather-io/issue/LEA-2976/in-privacy-mode-consider-displaying-amounts-in-sign-transaction-screen). Sendable amounts and fees are no longer masked.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -11,3 +11,4 @@ CHANGELOG.md
 **/locales/**/messages.ts
 storybook-static/
 coverage/
+apps/mobile/index.js
--- a/apps/mobile/index.js
+++ b/apps/mobile/index.js
@@ -1,5 +1,18 @@
-import 'react-native-get-random-values';
+// @formatjs polyfill import order matters, altering it can crash the app.
+// This file is in .prettierignore.
+import '@formatjs/intl-getcanonicallocales/polyfill.js';
+import '@formatjs/intl-locale/polyfill';
 
+// Intl.NumberFormat support differs between Android and iOS https://github.com/facebook/hermes/blob/main/doc/IntlAPIs.md
+// For consistency, use force to ensure polyfill is applied regardless of Intl availability.
+// TODO: import locale data dynamically once we add translations
+import '@formatjs/intl-numberformat/polyfill-force';
+import '@formatjs/intl-numberformat/locale-data/en';
+
+import '@formatjs/intl-pluralrules/polyfill';
+import '@formatjs/intl-pluralrules/locale-data/en';
+
+import 'react-native-get-random-values';
 import 'expo-router/entry';
 import { polyfillWebCrypto } from 'expo-standard-web-crypto';
 import 'fast-text-encoding';
@@ -8,3 +21,4 @@ polyfillWebCrypto();
 if (typeof Buffer === 'undefined') {
   global.Buffer = require('buffer').Buffer;
 }
+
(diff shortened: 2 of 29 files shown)


tigranpetrossian:
Moving back to draft, needs to be updated to the [new shape](https://trustmachines.slack.com/archives/C05LAP952E7/p1753777726362519) before merging.

apps/mobile/index.js
```diff
@@ -1,5 +1,18 @@
-import 'react-native-get-random-values';
+// @formatjs polyfill import order matters, altering it can crash the app.
+// This file is in .prettierignore.
+import '@formatjs/intl-getcanonicallocales/polyfill.js';
+import '@formatjs/intl-locale/polyfill';
+
+// Intl.NumberFormat support differs between Android and iOS https://github.com/facebook/hermes/blob/main/doc/IntlAPIs.md
+// For consisteny, use force to ensure polyfill is applied regardless of Intl availability.
```
pete-watters: *nit* Typo in `consisteny` 🫨

.prettierignore
```diff
@@ -11,3 +11,5 @@ CHANGELOG.md
 **/locales/**/messages.ts
 storybook-static/
 coverage/
+apps/mobile/index.js
```
pete-watters, line 14: This is duplicated?

apps/mobile/src/components/balance/balance.tsx
```diff
@@ -1,74 +1,53 @@
...
-import { currencyDecimalsMap, currencyNameMap } from '@leather.io/constants';
 import { Money } from '@leather.io/models';
-import { SkeletonLoader, TextProps } from '@leather.io/ui/native';
-import { formatMoneyWithoutSymbol, i18nFormatCurrency } from '@leather.io/utils';
+import { SkeletonLoader, Text, TextProps } from '@leather.io/ui/native';
+import { FormatAmountOptions } from '@leather.io/utils';
 
 const EmptyBalanceDisplay = '-.--';
 
-export function isQuoteCryptoCurrency(balance: Money) {
```
pete-watters, line 10: I'm glad to see the back of this nasty code, thanks for refactoring and improving it 👍

pete-watters (approved):
This code LGTM and is a great improvement. I couldn't test locally yet as I'm having build issues but I will do it soon. 

Great work @tigranpetrossian

apps/mobile/src/components/balance/balance.tsx
tigranpetrossian: Thanks @pete-watters I've spent a lot of time finding a way for the new formatter to be currency-type-agnostic, so we can use one function for everything. There's still a non-zero chance that at some point we might need to hint it about the currency type, but we'll deal with it when that happens.

pete-watters:
I've tested this manually and confirmed that all is well. Great work @tigranpetrossian 👏

apps/mobile/src/components/balance/balance.tsx
pete-watters: Time well spent 👍 the fact it looks so simple and elegant now highlights the thought put into it's design

PR #1499, opened 2025-07-28
Add "pad-decimals" preset to the currency formatter

tigranpetrossian:
Per https://github.com/leather-io/mono/pull/1493#pullrequestreview-3055443041, this preset should reduce the verbosity of padding decimals via `minimumFractionDigits`. 
As I'm migrating the extension to the new formatter, it became clear very quickly that this is necessary.

### DIFF
--- a/packages/utils/src/currency-formatter/currency-formatter.types.ts
+++ b/packages/utils/src/currency-formatter/currency-formatter.types.ts
@@ -14,7 +14,7 @@ export interface FormatAmountOptions extends FormatAmountCustomOptions {
   preset?: CurrencyFormatterPreset;
   numberFormatOptions?: Intl.NumberFormatOptions;
 }
-export type CurrencyFormatterPreset = 'token-price' | 'shorthand-balance';
+export type CurrencyFormatterPreset = 'token-price' | 'shorthand-balance' | 'pad-decimals';
 
 export type CurrencyFormatterPresetValue =
   | CurrencyFormatterPresetResult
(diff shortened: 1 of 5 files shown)


packages/utils/src/currency-formatter/currency-formatter.types.ts
```diff
@@ -14,7 +14,7 @@ export interface FormatAmountOptions extends FormatAmountCustomOptions {
   preset?: CurrencyFormatterPreset;
   numberFormatOptions?: Intl.NumberFormatOptions;
 }
-export type CurrencyFormatterPreset = 'token-price' | 'shorthand-balance';
+export type CurrencyFormatterPreset = 'token-price' | 'shorthand-balance' | 'pad-decimals';
```
kyranjamie: Just a note on the naming of these presets (not from this pr), wouldn't a more generic `price` formatter suit non-token prices well too, e.g. we show EUR/USD or other fee info? Is it necessary to specify that the price is for tokens?

kyranjamie (approved):

packages/utils/src/currency-formatter/currency-formatter.types.ts
tigranpetrossian: This is general purpose enough to be called "price" actually, I can no longer remember what was the reason for using "token" there. Let rename before merging.

tigranpetrossian: Done. Looking at usages so far, we mostly need this preset for displaying a very accurate price without compact notation or rounding, the use case is mostly things like token details page, send, onramps, etc.
Simply calling `formatAmount` with fiat `Money` usually does a great job in other places like fees, holdings out of the box.

PR #1510, opened 2025-07-31
ci: don't check bot commit signature

edgarkhanzadian:


kyranjamie (changes requested):
Are you sure this isn't a randomly generated number? Let's just delete the job that runs the signed commit check?

edgarkhanzadian:
@kyranjamie well it's been working well for the android/ios release for a few months already 😂

edgarkhanzadian:
i'd say let's give this a try, i think it's not bad idea to check commit signatures

kyranjamie:
It has been broken for the release build for months?

edgarkhanzadian:
No, currently we release android and ios only when a release please pr is merged. This is the check that we use to make sure we do the release only when the release please pr is merged. So far it worked well without issues

kyranjamie (approved):

PR #1514, opened 2025-07-31
ci: don't check fingerprint

edgarkhanzadian:
This job is currently too noisy as it triggers on any change, even if unrelated to mobile. Turning it off until we actually start using eas updates


edgarkhanzadian:
apparently, the job is still going to run on this PR until this PR is merged because the job is set on `pull_request_target`. So changing the yaml in PR is not going to stop the job from running on this PR until we merge it into dev. Wasted some time on force pushing into this pr 😂

pete-watters (approved):
Thanks @edgarkhanzadian 👍  I can take over looking into this.

Here we are skipping this job now but I believe it shouldn't even be triggered at all for non mobile related work

PR #1523, opened 2025-08-04
Re-visit i18n on mobile

tigranpetrossian:
This PR is the first step to improving our i18n setup updating to Lingui v5, converting our translations to a simpler format and minor setup upgrade.

More to come in followup PRs: 
* Language detection and switching
* Completely automatic the message extraction and compilation (currently fighting Metro to make this work)

### Simpler implicit message declarations

**`t` usage: Before (an intentionally complex example):**

```ts
t({ 
  id: "configure_wallet.delete_wallet.authentication_failed",
  message: "Authentication failed"
})
```

`t` usage: After
```
t`Authentication failed`
```

### No need to use a different function for interpolation/ICU

**Before**
Previously, we've relied on `i18n` function (either from `useLingui` or imported directly) for interpolation:
```ts
i18n._({
  id: 'fees-sheet.fee-rate-caption',
  message: '{feeRate} sats/vB · {quoteFee}',
  values: { feeRate, quoteFee: formatCurrency(quoteFee) },
})
```

**After**
No need for cryptic syntax, works out of the box.
```
t`${feeRate} sats/vB · ${formattedQuoteFee}`
```

Lingui automatically picks up variable names and does the interpolation, preserving variable names for translators:
```
#: src/features/approver/components/fees/bitcoin-fee-option.tsx:41
msgid "{feeRate} sats/vB · {formattedQuoteFee}"
msgstr "{feeRate} sats/vB · {formattedQuoteFee}"
```

### Rationale
Apps using internationalization typically resort to one of the two options:
1. Key-based: maintaining a set of developer-named IDs, and keeping translations in JSON/PO files
2. Extraction and implicit IDs: code contains the source copy, which is extracted and translated

Since we've been using both at once which came with many caveats, I've considered both options separately, and landed on `#2` For following reasons:

1. Lingui has become really good at automatic extraction, see the interpolation example above
2. Manually maintaining keys is difficult: naming (conventions), inability to easily preview and see text, searching for code by copy -- all of those things are lost. 

### Tooling updates

`messages.js` is now `messages.ts` and is excluded via .gitignore. It compile automatically both locally and on CI (to be verified).

You still need to run `pnpm run lingui:extract` to update the translation files before committing. Currently exploring ways to automate this as well.


apps/mobile/src/features/send/components/recipient/build-recipient-suggestions.spec.ts
```diff
@@ -12,7 +12,7 @@ import { btcAsset } from '@leather.io/constants';
 import { SendAssetActivity } from '@leather.io/models';
 
 // One of the imports references @lingui/macro, causing compile-time errors.
```
pete-watters: Not a big deal but maybe we can update / remove the comment above now as it's out-dated
`One of the imports references @lingui/macro...` 

Could be `One of the imports references @lingui/core` if it is still needed

apps/mobile/src/features/token/components/token-price-change.tsx
```diff
@@ -22,40 +21,35 @@ export function TokenPriceChange({
...
-        {/* FIXME LEA-3015: this is showing an operator for negative values (-), but we don't want that in the design */}
-        <Balance
-          balance={priceChangeFiat}
-          variant="label02"
-          lineHeight={16}
-          color={getPriceChangeColor(changePercent)}
-        />
-        )
+      <Text variant="label02" color={color}>
+        {`${changePercent}% `}
```
pete-watters, line 43: This is a great example of how much better and simpler this new implementation is. Great work 👏

apps/mobile/src/i18n/locales/en/messages.po
```diff
@@ -0,0 +1,1629 @@
...
+#: src/features/send/components/recipient/recipient-toggle.tsx:57
+msgid "Enter recipient"
+msgstr "Enter recipient"
+
+#: src/features/recover-wallet/recover-wallet.layout.tsx:40
+#: src/features/recover-wallet/recover-wallet.tsx:112
+msgid "Enter your Secret Key"
+msgstr "Enter your Secret Key"
+
+#. js-lingui-explicit-id
```
pete-watters, line 701: Do we need to manually  deprecate these ones? With `js-lingui-explicit-id`?

pete-watters (approved):
This looks great @tigranpetrossian , nice work! It simplifies the setup and makes the code a lot cleaner and easier to reason with. 

Keeping the text in the source code is really helpful

apps/mobile/src/i18n/locales/en/messages.po
tigranpetrossian: Not at all, this will go away once I refactor out few of the leftovers using `i18n. I descoped it from this PR for now because it's a bit of work.

apps/mobile/src/features/send/components/recipient/build-recipient-suggestions.spec.ts
tigranpetrossian: Oh good catch, this is no longer necessary. Removed.

PR #1527, opened 2025-08-05
(NO_MERGE) feat: add readonly wallets & accounts

edgarkhanzadian:
The main purpose of this PR is to enable creation of readonly wallets and accounts, creation of STX only and BTC only accounts without triggering any errors in the usual app flow. App also guards users from accessing Send flow using a readonly account and connecting to the dApp with a readonly account.

This PR also builds functionality for testing and reviewing readonly accounts. When restoring a wallet, user is able to toggle the "is Readonly" switch on the screen in the Advanced settings and create a readonly wallet. When creating a wallet, app would create both STX and BTC keychains for the first wallet. For the subsequent accounts, user can add either BTC only or STX only account in the wallets screen. This is mostly made for the testing purposes.

"is Readonly" switch in the restore wallet screen is feature flagged and is only going to be enabled per device id in dev environment. It is not going to be enabled for everyone in dev env.

If user doesn't toggle the switch and creates a usual read/write wallet, the app is going to proceed as usual without any changes.

The video is ~2mins long, but it depicts me trying to test out different readonly wallet flows

https://github.com/user-attachments/assets/aac27150-ac8f-4d55-bac6-93026c5ece2f

readOnlyAndReadWrite


apps/mobile/src/features/settings/wallet-and-accounts/wallet-card.tsx
```diff
@@ -113,29 +148,35 @@ export function WalletCard({ fingerprint, variant, name }: WalletCardProps) {
...
-                  setIsAddingAccount(false);
-                }
-              }}
-              buttonState="ghost"
-              disabled={isAddingAccount}
-              title={isAddingAccount ? t`Adding account...` : t`Add account`}
-              icon={isAddingAccount ? <SpinnerIcon /> : <PlusIcon variant="small" />}
-            />
+            <>
+              {!isReadonlyWallet && (
```
pete-watters, line 152: This could be done in a cleaner way, using a component to wrap the logic and avoid so much repetition

apps/mobile/src/features/browser/approver-sheet/stx/call-contract/call-contract.tsx
```diff
@@ -7,7 +7,7 @@ import { useBroadcastStxTransaction } from '@/queries/stacks/use-broadcast-stx-t
 import { useAccounts } from '@/store/accounts/accounts.read';
 import { App } from '@/store/apps/utils';
 import { useStacksSigners } from '@/store/keychains/stacks/stacks-keychains.read';
-import { assertStacksSigner } from '@/store/keychains/stacks/utils';
+import { assertReadWriteStacksSigner } from '@/store/keychains/stacks/utils';
```
pete-watters, line 10: Maybe we could instead keep this as `assertStacksSigner` and assert `if (!signer || !signer.isReadonly)` instead?

kyranjamie (changes requested):
I love the enthusiasm to work on this feature, Edgar, and the speed at which you've pumped out a working version. 

I know we spoke about it on the call as a possible feature, but as far as I understood, we didn't decide to prioritise this. How we integrate read only wallets hasn't been considered from a design or product perspective, and there are _many_ implications in adding this feature.

This PR adds a huge amount of complexity to the key store, baking in read only specific conditions globally, which I think are detrimental to the code. `isReadonly` and `assertReadWriteStacksSigner` now have to be added universally. There are other ways we build this feature, that don't require baking in to the existing account structure, removing the need for many of the changes here.

Given the complexity added to handle a use-case we haven't fully considered the need for yet, I don't think we should merge this PR.

apps/mobile/src/features/send/forms/stx/stx-loader.tsx
```diff
@@ -35,14 +35,15 @@ function useStxData({ fingerprint, accountIndex }: AccountId): FetchState<StxDat
...
-    data: isReady
-      ? {
-          availableBalance: balance.data.stx.availableUnlockedBalance,
-          quoteBalance: balance.data.quote.availableUnlockedBalance,
-          nonce: nextNonce.data?.nonce,
-          marketData: marketData.data,
-        }
-      : null,
+    data:
+      isReady && balance.data
```
pete-watters, line 39: Is this needed? I think if it's status `isReady` it should have `data`?

pete-watters:
Great work here @edgarkhanzadian! I agree with @kyranjamie in that I think we should plan this through a bit better first. It adds a lot of complexity and it's unclear what direction this will take in the future so I think it's better to wait. 

In the demo it seems like you are pasting a mnemonic key and then toggling it as read-only. I understand this is just a demo but it would not be able to work that way as this would be a security problem

PR #1530, opened 2025-08-07
feat: token details pages

pete-watters:
This PR finalises `Token Details` on `mobile` and refactors it to use pages instead of sheets. 

I tried to implement it with sheets but it was error prone and full of problems. It would take a long time to do and would not be maintainable or worth the effort involved to do it. Pages is a much more natural fit and more maintainable.

It has nested views for `Send` + `Receive` and I’d need to manage a lot of state and routing myself
    - Essentially I’d have to build pages inside the sheet with all its own routing
    - I’d need to load views from `Send` + `Receive` inside of there also or I would be opening sheets on top of sheets 
- Send and Receive work differently using router props so would need to re-write that
- Managing state of ‘Back’ etc based on context would be a nightmare

Here's a quick demo:

https://github.com/user-attachments/assets/f0588bc3-a769-46ea-8d29-c9ffd9f144b9

**Remaining**
- integrate SIP-10 sends and Rune token data 
- improve loading states


apps/mobile/src/components/screen/screen.tsx
```diff
@@ -21,12 +21,14 @@ function Body(props: BoxProps) {
 }
 
 function Title({ children }: HasChildren) {
-  return (
+  return typeof children === 'string' ? (
```
pete-watters, line 24: Modifying this slightly so the `TokenOverview` icon can trigger the header animation

https://github.com/user-attachments/assets/65a97cda-dee3-4d88-9237-8e389ca0e378

apps/mobile/src/features/account/components/account-avatar.tsx
```diff
@@ -28,39 +28,43 @@ import {
...
+  size?: number;
 }
 
 export function AccountAvatar(props: AccountAvatarProps) {
   const Icon = isString(props.icon) ? accountIconMap[props.icon] : props.icon;
   return (
     <SquircleBox
-      width={48}
-      height={48}
+      width={props.size ?? 48}
```
pete-watters, line 66: This adjustment allows for smaller icons on this address list 

Screenshot 2025-08-07 at 17 09 56

apps/mobile/src/features/account/account-list/account-list-item.tsx
```diff
@@ -12,7 +12,7 @@ interface AccountListItemProps extends PressableProps {
 }
 export function AccountListItem({
   accountName,
-  address,
+  secondaryAside,
```
pete-watters, line 15: I renamed this to be more generic so the component could be re-used to show quote balance in this view 

Screenshot 2025-08-11 at 11 54 14

apps/mobile/src/features/balances/balances.tsx
```diff
@@ -8,34 +7,31 @@ import { RunesBalance, RunesBalanceByAccount } from '@/features/balances/bitcoin
...
   const tokenDetailsFlag = useTokenDetailsFlag();
 
-  const tokenSheetRef = useRef<SheetRef>(null);
-
-  function onOpenToken({ asset, availableBalance, quoteBalance }: OnOpenTokenProps) {
-    setSheetData({ asset, availableBalance, quoteBalance });
-    // analytics.track('token_sheet_opened', { source: 'action_bar' });
-    tokenSheetRef.current?.present();
+  function onOpenToken({ tokenId }: OnOpenTokenProps) {
+    router.navigate({
```
pete-watters, line 21: Now we are using a page it's easier to just push a route and let the route params then dictate what data is fetched. 

Previously using a sheet I was setting state `sheetData` to persist balances for render

apps/mobile/src/features/receive/receive-sheet.tsx
```diff
@@ -31,7 +32,7 @@ export function ReceiveSheet() {
       onDismiss={handleDismiss}
     >
       <SheetNavigationContainer base="receive">
-        <Receive accountId={accountId} />
+        <Receive accountId={accountId} assetType={assetType} tokenId={tokenId} />
```
pete-watters, line 35: `Receive` needs `assetType` and `tokenId` as users can choose to open the sheet from `BTC` detail level or specifically `BTC` address level:
Screenshot 2025-08-14 at 16 24 40

apps/mobile/src/features/receive/screens/asset-details.tsx
```diff
@@ -26,6 +26,10 @@ import { useReceiveNavigation, useReceiveRoute } from '../navigation';
 export function AssetDetails() {
   const route = useReceiveRoute<'asset-details'>();
   const navigation = useReceiveNavigation();
+  // Don't allow going back from token-details screen
```
pete-watters, line 29: In the case of token details we are skipping right to asset-details by pre-providing our params

apps/mobile/src/features/send/send-sheet.tsx
```diff
@@ -28,14 +30,17 @@ export function SendSheet() {
...
+      <Send accountId={accountId} asset={asset} />
     </FullHeightSheet>
   );
 }
 
 function useInitialSendParams() {
   const params = useGlobalSearchParams();
   const accountId = isString(params.accountId) ? params.accountId : undefined;
+  const tokenId = isString(params.tokenId) ? params.tokenId : undefined;
+  // FIXME LEA-3125: we need to refactor this so we can also send SIP-10 tokens
```
pete-watters, line 42: Now we can send SIP-10 I need to improve this. I'll do it in a follow up and update some of the SIP-10 services as I go to make it easier to work with specific tokens.

apps/mobile/src/features/token/bitcoin/bitcoin-account-list.tsx
```diff
@@ -0,0 +1,36 @@
...
+  wallet,
+  accountIndex,
+  fingerprint,
+}: BitcoinAccountListItemProps) {
+  const { value } = useBtcAccountBalance(fingerprint, accountIndex);
+
+  const availableBalance = value?.btc.availableBalance;
+  const quoteBalance = value?.quote.availableBalance;
+  if (!availableBalance || !quoteBalance) {
+    // TODO LEA-3015: add better loading state
```
pete-watters, line 24: I'll add these loading states in a small follow up / to this PR soon

packages/services/src/activity/activity.service.ts
```diff
@@ -60,18 +59,29 @@ export class ActivityService {
...
   }
 
+  /*
+   * Gets activity list for an account restricted to a single asset
+  */
+  public async getTotalActivityByAsset(
+    accounts: AccountAddresses[],
+    asset: CryptoAsset,
+    signal?: AbortSignal
+  ): Promise<OnChainActivity[]> {
```
pete-watters, line 74: Changing this return type to avoid casting and narrowing. We only support `OnChainActivity` so thats what we return

alexp3y: Nice. Could almost see us just renaming the entire service to `OnChainActivityService` to make it super clear.

apps/mobile/src/app/(tabs)/(index)/token/[tokenId]/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const configureTokenParamsSchema = z.object({
+  tokenId: z.string(),
+});
+
+export default function TokenScreen() {
+  const params = useLocalSearchParams();
+  const { tokenId } = configureTokenParamsSchema.parse(params);
+
+  switch (tokenId) {
```
Comment (alexp3y) [inline apps/mobile/src/app/(tabs)/(index)/token/[tokenId]/index.tsx:15]: Using `assetProtocol` here will help us expand this to new asset types.

apps/mobile/src/features/receive/get-assets.ts
```diff
@@ -1,11 +1,18 @@
 import { t } from '@lingui/core/macro';
 
-enum AssetType {
+export enum AssetType {
   NativeSegwit = 'native_segwit',
   Taproot = 'taproot',
   Stacks = 'stacks',
 }
-
+export interface Asset {
```
alexp3y, line 8: Maybe this can be a bit more specifically named so it doesn't lose context or collide with the `CryptoAsset` model... Would something like `ReceiveAsset` / `ReceiveAssetType` make sense?

alexp3y (approved):
Looks great 🦾

packages/services/src/activity/activity.service.ts
pete-watters: Sounds good 👍 the other types could be useful soonish so decided to leave them although I wanted to release waitlist

pete-watters:
Thanks for the review @alexp3y. I'm going to merge this and will add the suggested improvements in a follow up. I've been stuck in rebase land for a while here and it's flagged so want to get the good chunk of work in now

PR #1535, opened 2025-08-07
fix: update @hookform/resolvers to v5.2.1 for zod v4 compatibility

tigranpetrossian:
The send forms were not properly validating after the zod v4. Needed to upgrade @hookform/resolvers as well. 
cc @kyranjamie @fbwoolf In case you hit this in the extension.

@kyranjamie updated for the `web` too. Doesn't seem to have any breaking changes, but can you please verify just in case.


tigranpetrossian:
Being worked on in #1536

PR #1536, opened 2025-08-08
refactor: upgrade react-hook-form

kyranjamie:
Web & mob changes.

Upgrading the form and schema resolver lib, given the zod errors being triggered since zod upgrade. Guess here is that exceptions from the schema lib are not being caught given the version is quite outdated


tigranpetrossian:
@kyranjamie I'll close #1535. 
Relevant breaking changes are from [this resolvers release](https://github.com/react-hook-form/resolvers/releases/tag/v5.0.0) if it's useful

kyranjamie:
Thanks @tigranpetrossian yeah fixing these now

tigranpetrossian (approved):

PR #1537, opened 2025-08-08
Add automatic language detection & language setting to mobile

tigranpetrossian:
This wraps up the i18n improvements for mobile. Adds automatic language detection, a language setting, and automates extraction.

### Automatic language detection
The app now detects and uses the device's language automatically on startup. User-overrideable, the override disables further detection.

### Automatic message extraction in commits
Running `lingui` commands manually is no longer required.
The extraction now runs automatically when committing changes to mobile source files. The pre-commit hook:
  - Detects changes in `apps/mobile/src/`
  - Runs `lingui:extract` automatically
  - Stages the updated translation files

### Switching languages

Added language selector to Display settings when multiple languages are available.
Note that this is for demo purposes only. The  PR does not actually include Spanish/German translations. These are placeholders I added for testing.

https://github.com/user-attachments/assets/521aa146-ccc7-439c-a146-34675310e9c6

### Dynamic locale polyfills

Locale data (NumberFormat, Plurals) is now imported dynamically based on the selected language. 

### Fixed translation issues
- Removed incorrect translation markings in few places
- Fixed pluralization in activity formatting
- Added ICU syntax where appropriate
- Fixed `any` type issue in raw store
  
### Updated `eslint-plugin-lingui`
  
Doesn't seem to be much different, except `no-unlocalized-strings` defaults were removed.

### DIFF
--- a/apps/mobile/README.md
+++ b/apps/mobile/README.md
@@ -95,8 +95,17 @@ names for translators:
 #: src/features/approver/components/fees/bitcoin-fee-option.tsx:41
 msgid "{feeRate} sats/vB · {formattedQuoteFee}"
 msgstr "{feeRate} sats/vB · {formattedQuoteFee}"
-
 ```
+
+**Important**: The app reloads when language changes, so we don't re-render components in real-time.
+This means you don't need `useLingui` hook — it adds unnecessary complexity without benefit.
+
+The preference hierarchy for translations is:
+
+1. **Use `t` macro with template literals for most scenarios** — it's the simplest and most readable
+2. **Use lazy `msg` when `t` cannot be used** (outside functions, module-level declarations) — then wrap with `i18n._()` when rendering
+3. **Avoid `useLingui` hook** — we don't need reactive re-rendering for language changes
+4. **Avoid directly importing `i18n` for translations** — use `t` macro instead
 
 #### Limitations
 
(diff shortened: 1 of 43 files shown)


kyranjamie (approved):

edgarkhanzadian (approved):

apps/mobile/README.md
```diff
@@ -119,19 +128,51 @@ const statusMessages = {
 
 #### Extraction
 
-After finishing working on the feature, run the following command to extract strings from code
-into .po files:
+Message extraction happens automatically when you commit changes to mobile — no manual action needed.
```
pete-watters, line 131: Great improvement 👍

apps/mobile/src/app/settings/display/index.tsx
```diff
@@ -47,52 +53,59 @@ export default function SettingsDisplayScreen() {
       <SettingsList>
         <SettingsListItem
           title={t`Theme`}
-          caption={i18n._({
-            id: 'display.theme.cell_caption',
-            message: '{theme}',
-            values: { theme: capitalize(themePreference) },
+          caption={t({
+            message: select(themePreference, {
```
pete-watters, line 57: Thanks for fixing and explaining best practice here 👍 

Maybe the theme listing could be shared from somewhere rather than repeating it?

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
```diff
@@ -61,7 +61,7 @@ function ConfigureAccount({ fingerprint, accountIndex, account }: ConfigureAccou
 
   return (
     <>
-      <SettingsLayout title={t`Configure account`}>
+      <SettingsLayout title={t`Configure\naccount`}>
```
Comment (pete-watters) [inline apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:64]: Just double checking this is correct? Is it adding a line break?

apps/mobile/src/features/earn/sbtc-card.tsx
```diff
@@ -13,8 +13,8 @@ export function SbtcCard() {
   return (
     <EarnCard
       title={t`Earn with SBTC`}
-      minYield={t`6`}
-      maxYield={t`8%`}
+      minYield="6"
```
pete-watters, line 16: Nice cleanup here 👍 we wanted to update these via CrowdIn at one point but not needed now

eslint.config.js
```diff
@@ -70,17 +70,60 @@ export default tseslint.config(
     rules: {
       'lingui/no-unlocalized-strings': [
         'error',
+        // https://github.com/lingui/eslint-plugin/blob/main/docs/rules/no-unlocalized-strings.md
         {
-          ignoreFunction: [
+          ignore: ['^(?![A-Z])\\S+$', '^[A-Z0-9_-]+$'],
```
pete-watters, line 75: Great improvement here too 👍

pete-watters (approved):
Great work 👏

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
Comment (tigranpetrossian) [reply apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:64]: Yep, this is by design. Lingui didn't support these in the older versions if I'm not mistaken.

apps/mobile/src/app/settings/display/index.tsx
tigranpetrossian: I think this is a fairly small use-case not worth creating an abstraction for.

eslint.config.js
tigranpetrossian: Can't take credit, these are from the plugin's docs.

apps/mobile/README.md
tigranpetrossian: There's a small uncovered edge case with partial commits/amends that this might occasionally fail on. I'll fix it in a followup some time soon.

PR #1539, opened 2025-08-08
Mobile production build and submit workflow

camerow:
This PR implements a fairly simple workflow for kicking off production builds on Expo and writing them to a release.

✨ Features:

- [x] Builds and auto-submits those builds to the respective stores
- [x] Creates a Github release with links to Expo for each build
- [x] Stores the build output and saves it to the release. We can use this in the future for subsequent actions on the builds and appending information to the release
- [x] 🐛 Fixes the issues around `postinstall` in mobile using the `CI=1` env var causing other builds (`web`, cc @kyranjamie) to do too much by using `PROVISION_MOBILE` as the env var instead and setting it explicitly in these jobs and the Expo env
- [x] Moves `.easignore` to `apps/mobile` which reduced the file size being uploaded by ~200MB
- [x] Converts `app.config.js` -> `app.config.ts`
- [x] Adds `apps/mobile/fingerprint.config.js` which will hopefully make fingerprinting more consistent, but this will take some iteration

🤔 Future Enhancements:

- Add some logic to create tags like `mobile-v2.63.0-rc.1` etc so we can run and test multiple production builds
- Add some logic to cancel EAS builds if another build is run at this version
- Generate changelogs in the GH Action
- Add QR codes to quickly access and install builds for testing
- Smoke tests, E2E tests to pass / fail production builds


apps/mobile/package.json
```diff
@@ -23,7 +23,7 @@
     "lingui": "pnpm run lingui:extract && pnpm run lingui:compile",
     "lingui:compile": "lingui compile --typescript --strict",
     "lingui:extract": "lingui extract --clean --overwrite",
-    "postinstall": "pnpm lingui:compile && if [ \"$CI\" = \"1\" ]; then pnpm -w build:mobile; fi",
+    "postinstall": "pnpm lingui:compile && if [ \"$PROVISION_MOBILE\" = \"true\" ]; then pnpm -w build:mobile; fi",
```
kyranjamie, line 26: It's not entirely clear to me why we need to use `postinstall` at all. Can't all jobs that require mobile to be built imperatively build it, rather than tying it to the package manager lifecycle hook?

kyranjamie (approved):
Others more involved with mobile build system should review as well, but otherwise LGTM

apps/mobile/package.json
pete-watters: I think it's required in EAS, @edgarkhanzadian is that right?

.github/workflows/mobile-production-deploy.yml
```diff
@@ -0,0 +1,92 @@
...
+          name: '@leather.io/mobile: v${{ needs.build-and-submit-mobile.outputs.build_version }}'
+          body: |
+            # iOS Build:
+            https://expo.dev/accounts/leather-wallet/projects/leather-wallet-mobile/builds/${{ needs.build-and-submit-mobile.outputs.ios_build_id }}
+
+            # Android Build:
+            https://expo.dev/accounts/leather-wallet/projects/leather-wallet-mobile/builds/${{ needs.build-and-submit-mobile.outputs.android_build_id }}
+          files: /tmp/build_output.json
+
+  # Useful for testing the workflow without having to run tons of builds to completion
```
pete-watters, line 72: Do you want to keep this code? If so maybe it's better to have most of the code un-commented and to skip the job with a more precise comment

pete-watters (approved):
Nice work @camerow

apps/mobile/package.json
camerow: I believe we need this because we don't have direct imperative control in Expos servers when not using workflows - and we don't want to use workflows because they don't give us enough metadata for linking to a build in a GH release.

PR #1540, opened 2025-08-08
Fix various minor issues

tigranpetrossian:
### Enabled fast refresh in the mobile app when editing the `ui` package

I had trouble enabling this for all packages, but `ui` is probably where this is most needed.

### Disabled Sentry and LD logging 

Don't see a point in adding this to a separate config at the moment, since it's pretty straightforward to temporarily enable for yourself.

### Addressed deprecation notices from firebase messages

This has been polluting the console as well for me.

### Removed void & optional chaining from analytics calls

A while ago I rewrote the analytics in a way that doesn't require optional chaining, and excluded `track` from `no-floating-promises` since all analytics calls in reality are fire & forget.

**Before**
```ts
void analytics?.track()
```
**After**
```ts
analytics.track()
```

### Added custom `useTheme` 

**Before**
```ts
import { useTheme } from '@shopify/restyle';
import { Theme } from '@leather.io/ui';

const theme = useTheme<Theme>()
```
**After**
```ts
import { useTheme } from '@leather.io/ui';

const theme = useTheme()
```


apps/mobile/src/features/notifications/use-notifications.ts
```diff
@@ -8,9 +8,11 @@ import { useBitcoinAddresses } from '@/store/keychains/bitcoin/bitcoin-keychains
 import { useStacksSignerAddresses } from '@/store/keychains/stacks/stacks-keychains.read';
 import { useSettings } from '@/store/settings/settings';
 import { useWallets } from '@/store/wallets/wallets.read';
-import messaging from '@react-native-firebase/messaging';
+import { deleteToken, getMessaging, getToken, onMessage } from '@react-native-firebase/messaging';
 import * as Notifications from 'expo-notifications';
 
+const messaging = getMessaging();
```
pete-watters, line 14: Just mentioning here as there are updates to this file. We need to make a decision on whether to release notifications or just deprecate it.

- It's a good feature and useful so we could release it
- If we aren't releasing it we should roll back the complications it introduces with google services as we aren't using them

pete-watters (approved):
Nice work @tigranpetrossian , some good improvements to the codebase here

kyranjamie (approved):

apps/mobile/src/features/notifications/use-notifications.ts
edgarkhanzadian: +1, we need to revisit it and preferably get it working

edgarkhanzadian (approved):

PR #1543, opened 2025-08-11
feat(web): advanced tools

kyranjamie:
This PR adds an "advanced" feature to the web app, and the signer key generation form that's currently hosted on earn. This is the first step towards deprecating earn. Originally I started directly on the independent stacking form, but backtracked to start on this form first, as it's a prerequisite of the stacking form.

I've built out several helper component that we'll need for the remaining forms. @fabric-8 some components used aren't in the design system (e.g. the select menus, the rounded buttons etc). Implementation isn't exactly to designs, beginning with a flat hierarchy or tools rather than grouped.

image

image

Todo:
- [x] Closing flow, once user has approved signature, what happens next?


fbwoolf (approved):

tigranpetrossian (approved):

fabric-8:
fyi, web build throws this:
CleanShot 2025-08-13 at 10 37 11@2x

kyranjamie:
I've noticed there's a dependency that only works when the user is signed in, and the form errors if the user doesn't have extension installed/connected.

Working on a test config that can preconfigure the test into various different states.

fabric-8:
@kyranjamie I assume this is what you were referring to RE: wallet not connected 
CleanShot 2025-08-13 at 11 32 03@2x

kyranjamie:
I've added the result box and the copy JSON feature also in Earn. @fabric-8 we have a `InfoBox` component I'm using that differs from the designs, but aim to refactor this internally to new designs next time I work on a form. 

I'm running into some errors I hadn't considered. For example, we require the user is connected to access the form, so I have to add different states for if the user isn't connected/doesn't have extension installed. If the user signs out on this page, then the address disappears causing an error, so we need fallback handling to at the very least, navigate to home page.

PR #1545, opened 2025-08-12
fix: flashlist performance

edgarkhanzadian:
- fix flashlist performance
- - we should make sure that flashlist is not a child component of scrollview or any other list. Keeping it high up the tree is a priority
- update flashlist to the latest version
- merge sip10s and runes into "assets", s.t. we have better control over number and priority of assets displayed in home / assets page 

https://github.com/user-attachments/assets/1d93bfbb-2fad-4739-962e-e5acda27b5fb


camerow (approved):
🚀

apps/mobile/src/features/balances/assets/assets-flashlist.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface AssetsFlashListProps {
+  sip10Data: ReturnType<typeof useSip10TotalBalance>;
+  runesData: ReturnType<typeof useRunesTotalBalance>;
+  header: ReactNode;
+}
+
+export function AssetsFlashList({ sip10Data, runesData, header }: AssetsFlashListProps) {
+  const runesFlag = useRunesFlag();
+
+  const sip10Memo = useMemo(() => {
```
pete-watters, line 27: Did you do some checks into the effectiveness of this memoization? Don't we already have it cached in the React Query cache?

I don't think we should really use `useMemo` as its not as effective as it seems. I was working on [this draft](https://github.com/leather-io/mono/pull/1392) of using React Compiler but it was blocked by our React Animated version. We can look into that again and then avoid using `memo` altogether

apps/mobile/src/features/balances/assets/assets-balance-by-account.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { useRunesFlag } from '../../feature-flags';
+import { OnOpenTokenProps } from '../balances';
+import { ASSETS_BALANCES_WIDGET_LIMIT } from '../constants';
+import { renderAsset } from './render-assets';
+
+interface AssetsBalanceProps {
+  onPress?: ({ asset, availableBalance, quoteBalance }: OnOpenTokenProps) => void;
+}
+export function AssetsBalanceByAccount({
```
pete-watters, line 16: Nice work here 👍  this gives us more flexibility with sorting also. I wonder should we do this for all assets and move away from always having them ordered as `BTC`, `STX`, `sBTC` then the rest. I will mention it soon.

pete-watters:
@edgarkhanzadian - was this refactor done to solve a specific problem with how we implemented `Flashlist`? 

@tigranpetrossian had already made some improvements via introducing `Screen.List`. 

In the past we had some problems with content disappearing on scroll of large lists with the latest flashlist version. 

I'm not sure about doing refactors like this without good reason given we don't have many tests

apps/mobile/src/features/balances/assets/assets-flashlist.tsx
tigranpetrossian: @pete-watters RQ cache and memoization are orthogonal concerns, one's caching data so that isn't unnecessarily refetched, the other ensures expensive computations aren't unnecessarily re-running on every render. I'm assuming Edgar added this to optimize the sorting, because sip-10 data is largely an unknown beast.

I'd be more worried about `.sort` here mutating the data. @edgarkhanzadian If this query is shared elsewhere, the list will likely be sorted there too, use `remeda`?

tigranpetrossian:
> @edgarkhanzadian - was this refactor done to solve a specific problem with how we implemented `Flashlist`?
> 
> @tigranpetrossian had already made some improvements via introducing `Screen.List`.
> 
> In the past we had some problems with content disappearing on scroll of large lists with the latest flashlist version.
> 
> I'm not sure about doing refactors like this without good reason given we don't have many tests

@pete-watters If I remember correctly sip-10s were already a nested Flashlist with ScrollViews, which was problematic. I descoped fixing that from `Screen` implementation because it was quite a bit of work. It was also used in a widget that showed only 5 items max. Happy to see those being ironed out.

In an ideal world I'd love to see these normalized to a single AssetList that doesn't necessarily know what SIP-10 is, the way it's done in the extension, but that's tricky given how asset list is being rendered both within Screen and in independent views.
We can look into that as a followup.

pete-watters:
> > @edgarkhanzadian - was this refactor done to solve a specific problem with how we implemented `Flashlist`?
> > @tigranpetrossian had already made some improvements via introducing `Screen.List`.
> > In the past we had some problems with content disappearing on scroll of large lists with the latest flashlist version.
> > I'm not sure about doing refactors like this without good reason given we don't have many tests
> 
> @pete-watters If I remember correctly sip-10s were already a nested Flashlist with ScrollViews, which was problematic. I descoped fixing that from `Screen` implementation because it was quite a bit of work. It was also used in a widget that showed only 5 items max. Happy to see those being ironed out.
> 
> In an ideal world I'd love to see these normalized to a single AssetList that doesn't necessarily know what SIP-10 is, the way it's done in the extension, but that's tricky given how asset list is being rendered both within Screen and in independent views. We can look into that as a followup.

OK great, I just wanted to double check this didn't affect your improvements there

apps/mobile/src/features/balances/assets/assets-flashlist.tsx
pete-watters: I just wonder if we have checked the gain from memoizing this. If the data is likely to change often anyway sometimes its less performant to memoize

PR #1552, opened 2025-08-14
fix: update zod patch version

kyranjamie:
Investigating zod issues in extension, upgrading patch version while at it


alexp3y (approved):

alexp3y:
Seems to be some linter errors.

PR #1556, opened 2025-08-15
feat: normalize account service request usage

alexp3y:
This PR broadens the usage of "Account" as a primary-driver of context when making service calls and adds new account-based look-ups for Stacks assets (instead of just "get by address").

The idea is that `getBtcAccountBalance(req)`, `getSip10AccountBalance(req)`, `getAccountUtxos(req)`, etc. should all expect a consistent request object representing the state of the current account.

In practice, the request account object is constructed centrally in each app (e.g. `use-account-addresses`), ensuring consistent usage across service queries / hooks.

Favoring accounts also help us support Ledger where either Bitcoin or Stacks account information can be absent. Services should be fully resilient to these states.

e.g. `getStxAddressBalance(address)` by definition expects a valid Stacks address. If an invalid address is passed, an exception is thrown.

`getStxAccountBalance(req)` by contrast expects a request object where `req.account.stacks` can possibly be undefined. This is still a valid request, and the service can safely return a 0 balance response.

### DIFF
--- a/packages/services/src/balances/sip10-balances.service.ts
+++ b/packages/services/src/balances/sip10-balances.service.ts
@@ -6,6 +6,7 @@ import {
   baseCurrencyAmountInQuote,
   createBaseCryptoAssetBalance,
   createMoney,
+  hasStacksAddress,
 } from '@leather.io/utils';
 
 import { Sip10AssetService } from '../assets/sip10-asset.service';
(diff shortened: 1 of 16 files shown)


packages/services/src/balances/sip10-balances.service.ts
```diff
@@ -63,15 +65,33 @@ export class Sip10BalancesService {
     };
   }
 
+  public async getSip10AccountBalance(
```
pete-watters, line 68: It would be great if we can extend this to also offer token specific account balances + addresses (for SIP-10 and Runes). 

This would save us having to filter the balance arrays client side.

pete-watters (approved):
Nice work @alexp3y  👏

PR #1558, opened 2025-08-15
feat: single account home page

edgarkhanzadian:
- Implement home page with flashlists for assets and collectibles
- Remove previous homepage related code

https://github.com/user-attachments/assets/e23c6b05-31af-4a1b-bfc4-c4be412991ba

https://github.com/user-attachments/assets/f599fec9-a154-4b23-8a8c-b5ed9bcf0f71


apps/mobile/src/features/send/send-flow-provider.tsx
```diff
@@ -4,25 +4,20 @@ import { Account } from '@/store/accounts/accounts';
 import { InputCurrencyMode } from '@/utils/types';
 
 import { FungibleCryptoAsset } from '@leather.io/models';
-import { isDefined } from '@leather.io/utils';
 
 interface SendState {
-  selectedAccount: Account | null;
   selectedAsset: FungibleCryptoAsset | null;
+  account: Account;
```
pete-watters: I'm not sure if changing this from `selectedAccount` is an improvement? 🤔

apps/mobile/src/app/(tabs)/(index)/index.tsx
```diff
@@ -1,77 +1,15 @@
...
-                </CollectiblesWidget>
-              )}
-            </>
-          )}
-        </Box>
-      </Screen.ScrollView>
-      <NotificationsSheet sheetRef={notificationSheetRef} />
-    </Screen>
-  );
+  const { currentFingerprint, currentAccountIndex } = useCurrentAccount();
```
pete-watters: Maybe we could add something like `hasWallets` here to make this slightly easier to read? 

Like:
`const hasAccount = currentFingerprint !== null && currentAccountIndex !== null`

apps/mobile/src/app/_layout.tsx
```diff
@@ -63,6 +65,7 @@ ErrorUtils.setGlobalHandler(error => {
 function App() {
   useWatchNotificationAddresses();
   usePageViewTracking();
+  const { currentAccountId } = useCurrentAccount();
```
pete-watters: Maybe we can standardise the names? 

`currentAccountId` is good but we renamed `selectedAccount` -> `account`. 

We could use `currentAccount` / `selectedAccount` / `activeAccount` everywhere for the current active account?

apps/mobile/src/components/home/components/asset-tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <Box>
+      <Box flexDirection="row">
+        <TabButton
+          isActive={listTab === 'tokens'}
+          title={t`Tokens`}
+          onPress={() => {
+            setListTab('tokens');
+          }}
+        />
+        <TabButton
```
pete-watters, line 30: Do we always want to show `NFTs` even if the user doesn't have any? 

Was there an empty state in the design? 🤔  
Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-08-18 at 07 45 26

apps/mobile/src/features/earn/stacking-card.tsx
```diff
@@ -1,26 +0,0 @@
...
-  const { openURL } = useOpenURL();
-  return (
-    <EarnCard
-      title={t`Stacking rewards`}
-      minYield="6"
-      maxYield="10%"
-      description={t`Acquire Stacks (STX) on to Bitcoin’s leading L2 to earn yield from staking`}
-      image={
-        <>
-          <Image source={require('@/assets/stickers/stacking.png')} width={177} height={228} />
```
pete-watters, line 20: Are we ditching all of these cards now? So if the user isn't stacking yet we don't encourage them to?

Screenshot 2025-08-18 at 06 38 59

If so we should also get rid of the assets used - `@/assets/stickers/stacking.png` etc.

apps/mobile/src/components/home/utils.ts
```diff
@@ -0,0 +1,3 @@
+export type ListTab = 'tokens' | 'collectibles';
```
pete-watters: This is more of a types / constants file than `utils`?

apps/mobile/src/features/collectibles/collectibles-serializer.tsx
```diff
@@ -9,35 +8,31 @@ import {
...
-            name: collectible.stamp.toString(),
-            type: collectible.protocol,
-            src: collectible.stampUrl,
-          };
-        default:
-          assertUnreachable(collectible);
-      }
-    })
-    .filter(Boolean) as CollectibleCardProps[];
+export function serializeCollectible(collectible: NonFungibleCryptoAsset) {
```
pete-watters, line 11: I'm not sure if its this change or others but in the new `collectibles` I am seeing some empty slots where they aren't rendering. 

**new**

https://github.com/user-attachments/assets/0a22c4f8-43c0-42a5-bfa8-7211a2ca381b




**old**


https://github.com/user-attachments/assets/19aeb275-1a69-4543-8fc6-624606618f50

pete-watters:
Great work here @edgarkhanzadian 👏  I took a look and did some testing and almost everything looks great. I found a [small issue](https://github.com/leather-io/mono/pull/1558#discussion_r2281483104) with NFTs rendering.

I merged `Token Details` today so this caused a big rebase here, apologies for that, but I've been stuck rebasing all these updates myself. Now we have made this move to single account view we will need to rework that a lot anyway

Merge wise, we cannot feature flag this so we need to make certain everything works and that we want to have this in the next release or else cut a release branch of work before it. I'm all for releasing it but we might need more time testing + reviewing. 

We need to make sure we are certain some things were planned like removing `Earn Cards` and maybe `All Activity` doesn't make sense as much now we are focused on a single account? 

https://github.com/user-attachments/assets/345dff58-7849-458b-a90b-4bc787e24bed

apps/mobile/src/app/(tabs)/(index)/index.tsx
```diff
@@ -1,77 +1,15 @@
...
+
+  if (currentFingerprint !== null && currentAccountIndex !== null) {
```
tigranpetrossian: This seems like excessive branching for what is simply an empty state representation. 
A single HomeScreen that does an early return of a separate component would be probably easier to reason about.

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -0,0 +1,106 @@
...
+    accountSelectorSheetRef.current?.close();
+    setCurrentAccountId(accountId);
+  }
+  const { onPressToken, sheetData, tokenSheetRef } = useTokenDetails();
+
+  return (
+    <Screen>
+      {/* TODO: research better way of switching between flashlists */}
+      {listTab === 'tokens' && (
+        <AssetsFlashList
```
tigranpetrossian: Names for those lists don't necessarily need to reflect the internal implementation. AssetList/Collectible~~s~~List should be just fine.

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -0,0 +1,106 @@
...
+          header={
+            <>
```
tigranpetrossian, line 66: Unless there where any issues without using the standard <Screen> subcomponents, I'd very much like to see this page being consistent with the way we do layouts anywhere else, otherwise this feels like abusing the FlashList header for half of the page's content. This also unnecessarily remounts that content (AccountDetails, AssetTabs), preventing those for from having any sort of internal state, i.e. normal tabs implementation.

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  function onAccountPress(accountId: string) {
+    accountSelectorSheetRef.current?.close();
+    setCurrentAccountId(accountId);
+  }
+  const { onPressToken, sheetData, tokenSheetRef } = useTokenDetails();
+
+  return (
+    <Screen>
+      {/* TODO: research better way of switching between flashlists */}
```
tigranpetrossian: This needs a well-made Tabs component with baked in accessibility and focus management. I can make a general purpose one to be used here and few other places (I think Settings?) where we use those.
This is blocked by the issue I raised about FlashList headers in the adjacent comment though.

apps/mobile/src/core/current-account-provider.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const context = useContext(CurrentAccountContext);
+  if (!context) throw new Error('`useCurrentAccount` must be used within `CurrentAccountProvider`');
+  return context;
+}
+
+export function CurrentAccountProvider({ children }: HasChildren) {
+  const [currentAccountId, setCurrentAccountId] = useState<null | string>(null);
+  const { list } = useAccounts();
+
+  useEffect(() => {
```
tigranpetrossian, line 25: This code isn't effectful, `useEffect` shouldn't be necessary.
I'm guessing this is to react to changes caused by `setCurrentAccountId`, if that's the case:

```ts
function handleAcconutIdChange(id: string) {
  // ... the logic for setting and safeguarding goes here
}
```

apps/mobile/src/core/current-account-provider.tsx
```diff
@@ -0,0 +1,58 @@
...
+    if (!currentAccountId) return;
+    const foundAccount = list.find(acc => acc.id === currentAccountId);
+    if (!foundAccount) {
+      setCurrentAccountId(null);
+    }
+  }, [list, currentAccountId]);
+
+  useEffect(() => {
```
tigranpetrossian: What's the case for this happening? Similar to the other `useEffect` comment, one of the problems of using useEffect` (apart from it being meant for side effects) is it obscures intent.

apps/mobile/src/features/account/components/account-actions.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { Box, Button } from '@leather.io/ui/native';
+
+export function AccountActions() {
+  const { hasWallets } = useWallets();
+  const { sendSheetRef, receiveSheetRef } = useGlobalSheets();
+
+  if (!hasWallets) return null;
+  return (
+    <Box px="5" pb="5" flexDirection="row" gap="2">
```
tigranpetrossian, line 13: I think @pete-watters made a similar refactor in his Tokens PR too, worth consolidating.

apps/mobile/src/features/collectibles/collectibles-flashlist.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function CollectiblesFlashList({ collectiblesData, header }: CollectiblesFlashListProps) {
+  const collectiblesMemo = useMemo(() => {
+    if (collectiblesData.state === 'success') return collectiblesData.value;
+    return [];
+  }, [collectiblesData]);
+
+  return (
+    <>
```
tigranpetrossian: nit: leftover fragment

apps/mobile/src/features/receive/receive-flow-provider.tsx
```diff
@@ -3,86 +3,55 @@ import { ReactNode, createContext, useContext, useReducer } from 'react';
 import { SelectedAsset } from '@/features/receive/screens/select-asset';
 import { Account } from '@/store/accounts/accounts';
 
-import { isDefined } from '@leather.io/utils';
-
 interface ReceiveState {
-  selectedAccount: Account | null;
   selectedAsset: SelectedAsset | null;
-  accounts: Account[];
+  account: Account;
```
tigranpetrossian: The most coveted line of change for me 🥹

apps/mobile/src/features/send/components/recipient/use-shameful-account-helpers.ts
```diff
@@ -12,10 +13,9 @@ interface UseAccountHelpersResult {
   getAddressByAccount: (fingerprint: string, index: number) => string | null;
 }
 
-export function useAccountHelpers(
-  accounts: Account[],
-  asset: FungibleCryptoAsset
-): UseAccountHelpersResult {
+export function useAccountHelpers(asset: FungibleCryptoAsset): UseAccountHelpersResult {
+  const { list: accounts } = useAccounts();
```
tigranpetrossian: I think this needs to be `useAccounts('active')` to filter out hidden accounts.

apps/mobile/src/features/send/send-flow-provider.tsx
tigranpetrossian: Within `Send` it is I suppose. There's no concept account selection within the feature anymore, since it's global. Probably worth looking into globally available `currentAccount` object now, not sure what's the LOE for including it in this PR.

apps/mobile/src/features/settings/description-sheet.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { descriptionSheetRef } = useGlobalSheets();
+  const { bottom } = useSafeAreaInsets();
+  const { themeDerivedFromThemePreference } = useSettings();
+  const [sheetData, setSheetData] = useState<DescriptionSheetData | null>(null);
+  const ref = useRef<SheetRef>(null);
+
+  const theme = useTheme();
+  useImperativeHandle(descriptionSheetRef, () => ({
+    present(newSheetData) {
+      setSheetData(newSheetData);
```
tigranpetrossian, line 26: This is so much better than the bizarre way the library handles data passing. Why didn't I think of this!
I might look into making this core `Sheet` functionality.

packages/ui/src/components/collectibles/native/collectible-card.native.tsx
```diff
@@ -16,29 +15,23 @@ export interface CollectibleCardProps {
...
-              return <CollectibleText src={src} size={size} />;
-            case 'html':
-            case 'gltf':
-            case 'svg':
-            case 'video':
-              return <CollectibleHtml src={src} size={size} />;
-            case 'image':
-              return <CollectibleImage source={src} alt={name} size={size} />;
-            default:
-              assertUnreachable(mimeType as never);
```
tigranpetrossian, line 36: There is no reason to use `assertUnreachable` if the mimeType isn't a discriminated union. `assertUnreachable` is meant for enforcing exhaustiveness and crash upon breached API contracts, type casting within it negates the former and increases the likelihood of the latter. Returning nothing or showing an "Unknown type" placeholder is the right way to handle this.

tigranpetrossian (changes requested):
Thanks @edgarkhanzadian,

This probably the single most important change in the app at least for me personally.
This is going to unlock lots of things we couldn't do before.

Requested few changes for things that I think need to be fixed rather sooner than later, namely:

1. `AssetFlashList`/`CollectibleFlashList`, apart from the names, these carry a large amount of the content in their headers, which I think is going to cause issues down the line. A more standard layout using Screen, and possibly even adding Screen.Flashlist replicating current Screen.Flatlist should make things easier to reason about and prevent possible transient state issues caused individual header content remounting when switching tabs.
2. `useEffect` for none effectful code that can be either derived to run in a handler
3. `assertUnreachable(mimeType as never);`. I don't know how often we get  unknown mimeTypes, if often - this warrants a placeholder design for "Unknown type", otherwise just returning null should be enough.
4. This technically isn't a blocker, but I'd love to see us doing a better job at basic accessibility. Current Tabs implementation in the homepage doesn't meet those standards. I can address this myself in a followup, but it will need (1.) to be resolved first.

apps/mobile/src/features/send/send-flow-provider.tsx
edgarkhanzadian: i did implement a globally available useCurrentAccount, but it always returns a nullable response (as sometimes we might not have the current account). So to decrease the amount of assertions in the send flow, i've asserted it's existence on the send root level and then propagated the account to the children via props/context

tigranpetrossian: Nice, thanks @edgarkhanzadian. I've been focused on UI code and missed that.
It's nullable for when the user hasn't added any wallets yet, right?

edgarkhanzadian: yeap yeap

pete-watters: Thanks Edgar, that makes sense 👍  I guess now we almost always have either:
- an empty wallet
- a wallet with one account selected? 

Maybe we can work on `useCurrentAccount` in that regard?

apps/mobile/src/app/(tabs)/(index)/index.tsx
pete-watters: I had written similar and removed, I agree with this

apps/mobile/src/components/home/home-with-account-screen.tsx
pete-watters: I began using this approach in token details to get around problems with nested scrolling Flashlists. In this case `header` is passed on to the `FlashList` [ListHeaderComponent prop](https://shopify.github.io/flash-list/docs/usage/#listheadercomponent). 

This allows the content to be placed above the list and for the Flashlist scroll to remain intact. Is there a better way we can do this? Does `Screen.List` handle this better? 

Perhaps we are making the header too heavy but it seems the best way to split the content when you have a flashlist below the fold

apps/mobile/src/features/send/components/recipient/use-shameful-account-helpers.ts
pete-watters: Maybe time to refactor this file and make it less `shameful` also 😄

packages/ui/src/components/collectibles/native/collectible-card.native.tsx
pete-watters: This is my original poor code, battling the compiler that Edgar has just moved around. We are very unlikely to hit this scenario of:
- type `inscription` 
- `mimeType` unknown 

Probably the safest bet is to just return nothing here as you say or if we want to be comprehensive come up with an `Unknown` placeholder

tigranpetrossian: Yeah sorry about that, it's easy to confuse moved/added code  in GitHub diffs. We can address this in a followup.

apps/mobile/src/app/(tabs)/(index)/index.tsx
edgarkhanzadian: yeap good idea, the current interface of useCurrentAccount could be quite improved

apps/mobile/src/components/home/components/asset-tabs.tsx
edgarkhanzadian: Good point! I'll follow up with the design

apps/mobile/src/features/earn/stacking-card.tsx
edgarkhanzadian: Yeap, those are not going to be used anymore. I'll delete the pngs

apps/mobile/src/features/collectibles/collectibles-serializer.tsx
edgarkhanzadian: Hmm yea could be an issue with flashlist rendering, taking a look

apps/mobile/src/core/current-account-provider.tsx
edgarkhanzadian: actually this useEffect is not needed anymore, removing

apps/mobile/src/core/current-account-provider.tsx
edgarkhanzadian: I don't think i'd be able to remove this one. I need to set the current account id here as soon as the account list gets updated with accountids

apps/mobile/src/features/send/send-flow-provider.tsx
edgarkhanzadian: i updated the context interface to be:
```
interface CurrentAccountContextValue {
  currentAccount: Account | null;
  setCurrentAccount(account: Account | null): void;
}
```
That should be easier to check existence of the current account and just generally retrieve fingerprint/accountIndex from it

apps/mobile/src/components/home/home-with-account-screen.tsx
edgarkhanzadian: Yeap, unfortunately we need to do it to avoid nesting a flashlist inside of a scrollview

apps/mobile/src/features/send/components/recipient/use-shameful-account-helpers.ts
tigranpetrossian: Not a chance!

apps/mobile/src/features/collectibles/collectibles-serializer.tsx
edgarkhanzadian: Seems to be an issue with duplicate bns nfts, should be resolved with services update

pete-watters: Sounds good 👍

apps/mobile/src/app/_layout.tsx
edgarkhanzadian: good point! Refactored to try to keep these namings as consisted as possible

apps/mobile/src/app/(tabs)/(index)/[tokenId]/index.tsx
```diff
@@ -1,22 +1,23 @@
...
   const params = useLocalSearchParams();
-  const { tokenId, accountId } = configureTokenParamsSchema.parse(params);
-  if (!accountId) {
-    throw new Error('accountId is required');
-  }
-  const { accountIndex, fingerprint } = deserializeAccountId(accountId);
+  const { tokenId } = configureTokenParamsSchema.parse(params);
+  const { currentAccount } = useCurrentAccount();
+
+  assertExistence(currentAccount, 'Current account is required for AccountTokenScreen');
```
Comment (pete-watters) [inline apps/mobile/src/app/(tabs)/(index)/[tokenId]/index.tsx]: It's good to start using `assertExistence` as it's clear this is mandatory 👍

apps/mobile/src/app/(tabs)/activity.tsx
```diff
@@ -1,21 +1,11 @@
...
-
   return (
-    <Screen>
-      <Screen.Header
-        leftElement={null}
-        centerElement={<Text variant="heading05">{t`All Activity`}</Text>}
-      />
-      <ActivityFlashList data={activity} header={<Screen.Title>{pageTitle}</Screen.Title>} />
-    </Screen>
+    <CurrentAccountLoader fallback={<ActivityScreenWithoutAccount />}>
```
pete-watters, line 7: It's debatable but I'm not sure of the value of letting people see this empty activity screen:

https://github.com/user-attachments/assets/f27767d6-ffcd-4ea9-a78c-38e46a0455fe

I would suggest to instead:
- disable the activity tab button if it's an empty wallet
- hide the button until there is activity 

I'm fine with this also but as mentioned on Slack, it's adding a complication checking for existence of account before calling hooks and I think it's a cleaner code solution to simply not call them without an account ID by not letting users access this area without the required info

apps/mobile/src/app/(tabs)/(index)/index.tsx
```diff
@@ -1,77 +1,11 @@
...
-                  <Collectibles collectibles={collectibles} mode="widget" />
-                </CollectiblesWidget>
-              )}
-            </>
-          )}
-        </Box>
-      </Screen.ScrollView>
-      <NotificationsSheet sheetRef={notificationSheetRef} />
-    </Screen>
+    <CurrentAccountLoader fallback={<HomeScreenWithoutAccount />}>
```
pete-watters, line 7: This works well but we should think of a better way / naming. This case of  `HomeScreenWithoutAccount` is only really when there is a new / empty wallet and quite a basic screen. 

I think a combined home screen with an early return if `!account` would work as well.

apps/mobile/src/app/_layout.tsx
```diff
@@ -76,11 +79,16 @@ function App() {
         <Stack screenOptions={{ headerShown: false }}>
           <Stack.Screen name="(tabs)" />
         </Stack>
-        <SendSheet />
-        <ReceiveSheet />
+        {currentAccount && (
```
pete-watters, line 82: I think this could be clearer, so `currentAccount` is mandatory for `Send` / `Receive` but it's more that they have to have a non-empty wallet to do that. 

Would it work if we moved these and related sheets to `HomeScreenWithAccount` / `HomeScreen` ?

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
```diff
@@ -79,12 +79,6 @@ function ConfigureAccount({ fingerprint, accountIndex, account }: ConfigureAccou
                   iconTestID={defaultIconTestId(account.icon)}
                   primaryTitle={account.name}
                   caption={wallet.name}
-                  onPress={() => {
```
Comment (pete-watters) [inline apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:82]: This is removing the ability to jump to that account on the `Home` screen from settings. Is that intended? 

Is there a way we can do that with the new setup? (Go to Home of a selected account)? 


https://github.com/user-attachments/assets/e9290832-f701-49d4-9135-2a3283a23695

apps/mobile/src/core/current-account-provider.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export function useCurrentAccount() {
+  const context = useContext(CurrentAccountContext);
+  if (!context) throw new Error('`useCurrentAccount` must be used within `CurrentAccountProvider`');
+  return context;
+}
+
+export function CurrentAccountProvider({ children }: HasChildren) {
+  const [currentAccount, setCurrentAccount] = useState<Account | null>(null);
+  const { list } = useAccounts();
```
pete-watters: Just checking, does this need to be active accounts? 

Or it needs to be aware of hidden account also

apps/mobile/src/features/collectibles/collectibles-serializer.tsx
pete-watters, line 11: Nice improvements here 👍

apps/mobile/src/features/send/components/recipient/use-shameful-account-helpers.ts
```diff
@@ -1,6 +1,7 @@
 import { useCallback, useMemo } from 'react';
```
pete-watters, line 1: We should rename this file as there's no need to carry shame around

apps/mobile/src/features/settings/description-sheet.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export interface DescriptionSheetInstance {
+  present(sheetData: DescriptionSheetData): void;
+  dismiss(): void;
+}
+
+export type DescriptionSheetRef = RefObject<DescriptionSheetInstance | null>;
+
+export function DescriptionSheet() {
```
pete-watters, line 19: What's this sheet for? General purpose information / announcements? Maybe it can have a better name?

I think it needs some styling also:
Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-08-20 at 07 04 54

apps/mobile/src/store/accounts/accounts.read.ts
```diff
@@ -75,3 +76,11 @@ export function useAccounts(status: AccountStatus = 'active') {
 export function useAccountByIndex(fingerprint: string, index: number) {
   return useSelector(selectAccountByIndex(fingerprint, index));
 }
+
+export function useAccountByAccoundId(accountId: string) {
```
pete-watters, line 80: Nice addition 👍

apps/mobile/src/features/activity/activity-list.tsx
```diff
@@ -12,14 +12,14 @@ interface ActivityFlashListProps {
   header: ReactNode;
 }
 
-export function ActivityFlashList({ data, header }: ActivityFlashListProps) {
+export function ActivityList({ data, header }: ActivityFlashListProps) {
```
pete-watters, line 15: I don't think we should use `useMemo` here or for `Collectibles` / `SIP-10`'s without good reason and having measured performance. I think this was changed in a previous PR. 

We are supposed to have a loading state here and this is bypassing `toFetchState` and just returning `data.value` if `data.state === 'success'`. We have other `data.state`s like `isLoading` and `error` and now we aren't using them at all. 

`useMemo` can be effective to improve rendering times but in these cases we are changing the data often so I question whether it's improving much. 

For example, here we are:
- caching the response of  the `useAccountActivity` hook
- that hook wraps `useAccountActivityQuery` in a to fetch state
- `useAccountActivityQuery` calls the `getAccountActivity` service
- the stale time is ` 1 * 5000`ms
- `getAccountActivity` combines `getBtcActivity` and `getStacksActivity`
- `getBtcActivity` fetches and combines BTC transactions with market data 
- market data includes price information so will get updated often 

I'm open to correction here but my understanding is that this data will change roughly every 5 seconds anyway as `activity` will become stale and re-fetch. The activity itself may not have changed but price information from `marketData` will likely have been changed and therefore this memo will be invalid. Is that right @alexp3y ? 

I am all for using `useMemo` to cache big not changing data but I think we need to really consider the use cases.

pete-watters:
This PR looks great, nice work @edgarkhanzadian ! I added some suggested improvements but they can be made in smaller follow ups. 

I need to block it temporarily as `Token details` has been broken and crashing in some cases (for `SIP-10`, sometimes for `BTC`). I'll branch off this branch and try and fix it quickly. It's a critical issue as it's non recoverable on `Home`. 

You can see it happening below:

https://github.com/user-attachments/assets/cc32213d-d4e4-4e71-8227-078b0443c6e4




Something else you can see in that video is with activity - the empty activity layout flashes before loading. We had loaders for this and other areas but they were removed by no longer using `FetchWrapper` in [this PR](https://github.com/leather-io/mono/pull/1545/files). We have `toFetchState` and we should use it to make a slicker experience for users showing loaders while data is being fetched and only showing `empty` as a last resort if at all.

pete-watters (changes requested):
Blocking this as token details is crashing. I'll try fix it and get it integrated here. 

For some reason on this branch, SIP-10 balances are not found with `useSip10BalanceByAssetId` returning the error `Sip10 Token Not Found` for all tokens. 

It's debatable to whether this should block this PR. It's a critical crash but for a flagged and non-released feature so we could try fix it after merging this

apps/mobile/src/features/receive/receive-sheet.tsx
```diff
@@ -32,19 +32,8 @@ export function ReceiveSheet() {
...
       </SheetNavigationContainer>
     </FullHeightSheet>
   );
 }
-
-function useInitialReceiveParams() {
-  const params = useGlobalSearchParams();
-  const accountId = isString(params.accountId) ? params.accountId : undefined;
-  const tokenId = isString(params.tokenId) ? params.tokenId : undefined;
-  const assetType = isDefined(params.assetType)
```
pete-watters, line 45: I was using this stuff to default to a selected asset in `TokenDetails`

apps/mobile/src/app/(tabs)/activity.tsx
edgarkhanzadian: I agree we can think about removing this part with design

apps/mobile/src/app/(tabs)/(index)/index.tsx
edgarkhanzadian: I need this because most of the hooks can't use nullable data, discussion in [thread](https://trustmachines.slack.com/archives/C05LAP952E7/p1755607009445409)

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
Comment (edgarkhanzadian) [reply apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:82]: I'm not sure we need the functionality here, let's discuss this with design

apps/mobile/src/core/current-account-provider.tsx
edgarkhanzadian: yea this could be an edge case. If user has 3 accounts and all of those are hidden, user would see an empty wallet screen.

Actually, if user's first account is hidden then they would also see an empty wallet screen, this needs fixing.

apps/mobile/src/features/settings/description-sheet.tsx
edgarkhanzadian: That's for when we click the tooltip button

apps/mobile/src/features/receive/receive-sheet.tsx
edgarkhanzadian: Ah crap! Looking into it! Nice catch!

apps/mobile/src/features/settings/description-sheet.tsx
edgarkhanzadian: This sheet. Actually, it's good thing you mentioned it, it seems like it got broken during a merge conflict

https://github.com/user-attachments/assets/a9433407-1060-4de8-ac92-c3bf315a89ab

apps/mobile/src/core/current-account-provider.tsx
edgarkhanzadian: I've made an issue for us to discuss hidden account handling later https://linear.app/leather-io/issue/LEA-3136/hidden-accounts-handling

apps/mobile/src/features/activity/activity-list.tsx
edgarkhanzadian: Sounds good, i've opened a linear issue for me to review the performance and handle error state after we merge this PR https://linear.app/leather-io/issue/LEA-3137/flashlists-handling

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
Comment (edgarkhanzadian) [reply apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:82]: https://linear.app/leather-io/issue/LEA-3138/account-press-handling

tigranpetrossian (approved):

apps/mobile/src/app/(tabs)/(index)/[assetId]/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  assertExistence(currentAccount, 'Current account is required for AccountTokenScreen');
+
+  switch (assetId) {
+    case 'BTC':
+      return <BitcoinTokenDetailsByAccount account={currentAccount} />;
+    case 'STX':
+      return <StacksTokenDetailsByAccount account={currentAccount} />;
+    default:
+      return <Sip10TokenDetailsByAccount account={currentAccount} assetId={assetId} />;
```
Comment (pete-watters) [inline apps/mobile/src/app/(tabs)/(index)/[assetId]/index.tsx:27]: Thanks for renaming this to `assetId` 👍  It was on my TODOs after seeing all the changes come in

apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx
Comment (pete-watters) [reply apps/mobile/src/app/settings/wallet/configure/[wallet]/[account]/index.tsx:82]: OK 👍 It was required previously and it's weird it can be clicked but no action follows

apps/mobile/src/features/balances/assets/assets-list.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const runesMemo = useMemo(() => {
+    if (runesData.state === 'success' && runesFlag) return runesData.value.runes;
+    return [];
+  }, [runesData, runesFlag]);
+
+  // TODO: Unnecessary memoization
+  const allAssetsMemo = useMemo(() => [...sip10Memo, ...runesMemo], [sip10Memo, runesMemo]);
+
+  return (
+    // TODO: Redundant fragment
```
pete-watters: 🫨  couple of TODO's

apps/mobile/src/features/balances/assets/assets-list.tsx
```diff
@@ -0,0 +1,60 @@
...
+        renderItem={({ item }) =>
+          renderAsset({
+            item,
+            onPress: () => {
+              if (item.asset.protocol === 'sip10') onPressToken?.(item.asset.assetId);
+            },
+          })
+        }
+        getItemType={item => item.asset.protocol}
+        showsVerticalScrollIndicator={false} // TODO: Reason?
```
pete-watters: Maybe we don't need this, lets try removing. I think I added it to stop the scroll bar showing but that was before we moved the top content to `ListHeader` so it could be OK to remove

apps/mobile/src/features/receive/receive-sheet.tsx
pete-watters: All fixed now, thanks!

apps/mobile/src/features/receive/use-select-assets.ts
```diff
@@ -3,37 +3,41 @@ import { Account } from '@/store/accounts/accounts';
...
-  if (selectedAccount && tokenId) {
-    const tokenAssetSymbol = tokenId !== 'BTC' && tokenId !== 'STX' ? 'STX' : tokenId;
-
-    // assetType is passed to receive sheet in specific cases e.g. taproot or native segwit
-    if (assetType) {
-      return assets.filter(asset => asset.type === assetType);
-    }
-
-    return assets.filter(asset => asset.symbol === tokenAssetSymbol);
+  switch (receiveType) {
```
pete-watters, line 28: Nice improvement here 😎  Thanks!

apps/mobile/src/features/balances/assets/assets-list.tsx
edgarkhanzadian: Addressed those!

apps/mobile/src/features/balances/assets/assets-list.tsx
edgarkhanzadian: Removed!

apps/mobile/src/features/token/bitcoin/bitcoin-token-details.tsx
```diff
@@ -22,52 +16,16 @@ export function BitcoinTokenBalance(props: BitcoinTokenBalanceProps) {
...
-}: BitcoinTokenDetailsByAccountProps) {
+export function BitcoinTokenDetailsByAccount({ account }: BitcoinTokenDetailsByAccountProps) {
+  const { fingerprint, accountIndex } = account;
   const { value } = useBtcAccountBalance(fingerprint, accountIndex);
-  const account = useAccountByIndex(fingerprint, accountIndex);
   const activity = useAccountActivityByAsset(fingerprint, accountIndex, btcAsset);
   const availableBalance = value?.btc.availableBalance;
   const quoteBalance = value?.quote.availableBalance;
-  if (!availableBalance || !quoteBalance || !account) {
+  if (!availableBalance || !quoteBalance) {
```
pete-watters, line 28: One of the problems I was seeing with TokenDetails that's my own fault 😰  

In some cases it can be slow to get the balance and then it returns `null` here. 

I'll add better loaders so that doesn't happen. I opened a task to follow up here - [LEA-3140](https://linear.app/leather-io/issue/LEA-3140/loading-states-btc-token-details-crashing)

apps/mobile/src/features/token/token.tsx
```diff
@@ -39,11 +45,21 @@ export function Token({
   quoteBalance,
   tokenId,
 }: TokenProps) {
+  const { sendSheetRef, receiveSheetRef } = useGlobalSheets();
   const { data: assetDescription } = useAssetDescriptionQuery(asset);
   const marketData = useMarketDataQuery(asset);
   const price = marketData.data?.price;
   const { data: assetPriceChange } = useAssetPriceChangeQuery(asset);
   const changePercent = assetPriceChange?.changePercent ?? 0;
+  const receiveType = protocolMatch<ReceiveType>(asset.protocol, {
```
pete-watters, line 54: This should probably be in `receive`? Lets move it later

pete-watters (approved):
I retested this and it's working well, nice job @edgarkhanzadian ! 

- SIP-10 token details is working
- Send + receive  from token details is working too

`BTC` token details is crashing intermittently as sometimes it doesn't get the balance and returns `null`. I can look into improving that in a follow up and add loaders.

We also should prioritise fixing the loading states for other areas like `activity`. 

Overall this is a great PR and a huge and welcome change to the app, great work 👏

PR #1560, opened 2025-08-17
Refactor the mobile Sheet, closes LEA-2086

tigranpetrossian:
This re-visits the native Sheet, fixing some long standing visual issues, better sensible defaults based on existing usages, and conforming to the [Component architecture](https://www.notion.so/trustmachines/Component-architecture-196d74b694c780c29c76f9c8d36c8309?source=copy_link)

**New API**

```tsx
<Sheet>
  <Sheet.ScrollView> // Sheet.View, Sheet.FlashList, Sheet.SectionList, etc.
   ...
  <Sheet.ScrollView>
</Sheet>
```

### Changes

* Lot's of minor tweaks bringing all sheet usages closer to original designs.
* Sheet subcomponents now support our style props, e.g., `<Sheet.View px="5" gap="3" />
* passing `themeVariant` is no longer necessary
* Two handle positions: `inside` and `outside`
* Based on existing use cases, Sheet.Header uses similar implementation to Screen.Header supporting left, center, and right elements.
* `SheetRef` has been renamed to `SheetInstance`, as it's not technically a ref, but a value used in refs.
* Exposed new `SheetRef` alias for `RefObject<SheetInstance>`


apps/mobile/src/app/(tabs)/(index)/index.tsx
```diff
@@ -20,11 +20,11 @@ import { useWallets } from '@/store/wallets/wallets.read';
 import { t } from '@lingui/core/macro';
 import { router } from 'expo-router';
 
-import { Box, LeatherLogomarkIcon, SheetRef } from '@leather.io/ui/native';
+import { Box, LeatherLogomarkIcon, SheetInstance } from '@leather.io/ui/native';
```
pete-watters, line 23: This is a great renaming

apps/mobile/src/features/wallet-manager/add-wallet/add-wallet-sheet.tsx
```diff
@@ -21,15 +21,13 @@ export function AddWalletSheet() {
...
-        // TODO: this should be set when we call ref.current.present
-        opensFully={false}
-      />
-    </>
+    <AddWalletSheetLayout
+      createWallet={createWallet}
+      restoreWallet={restoreWallet}
+      addWalletSheetRef={addWalletSheetRef}
+      themeVariant={themeDerivedFromThemePreference}
+      // TODO: this should be set when we call ref.current.present
```
pete-watters, line 26: Maybe we can ditch this comment now? Or do it? 🤔

packages/ui/src/components/sheet/README.md
```diff
@@ -0,0 +1,116 @@
+# Sheet
```
pete-watters, line 1: Thanks for adding this 👍

packages/ui/src/components/sheet/sheet.native.tsx
```diff
@@ -1,150 +1,92 @@
...
-    fullHeightWithoutNotch: [height - top - theme.spacing['5']],
-    none: undefined,
-  });
-  const handleComponentTop = snapPointVariantMatcher(snapPointVariant, {
-    fullHeightWithNotch: top,
-    fullHeightWithoutNotch: -theme.spacing['4'],
-    none: 0,
-  });
-
-  useEffect(() => {
```
tigranpetrossian, line 71: This needs to come back, but in a way that doesn't trigger warnings.

packages/ui/src/components/sheet/sheet-containers.native.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export function SheetSectionList({
+  style,
+  ...props
+}: ComponentProps<typeof BottomSheetSectionList>) {
+  const borderRadius = useSheetBorderRadius();
+  return <BottomSheetSectionList style={[borderRadius, style]} {...props} />;
+}
+
+export function SheetVirtualizedList({
```
pete-watters, line 47: Wow, so if I understand the sheet library has `BottomSheetVirtualizedList` and `BottomSheetFlashList` we weren't using? 

Whats the key difference/ use case you recommend for each? `FlashList` for data heavy?

pete-watters (approved):
Great work @tigranpetrossian, thanks for doing this. The extra attention to small details makes a big difference

packages/ui/src/components/sheet/sheet-containers.native.tsx
tigranpetrossian: @pete-watters Yes , those are gesture-aware versions of the same scrollables. VirtualizedList is pretty low-level, but FlashList does come in handy. We also use the BottomSheetSectionList in send's recipient flow.

kyranjamie (approved):
Nice refactor @tigranpetrossian

apps/mobile/src/features/wallet-manager/add-wallet/add-wallet-sheet.tsx
tigranpetrossian: I don't know what the comment means tbh, so I left it as is. Don't think it's related to any sort of sheet limitations that were resolved.

tigranpetrossian:
Need to bring back the "Back button" behavior, will also wait for #1558 to be merged not to wreck @edgarkhanzadian's PR.

edgarkhanzadian:
@tigranpetrossian that's alright, i'd rather have all of the big stuff in before changing the whole app structure with single account page pr

tigranpetrossian:
@edgarkhanzadian 👌 Merging this to unblock yours. Will add the BackHandler in a separate PR.

PR #1561, opened 2025-08-18
Update account icon and cell spacings in token details

tigranpetrossian:
@pete-watters This is a small followup to your #1530, mostly touching pressed states of list items and the account icon.
I fixed the account icon typings across the slice and usages, it should be easier to work with going forward.


pete-watters (approved):
Thanks for helping here @tigranpetrossian

PR #1562, opened 2025-08-18
fix(mobile): fix squeezed buttons in the earn card

tigranpetrossian:
image


pete-watters (approved):
Thanks @tigranpetrossian

kyranjamie (approved):

PR #1564, opened 2025-08-18
feat: add service calls by asset id, ref LEA-3125

alexp3y:
Adds some additional asset-specific methods to activity & balance services in order to call them by assetId (SIP10) or runeName.

Replaces some service usages in `mobile` that previously filtered full account list responses w/ the more targeted asset-specific calls.

@pete-watters covers most of the LEA-3125 TODOs. There's 1 left that _i think_ needs an additional `protocol` route param before the asset-specific call can be added.

@edgarkhanzadian I reworked the SIP10 loader used in the send flow to clean-up some data fetching. Might wanna take a look there and make sure everything is stable?

### DIFF
--- a/apps/mobile/src/features/send/forms/stx/sip10-loader.tsx
+++ b/apps/mobile/src/features/send/forms/stx/sip10-loader.tsx
@@ -1,92 +1,58 @@
 import { Error } from '@/components/error/error';
-import { type FetchState, toFetchState } from '@/components/loading/fetch-state';
 import { SendFormLoadingSpinner } from '@/features/send/components/send-form-layout';
-import { useSip10AccountBalance } from '@/queries/balance/sip10-balance.query';
+import { useSip10BalanceByAssetId } from '@/queries/balance/sip10-balance.query';
 import '@/queries/balance/stx-balance.query';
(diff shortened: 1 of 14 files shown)


apps/mobile/src/features/send/forms/stx/sip10-loader.tsx
```diff
@@ -1,92 +1,58 @@
...
+    useStacksSignerAddressFromAccountIndex(account.fingerprint, account.accountIndex) ?? '';
+  const queryClient = useQueryClient();
+  const sip10Balance = useSip10BalanceByAssetId(
+    account.fingerprint,
+    account.accountIndex,
+    asset.assetId
   );
-  if (!token) return null;
+  const marketData = useMarketDataQuery(asset);
+  const nextNonce = useNextNonce(address);
```
edgarkhanzadian, line 35: We should try to avoid empty strings and use null/undefined instead of it. Empty strings only trick the compiler into thinking that there is a value even tho there is none.

Fix can be done later as it is a systemic issue throughout the codebase, but we need to take a look into it

edgarkhanzadian (approved):
Looks great! Let's merge this asap as these service calls are going to be helpful in the single account PR

PR #1567, opened 2025-08-19
chore(mobile): refactor token details to add asset protocol to route

pete-watters:
This PR adds some improvements to token details to make it more scalable

✨ Features:

- [x] Adds `assetProtocol` to the routes for more scalable implementation 
- [x] Refactors `Asset` interface to `ReceivableAsset`

### DIFF
--- a/apps/mobile/src/app/(tabs)/(index)/account/[accountId]/token/[assetProtocol]/[tokenId]/index.tsx
+++ b/apps/mobile/src/app/(tabs)/(index)/account/[accountId]/token/[assetProtocol]/[tokenId]/index.tsx
@@ -5,25 +5,28 @@ import { deserializeAccountId } from '@/store/accounts/accounts';
 import { useLocalSearchParams } from 'expo-router';
 import { z } from 'zod';
 
+import { CryptoAssetProtocols } from '@leather.io/models';
+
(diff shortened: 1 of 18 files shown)


apps/mobile/src/app/(tabs)/(index)/account/[accountId]/token/[assetProtocol]/[tokenId]/index.tsx
```diff
@@ -5,31 +5,36 @@ import { deserializeAccountId } from '@/store/accounts/accounts';
...
-    default:
+    case CryptoAssetProtocols.sip10:
       return (
         <Sip10TokenDetailsByAccount
           accountIndex={accountIndex}
           fingerprint={fingerprint}
           tokenId={tokenId}
         />
       );
+    default:
```
Comment (pete-watters) [inline apps/mobile/src/app/(tabs)/(index)/account/[accountId]/token/[assetProtocol]/[tokenId]/index.tsx:37]: @tigranpetrossian : can you please help me understand what's best to do in this case? 

- We can't reach this screen without selecting an asset
- We shouldn't have a fallback here or an empty screen 
    - I can throw this error to satisfy the compiler
    - below I instead use `assertUnreachable(assetProtocol as never)` which is also incorrect

pete-watters:
I'll re-do this in the context of single account setup

PR #1568, opened 2025-08-19
fix: broken stacks fee estimation

kyranjamie:
Contract call stacks fee must always be greater than 1µSTX * tx bytes

cc/ @314159265359879

### DIFF
--- a/packages/query/src/stacks/fees/fees.utils.spec.ts
+++ b/packages/query/src/stacks/fees/fees.utils.spec.ts
@@ -1,4 +1,12 @@
-import { getFeeEstimationsBasedOnDefaultMinMaxValues } from './fees.utils';
+import { PayloadType } from '@stacks/transactions';
+
+import { createMoney } from '@leather.io/utils';
+
+import {
+  enforceContractCallMinimumFees,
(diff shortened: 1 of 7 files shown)


camerow (approved):
Love how well tested this is! Nice one @kyranjamie

packages/query/src/stacks/fees/fees.utils.spec.ts
```diff
@@ -42,3 +50,210 @@ describe(getFeeEstimationsBasedOnDefaultMinMaxValues.name, () => {
     expect(highFeeResult).toEqual(expectedResult[2]);
   });
 });
+
+describe('Contract call fee validation', () => {
+  // https://github.com/leather-io/extension/issues/6251
```
pete-watters, line 55: Maybe you can just tag this issue in the commit?

pete-watters (approved):
Nice 🚀

edgarkhanzadian (approved):

packages/query/src/stacks/fees/fees.utils.spec.ts
kyranjamie: Happy to tag, but will leave comment

fbwoolf (approved):

PR #1572, opened 2025-08-21
chore: mobile release version

edgarkhanzadian:
Fix mobile release version

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -5,7 +5,7 @@ export default ({ config }: ConfigContext): ExpoConfig => {
     name: 'Leather',
     owner: 'leather-wallet',
     slug: 'leather-wallet-mobile',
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

apps/mobile/app.config.ts
```diff
@@ -5,7 +5,7 @@ export default ({ config }: ConfigContext): ExpoConfig => {
     name: 'Leather',
     owner: 'leather-wallet',
     slug: 'leather-wallet-mobile',
-    version: '2.66.0', // x-release-please-version
+    version: '2.66.1', // x-release-please-version
```
pete-watters, line 8: Is this supposed to be auto updated? 🤔

edgarkhanzadian: Yeap, not sure what went wrong here

PR #1575, opened 2025-08-21
chore(mobile): token details and single account refactor - LEA-3015

pete-watters:
This PR adds some improvements to token details, using `assetProtocol` in routes and doing some more cleanup post single account

✨ Features:

- [x] `assetProtocol` added to token details route
- [x] Removed un-needed files and added deprecation warnings for others
- [x] Removes `QueryPreloader` which was still pre-loading total account data 😱 

🤔 Future Enhancements:

-  I added deprecation warnings to a lot of the queries for **total** (multi account) data. We can probably just remove them entirely as we still have them in our services 
- We have other components called `BitcoinBalanceByAccount` etc. I can refactor them to remove `ByAccount`?


apps/mobile/src/features/activity/activity-list.tsx
```diff
@@ -2,13 +2,13 @@ import { ReactNode } from 'react';
 
 import { Screen } from '@/components/screen/screen';
 import { RefreshControl } from '@/features/refresh-control/refresh-control';
-import { useTotalActivity } from '@/queries/activity/account-activity.query';
+import { useAccountActivity } from '@/queries/activity/account-activity.query';
```
pete-watters, line 5: We only use `useAccountActivity` now

apps/mobile/src/features/balances/assets/render-assets.tsx
```diff
@@ -1,27 +1,42 @@
+import { TokenDetailsProps } from '@/features/token/types';
+
+import { CryptoAssetProtocols } from '@leather.io/models';
 import { RuneBalance, Sip10Balance } from '@leather.io/services';
 
 import { RunesTokenBalance } from '../bitcoin/runes-token-balance';
 import { Sip10TokenBalance } from '../stacks/sip10-token-balance';
 
+function isSip10Balance(item: Sip10Balance | RuneBalance): item is Sip10Balance {
```
pete-watters, line 9: I added these guards to avoid casting in the `switch` - `item as Sip10Balance`

apps/mobile/src/features/balances/bitcoin/bitcoin-balance.tsx
```diff
@@ -10,37 +12,11 @@ export function BitcoinTokenBalance(props: BitcoinTokenBalanceProps) {
   return <TokenBalance ticker="BTC" icon={<BtcAvatarIcon />} tokenName={t`Bitcoin`} {...props} />;
 }
 
-interface BitcoinBalanceProps {
-  onPress?: (tokenId: string) => void;
-}
-
-export function BitcoinBalance({ onPress }: BitcoinBalanceProps) {
```
pete-watters, line 17: I am removing any of these 'legacy' components now. 

We can probably just rename `BitcoinBalanceByAccount` as `BitcoinBalance`? It's kind of inferred based on the props now?

apps/mobile/src/features/receive/get-assets.ts
```diff
@@ -5,7 +5,7 @@ export enum AssetType {
   Taproot = 'taproot',
   Stacks = 'stacks',
 }
-export interface Asset {
+export interface ReceivableAsset {
```
pete-watters, line 8: As promised @alexp3y

apps/mobile/src/app/(tabs)/(index)/[assetProtocol]/[assetId]/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { useCurrentAccount } from '@/core/current-account-provider';
```
Comment (pete-watters) [inline apps/mobile/src/app/(tabs)/(index)/[assetProtocol]/[assetId]/index.tsx:1]: Moving this down the route hierarchy to make things more scalable in future, as suggested by @alexp3y 👏 

Now we can use the route to determine protocol and asset

apps/mobile/src/app/(tabs)/(index)/[assetProtocol]/[assetId]/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { BitcoinTokenDetails } from '@/features/token/bitcoin/bitcoin-token-details';
+import { Sip10TokenDetails } from '@/features/token/stacks/sip10-token-details';
+import { StacksTokenDetails } from '@/features/token/stacks/stacks-token-details';
+import { useLocalSearchParams } from 'expo-router';
+
+import { CryptoAssetProtocols } from '@leather.io/models';
+import { assertExistence, assertUnreachable } from '@leather.io/utils';
+
+type SupportedAssetProtocol = 'nativeBtc' | 'nativeStx' | 'sip10';
```
Comment (pete-watters) [inline apps/mobile/src/app/(tabs)/(index)/[assetProtocol]/[assetId]/index.tsx:10]: @tigranpetrossian helped me improve this. I was using `assertUnreachable(assetProtocol as never)` as the type was too broad here. 

I can look into extending `SupportedAssetProtocol` when adding Runes/ collectibles and move it to services

apps/mobile/src/queries/activity/account-activity.query.ts
```diff
@@ -6,6 +6,11 @@ import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
 import { AccountAddresses, CryptoAsset } from '@leather.io/models';
 import { getActivityService } from '@leather.io/services';
 
+/**
```
pete-watters, line 9: I added a lot of deprecations here as these aren't used. Maybe now is the time to remove? 

We still have the services for total everything but don't use the queries

apps/mobile/src/features/balances/bitcoin/bitcoin-balance.tsx
tigranpetrossian: I think this definitely can be the default thinking now, for things like hooks as well. In future we can name wallet-level or totals separately.

apps/mobile/src/queries/balance/btc-balance.query.ts
```diff
@@ -5,11 +5,18 @@ import { QueryFunctionContext, useQuery } from '@tanstack/react-query';
 
 import { AccountRequest, getBtcBalancesService } from '@leather.io/services';
 
+/**
+ * @deprecated useBtcTotalBalance is not used now we have moved to single account view
```
tigranpetrossian, line 9: On one hand these are still going to be useful at some point, on the other hand I see how this can be mistakenly used. I think it's worth rethinking the data hook design at some point in a way that solves both.

apps/mobile/src/app/_layout.tsx
```diff
@@ -105,23 +104,21 @@ function RootLayout() {
                 <SafeAreaProvider>
                   <QueryClientProvider client={queryClient}>
                     <LeatherQueryProvider>
-                      <QueryPreloader>
```
tigranpetrossian, line 108: Thanks, forgot this existed. I have a task somewhere for improving the app bootstrap overall, I can revisit this there if it turns out it's still worth preloading some things.

tigranpetrossian (approved):

apps/mobile/src/queries/balance/btc-balance.query.ts
pete-watters: For now, I think this is a good interim step to keep them around but make it clear they shouldn’t be used without consideration.

If we have a portfolio with multi account balances this will be useful

pete-watters:
Thanks @tigranpetrossian . I’ll merge this now and  make more naming improvements in the next one

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -71,12 +73,16 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
               <BitcoinBalanceByAccount
                 fingerprint={fingerprint}
                 accountIndex={accountIndex}
-                onPress={onPressToken}
+                onPress={() =>
+                  onPressToken?.({ assetProtocol: CryptoAssetProtocols.nativeBtc, assetId: 'BTC' })
+                }
```
edgarkhanzadian, line 78: Any chance we could infer assetProtocol from assetId?
In case of nativeStx and nativeBtc protocols, assetId seems quite redundant

edgarkhanzadian: or maybe we could have a type like this:
```
{
  assetProtocol: 'nativeBtc'
} | 
{
  assetProtocol: 'sip10';
  assetId: string
}
```

apps/mobile/src/queries/balance/total-balance.query.ts
```diff
@@ -22,7 +22,10 @@ export interface TotalBalance {
   runes: FetchState<RunesAggregateBalance>;
   totalBalance: FetchState<Money>;
 }
-
+/**
+ * @deprecated useTotalBalance is not used now we have moved to single account view
+ * @see useAccountBalance
+ */
```
edgarkhanzadian, line 28: This one is going to be used in accounts list view, we are going to show the total balance of all accounts at the top of the list

apps/mobile/src/components/home/home-with-account-screen.tsx
pete-watters: I'll check and see if I can 👍

apps/mobile/src/queries/balance/total-balance.query.ts
pete-watters: OK 👍  I'll leave it here until it's in use, then we can remove the deprecation warning. Is there a task open for that work? I can link it to this section

apps/mobile/src/components/home/home-with-account-screen.tsx
pete-watters: I tried to made `assetId` optional as it's not needed at all for BTC / STX but then I hit problems with the app crashing due to the routing not working. 

I'll think about how I can improve it but I think it's not so bad to keep it simpler but pass two redundant props

apps/mobile/src/queries/balance/total-balance.query.ts
pete-watters: Thinking about this now and I think we shouldn't do that. 

Loading the total balance can take quite a long time. So long that we were preloading it. 

We can try but I think it will take a long time for the account list to render if we do this

PR #1577, opened 2025-08-25
feat: create swap service

alexp3y:
Multi-provider swap service supporting arbitrary concurrent integrations to power a unified swap experience - i.e. `bitflow-sdk`, `velar-sdk`, `sbtc-bridge`, `alex-sdk`, etc.

Mobile swap queries are meant to be illustrative and can/should be changed to suit actual application usage.

This version shouldn't be considered release-ready but rather supporting further iterations after connecting to the new mobile UI feature allows for more / easier testing.


packages/models/src/assets/asset.model.ts
```diff
@@ -8,22 +8,30 @@ export const CryptoAssetCategories = {
   fungible: 'fungible',
   nft: 'nft',
 } as const;
-export const CryptoAssetProtocols = {
+export const FungibleCryptoAssetProtocols = {
```
tigranpetrossian, line 11: If this isn't going to carry more metadata, would it make sense to convert it to an lowercase array similar to `swapProviderIds`?

packages/models/src/swap/swap.model.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface SwapProviderAsset {
+  providerId: SwapProviderId;
+  providerAssetId: string;
+  assetId: CryptoAssetId;
+}
+
+export interface SwapQuote {
+  executionType: SwapExecutionType;
+  providerId: SwapProviderId;
+  providerQuoteData: any;
```
tigranpetrossian: For this one and the few fields under `StacksContractCallSwapExecutionData`, I'd use `unknown` vs. `any` to force consumers narrow it down before using.

packages/services/src/swap/swap.service.ts
```diff
@@ -0,0 +1,228 @@
...
+      )
+    );
+    const targetAssets = await Promise.all(providerServiceCalls);
+    return await this.combineProviderAssets(targetAssets.flat());
+  }
+
+  private async combineProviderAssets(
+    providerSwapAssets: SwapProviderAsset[]
+  ): Promise<SwapAsset[]> {
+    const providerSwapAssetMap = new Map<SerializedCryptoAssetId, SwapProviderAsset[]>();
```
tigranpetrossian: Ultimately not a big deal, but this kind of thing can be a fluent remeda pipeline: (pseudocode-ish)
```ts
function combineProviderAssets(providerSwapAssets: SwapProviderAsset[]): Promise<SwapAsset[]> {
  return pipe(
    providerSwapAssets,
    groupBy(asset => serializeAssetId(asset.assetId)),
    entries,
    map(([serializedId, providerAssets]) =>
      this.fungibleAssetService
        .getAsset(deserializeAssetId(serializedId) as FungibleAssetId)
        .then(asset => ({ asset, providerAssets })
        .catch(() => null)
    ),
    Promise.all,
    results => filter(results, isNonNullish))
  );
}
```

tigranpetrossian (approved):
Thanks @alexp3y, this is really a big deal. Multi-provider support and all the business logic coming from the service enables all kinds of possibilities for the UI.
Don't see any issues except for the `any` types being a potential footgun on the usage side.

apps/mobile/metro.config.cjs
```diff
@@ -49,6 +49,15 @@ config.resolver.resolveRequest = (context, moduleName, platform) => {
...
   }
+
+  // Block @stacks/connect-ui which contains webpack-specific dynamic imports
+  // that Metro can't handle (used by @bitflowlabs/core-sdk)
+  if (moduleName.startsWith('@stacks/connect-ui')) {
+    return {
+      type: 'empty',
+    };
+  }
+
```
edgarkhanzadian, line 60: I wonder why stacks sdks package ui web components with them. Would be grade to see it separated as @bitflowlabs/core-sdk and @bitflowlabs/web-sdk

apps/mobile/src/services/init-app-services.ts
```diff
@@ -8,6 +8,15 @@ export function initAppServices() {
...
       leatherApiUrl: process.env.EXPO_PUBLIC_LEATHER_API_URL,
+      bitflow: {
+        bitflowApiHost: process.env.EXPO_PUBLIC_BITFLOW_API_HOST ?? '',
+        bitflowApiKey: process.env.EXPO_PUBLIC_BITFLOW_API_KEY ?? '',
+        bitflowProviderAddress: process.env.EXPO_PUBLIC_BITFLOW_PROVIDER_ADDRESS ?? '',
+        readonlyCallApiHost: process.env.EXPO_PUBLIC_BITFLOW_READONLY_CALL_API_HOST ?? '',
+        readonlyCallApiKey: process.env.EXPO_PUBLIC_BITFLOW_READONLY_CALL_API_KEY ?? '',
+        keeperApiHost: process.env.EXPO_PUBLIC_BITFLOW_KEEPER_API_HOST ?? '',
+        keeperApiKey: process.env.EXPO_PUBLIC_BITFLOW_KEEPER_API_KEY ?? '',
+      },
```
edgarkhanzadian, line 19: should we consider some kind of a console warning in case these env variables are not set?

edgarkhanzadian (approved):
Great work!! 🚀

packages/services/src/infrastructure/api/alex/alex-sdk.client.ts
```diff
@@ -0,0 +1,60 @@
...
+import { AMMRoute } from 'alex-sdk/dist/utils/ammRouteResolver';
+import { inject, injectable } from 'inversify';
+
+import { Types } from '../../../inversify.types';
+import { HttpCacheService } from '../../cache/http-cache.service';
+
+export type AlexSdkTokenInfo = TokenInfo;
+export type AlexSdkCurrency = Currency;
+export type AlexSdkTxToBroadCast = TxToBroadCast;
+export type AlexSdkAMMRoute = AMMRoute;
```
kyranjamie, line 12:
```suggestion
export type AlexSdkAmmRoute = AmmRoute;
```

packages/services/src/infrastructure/api/alex/alex-sdk.client.ts
```diff
@@ -0,0 +1,60 @@
...
+import { TxToBroadCast } from 'alex-sdk/dist/helpers/SwapHelper';
+import { AMMRoute } from 'alex-sdk/dist/utils/ammRouteResolver';
+import { inject, injectable } from 'inversify';
+
+import { Types } from '../../../inversify.types';
+import { HttpCacheService } from '../../cache/http-cache.service';
+
+export type AlexSdkTokenInfo = TokenInfo;
+export type AlexSdkCurrency = Currency;
+export type AlexSdkTxToBroadCast = TxToBroadCast;
```
kyranjamie, line 11:
```suggestion
export type AlexSdkTxToBroadcast = TxToBroadcast;
```

packages/services/src/infrastructure/api/alex/alex-sdk.client.ts
kyranjamie: casing consistency

packages/services/src/infrastructure/api/leather/leather-api.client.ts
```diff
@@ -28,6 +28,8 @@ export type LeatherApiTokenPriceHistory =
 export type LeatherApiLocale = Required<
   NonNullable<paths['/v1/tokens/native/{symbol}/description']['get']['parameters']['query']>
 >['locale'];
+export type LeatherApiSwapDex =
+  paths['/v1/swap/dexes']['get']['responses'][200]['content']['application/json'][string];
```
kyranjamie, line 32: nice lol

packages/services/src/swap/swap.service.ts
kyranjamie: Different flavour of the same jazz, but I loved working on services like this streaming all the data together with observables.

packages/stacks/src/stacks.utils.ts
```diff
@@ -178,3 +178,14 @@ export function getStacksAssetStringParts(identifier: string): {
...
+export function isValidStacksPrincipal(principal: string | undefined | null): boolean {
+  if (
+    !principal ||
+    (!principal.startsWith('SP') && !principal.startsWith('SM') && !principal.startsWith('ST'))
+  ) {
+    return false;
+  }
+  const splits = principal.split('.');
+  return splits[0].length > 39 && splits[0].length < 42;
+}
```
kyranjamie: Repeated logic, this exists in `packages/stacks/src/validation/address-validation.ts`

packages/services/src/infrastructure/api/alex/alex-sdk.client.ts
alexp3y: `TxToBroadCast` comes from the Alex SDK. Doesn't match our casing rules but it preserves the convention of prefixing an external type to add context when exported/used outside of the direct client: `<client prefix><external type>`.

packages/services/src/infrastructure/api/alex/alex-sdk.client.ts
alexp3y: Same as [here](https://github.com/leather-io/mono/pull/1577#discussion_r2348703379).

kyranjamie (approved):

apps/web/vite.config.ts
```diff
@@ -33,6 +33,8 @@ export default defineConfig(({ command, mode, isSsrBuild }) => ({
       'leather-styles': path.resolve(__dirname, 'leather-styles'),
       axios: path.resolve(__dirname, 'node_modules/axios/dist/esm/axios.js'),
       'msw/node': path.resolve(__dirname, 'node_modules/msw/lib/node/index.js'),
+      // Stub out @bitflowlabs/core-sdk for browser compatibility
+      '@bitflowlabs/core-sdk': path.resolve(__dirname, 'stubs/bitflow-sdk-stub.js'),
```
alexp3y, line 37: @kyranjamie I added this to fix some failing `web` playwright tests due to Node dependencies in the Bitflow SDK. It's a bit of heavy-handed but it does the trick. We don't currently need this dependency in web.

But maybe you can take a look? Perhaps there's a more simple polyfill we could add; I spent a couple hours trying more targeted things but wasn't able to get it working.

kyranjamie: I think this is fine, the kind of thing you need to do with poorly authored packages

PR #1578, opened 2025-08-26
chore(mobile): refactor token details to add more granular loading st…

pete-watters:
This PR refactors `token-details` to add better loading states and fix a bug with BTC token details crashing if no balance is available.

✨ Features:

- [x] Fixes crash of token details when no BTC balance available
- [x] Adds loaders and empty balance/price fallbacks  

📱 Demo:

https://github.com/user-attachments/assets/a593416f-7486-4943-a8c8-207c6b51d49a


apps/mobile/src/components/loading/fetch-wrapper.tsx
```diff
@@ -1,28 +0,0 @@
-import { Error, FetchState, Loading } from '@/components/loading';
```
pete-watters, line 1: This was un-used post [this flashlist refactor PR](https://github.com/leather-io/mono/pull/1545). 

That PR also broke the activity loading states so it may return but removing for now as it's not used

apps/mobile/src/components/balance/price-change.tsx
```diff
@@ -14,21 +15,32 @@ function getPriceChangeColor(changePercent: number) {
...
+  price?: Money;
+  changePercent?: number;
+  isLoading: boolean;
+}
+export function PriceChange({ price, changePercent, isLoading }: PriceChange) {
+  if (isLoading) {
+    return <SkeletonLoader height={16} width={100} isLoading />;
+  }
+
+  if (!price || !changePercent) {
```
pete-watters, line 28: @alexp3y For some SIP-10 tokens like `WELSH` we don't seem to get price change data so I am changing the UI here to match how we handle un-known balances rather than show `$0`. 

Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-08-26 at 08 19 09

I think that's better than showing a price change of `0` which is incorrect

apps/mobile/src/features/token/types.ts
```diff
@@ -8,3 +14,61 @@ export interface TokenDetailsProps {
...
+
+export type SupportedAssetProtocol = 'nativeBtc' | 'nativeStx' | 'sip10';
+
+export type TokenBalance =
+  | AccountQuotedBtcBalance
+  | AddressQuotedStxBalance
+  | Sip10Balance
+  | RuneBalance;
+
+export function isAccountQuotedBtcBalance(
```
pete-watters, line 26: I added these type guards for balance type as it makes it cleaner to handle multiple types of protocol balances more easily via `getBalance`

apps/mobile/src/features/token/use-get-token-details.ts
```diff
@@ -0,0 +1,28 @@
+import { toFetchState } from '@/components/loading';
+import {
+  useAssetDescriptionQuery,
+  useAssetPriceChangeQuery,
+} from '@/queries/assets/fungible-asset-info.query';
+import { useMarketDataQuery } from '@/queries/market-data/market-data.query';
+
+import { FungibleCryptoAsset } from '@leather.io/models';
+
+export function useGetTokenDetails({ asset }: { asset: FungibleCryptoAsset }) {
```
pete-watters, line 10: This hook groups the main token-details queries together and adds `fetchState` to help with more elegant loading. 

For SIP-10 - the `asset` needs to come from the `balance` query which sometimes hasn't completed when the user lands on the page.

apps/mobile/src/features/token/components/token-loading.tsx
```diff
@@ -0,0 +1,52 @@
+import { Screen } from '@/components/screen/screen';
```
pete-watters, line 1: For SIP-10 - the `asset` needed to fetch other data relies on `balance` being loaded so it can be a bit jumpy. It's strange as we should already have it cached from the home screen.

I added this improvised loading screen for SIP-10's to make it more smooth, I will check it with design but it's an improvement for now on a blank screen and jumpy content:

Simulator Screenshot - iPhone 16 iOS 18 4 - 2025-08-26 at 11 50 45

fbwoolf (approved):

pete-watters:
Thanks @fbwoolf ! I'll miss you 😢

PR #1579, opened 2025-08-26
feat: add asset id model and utils, ref LEA-3125

alexp3y:
Adds a new `CryptoAssetId` type to complement the existing Asset model. 

The goal of this type is to flatten and uniquely identify any asset object in the `CryptoAsset` type heirarchy, where ID fields are heterogenous. The associated `getAssetId(asset)` utils ensure that ID derivations are centrally defined and deterministic.


packages/utils/src/assets/asset-id.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export function buildSip9AssetId(asset: Sip9Asset): CryptoAssetId {
+  return {
+    protocol: asset.protocol,
+    id: `${asset.assetId}|${asset.tokenId}`,
+  };
+}
+
+export function readSip9IdFields(assetId: CryptoAssetId): {
```
tigranpetrossian: There would be a bit of an overhead with more types, but would be worth having `readSip9IdFields` require a narrower type instead of accepting a broader one and throwing an error? The name implies it expects Sip9, is there a case where an arbitrary assetId would be passed into this?

tigranpetrossian (approved):
Thanks @alexp3y,
I'd love us to have more resilient helpers that don't cause exceptions, but that can also be part of the `utils` refactor that I want to get to some time in this century.

packages/utils/src/assets/asset-id.ts
alexp3y: Hm, I just tried this out and tbh it didn't feel quite right.. I think the responsibility of the type starts to get lost as a composite. Not saying this isn't something we might want to do eventually, but I'd prefer to keep it simple for the time being. I removed the `readSip9IdFields`, which was probably my mistake to include. We shouldn't be tempted into trying to read additional information out of the ID over just using it as an ID.

PR #1580, opened 2025-08-26
feat(mobile): check in Google Services config files

camerow:
## Summary
- Check in `google-services.json` and `GoogleService-Info.plist` files to the repository
- Remove Firebase authentication steps from local development scripts (`pnpm ios` and `pnpm android`)
- Add `firebase:refresh` command for manual Firebase config updates when needed

## Test plan
- [x] Run `pnpm ios` - should build without fetching Firebase config
- [ ] Run `pnpm android` - should build without fetching Firebase config
- [x] Run `pnpm firebase:refresh` - should authenticate and refresh both config files


pete-watters (approved):
Nice work! The code LGTM, I’ll test it out soon

PR #1583, opened 2025-08-27
feat: flatten object util

kyranjamie:
A small utility to turn a nested array or object into a flattened object


tigranpetrossian (approved):

alexp3y (approved):
Pretty cool. Any specific use-cases in mind? The reverse of restoring objects from flattened would also be useful. Perhaps there are some existing packages out there already do this?

kyranjamie:
I vibed this, if I'm honest 😅  I feel like that's more appropriate for isolated, tested functions. My thinking was when we want to track lots of data in analytics, which doesn't support nesting.

PR #1584, opened 2025-08-27
feat(mobile): SIP-10 sends from token details

pete-watters:
This PR integrates SIP-10 sends to Token details.

✨ **Features**:

- [x] Allows users to send SIP-10 tokens from token details
- [x] Fixes a bug introduced the was breaking send of BTC / STX from token details

📱 **Demo**:

https://github.com/user-attachments/assets/e768d43b-b1ba-4b1e-80bf-de292644e199

📓 **Note**:

This PR is branched off [chore/LEA-3140/token-details-loading-states](https://github.com/leather-io/mono/tree/chore/LEA-3140/token-details-loading-states) which will need to be merged first

### DIFF
--- a/apps/mobile/src/features/send/send-sheet.tsx
+++ b/apps/mobile/src/features/send/send-sheet.tsx
@@ -1,21 +1,28 @@
+import { RefObject, useImperativeHandle, useRef, useState } from 'react';
+
 import { FullHeightSheet } from '@/components/sheets/full-height-sheet/full-height-sheet';
 import { useCurrentAccount } from '@/core/current-account-provider';
 import { useGlobalSheets } from '@/core/global-sheet-provider';
+import { SheetNavigationContainer } from '@/core/sheet-navigation-container';
 import { Send } from '@/features/send/send';
 import { analytics } from '@/utils/analytics';
-import { useGlobalSearchParams } from 'expo-router';
-import { isString } from 'remeda';
 
-import { btcAsset, stxAsset } from '@leather.io/constants';
 import { FungibleCryptoAsset } from '@leather.io/models';
-import { useHaptics } from '@leather.io/ui/native';
+import { SheetInstance, useHaptics } from '@leather.io/ui/native';
 import { assertExistence } from '@leather.io/utils';
+
+export interface SendSheetInstance {
+  present(asset?: FungibleCryptoAsset): void;
+  dismiss(): void;
+}
+export type SendSheetRef = RefObject<SendSheetInstance | null>;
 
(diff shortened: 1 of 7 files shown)


apps/mobile/src/features/send/send-sheet.tsx
```diff
@@ -28,23 +35,25 @@ export function SendSheet() {
...
     >
-      <Send currentAccount={currentAccount} asset={asset} />
+      <SheetNavigationContainer base="send">
+        <Send currentAccount={currentAccount} asset={asset} />
+      </SheetNavigationContainer>
     </FullHeightSheet>
   );
 }
-
-function useInitialSendParams() {
```
pete-watters, line 42: We already have access to the `asset` when the user is clicking `Send` from token details so it's easier to decouple this sheet from the routing and use `useImperativeHandle` similar to the receive sheet.

packages/services/src/balances/sip10-balances.service.ts
```diff
@@ -38,7 +38,7 @@ export class Sip10BalancesService {
     @inject(Types.SettingsService) private readonly settingsService: SettingsService,
     private readonly stacksApiClient: HiroStacksApiClient,
     private readonly marketDataService: MarketDataService,
-    private readonly sip10TokensService: Sip10AssetService
+    private readonly sip10AssetsService: Sip10AssetService
```
pete-watters, line 41: @alexp3y initially I thought I would need to extend this service to give me `asset` based on `tokenId` / `assetId`. 

I didn't need to do that but I kept this renaming, I think its a bit clearer if we just go with `asset` and banish `token`? 🤔

apps/mobile/src/features/token/token.tsx
```diff
@@ -86,7 +86,7 @@ export function Token({
               }
               actionButtons={
                 <ActionButtons
-                  onSend={() => sendSheetRef.current?.present()}
+                  onSend={() => asset && sendSheetRef.current?.present(asset)}
                   onReceive={() => receiveType && receiveSheetRef.current?.present(receiveType)}
                   canSend={canSend}
```
pete-watters, line 91: Right now `canSend` is not used but I will work on adding `TokenDetails` for **Runes** soon and will use it for that

kyranjamie (approved):
LGTM. Guess most the work is in the other branch?

pete-watters:
> LGTM. Guess most the work is in the other branch?

Yes this is just a simpler addition but I had to do more extensive refactoring in other areas in the other branch

PR #1585, opened 2025-08-27
docs: update mobile documentation and sample env files

camerow:


tigranpetrossian (approved):

apps/mobile/.env.development
```diff
@@ -1 +1,7 @@
 EXPO_PUBLIC_NODE_ENV="development"
+EXPO_PUBLIC_LAUNCH_DARKLY=""
```
pete-watters, line 2: @camerow : I think that adding `EXPO_PUBLIC_LAUNCH_DARKLY=""` here breaks the mobile build as it over-writes what you have in `.env` with an empty string.

I'm not sure what exactly happens but if I don't comment this out the build fails with this error.  

simulator_screenshot_D5FDB9F6-742A-4AB6-8416-B19B85BA2E37

It seems to load the `.env` files in order:
```
env: load .env.development .env
```

If I comment out this line I get all the variables:
```
env: export EXPO_PUBLIC_NODE_ENV EXPO_PUBLIC_LAUNCH_DARKLY EXPO_PUBLIC_ENVIRONMENT EXPO_PUBLIC_MAESTRO_CI
```

If I don't I just get:
```
env: export EXPO_PUBLIC_NODE_ENV EXPO_PUBLIC_ENVIRONMENT EXPO_PUBLIC_MAESTRO_CI
```

PR #1588, opened 2025-08-27
fix(mobile): full width balances warning

camerow:
Before: 
Screenshot 2025-08-27 at 11 11 32 AM

After:
Screenshot 2025-08-27 at 11 08 31 AM

### DIFF
--- a/apps/mobile/src/components/home/home-with-account-screen.tsx
+++ b/apps/mobile/src/components/home/home-with-account-screen.tsx
@@ -1,5 +1,6 @@
 import { useRef, useState } from 'react';
 
+import { FetchErrorCallout } from '@/components/error/fetch-error';
 import { Screen } from '@/components/screen/screen';
 import { useCurrentAccount } from '@/core/current-account-provider';
 import { AccountDetails } from '@/features/account/account-details';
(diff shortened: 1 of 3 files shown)


apps/mobile/src/features/account/components/account-total-balance.tsx
```diff
@@ -28,12 +27,11 @@ export function AccountTotalBalance({ account }: AccountTotalBalanceProps) {
     accountIndex: account.accountIndex,
   });
   const isLoadingTotalBalance = totalBalance.state === 'loading';
-  const isErrorTotalBalance = totalBalance.state === 'error';
+
```
pete-watters, line 30: Maybe we should just pass down `totalBalance` now? I'm not sure if it's better to call the `useAccountBalance` hook more often or else pass props down

pete-watters (approved):
Thanks for fixing this @camerow. 

We have a lot more cleanup to do with loading states as [this PR](https://github.com/leather-io/mono/pull/1545) was merged too hastily and broke a lot of the loading states we had for little good reason. We need to do better there as even tests won't stop that happening

apps/mobile/src/features/account/components/account-total-balance.tsx
camerow: I always prefer getting state local to where it's needed rather than prop drilling unless perf is actually a concern.

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -58,10 +60,17 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
   }
   const tokenDetailsFlag = useTokenDetailsFlag();
   const onPressToken = tokenDetailsFlag ? onOpenToken : undefined;
+  const { totalBalance } = useAccountBalance({
```
pete-watters, line 63: @camerow we should have tested this better but there are some issues here. 

Now this is shown anytime you add a new account / wallet as sometimes we get an error and then recover / false alarm. 

The alignment of the callout is also off https://linear.app/leather-io/issue/LEA-3168/qa-fresh-install-no-balances-loaded

PR #1589, opened 2025-08-28
Fix/mobile-local-dev

camerow:
- [x] Remove `updates` from app config until we need to use them and find a fix for local builds
- [x] Add utility script `pnpm prebuild` command for generating fresh native directories
- [x] Remove `.env.development` from version control


pete-watters (approved):

camerow:
cc @alexp3y and @pete-watters I've successfully run Android / iOS with these changes in place.

PR #1591, opened 2025-08-29
feat(mobile): rune token details, ref LEA-3140

pete-watters:
This PR allows users to view token details for Runes. It will need design review / guidance but it's feature flagged OFF along with other Rune balances. 

✨ Features:

- [x] Connects Rune balances to token details view

🤔 Future Enhancements:

- Add Rune activity feed per account - get activity by `runeName`
- Fix issues with asset description for Runes not loading
- Refactor to stop using the plural `Runes` in some cases and `Rune` elsewhere

https://github.com/user-attachments/assets/003ab3ab-b9d3-4071-9466-2ed8a3e3ace0

### DIFF
--- a/apps/mobile/src/features/balances/assets/render-assets.tsx
+++ b/apps/mobile/src/features/balances/assets/render-assets.tsx
@@ -36,7 +36,18 @@ export function renderAsset({
     );
   }
   if (isRuneBalance(item)) {
-    return <RunesTokenBalance key={item.asset.symbol} item={item} />;
+    return (
+      <RunesTokenBalance
+        key={item.asset.symbol}
+        item={item}
+        onPress={() =>
+          onPress?.({
+            assetId: item.asset.runeName,
+            assetProtocol: item.asset.protocol,
+          })
+        }
+      />
+    );
   }
   return null;
 }
--- a/apps/mobile/src/queries/balance/runes-balance.query.ts
+++ b/apps/mobile/src/queries/balance/runes-balance.query.ts
@@ -15,6 +15,21 @@ export function useRunesTotalBalance() {
   return toFetchState(useRunesAggregateBalanceQuery(accounts.map(account => ({ account }))));
 }
 
+export function useRuneBalanceByRuneName(
+  fingerprint: string,
+  accountIndex: number,
+  runeName: string
+) {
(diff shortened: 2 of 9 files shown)


apps/mobile/src/queries/balance/runes-balance.query.ts
```diff
@@ -15,6 +15,21 @@ export function useRunesTotalBalance() {
   return toFetchState(useRunesAggregateBalanceQuery(accounts.map(account => ({ account }))));
 }
 
+export function useRuneBalanceByRuneName(
+  fingerprint: string,
+  accountIndex: number,
+  runeName: string
+) {
+  const runesBalances = useRunesAccountBalance(fingerprint, accountIndex);
```
pete-watters: @alexp3y I added this for now but I just found that you already have `getRuneBalanceByRuneName` in services so I will change it. I was about to ask if I should add it 🫨  

I don't see any activity service for Runes so I have left it out for now. 

I am also not getting any description data returned by when I send the `asset`

apps/mobile/src/features/balances/assets/render-assets.tsx
```diff
@@ -36,7 +36,18 @@ export function renderAsset({
     );
   }
   if (isRuneBalance(item)) {
-    return <RunesTokenBalance key={item.asset.symbol} item={item} />;
+    return (
+      <RunesTokenBalance
```
pete-watters, line 40: We have a mixture of `<Runes...` vs `<Rune ` in the app. I think we should change it to singular. 

In this specific case we can also get rid of `Token` in the name so we would have:
- `RuneBalance`
- `SIP10Balance` etc

apps/mobile/src/queries/balance/runes-balance.query.ts
alexp3y: Yep, we don't have activity support for Runes yet. Infra for Rune descriptions is all there, just need to draft and upload descriptions for the top tokens.

alexp3y (approved):

apps/mobile/src/queries/balance/runes-balance.query.ts
pete-watters: OK great. The way it's setup is that once response start to be recieved, it will just start working then

PR #1592, opened 2025-09-01
feat: add bns types and app-layer service

alexp3y:
Add two new BNS model types (`BnsName`, `BnsProfile`) retrievable via app-layer service calls (`BnsService`). 

By adding a `callReadOnlyFunction()` method to `HiroStacksApiClient`, we can now fetch the "primary-name" of a Stacks Address directly from the BNS V2 contract, replacing any reliance on the problematic BNS SDK. All other data can be sourced and cached directly from the BNS V2 API.

The PR also replaces previous query-package BNS usage in `mobile`, which required more direct in-app logic, with the higher-level `BnsService` calls.


pete-watters (approved):
Nice work @alexp3y 👍

apps/mobile/src/queries/bns/bns.query.ts
```diff
@@ -0,0 +1,44 @@
+import { toFetchState } from '@/components/loading';
```
pete-watters, line 1: I have been taking a look at mobile collectibles and trying to get them up to scratch. 

I wonder if I would be best placed to use this query to get BNS names instead? 

Right now we use `getAccountCollectibles` which bundles ordinals + SIP-9s together.  Then we have some bad code (I wrote) to filter SIP-9s to see if the contract contains `BNS`. 

Maybe we would be better off splitting non-fungibles into protocol based services? 

In the mobile FE now we have a `serializeCollectible` function that transforms the data received but I think we could move that into the services instead?

alexp3y: Yeah, I would definitely use this service now to get BNS names / profiles instead of filtering Sip-9s.

I'll try to prioritize splitting the `CollectiblesService` into individual protocol-specific services later this week.

tigranpetrossian (approved):

PR #765, opened 2024-12-18
Revamp the ESLint setup, closes LEA-1866

tigranpetrossian:
**Lots of words. Please read thoroughly and review at your own pace. Code changes are intentionally in separate commits, with relevant rule mentioned in the commit message body.** 

This PR revamps our ESLint setup:

* Upgrade ESLint to v9 
* Update the `eslint-config` package to extract out monorepo specific configurations and make it re-usable outside the monorepo.
* Add optional, configurable pre-commit and pre-push hooks so developers can run scripts of their choice automatically.

## eslint-config

The `eslint-config` package now holds less opinions about monorepo and/or Extension-specific setup. It’s rather an extension of few recommended presets we already used, with few important rule additions.
The package ships with two configs:

1. Base - `@leather.io/eslint-config`, this is the generic TS/JS setup suitable for all packages/projects
2. React - `@leather.io/eslint-config/react`, an additional preset used for packages/projects that use React. 

## Monorepo setup

ESLint v9 introduced a new [flat configuration pattern](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file), and a different mental model for monorepo setups, where the setup for the entire monorepo lives in a single file, instead of each package adding its own config file and extending the base config.

#### References:
* [The new config file](https://eslint.org/blog/2022/08/new-config-system-part-2/#the-new-config-file%3A-eslint.config.js)
* [Discussion on GitHub](https://github.com/eslint/eslint/discussions/16960#discussioncomment-5212286)

The [eslint-config.js](https://github.com/leather-io/mono/pull/765/files#diff-a32a0887ed9d1d707bbb3b845b7df7fd40e673c47e7b60a3ebd896b68d3b8839) for the monorepo now lives in the root and has the following structure:

```js
export default tseslint.config(
  {
    files: ['{packages,apps}/**/src/**/*.{ts,tsx}'],
    extends: [baseConfig],
  },
  {
    ignores: [...], // omitted for brevity
  },
  {
    name: 'ui',
    files: ['packages/ui/src/**/*.{ts,tsx}'],
    extends: [reactConfig],
  },
  {
    name: 'query',
    files: ['packages/query/src/**/*.{ts,tsx}'],
    extends: [reactConfig, pluginQuery.configs['flat/recommended']],
  },
  {
    name: 'mobile',
    files: ['apps/mobile/src/**/*.{ts,tsx}'],
    extends: [reactConfig, pluginLingui.configs['flat/recommended']],
    rules: {
[...]

PR #1596, opened 2025-09-02
Sanity CMS Package

camerow:
This PR adds a Sanity Studio, an initial migration of the Posts from Notion and a typed client for use in our applications. This is the initial bootstrapping phase for stubbing out this functionality. I plan to migrate Posts to more discrete schemas depending on the content. Eg. `guides`, `blogs`, `developer-docs` should all get their owns schema and be migrated to that. There is also small bits of reused content across the website that can be extracted to small content models for use across our applications. 

I have updated the README.md to give a fairly thorough rundown of what this currently does and how to use it.


apps/mobile/package.json
```diff
@@ -123,8 +123,8 @@
     "lottie-react-native": "7.2.2",
     "p-memoize": "7.1.1",
     "punycode": "2.3.1",
-    "react": "19.0.0",
-    "react-dom": "19.0.0",
+    "react": "19.1",
+    "react-dom": "19.1",
```
kyranjamie: Sure this safe to do?

apps/web/package.json
```diff
@@ -18,7 +18,8 @@
     "test:coverage": "vitest run --coverage",
     "test:integration": "playwright test",
     "test:unit": "vitest run",
-    "typecheck": "pnpm cf-typegen && react-router typegen && tsc -b"
+    "typecheck": "pnpm cf-typegen && react-router typegen && tsc -b",
+    "typegen": "react-router typegen && tsc -b"
```
kyranjamie: Shouldn't wrangler types be in typegen too?

packages/cms/src/client/queries/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  _type == "post"
+  && defined(slug.current)
+]|order(publishedAt desc)[0...12]{_id, title, slug, publishedAt}`);
+
+export const STACKING_POST_QUERY = defineQuery(`*[
+  _type == "post"
+  && _id == "1ead74b6-94c7-8010-a551-c0ef436999df"
+][0]`);
+
+export const POST_BY_SLUG_QUERY = defineQuery(`*[
```
kyranjamie: If these render poor type names, can we just use PascalCase instead?

packages/cms/src/types.ts
```diff
@@ -0,0 +1,349 @@
...
+  }>;
+  website?: string;
+  disclaimer?: string;
+  order?: number;
+  platform?: 'All' | 'Extension' | 'Mobile' | 'None';
+  aliases?: string;
+  dataPointInstructions?: string;
+  dataPointSource?: string;
+  dataPointValue?: string;
+  views?: Array<string>;
```
kyranjamie: This should be an eslint error, we enforce `string[]` notation. On that note, can you make sure that eslint is installed and running on CI in this package?

packages/cms/sanity.config.ts
```diff
@@ -0,0 +1,19 @@
...
+import { defineConfig } from 'sanity';
+import { structureTool } from 'sanity/structure';
+
+import { schemaTypes } from './src/studio/schemaTypes';
+
+export default defineConfig({
+  name: 'cms',
+  title: 'Leather - CMS',
+
+  projectId: '70cnou7r',
```
kyranjamie: can we make a const/env var of this as it's used both here cli.ts file

release-please-config.json
```diff
@@ -143,4 +150,4 @@
   "extra-files": [
     "app.config.ts"
   ]
-}
+}
```
kyranjamie: Image

kyranjamie: editorconfig/Prettier should leave a trailing space

packages/cms/src/types.ts
```diff
@@ -0,0 +1,349 @@
...
+ * Command: `sanity typegen generate`
+ *
+ * Any modifications made directly to this file will be overwritten the next time
+ * the TypeScript definitions are generated. Please make changes to the Sanity
+ * schema definitions and/or GROQ queries if you need to update these types.
+ *
+ * For more information on how to use Sanity TypeGen, visit the official documentation:
+ * https://www.sanity.io/docs/sanity-typegen
+ * ---------------------------------------------------------------------------------
+ */
```
kyranjamie: Do we have to commit these? Can't they be bundled as a package output? This should be pretty straightforward as any package that depend on cms' workspace:* will have the package auto-built by turbo.

Sometimes there are legit reasons to commit generated files (pnpm-lock.yaml), but I'm a big advocate for avoiding where possible. It gets messy and bloats the (already pretty big) repo, PRs are harder to review, they can cause conflicts etc.

packages/cms/eslint.config.mjs
```diff
@@ -0,0 +1,3 @@
+import studio from '@sanity/eslint-config-studio'
+
+export default [...studio]
```
kyranjamie: Defer to @tigranpetrossian here, but is this the correct way to extend the config? Pretty sure we're not depending on our core rules this way.

packages/cms/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "nodemon": "3.1.10",
+    "prettier": "3.5.1",
+    "tsup": "8.4.0",
+    "typescript": "5.8.3"
+  },
+  "keywords": [
+    "cms",
+    "sanity"
+  ],
+  "main": "package.json",
```
kyranjamie: Why is package.json the main output?

packages/cms/package.json
```diff
@@ -0,0 +1,48 @@
...
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "require": "./dist/index.js"
+    }
+  },
+  "dependencies": {
+    "@sanity/client": "7.10.0",
+    "@sanity/vision": "4.6.1",
+    "groq": "4.6.0",
+    "react": "19.1",
```
kyranjamie: Should react be a dependency of the build output? This is just the resulting types and client, not the studio we deploy, so figure it isn't needed?

packages/cms/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "react-dom": "19.1",
+    "sanity": "4.6.1",
+    "styled-components": "6.1.18"
```
kyranjamie: also should be a dev dep, no?

kyranjamie (commented):
Great to see progress on CMS, thanks Will

apps/mobile/package.json
pete-watters: I made these changes and I cannot build the mobile app so I don't think it is. 

@camerow do you need this change for this PR?

pete-watters (changes requested):
If I upgrade the React version in mobile the build fails with some type errors

``` 
ERROR  Warning: TypeError: Cannot read property 'default' of undefined

This error is located at:

Call Stack
  KeyboardControllerView (<anonymous>)
  RootLayout (<anonymous>)
  RootApp(./_layout.tsx) (<anonymous>)
  ErrorBoundary(./_layout.tsx) (<anonymous>)
  RNCSafeAreaProvider (<anonymous>)
  App (<anonymous>)
  ErrorOverlay (<anonymous>)
 ERROR  Warning: TypeError: Cannot read property 'getScrollableNode' of null

This error is located at:

Call Stack
  RootLayout (<anonymous>)
  RootApp(./_layout.tsx) (<anonymous>)
  ErrorBoundary(./_layout.tsx) (<anonymous>)
  RNCSafeAreaProvider (<anonymous>)
  App (<anonymous>)
  ErrorOverlay (<anonymous>)
 ERROR  Warning: TypeError: Cannot read property 'default' of undefined
```

If this PR needs that React version in mobile I think it's better to first do that seperately

apps/mobile/package.json
tigranpetrossian: Yeah, 19.1 is not supported by RN yet.

packages/cms/eslint.config.mjs
tigranpetrossian: True, this won't extend our own config--ESLint uses flat file config since v9. It should go into root eslint.config.js scoped for `packages/cms`. Details [here](https://github.com/leather-io/mono/pull/765), but the config file itself is pretty self-explanatory.

apps/mobile/package.json
camerow: Sorry, just an oversight on my part.

packages/cms/package.json
camerow: Probably a find and replace error on my part.

packages/cms/package.json
camerow: Yeah, I believe this can and should be a `devDependency`

packages/cms/src/client/queries/index.ts
camerow: Yeah, updating those. Good call.

apps/web/package.json
camerow: Yeah, for sure. Added

apps/mobile/package.json
camerow: Fixed.

packages/cms/package.json
camerow: updated dev dependencies

packages/cms/package.json
camerow: fixed

packages/ui/package.json
```diff
@@ -47,8 +47,8 @@
     "prism-react-renderer": "2.4.1",
     "prismjs": "1.29.0",
     "radix-ui": "1.4.2",
-    "react": "19.0.0",
-    "react-dom": "19.0.0",
+    "react": "19.1",
```
pete-watters: I think this can't be upgraded either. Maybe we should keep everything at v 19.0.0 until we can upgrade everything?

packages/cms/eslint.config.mjs
camerow: Thanks ya'll. Some of this was just overlooked boilerplate Sanity created out of the box.

camerow:
@kyranjamie / @pete-watters this is ready for a final review

pete-watters (approved):
LGTM, nice work!

PR #1599, opened 2025-09-03
fix(mobile): improvements to collectibles

pete-watters:
This PR handles some refactors and improvements to **mobile** collectibles:

✨ Features:

- [x]  adds service level filtering of `bns` from `sip-9` and fetches them separately from `useAccountBnsNames`
    - this helps show all **BNS** % **.locker** names
    - it also helps globally exclude generic names given
    - that fixes issues with empty collectibles 
- [x]  UI fixes to remove gaps between collectibles
- [x]  refactor of collectible code to clean up, deprecate `serializer` and split by protocol
- [x]  handling of  images with empty src by adding a fallback
- [x]  removes some legacy and unused code for Hiro Ordinals

☎️  Demo:

https://github.com/user-attachments/assets/e8102dca-073a-47c2-83c7-6777fd9fa405

🤔 Future Enhancements:

- [] move text inscription content fetch to service + render/ classify HTML text inscriptions properly
- [] improve loading states / fallback image UI


packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -1,4 +1,12 @@
 import { useEffect, useState } from 'react';
+import WebView from 'react-native-webview';
+
+// fetchInscriptionTextContent is a simple axios get
+// maybe its better to use axios directly in UI?
+// better yet would be to add the text content to the collectible object
+// in the service and then we can use that here
+// i tried that but it was not working as expected
+import { fetchInscriptionTextContent } from '@leather.io/query';
```
pete-watters: I added the `'@leather.io/query'` package to the UI library but I'm not sure if that's a good fit. 

I want to use `fetchInscriptionTextContent` so I:
- if it's a text inscription
- I can fetch the text

`fetchInscriptionTextContent` is actually a simple `axios` call:
```
export async function fetchInscriptionTextContent(src: string) {
  const res = await axios.get(src, { responseType: 'text' });
  return res.data;
}
```

I tried making this call again client side but it got into a real mess in the `FlashList` `renderItem`. I tried to append this to the collectibles service but again it got quite messy. 

We will improve that service soon and when we do I believe the text content should be provided in the collectible query response by the service

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -30,15 +34,28 @@ export function CollectibleText({ src, size = 200 }: CollectibleTextProps) {
 
   if (!content) return null;
 
+  if (content.includes('<html')) {
```
pete-watters: We get here if it's a text inscription but the text is actually `html`. I'm cautiously:
- removing any script tags
- disabling JS 
- disabling DOM storage

I did some testing and this will render inscribed HTML content correctly. 

Or other inscriptions classified as HTML we load the source directly in a WebView. 

This changes means we can see text inscriptions like this that were previously not working on mobile:

Screenshot 2025-09-03 at 13 05 44

packages/query/src/bitcoin/ordinals/inscription.utils.ts
```diff
@@ -1,33 +1,11 @@
...
+import { Inscription } from '@leather.io/models';
 import { createInscriptionAsset } from '@leather.io/utils';
 
-import { InscriptionResponseHiro } from '../../../types/inscription';
 import {
   BestInSlotInscriptionResponse,
   BestinSlotInscriptionBatchInfoResponse,
 } from '../clients/best-in-slot';
 
-export function createHiroInscription(inscriptionResponse: InscriptionResponseHiro) {
```
pete-watters, line 11: All of this code was used only for generating mocks in the `extension`. 

I have [another PR ](https://github.com/leather-io/extension/pull/6357)in draft that refactors the `extension` to use BiS mocks instead as that's what we are actually using

packages/bitcoin/src/utils/bitcoin.utils.ts
```diff
@@ -359,7 +359,3 @@ export function getBitcoinInputValue(input: TransactionInput) {
   // logger.warn('Unable to find either `witnessUtxo` or `nonWitnessUtxo` in input. Defaulting to 0');
   return 0;
 }
-
-export function serializeSatPoint(txid: string, output: string | number, offset: string | number) {
```
pete-watters, line 363: @alexp3y  - refactoring this away as its no longer used but I can keep it if you think it could be useful

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
camerow: cc @tigranpetrossian but IMO it's a mistake to add network dependencies to the `ui` package. Would prefer inversion of control / dependency injection instead.

pete-watters: I'll try harder to get the inscription text in the service and just return that to the client

apps/mobile/src/features/collectibles/collectibles-list.tsx
```diff
@@ -1,31 +1,66 @@
...
-export function CollectiblesList({ collectiblesData, header }: CollectiblesListProps) {
-  const collectibles = collectiblesData.state === 'success' ? collectiblesData.value : [];
+export function CollectiblesList({ currentAccount, header }: CollectiblesListProps) {
+  const { fingerprint, accountIndex } = currentAccount;
+  const { value: collectibles, state: collectiblesState } = useAccountCollectibles(
+    fingerprint,
+    accountIndex
+  );
+  const { value: bnsNames, state: bnsNamesState } = useAccountBnsNames(fingerprint, accountIndex);
+  const collectibleData: NonFungibleCryptoAsset[] = [
```
pete-watters, line 27: I'm not happy with this, maybe there's a better way? 

I want to exclude `BNS` from `useAccountCollectibles SIP-9` and instead use the data from `useAccountBnsNames`. 

When I spread them back together I need to add a lot of junk props to satisfy it being a valid `SIP-9 - NonFungibleCryptoAsset` 

I experimented with a new `BnsAsset` type but that didn't make sense either. 

Maybe this data manipulation should be done in the service. We don't yet use any of these fields apart from:
- `name`
- `collection` -> I set this as `bns` to set them apart from other SIP-9's

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
```diff
@@ -1,26 +1,23 @@
...
 
-export function CollectibleText({ src, size = 200 }: CollectibleTextProps) {
+export function CollectibleText({ src, height = 200 }: CollectibleTextProps) {
   const [content, setContent] = useState<string | null>(null);
-
+  // FIXME: this content should be processed in the service and let us use HTML content
   useEffect(() => {
     async function fetchContent() {
       try {
         const response = await fetch(src);
```
pete-watters, line 17: @camerow I backed out of adding the query package here and left it how it was before. 

This is still bad as we are fetching the inscription text content in the UI library. I will do some more work on this to refactor it and also properly render text inscriptions with HTML content as their text. 

That will be done on the server. For now this simpler fix lets us correctly render text insciptions

packages/ui/src/components/collectibles/native/bns.native.tsx
```diff
@@ -0,0 +1,35 @@
+import { Image } from 'expo-image';
+
+import { Box, Text } from '../../../../native';
+import { CollectibleImageProps } from './collectible-image.native';
+
+export function BnsImage({ alt, height = 200 }: Pick<CollectibleImageProps, 'alt' | 'height'>) {
+  return (
+    <Box height={height} overflow="hidden" bg="ink.background-secondary" position="relative">
+      <Image
+        source={require('../../../assets-native/images/bnsv2.png')}
```
pete-watters, line 10: Before for BNS, we used to use the image src in the SIP-9 but it was all for the same image. Now we just host it ourselves

apps/mobile/src/features/feature-flags/index.ts
```diff
@@ -44,10 +44,6 @@ export function useNotificationsFlag() {
   return useBoolVariation('release_push_notifications', false);
 }
 
-export function useInscriptionsFlag() {
```
pete-watters, line 47: Getting rid of this to just have a single `collectiblesFlag`

packages/utils/src/assets/asset-id.ts
```diff
@@ -67,6 +67,14 @@ export function getAssetId(asset: CryptoAsset): CryptoAssetId {
...
 
 function buildSip9AssetId(asset: Sip9Asset): CryptoAssetId {
+  // BNS SIP-9's have a collection and name
+  if (asset.collection === 'bns') {
+    return {
+      protocol: asset.protocol,
+      id: `${asset.name}|${asset.collection}`,
+    };
+  }
+
```
edgarkhanzadian, line 77: Why assetId|tokenId doesn't work for bns?

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
edgarkhanzadian: What is the issue with passing fetch content through CollectibeText props? That way the network request can be at least handled on the mobile app side

apps/mobile/src/features/collectibles/collectibles-list.tsx
```diff
@@ -1,31 +1,66 @@
...
+      protocol: 'sip9' as const,
+      collection: 'bns',
+      name: bns.fullName,
+      assetId: '',
+      contractId: '',
+      tokenId: 0,
+      description: bns.namespace,
+      cachedImage: '',
+      cachedImageThumbnail: '',
+    })) ?? []),
```
edgarkhanzadian, line 41: suggestions:
1. move AccountBnsName -> Sip9Asset conversion to a separate function
2. technically, bns names should have assetid, contractId, tokenId. So i'd rather have those prefilled rather than empty strings. Or we should at least set it as undefined/null instead of an empty string (as typescript would consider that those properties have value even though we definitely know those don't)
3. If we are unable to move bnsNames to collectibles service call, I believe we should receive Sip9Asset type instead of AccountBnsName from useAccountBnsNames and generally from `getBnsService().getAccountBnsNames(request, signal)` service call

apps/mobile/src/components/loading/empty-layout.tsx
```diff
@@ -3,6 +3,19 @@ import { Box, HasChildren } from '@leather.io/ui/native';
 interface EmptyLayoutProps extends HasChildren {
   image?: React.ReactNode;
 }
+// TODO: LEA-3190  ask design for this empty layout
+export function EmptyLayoutTab({ children, image }: EmptyLayoutProps) {
+  return (
+    <Box height={648} gap="4" alignItems="center" flexShrink={0}>
```
edgarkhanzadian, line 9: Can you use relative height here?

For that you can do:

```
const {height, width} = Dimensions.get("screen")

<Box height={height / 3} .... >
```

And for the box
```
const {height, width} = Dimensions.get("screen")

<Box height={width / 2} width={width / 2} .... >
```

NOTE: the numbers are random here, use whatever you would consider fits better in the UI

apps/mobile/src/features/collectibles/collectibles-list.tsx
```diff
@@ -1,31 +1,66 @@
...
 
-export function CollectiblesList({ collectiblesData, header }: CollectiblesListProps) {
-  const collectibles = collectiblesData.state === 'success' ? collectiblesData.value : [];
+export function CollectiblesList({ currentAccount, header }: CollectiblesListProps) {
+  const { fingerprint, accountIndex } = currentAccount;
+  const { value: collectibles, state: collectiblesState } = useAccountCollectibles(
+    fingerprint,
+    accountIndex
+  );
+  const { value: bnsNames, state: bnsNamesState } = useAccountBnsNames(fingerprint, accountIndex);
```
edgarkhanzadian, line 26: I strongly believe we should receive bnsNames as part of the collectibles from useAccountCollectibles here. Is there any particular reason we want to receive it separately in this case?

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -123,13 +121,13 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
       )}
       {listTab === 'collectibles' && (
         <CollectiblesList
+          currentAccount={currentAccount}
```
edgarkhanzadian, line 124: We received collectiblesData through props here as sometimes we wanted to use all collectibles data across the accounts and sometimes we would only show collectibles of one account. Now that doesn't seem to be a use-case anymore, i wonder if moving collectibles fetching to CollectiblesList is a good way forward 🤔 if so, can you also update AssetsList for consistency?

apps/mobile/src/features/collectibles/render-collectible.tsx
```diff
@@ -1,19 +1,60 @@
...
-    />
+    <ImageUnavailable>
+      <Text textAlign="center">{t`Image currently unavailable`}</Text>
+    </ImageUnavailable>
   );
 }
+
+function Stamp({ item }: { item: StampAsset }) {
+  if (!item.stampUrl) return <FallbackImage />;
+  return <CollectibleImage source={item.stampUrl} alt={item.stamp.toString()} height={200} />;
```
edgarkhanzadian, line 28: Is setting height to 200 here makes it responsive across different devices?

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -69,6 +63,10 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
   });
   const isErrorTotalBalance = totalBalance.state === 'error';
 
+  const allSip10Data = useSip10AccountBalance(fingerprint, accountIndex, { returnAllAssets: true });
+  const allRunesData = useRunesAccountBalance(fingerprint, accountIndex, { returnAllAssets: true });
+  const hasAssets = !!allSip10Data.value?.sip10s.length || !!allRunesData.value?.runes.length;
+
```
edgarkhanzadian, line 69: nit: I'd prefer not moving code below unless there is a necessity for it s.t. we don't cause merge conflicts in other prs.

edgarkhanzadian (commented):
Great work, Pete! 

Left a few comments here and there. Generally, separating bns names from the rest of sip9s doesn't sit right with me, we should be able to handle it as part of the collectibles list.

apps/mobile/src/components/home/home-with-account-screen.tsx
pete-watters: That was a mistake, thanks I'll move it back. I think it was auto-format when I stopped passing down `collectiblesData`

apps/mobile/src/components/home/home-with-account-screen.tsx
pete-watters: Yes I can move the assets to get the data more locally 👍

apps/mobile/src/components/loading/empty-layout.tsx
pete-watters: Sure I can take a look. This whole UI will be updated as we got some designs now

apps/mobile/src/features/collectibles/collectibles-list.tsx
pete-watters: There were bugs with it not having the correct data. I can refactor it so that the correct data is returned from the service and that will fix [this](https://github.com/leather-io/mono/pull/1599/files/2a88421719781d4aea2480f2fa4ea123f7467541#r2340194791) too 🤞

apps/mobile/src/features/collectibles/collectibles-list.tsx
pete-watters: Sounds good 👍

apps/mobile/src/features/collectibles/render-collectible.tsx
pete-watters: I set it explicitly based on the design. Maybe it's better to dynamically calculate it?

packages/ui/src/components/collectibles/native/collectible-text.native.tsx
pete-watters: We need to fetch the content if it's text inscription. I tried this on the mobile side but it was a mess of async calls in the FlashList `render` function. 

I think a better solution will be to do this code in the collectibles service and return the data ready to go. Then it will be ready for both platforms

packages/utils/src/assets/asset-id.ts
pete-watters: Fetching it the new way there is no `tokenId` in the payload

apps/mobile/src/features/collectibles/render-collectible.tsx
edgarkhanzadian: using react-native Dimensions might be safer here, to make sure it works properly on smaller devices

PR #1600, opened 2025-09-03
fix: account design fixes [LEA-3157] [LEA-3153] [LEA-3158] [LEA-3156] [LEA-3155] [LEA-3150]

edgarkhanzadian:
https://github.com/user-attachments/assets/0ee80cc9-0079-4cc6-880c-0ec8aeacb4f7

### DIFF
--- a/apps/mobile/src/features/account/components/account-header.tsx
+++ b/apps/mobile/src/features/account/components/account-header.tsx
@@ -1,7 +1,9 @@
 import { AccountAvatar } from '@/features/account/components/account-avatar';
 import { Account } from '@/store/accounts/accounts';
+import { useAccounts } from '@/store/accounts/accounts.read';
+import { useWalletByFingerprint, useWallets } from '@/store/wallets/wallets.read';
 
-import { Box, ChevronDownIcon, Pressable, Text } from '@leather.io/ui/native';
+import { Box, Pressable, Text } from '@leather.io/ui/native';
 
 interface AccountHeaderProps {
   account: Account;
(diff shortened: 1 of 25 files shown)


edgarkhanzadian:
A few more edits after design review

apps/mobile/src/features/account/components/account-header.tsx
```diff
@@ -1,21 +1,39 @@
...
+      flexDirection="row"
+      alignItems="center"
+      gap="3"
+    >
       <AccountAvatar variant="sm" icon={icon} />
-      <Box flexDirection="row" alignItems="center" gap="1">
+      <Box>
         <Text variant="label01">{name}</Text>
-        <ChevronDownIcon variant="small" />
+        {!isOneWallet && wallet && (
```
pete-watters: This small change greatly improves my multi wallet experience 👏 


https://github.com/user-attachments/assets/ee5f2eb6-71fa-4bd1-ae9b-59dd15115d50

apps/mobile/src/features/account/components/account-header.tsx
```diff
@@ -1,21 +1,39 @@
...
   onPress?(): void;
 }
 
 export function AccountHeader({ account, onPress }: AccountHeaderProps) {
-  const { icon, name } = account;
+  const { icon, name, fingerprint } = account;
+  const { list: accounts } = useAccounts();
+  const { list: wallets } = useWallets();
+  const wallet = useWalletByFingerprint(fingerprint);
+
```
pete-watters, line 18: I think we normally say `hasOneAccount` instead of `isOneAccount` but it's not a big deal.

If you say `is` I would say `isSingleAccount` / `isSingleWallet`

apps/mobile/src/features/balances/total-balance.tsx
```diff
@@ -1,9 +1,17 @@
 import { Balance } from '@/components/balance/balance';
 import { useAccountBalance } from '@/queries/balance/account-balance.query';
+import { useTotalBalance } from '@/queries/balance/total-balance.query';
 import { AccountLookup } from '@/shared/types';
 
 import { TextProps } from '@leather.io/ui/native';
 
+export function TotalBalance(props: TextProps) {
+  const { totalBalance } = useTotalBalance();
```
pete-watters, line 9: Now we are using `useTotalBalance` we can un-deprecate the other hooks in there:
Screenshot 2025-09-03 at 17 06 16

Something is not working with the loading state too in that it should show a loader until it has the final total. It seems to just show the first total and then add them as they arrive:


https://github.com/user-attachments/assets/dd9ed5a7-9c4e-439a-a028-c1a73de9c63e

pete-watters (approved):
Great stuff @edgarkhanzadian 👍  I mentioned a couple of small improvements but overall it looks and works great

apps/mobile/src/features/account/components/account-header.tsx
edgarkhanzadian: Great point, updated to `hasOneAccount` and `hasOneWallet`!

apps/mobile/src/features/balances/total-balance.tsx
edgarkhanzadian: > Now we are using useTotalBalance we can un-deprecate the other hooks in there:

Updated!

> Something is not working with the loading state too in that it should show a loader until it has the final total. It seems to just show the first total and then add them as they arrive:

Looking into it

edgarkhanzadian: Found it ! :D 

balance

pete-watters: That's great, thanks for fixing! 🥇

PR #1602, opened 2025-09-04
fix(mobile): token details QA feedback updates

pete-watters:
This PR makes some visual updates to token details post QA

✨ Features:

- [x] Fixes spacing issue for `Activity` header
- [X] Only show address list for BTC
- [X] Capitalise BTC address in list and show more of it 

☎️  Demo: 

You can see a demo of the new changes below. **NOTE** this also displays a bug we are having now when fetching BTC balances - often they error and then disappear [LEA-3174](https://linear.app/leather-io/issue/LEA-3174/fix-occasionally-weird-token-balance-0-states)

https://github.com/user-attachments/assets/1ddaa45d-30f9-446d-b22e-320d39e320dc


apps/mobile/src/features/token/components/address-list.tsx
```diff
@@ -64,8 +60,13 @@ export function AddressListItem({
         <Cell.Label variant="primary">
           <Text variant="label02">{name}</Text>
         </Cell.Label>
-        <Cell.Label variant="secondary" onPress={() => void onCopyAddress(address)}>
-          {truncateMiddle(address)}
+        <Cell.Label
+          variant="secondary"
+          onPress={() => void onCopyAddress(address)}
+          textDecorationLine="underline"
+          textTransform="uppercase"
```
tigranpetrossian: Older bitcoin address formats are case-sensitive, but even with the new ones the convention is to keep them lowercase.

tigranpetrossian (approved):
Thanks @pete-watters, the capitalization for BTC addresses is unnecessary, otherwise looks good.

apps/mobile/src/features/token/components/address-list.tsx
pete-watters: Thanks @tigranpetrossian  👍  I was changing it to match the design but I will revert it.


Screenshot 2025-09-04 at 11 26 16

PR #1607, opened 2025-09-05
fix: error state should be with && instead of ||

edgarkhanzadian:
Only show error state if all of the queries errored out

### DIFF
--- a/apps/mobile/src/queries/balance/account-balance.query.ts
+++ b/apps/mobile/src/queries/balance/account-balance.query.ts
@@ -39,9 +39,9 @@ export function useAccountBalance(accountId: AccountId): AccountBalance {
     sip10AccountBalance.state === 'loading' ||
     runesAccountBalance.state === 'loading';
   const isError =
-    btcAccountBalance.state === 'error' ||
-    stxAccountBalance.state === 'error' ||
-    sip10AccountBalance.state === 'error' ||
+    btcAccountBalance.state === 'error' &&
+    stxAccountBalance.state === 'error' &&
+    sip10AccountBalance.state === 'error' &&
     runesAccountBalance.state === 'error';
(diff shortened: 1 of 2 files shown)


apps/mobile/src/queries/balance/account-balance.query.ts
```diff
@@ -39,9 +39,9 @@ export function useAccountBalance(accountId: AccountId): AccountBalance {
     sip10AccountBalance.state === 'loading' ||
     runesAccountBalance.state === 'loading';
   const isError =
-    btcAccountBalance.state === 'error' ||
-    stxAccountBalance.state === 'error' ||
-    sip10AccountBalance.state === 'error' ||
+    btcAccountBalance.state === 'error' &&
```
pete-watters, line 42: I'm not sure about this as if one of these has errored, that means you aren't showing the correct balance then. 

I think it's correct to show an error if any of the balance requests has failed

edgarkhanzadian: The idea is to revert this back to the previous state and discuss with design a way of surfacing to the user the partial error state: https://trustmachines.slack.com/archives/C05LAP952E7/p1757064288198169?thread_ts=1757063670.736659&cid=C05LAP952E7

pete-watters: I think we should keep this and focus on erroring less if we can. I'm open to being convinced otherwise though

camerow: Why not both? We're going to see errors, when we do we should handle them as well as reasonably possible and also optimize for less errors (while stil handling them well)

edgarkhanzadian: Ok, closing PR then

PR #1609, opened 2025-09-05
feat: add sensible shared balance query options

alexp3y:
Adds a shared balance query object with sensible refetch defaults. Allows 1 retry, which solves cases when AbortSignal errors are not handled silently by RQ.


pete-watters (approved):
Nice work fixing this @alexp3y 👍  A good side effect is to centralise the `balanceQueryOptions` now too so we an change them all more easily if needed

PR #1613, opened 2025-09-05
feat: token management

edgarkhanzadian:
- Fix state persistance for settings reducer

- Add token management details ui + redux reducer

https://github.com/user-attachments/assets/cc679671-8b44-4c21-af7e-c77b38a1f8a7


apps/mobile/src/queries/balance/runes-balance.query.ts
```diff
@@ -22,9 +28,51 @@ export function useRuneBalanceByRuneName(
   return toFetchState(useRuneBalanceByRuneNameQuery({ account }, runeName));
```
edgarkhanzadian, line 23: @alexp3y Would it be possible to pass runeId here from services s.t. we don't use runename as an identifier?

packages/services/src/balances/runes-balances.service.ts
```diff
@@ -130,7 +133,15 @@ export class RunesBalancesService {
       )
     )
       .filter(result => result.status === 'fulfilled')
-      .map(b => b.value);
+      .map(b => b.value)
+      .filter(ft => {
```
pete-watters: Maybe this filter can be extracted and shared between `runes` + `sip-10`? It seems to be doing the same thing.

I think it could also be refactored to:
```
.filter({asset} => {

          request.filters?.assetVisibility?.[serializeAssetId(getAssetId(asset))] || true
      });

```

pete-watters (changes requested):
I tested this and it works in hiding tokens but if I hide all of the tokens in the list, then the filter menu button disappears so I can't re-enable them 


https://github.com/user-attachments/assets/78814589-6b16-4924-9baa-8d39d2ef90b3

apps/mobile/src/store/settings/settings.write.ts
```diff
@@ -69,6 +71,19 @@ export const settingsSlice = createSlice({
...
       state.languagePreferenceSource = action.payload;
     },
+    userChangedAssetVisibility(
+      state,
+      action: PayloadAction<{
+        assetId: SerializedCryptoAssetId;
+        value: boolean;
+      }>
+    ) {
+      const updatedAssetVisibility = produce(state.assetVisibility, draftAssetVisibility => {
```
pete-watters: This is a bit tricky to follow. Maybe the `const updatedAssetVisibility = produce(` can be extracted from here? 

All the other settings have a crisper format:
```
    userChangedLanguagePreferenceSource(state, action: PayloadAction<LanguagePreferenceSource>) {
      state.languagePreferenceSource = action.payload;
    },
```

apps/mobile/src/queries/balance/sip10-balance.query.ts
```diff
@@ -12,9 +13,32 @@ export function useSip10TotalBalance() {
   return toFetchState(useSip10AggregateBalanceQuery(accounts.map(account => ({ account }))));
 }
 
-export function useSip10AccountBalance(fingerprint: string, accountIndex: number) {
+export function useSip10AccountBalance(
```
pete-watters, line 16: Do we need to have a separation between `useSip10AccountBalance` and `useManagedSip10Tools`? Maybe they can be merged so that `useSip10AccountBalance` also can return `isEnabled`.

apps/mobile/src/hooks/use-account-assets.ts
```diff
@@ -0,0 +1,18 @@
...
+export function useAccountAssets(
+  fingerprint: string,
+  accountIndex: number,
+  options?: {
+    returnAllAssets?: boolean;
+  }
+) {
+  const sip10Data = useSip10AccountBalance(fingerprint, accountIndex, options);
+  const runesData = useRunesAccountBalance(fingerprint, accountIndex, options);
+  return {
```
pete-watters: Maybe you can add `isLoading` in this response also? You could even wrap it in a `toFetchState` and also pass on any errors

apps/mobile/src/features/balances/manage-tokens.sheet.tsx
```diff
@@ -0,0 +1,113 @@
...
+            paddingTop: spacing['5'],
+            paddingBottom: bottom,
+          }}
+        >
+          <>
+            <Text variant="heading05" px="5">
+              {t`Manage tokens`}
+            </Text>
+            <Box pt="5">
+              {sip10s.value?.sip10s?.map(sip10 => (
```
pete-watters, line 77: In `AssetsList` we changed it to spread `sip10` + `runes` together. Maybe we should do that here also so we just have one map?

packages/services/src/balances/runes-balances.service.ts
edgarkhanzadian: Done!

apps/mobile/src/store/settings/settings.write.ts
edgarkhanzadian: yea using immer is a bit overkill here, changed!

apps/mobile/src/queries/balance/sip10-balance.query.ts
edgarkhanzadian: I think i'd rather not change the fetchstate that useSip10AccountBalance returns. Moreover, passing returnAllAssets option to useSip10AccountBalance would force it to bypass the filtering and that would not work with the isEnabled functionality (that expects to receive only enabled sip10s)

apps/mobile/src/hooks/use-account-assets.ts
edgarkhanzadian: Ended up removing this hook, it's probably redundant

apps/mobile/src/features/balances/manage-tokens.sheet.tsx
edgarkhanzadian: well, right before rendering we still check if it is a rune or sip10 and render those components differently. In case of AssetsList we had to do that as we can only pass one array to AssetsList.

I'd say there is not much need to mix these into one array unless we plan to mix runes and sip10s together and then sort out by USD value.

code

edgarkhanzadian:
@pete-watters thank you for catching that! Fixed

packages/services/src/balances/sip10-balances.utils.ts
```diff
@@ -34,3 +40,14 @@ export function sortByAvailableQuoteBalance(
 ) {
   return b.quote.availableBalance.amount.minus(a.quote.availableBalance.amount).toNumber();
 }
+
+export function filterUsingAssetVisibility(
```
alexp3y: This could probably be located somewhere a bit more central - It's not specific to SIP-10s and actually not even really to balances. It essentially implements the core business rule governing default token visibility filtering.

Would also be nice to have a couple unit tests, even though the logic is simple, just to help spec out exactly its behavior.

packages/utils/src/assets/asset-id.ts
```diff
@@ -66,3 +72,20 @@ function buildSip9AssetId(asset: Sip9Asset): CryptoAssetId {
     id: `${asset.assetId}|${asset.tokenId}`,
   };
 }
+
+export type SerializedCryptoAssetId = `${string}|${string}`;
```
alexp3y, line 76: 😻

alexp3y (approved):
Nice work! Feels really natural with the new service request token visibility options instead of repeated filtering in the balance queries.

pete-watters (approved):
I re-tested and it's working well, great work 🚀

PR #1614, opened 2025-09-05
feat: account balances service

alexp3y:
Refactors account total balances into its own service and cleans up `mobile` queries.

Edit: Removed performance improvements and query cancellation bug fixes to [another PR](https://github.com/leather-io/mono/pull/1615).

### DIFF
--- a/packages/services/src/balances/account-balances.service.ts
+++ b/packages/services/src/balances/account-balances.service.ts
@@ -1,0 +1,56 @@
+import { injectable } from 'inversify';
+
+import { Money } from '@leather.io/models';
+import { isDefined, sumMoney } from '@leather.io/utils';
+
+import { AccountRequest } from '../types';
+import { BtcBalancesService } from './btc-balances.service';
+import { RunesBalancesService } from './runes-balances.service';
+import { Sip10BalancesService } from './sip10-balances.service';
+import { StxBalancesService } from './stx-balances.service';
+
+@injectable()
+export class AccountBalancesService {
+  constructor(
+    private readonly btcBalancesService: BtcBalancesService,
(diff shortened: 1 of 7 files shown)


edgarkhanzadian (approved):
This greatly simplifies setup on mobile side, thank you!

alexp3y:
@edgarkhanzadian Just realized that this might make incorporating managed tokens into the account balance more complicated...

edgarkhanzadian:
> @edgarkhanzadian Just realized that this might make incorporating managed tokens into the account balance more complicated...

no worries, it might be different but not that much more complicated

packages/services/src/balances/account-balances.service.ts
```diff
@@ -0,0 +1,56 @@
...
+@injectable()
+export class AccountBalancesService {
+  constructor(
+    private readonly btcBalancesService: BtcBalancesService,
+    private readonly stxBalancesService: StxBalancesService,
+    private readonly sip10BalancesService: Sip10BalancesService,
+    private readonly runesBalancesService: RunesBalancesService
+  ) {}
+
+  public async getTotalBalance(request: AccountRequest, signal?: AbortSignal): Promise<Money> {
```
pete-watters, line 21: @alexp3y I was thinking about this last week and I wonder if we **need** two queries here. 

`useGetAccountTotalBalanceQuery` and `useGetAccountUnlockedBalanceQuery` are basically the same except we return `stxBalance.quote.availableBalance` for `total` and `stxAccountBalance.value?.quote.availableUnlockedBalance` for `unlocked` 

Unless I am missing something, we could just have one request and return a both `total` and `unlocked`?

pete-watters (approved):

PR #1615, opened 2025-09-07
fix: query cancellation issues

alexp3y:
Query cancellation and `AbortSignal` / `AbortError` handling issues have been causing RQ queries to fail unexpectedly and get stuck in infinite loading states.

There are two underlying causes:
1. The failure of React Query to always swallow an `AbortError` instance silently
2. Custom `PriorityQueue` implementation bugs

Improvements have been added here to ensure that the query `AbortSignal` is always propogated down to rate limiter `.add()` calls and Http client requests; And that any subsequent `AbortError` or `CancellationError` instances are propogated back up when the signal is thrown.

In the consuming client balance queries, configuring simple query retry count that targets Abort error types specifically help alleviate cases of uncaught (non-silently handled) aborts from failing permanently.

Our custom `PriorityQueue` class used for Rate Limiting also implemented a custom "abort" event listener that prevented aborted requests from resolving properly. Canceled queries would then get stuck in an infinite (and unrecoverable) loading/pending state.

The default `PQueue` implementation from the `p-queue` package is already robust. It implements `AbortSignal` handling/cancellation and priority-level based sorting out-of-the-box. The custom `PriorityQueue` class implementation is unnecessary and introduces a high risk of bugs and inefficiencies and has been removed.


alexp3y:
@kyranjamie Did some testing and I believe this will also solve the underlying issues in the `extension` services branch that lead to me using a non-QC HTTP cache service (discussion [here](https://github.com/leather-io/extension/pull/6337#discussion_r2297827508)). Guess I did just need to spend another day looking at it 🙃

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.ts
```diff
@@ -14,11 +14,17 @@ import { leatherApiLimiter } from './leather-rate-limiter';
 if (!AbortSignal.prototype.throwIfAborted) {
   AbortSignal.prototype.throwIfAborted = function throwIfAborted() {
     if (this.aborted) {
-      throw new Error('AbortError');
+      throwAbortError();
     }
   };
 }
```
kyranjamie, line 20: If this is only for RN, shouldn't there be a RN condition so we don't call on web?

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.ts
```diff
@@ -70,10 +76,22 @@ export class RateLimiterService {
...
+      }
+      throw error;
+    }
     if (result === undefined) {
-      throw new Error('Rate limited call aborted');
+      if (options?.signal?.aborted) {
+        throwAbortError();
+      } else {
+        throw new Error('Rate limited call undefined');
+      }
```
kyranjamie, line 93: Not having the easiest time following this. When would this ever execute, the above code'll either set `result` or throw?

kyranjamie (approved):
This is fantastic Alex, thanks for explanation. This reduces complexity here and in extension. Glad to see the back of the custom queue logic.

kyranjamie:
Also notice first issue on p-queue is an AbortSignal issue https://github.com/sindresorhus/p-queue/issues/219

edgarkhanzadian (approved):
Looks great, thanks!

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.ts
alexp3y: The `!AbortSignal.prototype.throwIfAborted` is the RN condition. On web this API is defined, but it needs to be polyfilled in RN to support `AbortSignal` 

https://github.com/facebook/react-native/issues/50015

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.ts
alexp3y: Ensures that the error thrown by `PQueue` when the signal is aborted is the `AbortError`... Saw several instances where `undefined` errors where being thrown in the case of cancellations.

alexp3y:
> Also notice first issue on p-queue is an AbortSignal issue https://github.com/sindresorhus/p-queue/issues/219

@kyranjamie Good find. I didn't realize this was happening but can confirm that we also see the issue. There is a fix PR that's been approved by the package maintainers, it just needs to be released. 

A forked version of `p-queue` containing the patch has been published by the PR author, which I've temporarily switched over to here here as it resolves the issue in a minimal way.

**EDIT**: The `p-queue` fix was merged and published (`8.1.1`) so switched back to the main package.

PR #1616, opened 2025-09-08
fix(mobile): pre load total balance

pete-watters:
This PR re-add's preloading of the users total balance to pre-populate that in the query cache during initial app loading. 

That makes it faster when initially opening the `Select Account` sheet

### DIFF
--- a/apps/mobile/src/app/_layout.tsx
+++ b/apps/mobile/src/app/_layout.tsx
@@ -14,6 +14,7 @@ import { CurrentAccountProvider, useCurrentAccount } from '@/core/current-accoun
 import { GlobalSheetProvider } from '@/core/global-sheet-provider';
 import { HapticsProvider } from '@/core/haptics-provider';
 import { LeatherQueryProvider } from '@/core/leather-query-provider';
+import { QueryPreloader } from '@/core/query-preloader';
 import { ThemeProvider } from '@/core/theme-provider';
 import { AddAccountSheet } from '@/features/account/sheets/add-account-sheet';
 import { featureFlagClient, setupFeatureFlags } from '@/features/feature-flags';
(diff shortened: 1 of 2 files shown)


tigranpetrossian (commented):
Can the balance be lazy loaded in that sheet while not blocking the rest of the content?
One of the reasons we moved to single accounts is to avoid the expensive total queries and rate limits, and this seems to negate at least some of that.

pete-watters:
> Can the balance be lazy loaded in that sheet while not blocking the rest of the content? One of the reasons we moved to single accounts is to avoid the expensive total queries and rate limits, and this seems to negate at least some of that.

We don't have to do this at all. It already lazy loads the `totalBalance` on demand on that sheet but can just take a few seconds. We can keep this on ice and assess if that's an issue or not. 

Pre-loading it just gets that data ready while the APP is still booting up anyway

tigranpetrossian:
> > Can the balance be lazy loaded in that sheet while not blocking the rest of the content? One of the reasons we moved to single accounts is to avoid the expensive total queries and rate limits, and this seems to negate at least some of that.
> 
> We don't have to do this at all. It already lazy loads the `totalBalance` on demand on that sheet but can just take a few seconds. We can keep this on ice and assess if that's an issue or not.
> 
> Pre-loading it just gets that data ready while the APP is still booting up anyway

Thanks @pete-watters, Aggregate balance queries are pretty heavy-4 requests per account, so I think that's a trade-off we should be willing to make, especially if it lazy loads and doesn't block the rest of the account sheet content.

pete-watters:
> > > Can the balance be lazy loaded in that sheet while not blocking the rest of the content? One of the reasons we moved to single accounts is to avoid the expensive total queries and rate limits, and this seems to negate at least some of that.
> > 
> > 
> > We don't have to do this at all. It already lazy loads the `totalBalance` on demand on that sheet but can just take a few seconds. We can keep this on ice and assess if that's an issue or not.
> > Pre-loading it just gets that data ready while the APP is still booting up anyway
> 
> Thanks @pete-watters, Aggregate balance queries are pretty heavy-4 requests per account, so I think that's a trade-off we should be willing to make, especially if it lazy loads and doesn't block the rest of the account sheet content.

Yeah it makes sense. Adding it is causing more harm than good 👍

PR #1617, opened 2025-09-08
fix(mobile): remove half sheet from secure your account, ref LEA-3149

pete-watters:
This PR fixes a visual bug with half a screen appearing to the right on the `Secure your wallet` step of wallet creation.

I also removed the `ScrollView` wrapper as I don't think this page needs to be scrollable. 

**Demo of it fixed**

https://github.com/user-attachments/assets/db3d965c-a96d-425d-90b9-841aa261a6cc

**Previous behaviour**

https://github.com/user-attachments/assets/79f954d7-5d30-4b4b-ae2f-6676bda10764

### DIFF
--- a/apps/mobile/src/app/secure-your-wallet.tsx
+++ b/apps/mobile/src/app/secure-your-wallet.tsx
@@ -19,23 +19,21 @@ export default function SecureYourWalletScreen() {
   return (
     <Screen>
       <Screen.Header />
-      <Screen.ScrollView>
-        <Screen.Title>{pageTitle}</Screen.Title>
-        <Box px="5">
-          <Box gap="3">
-            <Text variant="label01">
-              {t`Use your device PIN, Face ID or other biometrics to secure your wallets`}
-            </Text>
-          </Box>
-          <Box justifyContent="center" alignItems="center" aspectRatio={1}>
-            <Image
-              style={{ height: 270, width: 270 }}
-              source={require('@/assets/stickers/lock.png')}
-              contentFit="contain"
-            />
-          </Box>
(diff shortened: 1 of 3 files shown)


apps/mobile/src/app/secure-your-wallet.tsx
```diff
@@ -19,23 +19,21 @@ export default function SecureYourWalletScreen() {
   return (
     <Screen>
       <Screen.Header />
-      <Screen.ScrollView>
```
edgarkhanzadian, line 22: Can you please test the screen on smaller devices, like iphone SE? You can also try using bigger font size to check the accessibility of it. I remember using scrollviews here s.t. the title and images don't push the "continue" button out of the screen.

But that was ages ago, so it might not be relevant anymore

apps/mobile/src/components/screen/screen-header/screen-header.tsx
```diff
@@ -58,9 +58,7 @@ export function ScreenHeader({
           bottom={0}
           left={0}
         >
-          {centerElement ? (
-            <FadingView opacity={headerVisibility}>{centerElement}</FadingView>
-          ) : null}
```
edgarkhanzadian, line 63: even if centerElement is undefined, we still need to show the fading view?

pete-watters: This bug was weird and mysterious 😂  I'm not sure why but without having this `FadingView` added even if `centerElement` is undefined, we see that weird shadow to the right.

I'll spend more time double checking what exactly is happening, I was going for a quick fix here

PR #1619, opened 2025-09-09
fix(mobile): fix incorrect fallback nonce loader, closes LEA-3183

tigranpetrossian:
Fixes an issue with nonce loader incorrectly failing and in turn incorrectly falling back to 1.
This addresses the problem at hand, but I think decision needs to be made on how the flow behaves when the nonce actually fails to load, so marking it as draft for now.

### DIFF
--- a/apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
+++ b/apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
@@ -5,6 +5,7 @@ import { useNextNonce } from '@/queries/stacks/nonce/account-nonces.hooks';
 import { useStacksSigners } from '@/store/keychains/stacks/stacks-keychains.read';
 import { assertStacksSigner } from '@/store/keychains/stacks/utils';
 import { t } from '@lingui/core/macro';
+import { captureException } from '@sentry/react-native';
 
 export function NonceLoader({
   accountId,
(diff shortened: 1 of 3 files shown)


apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
```diff
@@ -19,23 +20,19 @@ export function NonceLoader({
...
         type: 'error',
       });
     }
-  }, [displayToast, isError, nonceResponse?.nonce, isNonceLoading]);
+  }, [displayToast, error]);
 
-  if (isNonceLoading) return null;
+  if (isLoading) return null;
 
-  return children(nonceResponse?.nonce || 1);
```
edgarkhanzadian, line 40: damn i made 2 mistakes on 1 line. 0 is a falsy value so it would default to 1 and 1 shoudn't be a default here. Amazing 😂

edgarkhanzadian (approved):
Nice catch, thank you!

apps/mobile/src/features/browser/approver-sheet/stx/nonce-loader.tsx
tigranpetrossian: Happens to the best of us

tigranpetrossian:
This can be merged as it technically improves the existing code, but I'm open to suggestions on what the behavior should be when nonce retrieval fails. Falling back to 0 will cause the transaction to fail.

edgarkhanzadian:
@tigranpetrossian maybe forcing user to update nonce manually? Or putting a little "reload" button near the nonce value?

tigranpetrossian:
> @tigranpetrossian maybe forcing user to update nonce manually? Or putting a little "reload" button near the nonce value?

This will need a bit of design to get right, although it's quite an edge case, so I'm unclear on the priority. Going to merge and open up a discussion.

PR #1623, opened 2025-09-10
fix: current account persist

edgarkhanzadian:
- Persist current account
- Preselect first account of a newly added wallet

https://github.com/user-attachments/assets/47749786-b8d7-41ed-b66a-3389136ea8b3


apps/mobile/src/features/account/account-selector/account-selector-sheet.layout.tsx
```diff
@@ -74,7 +75,14 @@ export function AccountSelectorSheetLayout({
                 <WalletLoader fingerprint={account.fingerprint} key={account.id}>
                   {wallet => (
                     <AccountCard
-                      isSelected={account.id === currentAccount?.id}
+                      isSelected={
+                        !!currentAccount &&
```
pete-watters, line 79: Can you add a well named `const` for this so it makes more sense what's happening here? 

So it's if there is a `currentAccount` and it matches the settings ID?

apps/mobile/src/features/account/components/account-header.tsx
```diff
@@ -26,9 +28,9 @@ export function AccountHeader({ account, onPress }: AccountHeaderProps) {
       alignItems="center"
       gap="3"
     >
-      <AccountAvatar variant="sm" icon={icon} />
+      <AccountAvatar variant="sm" icon={accountData?.icon ?? 'home'} />
```
pete-watters: Are these fallbacks in case you have an empty wallet?

apps/mobile/src/hooks/use-migrate-current-account.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+
+import { useAccounts } from '@/store/accounts/accounts.read';
+import { useSettings } from '@/store/settings/settings';
+
+// This hook is temporary
```
pete-watters: The hook is here to stay but we need to set the data temporarily? 🤔 

So if users update, and we don't have a `currentAccount` selected, this pre-sets it for them? 

Does `handleSettingsRehydration` not handle this?

apps/mobile/src/store/accounts/accounts.read.ts
```diff
@@ -4,7 +4,7 @@ import { createSelector } from '@reduxjs/toolkit';
 
 import { RootState } from '..';
 import { useAppSelector } from '../utils';
-import { initalizeAccount } from './accounts';
+import { initializeAccount } from './accounts';
```
pete-watters, line 7: Nice 🕶️

pete-watters (approved):
I tested this and it's working well 👍  I added some Q's but no blockers

I wonder can we also use this persisted current account to default to when we are connecting to apps in the browser? I think users still need to select an account now

edgarkhanzadian:
> I wonder can we also use this persisted current account to default to when we are connecting to apps in the browser? I think users still need to select an account now

Very good point. I'm currently looking into a way of:
1. removing need for selecting an account as we can just use the current account
2. using separate local storage for webview for every account.

apps/mobile/src/features/account/account-selector/account-selector-sheet.layout.tsx
edgarkhanzadian: Done!

apps/mobile/src/features/account/components/account-header.tsx
edgarkhanzadian: well, probably defaulting here to random icon wasn't the best option here. It would probably be better to assert existence of accountData in this component.

apps/mobile/src/hooks/use-migrate-current-account.ts
edgarkhanzadian: nope, handleSettingsRehydration handles rehydration from redux persisted storage to redux state.

Users who already have a wallet setup in the app but don't have a persisted currentAccount would need something setup in redux, otherwise they would get locked out of their wallet.

edgarkhanzadian: Oh wait you're actually right! I did setup that part of the rehydration awhile ago and just forgot about it. It's a little broken, but if fixed should make better solution than this useEffect. Thanks for pointing it out!

PR #1631, opened 2025-09-17
feat: app config and asset visibility services

alexp3y:
This PR reworks asset visibility logic in `@leather.io/services` to support the updated token management feature from @edgarkhanzadian .

Three changes affecting visibility logic are implemented:

1. App config can now be retrieved from the backend API via an `AppConfigService` and used to incorporate a list of default enabled / visible assets
2. Live pricing data from the backend is incorporated in the visibility calculation (i.e. if an asset has a price, enable it) supporting dynamic addition of new assets to the default visible list
3. The `assetVisibility` user setting (updated via the token management UI) is now injected directly into the application layer via the `SettingsService` (no longer requires being passed down via the service call interface).

These 3 factors come together in a new `FungibleAssetVisibilityService`, which can determine the visibility of an asset w/o any additional parameters - e.g.: `isAssetVisibile(asset)` / `isAssetVisibleById(assetId)`

Visibility determination is used directly in the asset services (`RuneAssetService` / `Sip10AssetService`) to add a new `getVisibleAsset`, which optionally returns an asset if visible. This is in turn used in balance services to implement visibility filtering.

Replacing the full user `assetVisibility` list in the service request is a simple `includeHiddenAssets` boolean that allows callers/queries to override the default visibility filtering and return all assets.


packages/services/src/infrastructure/api/leather/leather-api.types.ts
```diff
@@ -1838,7 +1889,7 @@ export interface paths {
             'application/json': {
               [key: string]: {
                 /** @enum {string} */
-                id: 'bitflow' | 'sbtc-bridge' | 'alex' | 'velar';
+                id: 'bitflow' | 'sbtc-bridge' | 'alex' | 'velar' | 'arkadiko';
```
pete-watters, line 1892: Is this a rogue `'arkadiko'` here or is it needed for this PR?

packages/services/src/infrastructure/api/leather/leather-api.client.ts
```diff
@@ -667,4 +669,17 @@ export class LeatherApiClient {
       ? await fetchFn()
       : await this.cacheService.fetchWithCache(['leather-api-swap-dexes'], fetchFn);
   }
+
+  async fetchAppConfig({ signal, skipCache }: ApiRequestOptions = {}) {
```
pete-watters, line 673: Nice work adding this. Is this used/ can it be used to fetch a hosted our token list here somewhere too along with other config? 
 
That way we can decouple from the `extension` repo?

pete-watters (approved):
Nice work @alexp3y 🥇

packages/services/src/infrastructure/api/leather/leather-api.types.ts
alexp3y: This `'arkadiko'` was added to the backend OpenAPI schema since the last time we synced. This file gets generated by the `gen-api-types` script in `@leather.io/services` and it got picked up.

packages/services/src/infrastructure/api/leather/leather-api.client.ts
alexp3y: Yep. This call is being used now to power the `AppConfigService` (which is used to provide our default token list to the `FungibleTokenVisibilityService`).

Exactly, this "App Config" via the Backend API is totally decoupled from existing config and can start to replace / remove reliance on `extension`.

packages/services/src/assets/fungible-asset-visibility.service.ts
```diff
@@ -0,0 +1,50 @@
...
+      const runePrices = await this.leatherApiClient.fetchRunePriceMap({ signal });
+      return isDefined(runePrices[assetId.id]);
+    } else if (assetId.protocol === 'sip10') {
+      const sip10Prices = await this.leatherApiClient.fetchSip10PriceMap({ signal });
+      return isDefined(sip10Prices[getPrincipalFromAssetString(assetId.id)]);
+    } else {
+      return false;
+    }
+  }
+}
```
edgarkhanzadian, line 50: Awesome!

packages/services/src/balances/runes-balances.service.ts
```diff
@@ -151,15 +154,27 @@ export class RunesBalancesService {
...
+        return {
+          asset,
+          quote: createBaseCryptoAssetBalance(
+            baseCurrencyAmountInQuote(totalBalance, runeMarketData)
+          ),
+          crypto: createBaseCryptoAssetBalance(totalBalance),
+        };
+      }
+    } catch {
+      // ignored
```
edgarkhanzadian: Should we just remove the try catch block here if we don't process the error here?

edgarkhanzadian (approved):
Great work!

packages/services/src/balances/runes-balances.service.ts
alexp3y: I had replaced the `Promise.allSettled` with a `Promise.all` and so added this to handle exception filtering. But yeah, looking again I think `Promise.allSettled` is cleaner here, I'll change it back.

PR #1634, opened 2025-09-18
Minor tweaks and improvements

tigranpetrossian:
These are some of the fixes/improvements I did as part of the Swap implementation. Submitting separately to keep the upcoming Swap PRs focused.

1. Make pre-made asset avatars customizable
2. Add default bottom safe padding to sheet scrollables
3. Tweak the numpad styling 
4. Extract the search input from Send into a generic one
5. Add basic type guard helpers against the `CryptoAsset` type


pete-watters (approved):
Nice work 👍

PR #1635, opened 2025-09-19
feat(mobile): add gamma api to fetch SIP-9 metadata

pete-watters:
This PR adds the Gamma api to fetch SIP-9 metadata. This means we have more reliable data for SIP-9 assets and can show a lot more than we could before. 

✨ Features:

- [x] Sets up gamma API service
- [x] extends SIP-9 asset creation to add a mix of gamma and hiro data
- [X] Deprecates legacy code 

⚠️ **WARNING**:

This PR adds infrastructure in a clean way but impacts the visual state of collectibles. This is OK as they are feature flagged **off** and the UI is coming soon


packages/query/src/bitcoin/ordinals/inscription.utils.ts
```diff
@@ -1,33 +1,11 @@
...
+import { Inscription } from '@leather.io/models';
 import { createInscriptionAsset } from '@leather.io/utils';
 
-import { InscriptionResponseHiro } from '../../../types/inscription';
 import {
   BestInSlotInscriptionResponse,
   BestinSlotInscriptionBatchInfoResponse,
 } from '../clients/best-in-slot';
 
-export function createHiroInscription(inscriptionResponse: InscriptionResponseHiro) {
```
pete-watters, line 11: All of this code was used only for generating mocks in the extension.

I have [an extension PR](https://github.com/leather-io/extension/pull/6357) in _draft_ that refactors the extension to use BiS mocks instead as that's what we are actually using

packages/bitcoin/src/utils/bitcoin.utils.ts
```diff
@@ -359,7 +359,3 @@ export function getBitcoinInputValue(input: TransactionInput) {
   // logger.warn('Unable to find either `witnessUtxo` or `nonWitnessUtxo` in input. Defaulting to 0');
   return 0;
 }
-
```
pete-watters, line 362: @alexp3y - refactoring this away as its no longer used but I can keep it if you think it could be useful

packages/services/src/infrastructure/api/gamma/gamma-api.types.ts
```diff
@@ -0,0 +1,5 @@
+export interface GammaNftMetadataResponse {
```
alexp3y, line 1: Looks like this is unused.

packages/services/src/assets/sip9-asset.service.ts
```diff
@@ -23,8 +27,9 @@ export class Sip9AssetService {
   ): Promise<Sip9Asset> {
     const principal = getContractPrincipalFromAssetIdentifier(assetIdentifier);
     const tokenId = getNonFungibleTokenId(tokenHexValue);
-    const metadata = await this.stacksApiClient.getNftMetadata(principal, tokenId, { signal });
-    if (!metadata) throw new Error(`Sip9 Metadata Not Found: ${assetIdentifier}`);
-    return createSip9Asset(assetIdentifier, tokenId, metadata);
+    const hiroMetadata = await this.stacksApiClient.getNftMetadata(principal, tokenId, { signal });
```
alexp3y, line 30: If these are both always called we could bundle them in a `Promise.all` for concurrent execution.

packages/models/src/assets/asset.model.ts
```diff
@@ -136,11 +136,22 @@ export interface Sip9Asset extends BaseNonFungibleCryptoAsset {
   readonly assetId: string;
   readonly contractId: string;
   readonly tokenId: number;
-  readonly collection: string;
   readonly name: string;
   readonly description: string;
   readonly cachedImage: string;
   readonly cachedImageThumbnail: string;
+  readonly contentType: string;
+  readonly collection: {
```
alexp3y, line 144: Even though we're getting Collection data from Gamma directly, we should try not to borrow too many specific concepts when defining our own internal `Sip9Asset.collection` model, which should be as provider-agnostic as possible.

e.g. `locationUrl` is ambiguous but could maybe be replaced by something like `providerUrl` that holds a full provider link (Gamma or otherwise) if we ever want to direct users externally. `id` is specific to the Gamma Collection API rather than something that defines a collection here. We already know these are Sip9 collections, so maybe `contract` is sufficient? Also, the `unit` is an external unit that shouldn't have to be understood by apps. Maybe we can interpret that and capture the entire `floorPriceAmount` using a `Money` instance?

packages/services/src/assets/stacks-asset.utils.ts
```diff
@@ -52,21 +53,35 @@ export function getNonFungibleTokenId(hex: string): number {
...
+  const name = hiroMetadata?.name || gammaMetadata.item.name || assetName;
+  const description = hiroMetadata?.description || gammaMetadata.item.description || '';
+  const cachedImage =
+    hiroMetadata?.cached_image || gammaMetadata.item.asset_content?.content_url || '';
+  const cachedImageThumbnail =
+    (hiroMetadata as any)?.cached_thumbnail_image ||
+    gammaMetadata.item.asset_content?.content_url ||
+    '';
+  const contentType = gammaMetadata.item.asset_content?.content_type || '';
+  const collection = gammaMetadata.item.collection || hiroMetadata?.properties?.collection || {};
```
alexp3y, line 70: We should probably have a separate util here for mapping gamma collection objects to our internal `Sip9Asset.collection` type. The raw API response is snake case so pure assignment won't work for many of the fields. Also for `hiroMetadata.properties.collection`.

alexp3y (approved):
Nice 🦾

packages/models/src/assets/asset.model.ts
pete-watters: Thanks @alexp3y 👍  I'll clean this up as I work through populating the collectible details screen. It was easier to just use their object for now but I will curate it better

packages/services/src/infrastructure/api/gamma/gamma-api.types.ts
pete-watters: It is, good catch 🫨

pete-watters:
Thanks @alexp3y 👍  I'm going to merge this to get the wheels in motion for further updates but I will address the feedback in the next PR

PR #1636, opened 2025-09-19
feat(mobile): collectibles UI 

pete-watters:
This PR is a re-opened instance of [this PR](https://github.com/leather-io/mono/pull/1599) but with Gamma API infrastructure in place to fill gaps in data. 

✨ Features:

- [x]  adds fixes to render missing SIP-9 and BNS names correctly
- [x]  UI fixes to remove gaps between collectibles + add loading states
- [x]  refactor of collectible code to clean up, deprecate `serializer` and split by protocol
- [x]  handling of  images with empty src by adding a fallback
- [x]  move fetch of ordinal text source client side

https://github.com/user-attachments/assets/ebc926a6-b852-4610-844c-dfbdbc21be44


apps/mobile/src/features/collectibles/components/sip9.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+interface Sip9Props {
+  item: Sip9Asset;
+  height: number;
+}
+export function Sip9({ item, height }: Sip9Props) {
+  if (!item.cachedImage || item.cachedImage.trim() === '') return <FallbackImage />;
+
+  if (isBns(item.collection.name)) {
+    return <BnsImage alt={item.name} height={height} />;
```
alexp3y, line 17: With the now more reliable API data, do we still need this special BNS handling here?

apps/mobile/src/features/collectibles/render-collectible.tsx
```diff
@@ -1,19 +1,23 @@
...
-      mimeType={collectible.mimeType}
-      size={thumbnailSize}
-      src={collectible.src}
-    />
-  );
+interface RenderCollectibleProps {
+  item: NonFungibleCryptoAsset;
+  height: number;
+}
+export function renderCollectible({ item, height }: RenderCollectibleProps) {
```
alexp3y, line 12: Awesome 🦾

alexp3y (approved):

apps/mobile/src/features/collectibles/components/sip9.tsx
pete-watters: For `BNS` we render the `alt` overlaid on the image and they are never clickable so it's slightly different 

Screenshot 2025-09-23 at 17 07 18

PR #1637, opened 2025-09-19
feat: use sanity CMS for help center

camerow:
This is a large PR but was generally somewhat hard to do in a small change. This adds slight updates to the guides/help-center ui but mainly adds content and types for Sanity to power our support guides. 

https://github.com/user-attachments/assets/f6117eb5-5847-449f-8481-1ec9729fcec7


apps/web/app/components/markdown/markdown-components.tsx
```diff
@@ -0,0 +1,345 @@
...
+  // Robust host check for 'leather.io' rather than substring match
+  let isLeatherLink = false;
+  let normalizedHref = href;
+  if (href) {
+    // Use the URL constructor for absolute URLs. For relative URLs, fallback to default behavior.
+    const url = href.startsWith('http://') || href.startsWith('https://') ? new URL(href) : null;
+    // Check if the host is 'leather.io' or a valid subdomain (e.g. www.leather.io)
+    if (url && (url.hostname === 'leather.io' || url.hostname.endsWith('.leather.io'))) {
+      isLeatherLink = true;
+      normalizedHref = href.replace(url.hostname, '');
```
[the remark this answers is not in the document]
pete-watters: @camerow is CoPilot right here?

apps/web/app/layouts/footer/footer.tsx
```diff
@@ -7,9 +7,11 @@ function AppFooter() {
     <Footer>
       <Footer.Grid>
         <Footer.Column title="Dive deeper">
-          <Footer.Link href={externalLeatherNavigator.guides}>Guide</Footer.Link>
+          <Footer.Link href="/help-center">Help Center</Footer.Link>
```
pete-watters, line 10: Should we put this into a `const` or is that only for external links?

apps/web/app/pages/help-center/components/support-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+          <br />
+          <br />
+          Never share your Secret Key or personal information—not even with Leather staff. We will
+          never ask for it to resolve any issue. Keep it private and secure at all times.
+        </styled.p>
+      </Box>
+      <Box border="default" borderRadius="md" pt="space.01">
+        <styled.form
+          onSubmit={handleSubmit(onSubmit)}
+          action="https://webhook.frontapp.com/forms/50d1c719ca0a60c39f4a/B9hyvGi0oIyoqcSVKcYibYtptDsvXCCqTHtRjl1bdn2MW2GKAjWPVkQPNuKt3h1VqBXc0Rh0g7b9zMk8aHrMfFGRbgdId6Is5_Ti_DTzbfLgf7yWT3o1lES_JhnrfA"
```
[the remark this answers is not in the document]
pete-watters: @camerow  Maybe this should be in a git secret?

apps/web/app/pages/help-center/components/support-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+export function SupportForm() {
+  const {
+    register,
+    handleSubmit,
+    formState: { errors, isSubmitting },
+    reset,
+  } = useForm<SupportFormData>();
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async function onSubmit(data: SupportFormData) {
```
pete-watters: You're better than `// eslint-disable-next-line`. Should we omit the `data` if we aren't using it?

pete-watters (approved):
Nice work!

apps/web/app/layouts/nav/nav-item.layout.tsx
```diff
@@ -32,12 +35,27 @@ export function NavItem({ children, icon, href }: NavItemProps) {
...
+        {children}
+        {newTab && (
+          <Box
+            width="16px"
+            height="16px"
+            alignItems="center"
+            justifyContent="center"
+            mr="space.04"
+            _groupHover={{ display: 'flex' }}
+            display="none"
```
kyranjamie, line 48: intended?

apps/web/app/pages/help-center/components/support-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+    await new Promise(resolve => setTimeout(resolve, 1000));
```
kyranjamie: import `delay`

apps/web/app/pages/help-center/components/support-form.tsx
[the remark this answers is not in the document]
kyranjamie: even a const outside jsx would be better

apps/web/app/pages/help-center/components/support-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+  return (
+    <Box width="100%" maxW="600px">
+      <Box mb="space.07" bg="ink.background-secondary" p="space.04" borderRadius="md">
+        <Box display="flex" justifyContent="space-between" mb="space.02" pr="space.03">
+          <styled.h4 textStyle="label.01">Stay safe from scams</styled.h4>
+          <InfoCircleIcon variant="medium" />
+        </Box>
+        <styled.p textStyle="caption" color="ink.action-primary-hover">
+          Leather will never contact you first via direct messages on any platform. If someone
+          reaches out claiming to be from Leather and offering help, they're almost certainly a
```
kyranjamie:
```suggestion
          reaches out claiming to be from Leather and offering help, they're a
```

apps/web/app/pages/help-center/components/support-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+              type="text"
+              {...register('name', {
+                required: 'Name is required',
+                minLength: { value: 2, message: 'Name must be at least 2 characters' },
+              })}
+              width="100%"
+              p="space.03"
+              fontSize="14px"
+              bg="ink.background-primary"
+              outline="none"
```
kyranjamie, line 49: We have an `Input` in UI library? can we use that

apps/web/app/pages/help-center/components/support-form.tsx
```diff
@@ -0,0 +1,144 @@
...
+            {errors.email && <ErrorText message={errors.email.message} />}
+
+            <styled.input
+              placeholder="Subject"
+              id="subject"
+              type="text"
+              {...register('subject', {
+                required: 'Subject must be at least 5 characters',
+                minLength: { value: 5, message: 'Subject must be at least 5 characters' },
+              })}
```
kyranjamie: using zod schema to validate form would be more consistent with how we do elsewhere

kyranjamie (approved):
Thanks WIll

apps/web/app/pages/help-center/components/support-form.tsx
camerow: Haha, thanks for the vote of confidence. Yeah we can omit it - I don't recall why I even put that there. I always like having the variables for function arguments around as a way of kind of showing what's available - maybe sometime we use the `{ args: all }` setting for `eslint`...
Screenshot 2025-09-24 at 6 38 36 AM

apps/web/app/layouts/nav/nav-item.layout.tsx
camerow: The `display: "none"`? Yeah, this was the method I was using to hide/show the image when the element is hovered.

apps/web/app/components/markdown/markdown-components.tsx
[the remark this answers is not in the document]
camerow: Not sure but checking that it also works, so i'll go with it's implementation.

apps/web/app/layouts/footer/footer.tsx
camerow: That's just for external links.

fabric-8:
Thanks for pushing this forward @camerow ! Some more feedback (would have opened a PR, but I'm not able to getting local dev to work right now :(

**Questions**
- Category pages
  - Shouldn't we just keep the contact form on the category subpages as well? Idea being: People look into a topic, people figure that they don't see their issue being addressed → Easy to reach out to support
- Article pages
  - We dropped disclaimers on the left. Was that a concious move?

**Feedback**
- Navigation
  - Use [correct icon](https://www.figma.com/design/2MLHeIeL6XPVi3Tc2DfFCr/%F0%9F%A7%B1-Leather-%E2%80%93-Design-System?node-id=12599-48048&t=Dq8Btshay90ed7FY-1) for help center. Should be 'support'
  
- Stay safe from scams container
  - Color for title + info icon off
  - caption font style for body doesn't seem to get applied correctly
- Contact form 
  - See [thread](https://trustmachines.slack.com/archives/C05LRS7G44R/p1758825203149869) and @314159265359879 comments; All fields should be mandatory. Let's indicate that by adding a `*` at the end of each placeholder text
  - Claude is warning that `onsubmit` isn't actually sending any data?

- Can we do category card hover like [this](https://www.figma.com/design/axKU1wP6z2p77QIEOJ4idy/%F0%9F%8F%97%EF%B8%8F-Web-App---Content-Restructure?node-id=2386-5089&t=TbdSwPcpk6wLweAa-1)? so instead of turning the icon container white, just add a border on hover as the overall contrast changes are a bit intense now
image

- Article page
  - category chip font color needs to be default
  - date font size wrong

apps/web/app/pages/help-center/components/support-form.tsx
kyranjamie: Downvote this. Typescript tells us what's available, why leave it there if it isn't used?

.dependency-cruiser.cjs
```diff
@@ -249,6 +249,7 @@ module.exports = {
           'theme-web',
           'tsup',
           '.stories.',
+          'packages/cms/src/studio', // Sanity Studio is development-only
```
kyranjamie, line 252: Even if it's only used in dev, if the dependency is used in the studio, it should be a regular dependency, no?

kyranjamie:
Some of the markdown changes added here are needed for the changelog feature

apps/web/app/components/markdown/markdown.tsx
```diff
@@ -0,0 +1,17 @@
+import ReactMarkdown from 'react-markdown';
+
+import rehypeRaw from 'rehype-raw';
+import remarkGfm from 'remark-gfm';
+import { markdownComponents } from '~/components/markdown/markdown-components';
+
+export default function CustomMarkdown({ children }: { children: string | null | undefined }) {
```
kyranjamie, line 7:
```suggestion
export function CustomMarkdown({ children }: { children: string | null | undefined }) {
```

named exports by convention

kyranjamie:
In changelog work, I started with the same approach as you have here: markdown plugin & `react-markdown`.

Digging further, I realise the "sanity way" to do this is with the `block` field type, and https://github.com/portabletext/react-portabletext (the lib replacing @sanity/block-content-to-react).

Rather than markdown directly, you can describe blocks, and images, which'll render a WYSIWYG like the markdown plugin. This way you can add images via the editor directly, you can't as it is here.

```ts
    defineField({
      name: 'summary',
      title: 'Summary',
      type: 'array',
      of: [
        {
          type: 'block',
        },
        {
          type: 'image',
        },
      ],
    }),
```

Docs here https://www.sanity.io/docs/studio/block-content. I think using markdown is a bit of an anti-pattern in sanity, given docs are primarily based on a block/portable text structure. You can convert portable text -> markdown if needed

.dependency-cruiser.cjs
camerow: I don't believe so - otherwise we'd bundle it with the exported client and there is no need to do that.

camerow:
> In changelog work, I started with the same approach as you have here: markdown plugin & `react-markdown`.
> 
> Digging further, I realise the "sanity way" to do this is with the `block` field type, and https://github.com/portabletext/react-portabletext (the lib replacing @sanity/block-content-to-react).
> 
> Rather than markdown directly, you can describe blocks, and images, which'll render a WYSIWYG like the markdown plugin. This way you can add images via the editor directly, you can't as it is here.
> 
> ```ts
>     defineField({
>       name: 'summary',
>       title: 'Summary',
>       type: 'array',
>       of: [
>         {
>           type: 'block',
>         },
>         {
>           type: 'image',
>         },
>       ],
>     }),
> ```
> 
> Docs here https://www.sanity.io/docs/studio/block-content. I think using markdown is a bit of an anti-pattern in sanity, given docs are primarily based on a block/portable text structure. You can convert portable text -> markdown if needed

Yeah @kyranjamie I did see that - the reason I went with this is to simplify the legacy stuff and take it as is. We can refactor later and expand/iterate upon all of this.

PR #1638, opened 2025-09-24
fix: testnet approval flow [LEA-3187]

edgarkhanzadian:
We didn't pass the network setting when generating the txHex for the first time, so it defaulted to mainnet.
This PR fixes that.

Displaying postconditions would be implemented in a separate PR

https://github.com/user-attachments/assets/1c15f026-5003-43d6-aecc-67130017da3d

### DIFF
--- a/apps/mobile/src/features/browser/approver-sheet/stx/call-contract/call-contract.tsx
+++ b/apps/mobile/src/features/browser/approver-sheet/stx/call-contract/call-contract.tsx
@@ -43,7 +43,7 @@ export function CallContractApprover({
   const network = getNetworkFromRequestParams({ params: request.params, defaultNetwork });
   const stxRequestParams = getStxRequestParams(request.params, nonce);
   const [txHex, setTxHex] = useState<null | string>(null);
-  useCallContractTxHex({ request, stxRequestParams, setTxHex, accountId });
+  useCallContractTxHex({ request, stxRequestParams, setTxHex, accountId, network });
 
   const { list: accounts } = useAccounts();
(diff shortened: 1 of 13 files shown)


apps/mobile/src/features/browser/approver-sheet/stx/call-contract/call-contract.tsx
```diff
@@ -43,7 +43,7 @@ export function CallContractApprover({
   const network = getNetworkFromRequestParams({ params: request.params, defaultNetwork });
   const stxRequestParams = getStxRequestParams(request.params, nonce);
   const [txHex, setTxHex] = useState<null | string>(null);
-  useCallContractTxHex({ request, stxRequestParams, setTxHex, accountId });
+  useCallContractTxHex({ request, stxRequestParams, setTxHex, accountId, network });
```
kyranjamie, line 46: This is a peculiar API choice imo, why not have the hook return the state and setter, rather than the hook accept a setter.

kyranjamie (approved):
Thanks for fixing @edgarkhanzadian

pete-watters (approved):
Thanks @edgarkhanzadian

apps/mobile/src/features/browser/approver-sheet/stx/call-contract/call-contract.tsx
edgarkhanzadian: hmm yes good point, implemented it a while ago and would be nice to spend some time refactoring this

PR #1640, opened 2025-09-25
Swap state, closes LEA-3208

tigranpetrossian:
This is the implementation of the core swap functionality as a platform-agnostic state machine, that we can hopefully use to power Swap on both platforms.

It's not yet complete feature-wise: validation and execution are still work in progress. Submitting in chunks to make reviewing a bit easier. UI for most of this functionality is ready as well, coming in the next PR.


apps/mobile/package.json
```diff
@@ -79,6 +79,7 @@
     "@stacks/stacks-blockchain-api-types": "7.14.1",
     "@stacks/transactions": "7.0.5",
     "@tanstack/react-query": "5.59.16",
+    "@testing-library/react": "16.3.0",
```
edgarkhanzadian: should be a dev dependency

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    .shiftedBy(-money.decimals)
+    .toFixed(money.decimals, BigNumber.ROUND_HALF_UP)
+    .replace(/\.0+$/, '')
+    .replace(/(\.\d*?)0+$/, '$1');
+}
+
+export function areSameAssets(first: FungibleCryptoAsset, second: FungibleCryptoAsset) {
+  const firstAssetId = getAssetId(first);
+  const secondAssetId = getAssetId(second);
+  return firstAssetId.protocol === secondAssetId.protocol && firstAssetId.id === secondAssetId.id;
```
edgarkhanzadian: we have `isSameAsset` in utils package

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export function adjustAmountForDecimals(amount: string, maxDecimals: number): string {
+  if (!amount || amount === '') return amount;
```
edgarkhanzadian: nit: empty string is a falsy value, so it is going to be caught by `!amount` check anyway
```suggestion
  if (!amount) return amount;
```

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
```diff
@@ -0,0 +1,136 @@
...
+) {
+  return (data: AccountSwapAsset[]): AccountSwapAsset[] => {
+    return pipe(
+      data,
+      filter(swapAsset => isRelevantSwapAsset(swapAsset, assetSelectionType, isAssetAllowed)),
+      sortBy(
+        getCurrencyPriority,
+        swapAsset => -getAvailableQuoteBalance(swapAsset),
+        swapAsset => swapAsset.asset.symbol
+      )
```
edgarkhanzadian, line 90: A little bit confused by the sortBy functionality.
Does this mean that it:
1. first, sorts by currency priority
2. then sort by the quote balance (minus sign because lower values show up first in the sort)
3. sort alphabetically by the symbol

Should we have a comment here describing the sorting functionality? wdyt?

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function hasPositiveCryptoBalance(swapAsset: AccountSwapAsset): boolean {
+  const cryptoBalance = swapAsset.balance?.crypto.availableBalance.amount.toNumber() ?? 0;
+  return cryptoBalance > 0;
+}
+
+function isAllowedZeroBalanceAsset(asset: FungibleCryptoAsset) {
+  return isBtcAsset(asset) || isStxAsset(asset);
+}
```
edgarkhanzadian, line 129: Why should we allow zero balance for btc and stx as a base asset?

apps/mobile/src/features/swap/swap-state/use-derived-amounts.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import { currencyDecimalsMap } from '@leather.io/constants';
+import { Currency, MarketData, Money } from '@leather.io/models';
+import {
+  baseCurrencyAmountInQuote,
+  createMoneyFromDecimal,
+  quoteCurrencyAmountToBase,
+} from '@leather.io/utils';
+
+export function useDerivedAmounts(state: SwapInternalState, marketData: MarketData | undefined) {
```
edgarkhanzadian, line 25: nit: I'd suggest writing a little description on top of the hook to make it easier for other devs to understand its complex login
```
/**
*
*
*/
```

edgarkhanzadian (approved):
Great work and thank you for setting up proper tests for hooks!

alexp3y (approved):
Awesome implementation and great use of the state reducer pattern. Can't wait to see this in action.

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
tigranpetrossian: On one hand if you need a comment for something this small, it usually indicates that the code isn't self-explanatory enough.
On the other hand you inferred it correctly, plus it's in the PRD, and in the test spec.

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
tigranpetrossian: I blame the Frenchman.

apps/mobile/src/features/swap/swap-state/swap-state.utils.ts
tigranpetrossian: We've had lots of conversations about this back in the day for Send, and landed on a consensus that it's probably useful to show at least the native assets without holdings, just for exploratory purposes.
This is just maintaining consistency with Send.

Ironically I remembered this only after designing and implementing an empty state for the listing 🫠

apps/mobile/src/features/swap/swap-state/use-derived-amounts.ts
tigranpetrossian: Now this one is definitely worthy of a comment. Will add.

apps/mobile/package.json
```diff
@@ -160,12 +160,14 @@
     "@lingui/conf": "5.4.0",
     "@lingui/format-po": "5.4.0",
     "@lingui/macro": "5.4.0",
+    "@testing-library/react": "16.3.0",
```
pete-watters, line 163: 😮

apps/mobile/src/features/swap/swap-state/swap-state.reducer.ts
```diff
@@ -0,0 +1,174 @@
...
+        ...state,
+        selectingAsset: action.payload,
+      };
+    }
+    case 'CLOSE_ASSET_SELECTOR': {
+      return {
+        ...state,
+        selectingAsset: null,
+      };
+    }
```
kyranjamie, line 24: I presume this is done this way because it's not global redux state, its just a feature-level reducer? 

Still, would it be worth using the helper fns of RTK for a more consistent, minimal syntax? 

`createAction` etc

apps/mobile/src/features/swap/swap-state/swap.queries.ts
```diff
@@ -0,0 +1,128 @@
...
+    const { baseAsset, targetAsset, baseAmount, queryOptions } = params;
+    return useQuery({
+      queryKey: ['swap-quotes', { baseAsset, targetAsset, baseAmount }],
+      queryFn: ({ signal }) => service.getSwapQuotes(baseAsset, targetAsset, baseAmount, signal),
+      ...queryOptions,
+    });
+  };
+}
+
+export function createSwapExecutionDataQuery(service: SwapService, request: AccountRequest) {
```
kyranjamie, line 96: Where are these higher order hooks used?  This one I couldn't see used anywhere.

apps/mobile/src/features/swap/swap-state/use-swap-state.ts
```diff
@@ -0,0 +1,202 @@
...
+    dispatch({ type: 'CLEAR_ASSET_SELECTION' });
+  }, []);
+
+  const flipAssets = useCallback(() => {
+    dispatch({ type: 'FLIP_ASSETS' });
+  }, []);
+
+  const setBaseAmount = useCallback((amount: string) => {
+    dispatch({ type: 'SET_BASE_AMOUNT', payload: amount });
+  }, []);
```
kyranjamie, line 138: Is `useCallback` necessary here? These are such simple dispatch events. In [this article](https://kentcdodds.com/blog/usememo-and-usecallback) a more conservative use of them is recommended

apps/mobile/src/features/swap/swap-state/swap-state.reducer.ts
tigranpetrossian: I've contemplated not just using the helpers, but implementing this as an RTK slice altogether, even though this isn't global state. But the idea us to _potentially_ be able to use the exact same hook (not a copy of it), in the extension by placing it in a shared package, and the mechanics of sharing a slice across-apps are still unclear to me, especially with us using react-query for remote data. But I'd keep that option open--the conversion is fairly easy.

As for just `createAction` type helpers, I think the reduction of boilerplate is minimal there without the whole RTK abstraction.

apps/mobile/src/features/swap/swap-state/swap.queries.ts
```diff
@@ -0,0 +1,128 @@
...
+    queryOptions?: CustomQueryOptions<SwapAsset[]>;
+  }) {
+    const { queryOptions } = params ?? {};
+    return useQuery({
+      queryKey: ['base-swap-assets'],
+      queryFn: ({ signal }) => service.getBaseSwapAssets(signal),
+      ...queryOptions,
+    });
+  };
+}
```
kyranjamie, line 64: Perhaps a more "React" way to inject the `SwapService` would be to use a Context and read via another hook, avoiding the need for the HOH. 

What's your thinking behind this choice?

apps/mobile/src/features/swap/swap-state/use-swap-state.ts
tigranpetrossian: In my experience, for non-trivial state machines rapidly updating on user input stable callbacks make quite a bit of difference even in the browser, let alone in lower-end mobile devices. 
It's largely anecdotal, but I'm making an internal note to measure and re-validate this. Just saw a tweet demoing a pretty cool experimental RN perf profiler, good excuse to play with it.

apps/mobile/src/features/swap/swap-state/swap.queries.ts
tigranpetrossian: I'm not very fond of the approach I took here, the idea mostly being to be able to inject the services into the main hook in different platforms and in tests. This is mostly stems from my frustration with how we still don't follow a single approach for RQ hooks that tick all of the following:

* Services and other app-level dependencies are injected
* Hooks are customizable
* Query result is fully available (no opinionated/contrived wrappers)

That being said, this is very likely to change, but that will require us carefully designing and enforcing a single approach once and for all.

apps/mobile/package.json
tigranpetrossian: In our case can't use this for anything other than testing hooks. Which I consider a feature.

apps/mobile/src/features/swap/swap-state/use-swap-state.ts
kyranjamie:
```ts
const actions = useMemo(() => ({
  setSlippage: (slippage: number) => {
    dispatch({ type: 'SET_SLIPPAGE', payload: slippage });
  },
  toggleSlippageEditing: (allowed: boolean) => {
    dispatch({ type: 'TOGGLE_SLIPPAGE_EDITING', payload: allowed });
  },
  setNonce: (nonce: number) => {
    dispatch({ type: 'SET_NONCE', payload: nonce });
  },
  openAssetSelector: (type: 'base' | 'target') => {
    dispatch({ type: 'OPEN_ASSET_SELECTOR', payload: type });
  },
  closeAssetSelector: () => {
    dispatch({ type: 'CLOSE_ASSET_SELECTOR' });
  },
}), [dispatch]);
```

a single memo would have same perf benefits with slightly tidier syntax

tigranpetrossian: That's clever, didn't think of it. In my brain useMemo === data. Let me validate if we need this in the first place then refactor. Might do it in the next PR if profiling doesn't work out of the box.

apps/mobile/src/features/swap/swap-state/swap.queries.ts
tigranpetrossian: This particular one isn't used yet, the rest are used in the mother hook. It's a valid approach for decoupling from services and app-level account metadata, but questionable like I mentioned in the other thread. Will likely re-write once we have the whole thing working.

apps/mobile/package.json
pete-watters: 😄  Just some Friday joking. Take some Kent D into your life 🤗

tigranpetrossian:
Merging as I have changes stacked on top of it. Happy to address anything in the followup.

PR #1642, opened 2025-09-26
chore(mobile) additional collectible formats + base for details

pete-watters:
This PR adds more improvements to collectibles and adds the base setup for the details pages.

✨ Features:

- [x] Adds support for audio + video collectibles + SVG ordinals
- [x] Adds foundation for collectible details pages 
- [x] Fixes a bug with SIP9 not rendering

🤔 Future Enhancements:

- Add more detail to collectible details pages + finalise UI style
- Get ordinal audio loading properly

☎️  Demo:

https://github.com/user-attachments/assets/6c2ded5a-dc93-4b66-a50b-2fb66471d3af


packages/models/src/assets/asset.model.ts
```diff
@@ -130,6 +130,80 @@ export interface StampAsset extends BaseNonFungibleCryptoAsset {
   readonly stamp: number;
   readonly stampUrl: string;
 }
+
+export type Sip9ContentType =
```
pete-watters, line 134: Adding this AI generated comprehensive list of possible types. It's likely we can support a lot of them but not all of them are tested yet.

packages/models/src/assets/asset.model.ts
```diff
@@ -130,6 +130,80 @@ export interface StampAsset extends BaseNonFungibleCryptoAsset {
...
+  | 'text/javascript'
+  | 'application/javascript'
+
+  // Archives
+  | 'application/zip'
+
+  // Unknown/Fallback
+  | 'unknown';
+
+export type SupportedSip9ContentType =
```
pete-watters, line 187: This is a subset of the SIP9s we have available for testing that are confirmed as working

packages/ui/src/components/collectibles/native/collectible-audio.native.tsx
```diff
@@ -1,14 +1,80 @@
...
 interface CollectibleAudioProps {
+  src: string;
+  alt: string;
   size?: number;
+  onPress?: () => void;
 }
 
-export function CollectibleAudio({ size = 200 }: CollectibleAudioProps) {
+export function CollectibleAudio({ src, alt, size = 200, onPress }: CollectibleAudioProps) {
+  const htmlContent = `
```
pete-watters, line 15: For video / audio I am using a Webview with HTML audio / video. 

I tried using a few packages - `expo-av` / `expo-video` but couldn't get them to work properly. In the end my pal Claude came up with this approach which seems to work consistently well. I'm hopeful we can refactor it to use the same code for `web` also.

packages/services/src/infrastructure/api/gamma/gamma-api.client.ts
```diff
@@ -29,7 +29,7 @@ export class GammaApiClient {
         }
       );
 
-      return gammaNftMetadataSchema.parse(JSON.parse(res.data));
+      return gammaNftMetadataSchema.parse(res.data);
```
pete-watters, line 32: This was a mistake I made previously, over parsing. This meant no SIP-9's were showing 🫨

packages/ui/src/components/collectibles/native/inscription.native.tsx
```diff
@@ -5,28 +5,31 @@ import { CollectibleAudio } from './collectible-audio.native';
...
   src: string;
+  onPress?: () => void;
 }
 
-export function Inscription({ mimeType, name, height = 200, src }: InscriptionProps) {
+export function Inscription({ mimeType, name, height = 200, src, onPress }: InscriptionProps) {
   switch (mimeType) {
     case 'audio':
-      return <CollectibleAudio size={height} />;
+      return <CollectibleAudio alt={name} src={src} size={height} onPress={onPress} />;
```
pete-watters, line 21: I inscribed an audio ordinal and got it to render correctly however it times out at loading when trying to play it. I'll return to it

packages/services/src/assets/stacks-asset.utils.ts
```diff
@@ -54,20 +55,21 @@ export function createSip9Asset(
...
   tokenId: number,
   metadata: HiroNftMetadataResponse | null,
-  gammaMetadata: GammaNftMetadata
+  gammaMetadata?: GammaNftMetadata | null
 ): Sip9Asset {
   const assetName = getAssetNameFromIdentifier(assetIdentifier);
   const { metadata: hiroMetadata } = metadata || {};
-  const name = hiroMetadata?.name || gammaMetadata.item.name || assetName;
-  const description = hiroMetadata?.description || gammaMetadata.item.description || '';
+  const name = hiroMetadata?.name || gammaMetadata?.item.name || assetName;
```
pete-watters, line 62: @alexp3y: I'm going to refactor this next and apply [your suggestion](https://github.com/leather-io/mono/pull/1635#discussion_r2362716204) as I work on adding more details to the collectible details pages.

I'll also re-write the `SIP9Asset` - `collection` type then

packages/ui/src/components/collectibles/native/sip9.native.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { CollectibleImage } from './collectible-image.native';
+import { CollectibleVideo } from './collectible-video.native';
+
+export interface Sip9Props {
+  item: Sip9Asset;
+  height: number;
+  onPress?: () => void;
+}
+
+async function checkContentType(url: string): Promise<MediaInfo> {
```
pete-watters, line 16: I'm adding this code so that we can infer the content type from the headers of the source as sometimes that's more reliable than the metadata

apps/mobile/src/features/collectibles/collectibles-list.tsx
```diff
@@ -42,7 +44,16 @@ export function CollectiblesList({ currentAccount, header }: CollectiblesListPro
     <Screen.FlashList
       numColumns={2}
       data={collectibles}
-      renderItem={isSuccess ? ({ item }) => renderCollectible({ item, height }) : undefined}
+      renderItem={
+        isSuccess
```
tigranpetrossian, line 48: Just thoughts out loud here: I frequently struggle with presenting various query states with FlashList and friends, I think most of if this stems from those virtualization components trying to do too many things: empty state, header, etc.

Rendering things like page headers & tabs with `ListHeaderComponent` is less then ideal, we do this on the homepage too, and I'm exploring ways to render these separately but maintain the scrollability with some partial stickiness, i.e. header scrolls away, but tabs stick to the top.

tigranpetrossian (approved):
Previewable collectibles are 🔥

apps/mobile/src/features/collectibles/collectibles-list.tsx
pete-watters: Thanks for sharing. I agree it’s less than ideal and weird on those other pages having so much in the header. 

I’ll think about how to improve it also. A small change to how we show loading state for example could add a nice improvement

PR #1644, opened 2025-09-26
Mobile Swaps asset selection, closes LEA-3201

tigranpetrossian:
This sets up the Swap sheet, adds asset selection flow, as well as some UI building blocks.

https://github.com/user-attachments/assets/b7666356-5984-4212-a4e4-a615617cfdf1

Additionally, some minor refactoring:
* Moved the separate sBTC avatar icon into SIP-10 avatar.
* Recovered some long lost analytics tracking for opening the global sheets.


apps/mobile/src/features/swap/components/asset-selector/asset-selector-empty-state.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Image style={{ height: 180, width: 180 }} source={require('@/assets/stickers/wallet.png')} />
+      <Title>{t`No assets`}</Title>
+      <Description>
+        {t`Add funds to your wallet to start\nswapping between currencies.`}
+      </Description>
+    </Root>
+  );
+}
+
+function Root(props: BoxProps) {
```
pete-watters, line 60: Nice style here using `Root` as a name

apps/mobile/src/features/swap/components/currency-mode-switcher.tsx
```diff
@@ -0,0 +1,88 @@
...
+    switch (secondaryAmount.status) {
+      case 'idle':
+        return <Box width={56} height={20} aria-hidden />;
+      case 'pending':
+        return <PendingIndicator />;
+      case 'error':
+        // TODO: Extremely rare, needs better design.
+        return (
+          <Text variant="label02" color="ink.text-subdued">
+            –
```
pete-watters, line 36: Maybe we can render the `EmptyBalanceDisplay` from balance here? `-.--`

apps/mobile/src/features/swap/components/balance.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { Money } from '@leather.io/models';
+import { Text } from '@leather.io/ui/native';
+
+interface BalancePreviewProps {
+  balance: Money;
+}
+
+export function Balance({ balance }: BalancePreviewProps) {
+  // TODO: Use the Balance component
```
pete-watters, line 11: Maybe swap the `Balance` component in now?

apps/mobile/src/features/swap/components/panel.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return <Box px="5">{children}</Box>;
+}
+
+interface CardProps {
+  type: 'pay' | 'receive';
+  children: ReactNode;
+}
+
+export function Card({ type, children }: CardProps) {
+  const typeStyle = {
```
pete-watters, line 19: Nice!

pete-watters (approved):
Great work @tigranpetrossian . Really clean and easy to follow and also great 🙇

apps/mobile/src/features/swap/components/currency-mode-switcher.tsx
tigranpetrossian: That would nicer, 100%. I want to add a little hint there as well for when market data fails to load, though it is pretty rare. I'll add this to next PR which is all about amounts.

apps/mobile/src/features/swap/components/balance.tsx
tigranpetrossian: 👌 Next PR replaces this

apps/mobile/src/hooks/use-current-account.ts
```diff
@@ -0,0 +1,14 @@
+import { useSettings } from '@/store/settings/settings';
+
+import { AccountId } from '@leather.io/models';
+import { assertExistence } from '@leather.io/utils';
+
+export function useCurrentAccount(): AccountId {
+  const { currentAccount } = useSettings();
```
kyranjamie, line 7: Seems odd to me that we're pulling current account from settings?

apps/mobile/src/features/swap/components/asset-selector/asset-selector-header.tsx
```diff
@@ -0,0 +1,13 @@
+import { Box } from '@leather.io/ui/native';
+
+interface AssetSelectorHeaderProps {
+  children: React.ReactNode;
```
kyranjamie, line 4: there's `HasChildren` in ui lib

apps/mobile/src/features/swap/components/asset-selector/asset-selector-item.tsx
```diff
@@ -0,0 +1,73 @@
...
+        name={asset.name}
+        contractId={asset.contractId}
+        imageCanonicalUri={asset.imageCanonicalUri}
+        indicator
+      />
+    );
+  }
+
+  return null;
+}
```
kyranjamie, line 73: In several Swap feature child components there's conditional code like this for all asset types. Such that, were we to need to add another in future, the implementation will involve code changes across a wide surface area in many nested children.

Would a more flexible implementation not avoid such conditionals? Favouring some kind of component injection, perhaps at a root container level?

Could the swap feature remain entirely chain/protocol agnostic, where all chain-specific concerns are composed at a root/container component layer, rather than spread throughout the component tree?

kyranjamie (commented):
Looks great @tigranpetrossian 

When landing on the swap page, could the assets be pre-loaded to reduce loading time of the assets when opening the asset picker?

I wonder if it'd be handy to have "common pairs" listed below form, e.g. btc → sbtc so you can pre-fill the swap with that.

apps/mobile/src/hooks/use-current-account.ts
tigranpetrossian: Agreed, I guess this was done just because the rest of the account management is in settings too (both in UI and in the store). I intend to revisit that at some point.
For now I just made a safe wrapper hook around that not to deal with nullability, because account is guaranteed to exist in most flows.

apps/mobile/src/features/swap/components/asset-selector/asset-selector-header.tsx
tigranpetrossian: `HasChildren` is more of a `MaybeHasChildren`, so tend to avoid it when `children` must be required to be honest.

apps/mobile/src/features/swap/components/asset-selector/asset-selector-item.tsx
tigranpetrossian: This is very true. I was able to avoid this within the hook itself, less so in the UI code. At least two reasons:

1. Some of the app-level UI primitives need to do this, but those got pretty tangled up over time and live in the UI. Avatar here is a good example: those token-specific Avatars live in the `ui` package as separate components, it's problematic, but also quite a bit of work to refactor, so I ended up probably with 3-4 separate instances of doing this `if(someAsset)` just to render some icons here and there. 
I'll need to log a task to make single `AssetAvatar` at app-level. It would probably lead to deleting couple thousand lines of random icon usage.

2. More broad strategy-based approach is definitely worth it too, I've been thinking about this since working on send. I found that the best way to do this kind of thing is after the fact when all the weak points are clear, especially for someone like me piecing knowledge together as I go.

[Drafted an issue](https://linear.app/leather-io/issue/LEA-3219/use-strategy-based-approach-for-chain-specific-code) next to other de-scoped items to address once we have the whole picture.

tigranpetrossian:
> Looks great @tigranpetrossian
> 
> When landing on the swap page, could the assets be pre-loaded to reduce loading time of the assets when opening the asset picker?
> 
> I wonder if it'd be handy to have "common pairs" listed below form, e.g. btc → sbtc so you can pre-fill the swap with that.

Absolutely, both are planned. I de-scoped common pairs and recent suggestions from the initial version due to time constraints, but preloading is a low hanging fruit that I plan to add before the release.

There's a growing list of future improvements that I'm slowly adding to [this project](https://linear.app/leather-io/project/swap-post-mvp-e45cf1f5e3d2), 

image

apps/mobile/src/features/swap/components/asset-selector/asset-selector-header.tsx
kyranjamie: `ReactNode` can be `undefined` though. Did you mean ReactElement? 🤔

tigranpetrossian: [children is optional in HasChildren](https://github.com/leather-io/mono/blob/b765ced072fc147d029492927490cb27e6c130e3/packages/ui/src/utils/has-children.shared.tsx#L6-L5), always found that odd, but now I'm realizing this was probably a mistake.

tigranpetrossian: There's only about 40 usages, I can have the Frenchman check and fix this in a separate PR.

kyranjamie: Probs a mistake on my part, but if we remove the `?` the signature should be the same? 🤔

tigranpetrossian: It would error without children because of the [subtle difference](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAJQKYEMDGMA0cDez0wByEAJkgL5wBmUEIcA5FKhgwFBvAB2MSUV6JHACyKAJ4AjJAAkUAZwDCAC2AAbEsy642cXXDQr1mgPwAuRCyKkkbch269+guLMWGNSLTh16Daj1zm+BjEZLYcVACuXBjAEFqikkgAFGC0YHLmiVKuyv6aAJS4zDCRUFpckaqqdlExMHFayACOkcDMJKnpmS7yeUaeRTglZRVVNRxo8XLwIOJScAC8cAA82UgA9AB8bFNcM3DMre1IJEurLW0d20A) between `?` and explicit `X | undefined`

tigranpetrossian:
Merging to be able to PR the next chunk. Post-merge feedback is welcome, can address in any of the next ones.

PR #1645, opened 2025-09-29
Add amount entry to Mobile Swap, closes LEA-3202

tigranpetrossian:
Adds amount entry, live balance preview, amount presets, and currency mode switching.

https://github.com/user-attachments/assets/b44b9cd7-8330-4874-8d03-328b356d446e


apps/mobile/src/components/balance/constants.ts
```diff
@@ -1 +1 @@
-export const EmptyBalanceDisplay = '-.--';
+export const EmptyAmountPlaceholder = '-.--';
```
pete-watters, line 1: Thanks for renaming this also, much better 👍

pete-watters (approved):
Nice work!

PR #1653, opened 2025-10-01
fix(web): frontapp csp block

kyranjamie:
Add frontapp to web csp, connect to Front app API, add success dialog

### DIFF
--- a/apps/web/app/utils/support/support-form-action.ts
+++ b/apps/web/app/utils/support/support-form-action.ts
@@ -1,0 +1,32 @@
+import {
+  frontSupportMessageSchema,
+  postFrontAppSupportMessage,
+} from '~/utils/support/front-app-integration';
+
+export async function handleSupportFormAction(request: Request) {
+  const form = await request.formData();
+
+  const payload = {
+    name: form.getAll('name')[0],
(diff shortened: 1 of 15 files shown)


314159265359879:
@kyranjamie 

looks like there are some hardcoded values in the form with this version

image

tigranpetrossian (approved):

apps/web/app/utils/support/support-form-action.ts
```diff
@@ -0,0 +1,32 @@
...
+  const parsedData = frontSupportMessageSchema.safeParse(payload);
+
+  if (!parsedData.success) {
+    // eslint-disable-next-line no-console
+    console.error('Error submitting form:', parsedData.error);
+    return;
+  }
+
+  const resp = await postFrontAppSupportMessage(parsedData.data).catch(error => {
+    // fail silently
```
tigranpetrossian, line 25: Not sure what happens in the UI (reset?), but is this worth tracking?

kyranjamie: To be honest this implementation isn't very robust. I've just about picked it up and got it to a working state. This is a flaw I haven't addressed.

PR #1655, opened 2025-10-01
feat(mobile): collectible details content

pete-watters:
This PR updates collectible details pages to show more relevant information and manipulate the data to make it look better in the UI

✨ Features:

- [x] Adds a new feature flag for collectible details specifically
- [x] Adds code to show relevant metadata in a nicely formulated way
- [x] Fixes some problems raised by @alexp3y in a previous PR

☎️ Demo:

https://github.com/user-attachments/assets/fc450a4d-6434-4f1d-ac3c-119997608bc3

📔 Improvements:

I made some judgements on what to show and where to show it. This is open for interpretation and can be changed or improved but for now it's functional


apps/mobile/src/features/token/utils/render-collectible-details.tsx
```diff
@@ -0,0 +1,120 @@
...
+          </Text>
+        )}
+      </Pressable>
+    );
+  }
+
+  // If value is a URL, make it clickable
+  if (typeof stringValue === 'string' && stringValue.startsWith('https://')) {
+    return (
+      <Pressable onPress={() => Linking.openURL(stringValue)}>
```
pete-watters, line 53: I am using `Linking` for this as when I tried to use `useOpenURL` I was getting a warning as this isn't a hook/ function component.  I can refactor it if needed but I think it's OK for now as it works the same. 

```
 React Hook "useOpenURL" is called in function "renderValue" that is neither a React function component nor a custom React Hook function
```

packages/services/src/infrastructure/api/gamma/gamma-api.utils.ts
```diff
@@ -0,0 +1,71 @@
+import { z } from 'zod';
```
pete-watters: @alexp3y I updated this as per recommendation [here](https://github.com/leather-io/mono/pull/1635#discussion_r2362743029). Let me know if you recommend more improvements

apps/mobile/src/features/token/utils/render-collectible-details.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const items: React.ReactNode[] = [];
+  const nestedCards: React.ReactNode[] = [];
+
+  // Defined sort order for the collectible details
+  const sortOrder = ['name', 'tokenId', 'assetId', 'contractId', 'mimetype', 'contentType'];
+  getSortedEntries(obj, sortOrder).forEach(([key, value]) => {
+    // Skip undefined or empty string values
+    if (value === undefined || value === '') return;
+
+    const keysToSkip = ['description', 'floorPrice', 'floor_price_amount'];
```
pete-watters: @alexp3y I left out `floor_price` for now. We were discussing it [here](https://github.com/leather-io/mono/pull/1635#discussion_r2362716204). If we really want it I can convert it to `Money` but for now I've just filtered it out

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
       | 'model/gltf-binary'
     >
   | '';
+
+export interface Sip9Collection {
```
alexp3y: All of these `Sip9Asset` subtypes (`Sip9Collection`, `Sip9Owner`, etc...) could be moved out to a separate `sip9-asset.model.ts` file to reduce noise here.

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
...
+  id: string;
+  type?: string;
+  name: string;
+  isVerified: boolean;
```
alexp3y: Is verified by whom?

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
...
+  locationUrl: string;
```
alexp3y: Location of what?

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
...
+  totalItems?: number;
+  floorPrice?: {
+    amount: number;
+    unit: string;
+  };
+}
+
+export interface Sip9Owner {
```
alexp3y: This seems more like Gamma profile data than purely representing a SIP-9 owner?

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
...
+}
+
+export interface Sip9AssetContent {
+  contentUrl: string;
+  contentType: string;
+}
+
+export interface Sip9Attribute {
+  traitType: string;
+  displayType?: string;
```
alexp3y: What is a `traitType` or `displayType`?

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
...
+  value: any;
+  rarityPercent?: number;
```
alexp3y: Rarity according to what/whom?

apps/mobile/src/features/token/utils/render-collectible-details.tsx
```diff
@@ -0,0 +1,120 @@
...
+  // Special case for locationUrl in collection context
+  if (key === 'location_url') {
+    return t`Collection URL`;
+  }
+
+  // Replace underscores with spaces
+  let formatted = key.replace(/_/g, ' ');
+
+  // Add spaces between PascalCase words
+  formatted = formatted.replace(/([a-z])([A-Z])/g, '$1 $2');
```
kyranjamie: why not chain? wrapped in fns for readability


```ts
function replaceUnderscores(str: string) {
  return str.replace(/_/g, ' ');
}

function addPascalCaseSpaces(str: string) {
  return str.replace(/([a-z])([A-Z])/g, '$1 $2');
}

const formatted = addPascalCaseSpaces(replaceUnderscores("My_TestKey"));
```

packages/services/src/infrastructure/api/gamma/gamma-api.utils.ts
alexp3y: Looks good but this probably isn't the best spot for these map utils. 

API clients shouldn't really be concerned with anything outside of network interactions and their own API types.

The service layer acts as the adapter that takes the output from this (and any other) client calls and creates the `Sip9Asset`. So the asset services (and their utils) should be responsible for knowing how to map from the provider API types to our internal domain model.

apps/mobile/src/features/token/utils/render-collectible-details.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  // Capitalize first letter
+  return formatted.charAt(0).toUpperCase() + formatted.slice(1);
+}
+
+function renderValue(value: any, key: string): React.ReactNode {
```
kyranjamie:
```suggestion
function renderValue(value: unknown, key: string): React.ReactNode {
```

better to use unknown if narrowing

apps/mobile/src/features/token/utils/render-collectible-details.tsx
```diff
@@ -0,0 +1,120 @@
...
+export function renderCollectibleDetailsRecursively(
+  obj: Record<string, any>,
+  title?: string
+): React.ReactNode[] {
+  const items: React.ReactNode[] = [];
+  const nestedCards: React.ReactNode[] = [];
+
+  // Defined sort order for the collectible details
+  const sortOrder = ['name', 'tokenId', 'assetId', 'contractId', 'mimetype', 'contentType'];
+  getSortedEntries(obj, sortOrder).forEach(([key, value]) => {
```
kyranjamie: Could use a map?

apps/mobile/src/features/token/utils/render-collectible-details.tsx
```diff
@@ -0,0 +1,120 @@
...
+    if (value === undefined || value === '') return;
+
+    const keysToSkip = ['description', 'floorPrice', 'floor_price_amount'];
+    if (keysToSkip.includes(key.toLowerCase())) return;
+    // Filter out any keys that begin with 'genesis' (ordinals related keys)
+    if (key.toLowerCase().startsWith('genesis')) return;
+
+    const label = formatLabel(key);
+
+    if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
```
kyranjamie:
```suggestion
    if (isObject(value) && value !== null && !Array.isArray(value)) {
```

packages/services/src/infrastructure/api/gamma/gamma-api.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Sip9AssetContent, Sip9Attribute, Sip9Collection, Sip9Owner } from '@leather.io/models';
+
+import { gammaNftMetadataSchema } from './gamma-api.schema';
+
+type GammaNftMetadata = z.infer<typeof gammaNftMetadataSchema>;
+
+export function mapGammaCollection(
+  collection?: GammaNftMetadata['item']['collection']
+): Sip9Collection | undefined {
+  if (!collection) return undefined;
```
kyranjamie: is it legit that collection can be `undefined`, or is this just to handle async data that's missing?

IMO map is a slightly confusing term as it could indicate an iteration.

If absolutely necessary, wrapping, vibe coded example

```ts
export const optionalize =
  <A, R>(fn: (a: A) => R) =>
  (a: A | undefined): R | undefined =>
    a === undefined ? undefined : fn(a);

// Example
export function mapGammaCollection(
  collection: GammaNftMetadata['item']['collection']
): Sip9Collection {
  return {
    id: collection.id,
    type: collection.type,
    name: collection.name,
    isVerified: collection.is_verified,
    locationUrl: collection.location_url,
    totalItems: collection.total_items,
    floorPrice: collection.floor_price_amount
      ? { amount: collection.floor_price_amount.amount, unit: collection.floor_price_amount.unit }
      : undefined,
  };
}

export const mapGammaCollectionSafe = optionalize(mapGammaCollection);
```

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
       | 'model/gltf-binary'
     >
   | '';
+
```
alexp3y: Didn't catch the addition of the content type model above, but the `Sip9ContentType` seems very exhaustive - Where exactly does this list come from? What's the purpose of the `SupportedSip9ContentType`? Doesn't look like we reference it or do any validation checks anywhere.

packages/services/src/infrastructure/api/gamma/gamma-api.utils.ts
pete-watters: The response data is sometimes inconsistent, so there could be an `undefined` collection. I can update it as per recommendation, thanks!

apps/mobile/src/features/token/utils/render-collectible-details.tsx
pete-watters: I'll change this to make it easier to follow. I let Claude create this whole section. 

These days I am veering less towards human readability and increased code generation

packages/models/src/assets/asset.model.ts
```diff
@@ -209,6 +209,59 @@ export type SupportedSip9ContentType =
...
+}
+
+export interface Sip9Attribute {
+  traitType: string;
+  displayType?: string;
+  value: any;
+  rarityPercent?: number;
+}
+
+export interface Sip9Details {
```
alexp3y: Solid foundation here on the `Sip9Asset` model, but it still feels pretty coupled to Gamma's API (e.g., `rarityRank`, `floorPrice`, `isVerified` still directly mirror their structure).

We should try to delineate between the on-chain fields that represent the essence of a SIP-9 (e.g., `name`, `contractId`, `tokenId`, `imageUri`) and any enriched data computed by a specific platform/marketplace (e.g., `rarityRank`, `floorPrice`, `isVerified`).

One idea could be to prefix enriched fields with something like "marketplace" or "provider" (e.g., `marketplaceRarityRank`, `marketplaceFloorPrice`), or just encapsulate them all in a parent structure like `Sip9ProviderData`. 

Even then, we should approach the "marketplace"/"provider" data like a universal remote - We should design it to work for Gamma now but also, in theory, so that it can plug into any provider in the future without needing changes.

alexp3y (commented):
Great start pulling together the additional SIP-9 data. Added some comments that build on the idea of making the `Sip9Asset` more provider-agnostic.

Also think we could refine which fields get presented in the UI before releasing, feels like a bit of a data dump atm. Fields we want to present and naming conventions there could also be a good guiding force for the model and help reduce any Gamma specificity.

packages/models/src/assets/asset.model.ts
pete-watters: I added that list in [this PR](https://github.com/leather-io/mono/pull/1642/files#r2382514401). It is very exhaustive, I added it as a reference to all possible types but maybe I should remove it? I got claude to tell me all the possible permutations a SIP9 could be. 

`SupportedSip9ContentType` is the ones I was able to actually test myself and make sure they work OK. The thinking was that if someone has a collectible that's not showing then it would be easy to extend `SupportedSip9ContentType` by adding the new type. 

I use `SupportedSip9ContentType` as a type in `sip9.native.tsx` 
```
interface MediaInfo {
  contentType: SupportedSip9ContentType;
  isVideo: boolean;
  isImage: boolean;
  isAudio: boolean;
}
```

In there it's used to help decide which component to render based on `ContentType`. 

I'm happy to go with your recommendation on what's best here

packages/models/src/assets/asset.model.ts
pete-watters: My thinking here was to get something working quickly that we can refine later. I'll see what I can do to improve it

pete-watters:
> Great start pulling together the additional SIP-9 data. Added some comments that build on the idea of making the `Sip9Asset` more provider-agnostic.
> 
> Also think we could refine which fields get presented in the UI before releasing, feels like a bit of a data dump atm. Fields we want to present and naming conventions there could also be a good guiding force for the model and help reduce any Gamma specificity.

Thanks @alexp3y . I'll see what I can do to clean it up. 

My idea here was that this is good enough for now and we could have it added behind a feature flag. Then we can test it out and see what fields we actually want to show and what to exclude. I already filtered out a lot of them based on a short conversation with @camerow but had planned to discuss it more with the design team. 

It is 100% a data dump for now as I believe that's good enough for a first iteration until we decide how best to refine it or if we even want to dedicate more time to that at all

packages/models/src/assets/asset.model.ts
pete-watters: It's the collection attribute group rarity percentage according to Gamma

packages/models/src/assets/asset.model.ts
pete-watters: I have no clue, by Gamma I assume 🤷

packages/models/src/assets/asset.model.ts
pete-watters: It's the collection main page location on `Gamma`

packages/models/src/assets/asset.model.ts
pete-watters: So to elaborate using [this example](https://github.com/leather-io/mono/pull/1655#discussion_r2413157714)

That monkey is one of `4.61%` in the collection with a Green t-shirt 

Screenshot 2025-10-08 at 13 22 33

packages/models/src/assets/asset.model.ts
pete-watters: These map to `attribute_groups.attributes.type` and `attribute_groups.attributes.label` in the Gamma API. 

Different NFTs can have individual traits for example a monkey with a hat. 

In the Hiro API these fields are called `trait_type` and `display_type` so I used those names.  

- `displayType` / `label` is the type of attribute the NFT has
- `traitType` is a string usually on Gamma and is often set as `simple`. 

So for example for this [Mutant Monkey](https://gamma.io/stacks/nfts/SP125J1ADVYWGWB9NQRCVGKYAG73R17ZNMV17XEJ7.mutant-monkeys_2396) 

This is what we get for attribute groups:

```
[
    {
        "title": "Attributes",
        "attributes": [
            {
                "type": "simple",
                "label": "Background",
                "value": "MM1 Yellow",
                "rarity_percent_of_100": 7.65
            },
            {
                "type": "simple",
                "label": "Clothes",
                "value": "MM1 Green Tshirt",
                "rarity_percent_of_100": 4.61
            },
            {
                "type": "simple",
                "label": "Eyes",
                "value": "MM1 Eye Patch",
                "rarity_percent_of_100": 6.66
            },
            {
                "type": "simple",
                "label": "Fur",
                "value": "MM1 Pink",
                "rarity_percent_of_100": 6.08
            },
            {
                "type": "simple",
                "label": "Hat",
                "value": "MM1 Viking Helmet",
                "rarity_percent_of_100": 3.6
            },
            {
                "type": "simple",
                "label": "Level",
                "value": "1",
                "rarity_percent_of_100": 45.42
            },
            {
                "type": "simple",
                "label": "Mouth",
                "value": "MM1 Pizza",
                "rarity_percent_of_100": 7.29
            },
            {
                "type": "simple",
                "label": "SGR",
                "value": "MM1 100%",
                "rarity_percent_of_100": 67.19
            },
            {
                "type": "simple",
                "label": "Trait Count",
                "value": "6",
                "rarity_percent_of_100": 40.55
            }
        ]
    }
]
```

`displayType` seems to have more potential values but I didn't handle them exhaustively:
```
displayType

Defines how the trait value should be rendered or interpreted by marketplaces and wallets. Common display types include:

string (default) - Regular text value
number - Numeric value
boost_number - Numeric boost/modifier (often shown with +/-)
boost_percentage - Percentage boost (shown with %)
date - Unix timestamp to be displayed as a date
```

apps/mobile/src/features/token/utils/render-collectible-details.tsx
pete-watters: I decided to just exclude this from our model altogether for now until we decide we want to show it

packages/models/src/assets/asset.model.ts
pete-watters: Ditching this

packages/models/src/assets/asset.model.ts
pete-watters: I've changed this to be `collectionExplorerUrl` and made it work as a clickable link now. That was a bug you helped me fix

https://github.com/user-attachments/assets/d1371651-ce79-45c2-a35a-55ee7036259e

packages/models/src/assets/asset.model.ts
pete-watters: I am ditching this and not showing it

packages/models/src/assets/asset.model.ts
pete-watters: I've renamed these as:

- attributeTraitType
- attributeValue
- attributeRarityPercent

Hopefully that gives better semantic meaning

packages/models/src/assets/asset.model.ts
pete-watters: I fixed a bug where these weren't actually rendering properly. 

Now the NFT attributes list properly if available

https://github.com/user-attachments/assets/454bb898-ef9b-481f-b677-a4a1ecf05da7

They look kind of ugly though. I will get a design review and try get some help rendering them in a nicer way using labels or something

pete-watters:
@alexp3y & @kyranjamie - thanks for the feedback and patience here. 

I have made changes as requested and I think it's looking a lot better now so I appreciate the tips.  If you spot anything else you think I can improve on let me know!

@alexp3y : 
- I moved the data manipulation to the `asset` service and renamed them from `map` to `transform`
- I got rid of a lot of fields we probably don't need
- Others I re-labelled so they are more intuitive
- I wrapped the additional fields in their own interface - `Sip9ProviderData`

kyranjamie (approved):

packages/models/src/assets/asset.model.ts
alexp3y: Nice, thanks for explaining. Didn't realize this was actually part of the SIP-016 standard.

packages/models/src/assets/asset.model.ts
alexp3y: Makes sense. I think the thing that feels off for me is that in the mapper we cast any value we get to `SupportedSip9ContentType`, whether it's supported or not.

Maybe whether something is "supported" is actually a concern of the specific UI that's displaying it? So, the model `contentType` field just holds any possible `Sip9ContentType` and then the supported determination is made elsewhere.

packages/models/src/assets/sip9-asset.model.ts
```diff
@@ -0,0 +1,128 @@
...
+  attributes?: Sip9Attribute[];
+}
+
+interface Sip9ProviderData {
+  attributes?: Sip9Attribute[];
+  contentType: SupportedSip9ContentType;
+  details: Sip9Details;
+}
+
+export interface Sip9Asset extends BaseNonFungibleCryptoAsset {
```
alexp3y, line 117: Awesome improvement. Looks a lot cleaner. Seeing this now I'm kind of second-guessing the Provider-speicifc structure and wondering if a lot of this could be consolidated further (as there's a lot of repitition). Happy to keep as-is for now but lmk if you have time for a quick chat later and we can discuss.

alexp3y (approved):

pete-watters:
Thanks @alexp3y  and @kyranjamie 👍 

@alexp3y and I had a pair session today and discussed further improvements to the models. I am working on that now but am going to merge this PR first as it's already pretty big. I'll add a fast follow up with model changes.

PR #1658, opened 2025-10-06
feat: add balance analytics

edgarkhanzadian:
- Send track events with scaled balances
- id users by device id

### DIFF
--- a/apps/mobile/src/utils/analytics-hooks.ts
+++ b/apps/mobile/src/utils/analytics-hooks.ts
@@ -1,0 +1,63 @@
+import { useEffect } from 'react';
+
+import { useAccountTotalBalance } from '@/queries/balance/account-balance.query';
+import { useBtcAccountBalance } from '@/queries/balance/btc-balance.query';
+import { useStxAccountBalance } from '@/queries/balance/stx-balance.query';
+import { makeAccountIdentifer } from '@/store/utils';
+import { analytics } from '@/utils/analytics';
+
+import { AccountId, Money } from '@leather.io/models';
+import { convertAmountToBaseUnit, isDefined, scaleValue } from '@leather.io/utils';
+
+export function useAccountScaledBalanceAnalytics({
+  currentAccount,
+}: {
+  currentAccount: AccountId;
+}) {
+  const { fingerprint, accountIndex } = currentAccount;
+  const accountId = makeAccountIdentifer(fingerprint, accountIndex);
+  const btcBalance = useBtcAccountBalance(fingerprint, accountIndex);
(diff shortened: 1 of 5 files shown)


apps/mobile/src/utils/analytics.ts
```diff
@@ -16,8 +23,30 @@ const segmentClient = createClient({
...
+ * */
+async function identify(userId?: string, traits?: UserTraits): Promise<any> {
+  await segmentClient.identify(userId, traits);
+  if (userId) {
+    await mixpanel.identify(userId);
+  }
+  if (traits) {
+    mixpanel.getPeople().set(traits);
+  }
+}
```
kyranjamie: I don't quite get what's going on here. Data passed to segment goes to mixpanel, so why would we need to handle both?

apps/mobile/src/utils/analytics-hooks.ts
```diff
@@ -0,0 +1,63 @@
...
+        btcBalance: scaledBtcAvailableBalance,
+      });
+    }
+  }, [
+    accountId,
+    scaledBtcAvailableBalance,
+    scaledStxAvailableBalance,
+    scaledStxLockedBalance,
+    scaledUsdBalance,
+  ]);
```
kyranjamie, line 62: Given that there are so many variables here, esp. market data ones, isn't this likely to trigger absolutely loads of balance updates events?

edgarkhanzadian: i made sure i'm providing numbers to useEffect instead of objects + i'm checking for all values existing before sending the event. While testing it seemed to me to happen only when changing accounts or when balance change, which seemed to be ok for me

apps/mobile/src/utils/analytics.ts
edgarkhanzadian: i want to make sure that when user updates the app we wouldn't just create a new id for it in mixpanel and lose all of the tracking with anonymous id we did with segment.

But yea you're right, just providing deviceId to segment should work for now as that would also set the device id in mixpanel. Will remove the mixpanel sdk for now

edgarkhanzadian: i will make the segment -> mixpanel migration in a separate PR after extension is moved to mono

camerow (approved):
Looking good to me Edgar, let's make sure this gets to extension as well.

kyranjamie (approved):

edgarkhanzadian:
@camerow yeap, i've made some changes to it already but currently waiting for it to get merged into mono first

PR #1660, opened 2025-10-07
feat(mobile): improve swap amount value positioning, add faux caret

tigranpetrossian:
This tweaks the positioning of the Swap amount field and adds a fake caret

https://github.com/user-attachments/assets/ade04e9b-0962-4c2b-9613-53e4d8dd10bf

### DIFF
--- a/apps/mobile/src/features/swap/components/amount-field/amount-field-caret.tsx
+++ b/apps/mobile/src/features/swap/components/amount-field/amount-field-caret.tsx
@@ -1,0 +1,69 @@
+import Animated, {
+  cancelAnimation,
+  useAnimatedReaction,
+  useAnimatedStyle,
+  useSharedValue,
+  withDelay,
+  withRepeat,
+  withSequence,
+  withTiming,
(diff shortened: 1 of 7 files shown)


kyranjamie (approved):
Only UX question is whether this makes users think they can reposition it?

I like the look, though. Could be bigger I reckon

tigranpetrossian:
> Only UX question is whether this makes users think they can reposition it?
> 
> I like the look, though. Could be bigger I reckon

Thanks Kyran. It probably does to an extent, but then I doubt users would do a lot of precise editing, especially for an amount and without a fine cursor/keyboard. Considered plugging the numpad into a real text input but that comes with all kinds of a11y limitations, inability to adjust vale to fit, and digit animations that I hope to add sometime this century.

tigranpetrossian:
@kyranjamie  I'll look into adjusting the size too. It's a bit of a pain, because RN's TextLayout often misreports metrics so it's mostly pray & spray.

PR #1665, opened 2025-10-07
refactor(web): fully remove the notion based build steps from app

camerow:
This PR removes the dependency on the Notion based content system from the web app completely. I opted for keeping some JSON files with the strings around simply because it made the refactor a bit easier. We could either go with a CMS option in the future or a more hardcoded option (which I think makes sense for the app type features). 

Another temporary trade off i've made is hard-coding some post slugs - better that they were dynamic but I also see this stuff changing quite a bit in the nearish future so I didn't think the investment into something more dynamic and scalable was currently worth it.


apps/web/app/pages/stacking/components/stacking-provider-table.tsx
```diff
@@ -498,10 +500,10 @@ export function LiquidStackingProviderTable(props: HTMLStyledProps<'div'>): Reac
...
     ],
     [
-      posts.stackingRewardsTokens,
-      posts.stackingPoolFees,
-      posts.historicalYield,
-      posts.totalLockedValueTvl,
+      // Removed deps: stackingRewardsTokens,
+      // Removed deps: stackingPoolFees,
+      // Removed deps: historicalYield,
+      // Removed deps: totalLockedValueTvl,
```
kyranjamie: remove not just comment out?

apps/web/app/pages/stacking/components/stacking-provider-table.tsx
```diff
@@ -235,11 +234,11 @@ export function StackingProviderTable(props: HTMLStyledProps<'div'>): ReactEleme
...
-      posts.stackingRewardsTokens,
-      posts.stackingPoolFees,
-      posts.historicalYield,
-      posts.totalLockedValueTvl,
-      posts.stackingMinimumCommitment,
+      // Removed deps: stackingRewardsTokens,
+      // Removed deps: stackingPoolFees,
+      // Removed deps: historicalYield,
+      // Removed deps: totalLockedValueTvl,
+      // Removed deps: stackingMinimumCommitment,
```
kyranjamie: Same here

apps/web/app/components/learn-hover-card.tsx
```diff
@@ -0,0 +1,110 @@
...
+          textDecoration: 'underline' as const,
+          textDecorationStyle: 'dotted' as const,
+          textDecorationColor: 'ink.border-default',
+          textUnderlineOffset: '2px',
+          cursor: 'help' as const,
+        }
+      : { textStyle };
+
+    return <StyledTag {...props}>{label}</StyledTag>;
+  }
```
kyranjamie, line 52: Not good react to do this (components written as function inside render of other component), but hopefully this is code that'll be replaced

apps/web/app/components/learn-info-hover-icon.tsx
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Handle click on the info icon button
+   * Navigates to the same URL as the "Learn more" link in the tooltip
+   */
+  function handleIconClick(e: React.MouseEvent): void {
+    e.stopPropagation();
+    if (article?.slug) {
+      const isUrl = URL.canParse(article.slug);
+      const href = isUrl ? article.slug : getPostHref(article.slug);
+      window.location.href = href;
```
kyranjamie: Should use `openExternalLink` helper with better validation

apps/web/app/components/learn-hover-card.tsx
```diff
@@ -0,0 +1,110 @@
...
+   * Handle click on the info icon button
+   * Navigates to the same URL as the "Learn more" link in the tooltip
+   */
+  function handleIconClick(e: React.MouseEvent): void {
+    e.stopPropagation();
+    if (article?.slug) {
+      const isUrl = URL.canParse(article.slug);
+      const href = isUrl ? article.slug : getPostHref(article.slug);
+      window.location.href = href;
+    }
```
kyranjamie, line 62: same here re `openExternalLink`

kyranjamie (commented):
Thanks Will. Hard coding is fine if it removes the dependency pulling in data from notion. We can also bring down whatever services were running to serve the posts data.

apps/web/app/components/learn-hover-card.tsx
camerow: yeah it's a bit messy agreed, low impact component that likely does go away so i do imagine this gets replaced.

apps/web/app/pages/stacking/components/stacking-provider-table.tsx
camerow: done

apps/web/app/pages/stacking/components/stacking-provider-table.tsx
camerow: done

apps/web/app/components/learn-info-hover-icon.tsx
camerow: Yeah good call - used `useNavigation` actually since this can all be rendered by the cms.

kyranjamie (approved):

PR #1668, opened 2025-10-08
fix: uneconomical utxos threshold

edgarkhanzadian:
Lower the threshold a lil'


alexp3y (approved):
🔥

kyranjamie (changes requested):
Should be 547, 500 can be rejected by network https://bitcoin.stackexchange.com/questions/86068/how-was-the-dust-limit-of-546-satoshis-was-chosen-why-not-550-satoshis

kyranjamie (approved):

PR #1670, opened 2025-10-09
Proposal: `@leather.io/state` package

kyranjamie:
> Try out Leather build 4144880 — [Extension build](https://github.com/leather-io/mono/actions/runs/18717822530), [Test report](https://leather-io.github.io/playwright-reports/experimental/state-pkg), [Storybook](https://experimental/state-pkg--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=experimental/state-pkg)

<details>

<summary>Original proposal</summary>

I'm not married to this idea, but putting it forward to start a discussion.

Next on my plate is to add multi-wallet support to the extension. I want to follow a similar design to the mobile implementation, based on a flat hierarchy of wallets and accounts. Following the same model in both apps enables us to share the underlying implementation. How should we do this?

With the reducer pattern, the state is self-contained and side-effect free. One solution, proposed here, is to have a `state` package, responsible for the shared state management between separate apps. Such that both can compose together their respective state trees, some of which is shared. An assumption here is that the functionality of these reducers will be the same cross-platform.

I'm not sure however one feels about this but, as I understand, the properties of reducer state doesn't lend itself to living in the services package. This would indicate it's a unit of functionality that can coexist with, and depend on, other services, and that these should saved for wiring up complex application abstractions rather than a stateless functions. Speak up if you think differently.

Thinking is that it'd live somewhere here in the architecture

image

</details>

In this PR I've created a package to home reusable state slices between the extension and mobile apps.


alexp3y:
This sounds pretty great. Would be nice to have more unified state definitions and standardized logic. Currently services rely on app-specific adapters (e.g. `useAccountAddresses`) to read app state consistently.

Agree that services aren't appropriate for state mutations and a parallel package like this would be a solid addition.

We do have a pattern of injecting some read-only state references into services (e.g. current user quote currency) via app-specific service implementations (e.g. `MobileSettingsService`).

Just an idea, but this could allow us to expand on that in some nice / interesting ways and prevent the need for adapters and passing tons of state (e.g. current account data) into service calls.

Would be curious to know how you see this package interacting with existing state definitions and how we could migrate, but I think it's a great idea to keep exploring.

kyranjamie:
@alexp3y I can imagine we'll want more state services, binding some selectors-driven services in each app.

Is your thinking that say, to get the current account balance, rather than pass the state to `BalanceService.getBalance(account)` everytime, we read state from a `CurrentAccountStateService`? I'm not familiar enough with inversify syntax but vibed example:

```ts
// mobile
// A Redux selector of account state
export const currentAccountSelector = (s: RootState) => s.currentAccount;
// Stream of account data, internally subscribes to store.subscribe()
const currentAccount$ = selectorToObservable(currentAccountSelector);

// Bind account state stream to inversify set up
container
  .bind(Types.CurrentAccountStateService)
  .toDynamicValue((ctx) => currentAccount$)
  .inSingletonScope();

function useCurrentAccountBalance() {
  const { currentAccountBalance$ } = getCurrentAccountBalanceService();
  // Note: Obsevable doesn't have `getValue`, `BehaviourSubject`'s do so not perfect example
  return useSyncExternalStore(currentAccountBalance$.getValue, currentAccountBalance$.subscribe);
}

// services
@injectable()
export class CurrentAccountBalanceService {
  constructor(
    @inject(Types.CurrentAccountStateService) private readonly accountState$: Observable<AccountState>,
    private readonly stxBalancesService: StxBalancesService,
  ) {}

  currentAccountBalance$(): Observable<string> {
    return this.currentAccount$.pipe(
      switchMap(account => this.stxBalancesService.getStxAccountBalance(account))
    );
  }
}
```

I've used `Observable`s in this example because I think as we continue to adopt services, especially if they depend on state, we have no other choice. `Promise`s don't suit tracking data over time. IMO the selling point of the services becomes much more apparent to me considering them as exposing streams of data, rather than just flat function calls (that could equally be wrapped up in a manner of other ways).

Currently, our services are API wrappers to aggregate data. No React compatibility issues there. If we move in this direction, we'll run into the issue that either the services don't know to response to state changes, or React doesn't know to re-render when services state updates. Observables + useSyncExternalStore is the best solution I think. [Demo Observable service here](https://github.com/leather-io/mono/compare/experimental/observable-services?expand=1).

edgarkhanzadian:
I think separating core redux slices into a separate package is a smart idea. And then we could compose it with more platform specific slices in the each app.

1. Are we going to share only slices? I think it would be better to keep it that way to not introduce react specific code in the store package.
2. How are we going to handle migrations in extension?
3. I might not be that much familiar with observables, but why do we not have the choice of not using those in the store package?

kyranjamie:
> Are we going to share only slices? I think it would be better to keep it that way to not introduce react specific code in the store package.

Yeah there should be no dependency on react, just redux toolkit.

> How are we going to handle migrations in extension?

With great difficulty, care and testing

> I might not be that much familiar with observables, but why do we not have the choice of not using those in the store package?

Sorry, I didn't make the best justification in last post. Currently the services are just async fn calls wrapped in `useQuery`, operating in this mode, there's no real issue with Promises or how the set up works. If we want to pursue more tightly integrated use of the services layer, such as injecting and reading state, then I think we'll run into issues with `Promise`s. 

Consider the example I've shown where we read `currentAccount` without Observables. When user switches account, how does services layer know that data has changed, and that all the data in `CurrentAccountBalanceService` needs to be refreshed? It doesn't if it's a promise until it's called again. If all the queries have the account in the `queryKey`, the UI would update, but this is brittle. The service is trusting that the consumer knows to refresh at the right time. For me, having worked with Angular apps and Rxjs in the past, the model of consuming streams becomes quite intuitive, even if some APIs have a learning curve. `balance$` becomes a composition of all the streams that it needs to derive its value. It's kinda the same, just that all values auto-update as any stream above it changes.

Websockets are another great example. How would we consume a websocket in the current services layer? The client only gets new data if it refetches. Our current design assumes that data always comes in a request-response format (http). Using Observables, no assumption is made. 

Encourage everyone to read https://this-is-learning.github.io/rxjs-fundamentals-course/docs/part-5 Much of our code already uses observables, all the Ledger code, and [Redux](https://redux.js.org/api/store#subscribelistener). I think Rxjs comes across as a scary library, but it's just an implementation of Observables. `Promise.then` vs `Observable.subscribe`.

alexp3y:
> If all the queries have the account in the queryKey, the UI would update

This is how apps already achieve reactive service data refreshing. In this example, the account balance query is responsive to changes in account, user currency setting, asset visibility configuration, etc...

The drawback is obviously that you have to anticipate changeable state upfront when creating the service query key, but it makes the promise-based service calls declarative and imo it holds up pretty well.

That said, it's probably best not to get too carried with injected state in services to keep this manageable.

```ts
export function useGetAccountTotalBalanceQuery(request: AccountRequest) {
  const { fiatCurrencyPreference, assetVisibility } = useSettings();
  return useQuery({
    queryKey: [
      'account-balances-service-get-total-balance',
      request,
      fiatCurrencyPreference,
      assetVisibility,
    ],
    queryFn: ({ signal }: QueryFunctionContext) =>
      getAccountBalancesService().getTotalBalance(request, signal),
    ...balanceQueryOptions,
  });
}
```

kyranjamie:
> The drawback is obviously that you have to anticipate changeable state upfront when creating the service query key

That's what I was alluding to. What happens when the service layer grows in complexity, and there are vastly more dependencies, requiring tens of parameters added to queryKey?

Perhaps I was too quick to propose Observables as a solution, and should've instead focused on what I think is the limitation we're going to run in to.

Happy to leave this discussion there and focus on the state changes in this PR

pete-watters:
I don't have much feedback apart from that this sounds like a good idea and I am happy to get behind it. I used RXJS a bit but a long time ago (7 years or so).  I think it's a nice way of managing incoming data and I would like if we could use websockets instead of queries. 

The main benefit I see here is having a separate state specific package we can use to share between apps in a consistent way. I think it makes sense to make this change now pre-multi wallet on web. I would also consider implementing your[ flat hierarchy account](https://www.notion.so/trustmachines/Wallet-account-hierarchy-84decafb79a84cae981d0aef4449dd30#84decafb79a84cae981d0aef4449dd30) suggestion now too if it's not a whole lot of extra work. It would be nice to align our structure with ledger a bit better

kyranjamie:
Closing in favour of #1725

PR #1671, opened 2025-10-09
chore(mobile): rework SIP9 types

pete-watters:
This PR is a follow up to [this one](https://github.com/leather-io/mono/pull/1655) and adds some changes to the SIP-9 model structure. 

@alexp3y I thought about adding `floor_price` but it really only makes sense to me in the context of collections rather than individual items. Gamma show it at the collection base. I added a service call for collections (`getStacksCollection`)  but we aren't using it yet.

### DIFF
--- a/packages/models/src/assets/sip9-asset.model.ts
+++ b/packages/models/src/assets/sip9-asset.model.ts
@@ -1,83 +1,46 @@
 import { BaseNonFungibleCryptoAsset } from './asset.model';
 
-export type Sip9ContentType =
-  // Images
-  | 'image/jpeg'
-  | 'image/jpg'
-  | 'image/png'
-  | 'image/gif'
-  | 'image/webp'
-  | 'image/svg+xml'
-  | 'image/bmp'
-  | 'image/tiff'
-  | 'image/avif'
+export const sip9ContentTypes = [
+  'image/jpeg',
+  'image/jpg',
+  'image/png',
+  'image/gif',
+  'image/webp',
+  'image/svg+xml',
+  'image/bmp',
+  'image/tiff',
+  'image/avif',
+  'video/mp4',
+  'video/webm',
+  'video/mov',
+  'video/quicktime',
+  'video/avi',
+  'video/x-msvideo',
+  'video/ogg',
+  'audio/mpeg',
+  'audio/mp3',
+  'audio/wav',
+  'audio/x-wav',
+  'audio/ogg',
+  'audio/aac',
+  'audio/flac',
+  'audio/webm',
+  'model/gltf+json',
+  'model/gltf-binary',
+  'application/octet-stream',
+  'text/plain',
+  'text/html',
+  'text/markdown',
+  'application/pdf',
+  'application/json',
+  'text/javascript',
+  'application/javascript',
+  'application/zip',
+  'unknown',
+  '',
+] as const;
 
-  // Videos
-  | 'video/mp4'
-  | 'video/webm'
-  | 'video/mov'
-  | 'video/quicktime'
-  | 'video/avi'
-  | 'video/x-msvideo'
-  | 'video/ogg'
-
-  // Audio
-  | 'audio/mpeg'
-  | 'audio/mp3'
-  | 'audio/wav'
-  | 'audio/x-wav'
-  | 'audio/ogg'
-  | 'audio/aac'
-  | 'audio/flac'
-  | 'audio/webm'
-
-  // 3D Models
-  | 'model/gltf+json'
-  | 'model/gltf-binary'
-  | 'application/octet-stream' // GLB files sometimes use this
-
-  // Documents/Text
-  | 'text/plain'
-  | 'text/html'
-  | 'text/markdown'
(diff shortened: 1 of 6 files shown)


packages/services/src/assets/sip9-asset.utils.ts
```diff
@@ -43,64 +42,53 @@ function transformGammaCollectionData(
...
 
-  const cachedImageThumbnail =
-    gammaMetadata?.item.asset_content?.content_url ||
-    (hiroMetadata as any)?.cached_thumbnail_image ||
-    hiroMetadata?.cached_image ||
-    '';
-
-  const contentType = (gammaMetadata?.item.asset_content?.content_type ||
-    '') as SupportedSip9ContentType;
+  const contentType = gammaMetadata?.item.asset_content?.content_type || '';
```
pete-watters, line 90: Hiro doesn't give any concept of content type I can see. I have code to figure it out in `sip9.native.tsx` in case Gamma doesn't give it either

packages/models/src/assets/sip9-asset.model.ts
```diff
@@ -122,7 +79,7 @@ export interface Sip9Asset extends BaseNonFungibleCryptoAsset {
   readonly tokenId: number;
   readonly name: string;
   readonly description: string;
-  readonly cachedImage: string;
```
pete-watters, line 125: @alexp3y  and I discussed this and `cachedImage` / `cachedImageThumbnail` are HIRO API specific so we renamed to `content` and are using `contentUrl` instead

packages/services/src/infrastructure/api/gamma/gamma-api.schema.ts
```diff
@@ -8,7 +10,7 @@ export const gammaNftMetadataSchema = z.object({
     description: z.string(),
     asset_content: z.object({
       content_url: z.string(),
-      content_type: z.string(),
+      content_type: z.string() as z.ZodType<Sip9ContentType>,
```
kyranjamie: Why not make a schema for this rather than casting ?

pete-watters: I didn't think of that, I will give it a try thanks! I was just trying to avoid casting `as Sip9ContentType` elsewhere

pete-watters: I updated it

packages/services/src/infrastructure/api/gamma/gamma-api.schema.ts
```diff
@@ -8,7 +12,7 @@ export const gammaNftMetadataSchema = z.object({
     description: z.string(),
     asset_content: z.object({
       content_url: z.string(),
-      content_type: z.string(),
+      content_type: sip9ContentTypeSchema,
```
alexp3y, line 15: Are we sure that Gamma will definitely send one of these types?

packages/ui/src/components/collectibles/native/sip9.native.tsx
```diff
@@ -1,12 +1,47 @@
...
+  'text/plain',
+  'application/octet-stream',
+  'model/gltf+json',
+  'model/gltf-binary',
+  '',
+] as const;
+
+export type SupportedSip9ContentType = (typeof supportedContentTypes)[number];
+
+const supportedContentTypesSet = new Set(supportedContentTypes);
```
alexp3y, line 36: Might require some TS-magic, but if possible it could be cool to make this specifically a subset of `Sip9ContentType` to enforce a relationship back to the model.

packages/services/src/assets/sip9-asset.utils.ts
```diff
@@ -43,64 +41,53 @@ function transformGammaCollectionData(
...
   return clarityValue.type === 'uint' ? Number(clarityValue.value) : 0;
 }
 
+interface Sip9Details {
+  name: string;
+  description: string;
+  content: Sip9AssetContent;
+  collection: Sip9Collection | undefined;
+  attributes: Sip9Attribute[] | undefined;
+}
```
alexp3y, line 71: Now that its a bit more streamlined, I wonder if we still need this intermediate "details" interface and `transformToSip9Details` function or if we could just move all this stuff directly into `createSip9Asset`?

alexp3y (approved):
Very nice! Looks super clean and should be a solid base to build out from.

packages/services/src/infrastructure/api/gamma/gamma-api.schema.ts
pete-watters: I'm not certain but I generated that list of exhaustive types for `Sip9ContentType` of all the possible types there could reasonably be so I think it's likely to be one of them. 

I also added `''` as a type in case it's empty. If it's not one of these types it could fail. I had been just casting `as Sip9ContentType` as even if the type is wrong / a new one we can handle it in the UI. I'm not sure whats best here

pete-watters:
Thanks @alexp3y! I'll make the further updates mentioned in the next PR along with Stamps. Gonna merge this for now

PR #1672, opened 2025-10-09
Swap validation

tigranpetrossian:
This PR adds validation to swap state, accompanied with some prerequisite refactoring.

### Refactoring
 * Added per protocol and and per swap execution type strategies. Helps resolving various metadata based on asset or swap type, e.g. min/max spendable, min receive amount, etc.
 * Slightly altered ActionButtons to enable displaying actions conditionally. 
 * Removed `isValidPrecision(input: number)` in favor it's underlying helper, `number` can't be trusted with precision.
 * Introduced a subset of `FungibleCryptoAssetType` type to indicate which assets support swapping.

### Validation

The validation itself is a purely derived result from the state, I'm intentionally not involving form/validation libraries here, as flows like Swap aren't conventional forms, and these libraries just add more friction for little value. 
While it tracks the validity of the flow at any given moment, I expect to ideally use this very little (perhaps as a final check), and instead rely on UI guards as much as possible to help users keeping the form valid, e.g. amount capped to correct decimals, can't enter incorrect slippage, etc.

Probably easier reviewed commit by commit to capture changes meaningfully.


apps/mobile/src/features/swap/strategies/execution-type/execution-type.utils.ts
```diff
@@ -0,0 +1,18 @@
+import { PER_DEX_FEE_PERCENTAGE } from '@/features/swap/swap.constants';
+import BigNumber from 'bignumber.js';
+
+import { Money, SwapDex } from '@leather.io/models';
+import { createMoney } from '@leather.io/utils';
+
+export function calculateMinToReceiveAmount(quoteAmount: Money, slippage: number): Money {
+  const minReceiveAmount = quoteAmount.amount.times(1 - slippage);
```
kyranjamie: the `1 - slippage` part of this happens outside BN and could have imprecision, e.g. 7% slippage

apps/mobile/src/features/swap/swap-state/validation/swap-validation.utils.ts
```diff
@@ -0,0 +1,26 @@
...
+import { countDecimals } from '@leather.io/utils';
+
+export function isParsableNumber(value: string) {
+  const numericValue = parseFloat(value);
+  return !isNaN(numericValue) && isFinite(numericValue);
+}
+
+export function isPresent(value: unknown): boolean {
+  if (value === null || value === undefined) return false;
+  if (typeof value === 'string') return value.trim().length > 0;
```
kyranjamie:
```suggestion
  if (isString(value)) return value.trim().length > 0;
```

apps/mobile/src/features/swap/swap-state/swap-quote-selector.ts
```diff
@@ -1,41 +1,36 @@
...
+  fairMarketRate: number | null,
+  slippage: number
 ): SwapQuoteSelectionResult {
   const quotes = pipe(
     swapQuotes,
     map(swapQuote => {
-      const selector = selectors[swapQuote.executionType];
-      if (!selector) return;
-      return selector(swapQuote, fairMarketRate);
+      const enricher = getExecutionTypeStrategy(swapQuote.executionType).enrichQuote;
```
kyranjamie, line 21: How do you mean by "enricher"? Enriches with what?

kyranjamie (approved):
Tidy work @tigranpetrossian

tigranpetrossian:
One thing I'm on the fence about is whether to include error messages in the validation, assuming this is cross-platform state. This way we centralize messages, but not sure if that's a pro or a con. Extension might need to word certain things differently if interfaces look different? @kyranjamie what do you think?

kyranjamie:
Also localisation concerns, and that we might not necessarily use the same library on each platform? What if we justpass in a copy object that we wire up in each app?

tigranpetrossian:
Localization shouldn't be an issue, you'd simply wrap as `` t`issue.message` `` Passing it in seems convoluted tbh, if you're creating your own copy object, might as well just use it in place, e.g. `myCustomMessages[field][code]`. It's the editorial aspect of it that I'm not sure of.

apps/mobile/src/features/swap/swap-state/use-swap-state.ts
```diff
@@ -53,12 +58,12 @@ function initializeState({
...
   accountRequest: { account: AccountAddresses };
-  baseAsset?: FungibleCryptoAsset;
-  targetAsset?: FungibleCryptoAsset;
+  baseAsset?: SupportedAsset;
+  targetAsset?: SupportedAsset;
   marketDataService: MarketDataService;
   swapService: SwapService;
   // TODO: Come up with a better name. This is an adapter for assetVisibility.
-  isAssetAllowed?: (asset: FungibleCryptoAsset) => boolean;
+  isAssetAllowed?: (asset: SupportedAsset) => boolean;
```
pete-watters: Maybe we can ditch the `// TODO: ` if you don't have a better name? 

`isAssetAllowed` seems OK to me

pete-watters: Sorry just saw the new comment now reviewing commit by commit

apps/mobile/src/features/swap/swap-state/validation/swap-validation.utils.ts
pete-watters: I think it's good to use these helpers if we have them and remember, but also in this new world of AI code, isn't it the same basically but without having to import `isString`? 

In the olden days, of course we don't want the same implementation everywhere but now if we are tabbing through features does it really matter?

pete-watters (approved):
Nice work @tigranpetrossian 👍 well structured PR too so easy to review

packages/services/src/swap/swap.service.ts
```diff
@@ -166,20 +167,26 @@ export class SwapService {
...
       )
-    )
-      .filter(result => result.status === 'fulfilled')
-      .map(result => ({
-        asset: result.value,
-        providerAssets: groupedAssets[serializeAssetId(getAssetId(result.value))],
-      }));
+    );
+
+    function assetsToSwapAssets(result: PromiseSettledResult<FungibleCryptoAsset>) {
```
alexp3y, line 177: Overall much cleaner. This whole function could be a single ternary statement for a small optimization.

alexp3y (approved):

tigranpetrossian:
Simplified the `ValidationResult` to carry a single issue per field vs. all possible issues at once. As I've been using this for the past couple days it was clear that it was unnecessary. Will have more clarity on how to handle messages once I test drive this for a little longer.

PR #1676, opened 2025-10-11
feat(mobile): google wallet backup

BowTiedBlox:
## Description
This PR adds Google Drive backup integration for wallet accounts in the Leather mobile app.

### Key Features
- Automatic cloud backup: Wallet data can now be securely backed up to Google Drive.
- Seamless restoration: When signing in on a new device, users can restore their wallet directly from the cloud.
- Flexible management: Users can choose to delete an existing backup or replace it with a new wallet backup at any time.
- Seamless Onboarding: New users can simply create an account with google sign up/sign in.

## User Stories
1. Sign-up flow
- user press `Continue with google` →2. app request access to user’s google account scopes (appDataFolder, email,…) → 3. user sets & confirms password → 4.wallet created

2. Sign-in flow
- user press `Continue with google`  → 
    a. user enter the password → wallet restored
    b. user press forgot password (back→ 

3. Delete from cloud
- user press delete from cloud →2. wallet delete from cloud & device

4. Backup wallet ( only for wallets created locally )
- user press backup wallet → 
    - user already has a backup on cloud  →  
      1. user enter the already existing backup password →1. user sets & confirms password → 2. backup created
      2. user press forgot password( backup gets deleted from cloud ) → 1. user sets & confirms new password → 2. backup created
    - user doesn’t have a backup on cloud → 1. user sets & confirms password →2.backup created
    
    
## Google Cloud Console

In Google Cloud Console, we set up the OAuth consent screen in Testing mode and added our dev Google accounts as Test users.
Then, under APIs & Services → Credentials, we created an Android OAuth client for the package name `io.leather.mobilewallet` and added the SHA-1 fingerprint from the local `debug.keystore`.

We also created an OAuth 2.0 “Web application” client and used its Client ID as the value for EXPO_PUBLIC_GOOGLE_CLIENT_ID. The same Web Client ID is passed to GoogleSignin.configure() in the app.

## Firebase Configuration

During local development, we used an OAuth 2.0 Web client ID in Google Cloud Console to implement the sign-in flow without access to the Leather team’s Firebase project. For production and shared environments, these settings should be consolidated under the same Firebase project used for push notifications (FCM) to prevent authentication errors.

A mismatch can occur between the client ID injected by google-services.json at build time and the client ID passed to `GoogleSignin.configure` if there are different projects used on google for Firebase notifications and wallet backup.  This may result in a Google DEVELOPER_ERROR. While it won’t happen in every case, if it does, temporarily commenting out the Firebase imports related to notifications and rebuilding the project resolves the conflict, indicating a configuration mismatch between the two client IDs.


kyranjamie:
Thank you for the PR @BowTiedBlox. Your implementation looks clean and well organised. We weren't aware of this grant, and need to first discuss these product changes as team.

BowTiedSirJonathan:
Hi @kyranjamie ,

We tested the implementation on top of your codebase using multiple Android Studio emulators. We also attempted to test it on iOS directly on multiple devices, but the Expogo version in your monorepo was more outdated than the one available on the App Store, resulting in an error that prevented it from opening.

The tested flows correspond to the user stories shared above.

cuevasm:
Hey @kyranjamie the grant was public and I spoke with Rena about it. I think to clarify the ask here, it was just that you'd hopefully be willing to run a beta test with a subset of your users with the tooling being built here. We wanted to avoid the problem of having to bootstrap a bunch of users from scratch to test with and rather tap into an existing apps users who may be excited to try out something like this. There's no expectation that you fully integrate, but it would be awesome if you could help the team get more feedback in this manner or some alternative way that we're open to hearing about. Thanks for your consideration!

BowTiedSirJonathan:
Hi @kyranjamie , are there any updates on this? Thank you!

kyranjamie:
Sorry for the delayed response @cuevasm @BowTiedBlox @BowTiedSirJonathan 

We want to be supportive of community efforts—and definitely see the value in offering alternative ways for users to back up their keys—but we're finding it difficult to commit to a path forwards. 

It just isn't as simple as to deploy this to a subset of our users for a beta test. We don't have A/B testing infra in place, and if we did ship it to users, we'd have to maintain the feature henceforth. Exporting key back ups is of course something we have to be extremely careful with. Were something to go wrong, it's our brand attached to an effort we've had no involvement in.

To invest in this feature properly, we'd need to assign many developer days/weeks we haven't accounted for, distracting us from existing priorities. Our team was unaware of this grant until this pull request. It's a big out-the-blue ask.

We're open to helping in other ways, and happy to jump on a call with the team to learn more about what you're trying to do. We’d also be glad to collaborate on future grant projects involving Leather **provided the core team is involved early, with a chance to plan and scope potential changes**.

cuevasm:
Thanks Kyran, I totally appreciate what you're saying, but I did speak with Rena about this before the grant was even approved. Maybe some comms broke down, no harm no foul and while I disagree it was 'out of the blue', I accept it could have felt that way to you. All teams are more than included and welcome to comment, weigh in on scope, offer feedback, etc. on grants as they are all done publicly, we welcome it and saw a lot of teams do that.

Anyway, I completely understand the need to be careful, but I'll also make an ask to consider for the future: When you can, I'd love to see Leather invest more time into being able to take meaningful contributions/do tests/etc. After all, it is supposed to be open-source. This feels a lot more like 'we have our priorities and even though the community has expressed clear interest in a feature by supporting this grant, we're not setup to even try it out'. I think you can probably understand how it feels like a bit of a bummer or a brush off (not the intent I know, I understand your position). Anyhow, there was no expectation to fully integrate it or take it as-is, but just to help them test/refine, etc - maybe we can connect on something minimal here still? 

I assume getting to a point where engaging with builders like this to expand Leather's feature set is a goal of the team? Is that something we (Stacks Foundation) can help with somehow? I think it'd be really cool if Leather could support the community with a way to test and contribute new potential features and it would expand the bandwidth you guys have to ship by allowing for the activation of contributors. I know people want to help :).

Ok so wrapping up the long message 🤣, I know we're all hustling and we all wish we had setups that were more permissible and we all deal with limitations, but flagging this as something cultural beyond just one feature. We have a high-level proven builder here trying to make a contribution we have every indication the community is interested in having, but we don't seem setup to act on it in a way (with all due empathy) I'd hope for an open-source project. If we don't have that at some point, people won't continue to be that interested in contributing to Leather and it will probably never really feel open-source as a result.

Super appreciate the meaningful engagement with the discussion and consideration. Obviously bummed it didn't work out, but I appreciate that you were open-minded and honest with us!

@BowTiedSirJonathan - I'll connect with you about other applications where we could test with existing userbases, thanks for keeping up on this and great to see the work moving forward <3

PR #1685, opened 2025-10-14
fix(mobile): fix bug with version 2.76 resetting the state

pete-watters:
> Try out Leather build 2f0a79a — [Extension build](https://github.com/leather-io/mono/actions/runs/18490343892), [Test report](https://leather-io.github.io/playwright-reports/fix/mobile-state-reset), [Storybook](https://fix/mobile-state-reset--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/mobile-state-reset)

I ran Claude through the changelog and it said this bug is happening as there have been some new fields added to settings which were causing it to reset the state. 

**Root cause**
Users' state was being reset on upgrade to version 2.76 due to failed Zod schema
validation in the state rehydration process.

The problem was in `apps/mobile/src/store/storage-persistors.ts:14-24`:
```
    stateReconciler: (inboundState, originalState, reducedState, config) => {
      try {
        stateSchema.parse(inboundState);
        return autoMergeLevel2(inboundState, originalState, reducedState, config);
      } catch (e) {
        console.warn(e);
        // set state on fire if zod doesn't pass
        return originalState;  // ← Returns empty state, resetting everything!
      }
    },
```

**Specific issues**
In `apps/mobile/src/store/settings/utils.ts`:

1. The settingsSchema was MISSING the hapticsPreference field entirely
   - Field exists in SettingsState interface (line 41)
   - Field was missing from Zod schema (lines 51-68)

2. Several fields were added as REQUIRED in recent versions:
   - hapticsPreference - added in v1.52.0
   - notificationsPreference - added in v2.11.0
   - assetVisibility - added in v2.73.0
   - currentAccount - added in v2.74.0

When users upgraded from older versions, their persisted state was missing these required fields, causing validation to fail and their entire state to be reset.

**Changes made**
1. `apps/mobile/src/store/settings/utils.ts`: ADDED missing hapticsPreference field and made newer fields optional
2. `apps/mobile/src/store/settings/settings-rehydration.ts` ADDED explicit fallback handling for optional fields

**How this PR fixes the issue**
**BEFORE**:
1. User upgrades to v2.76
2. Persisted state is missing hapticsPreference and other newer fields
3. Zod schema validation fails at storage-persistors.ts:17
4. Catch block at line 23 returns originalState (empty state)
5. ALL USER DATA IS WIPED

**AFTER**:
1. User upgrades to v2.76+
2. Persisted state may be missing newer fields
3. Zod schema validation PASSES (fields are optional)
4. Rehydration handler fills in default values for missing fields
5. USER DATA IS PRESERVED across upgrades


pete-watters:
@edgarkhanzadian it seems like the state was being reset as Zod validation fails. I am not sure if this fix is correct as I found another PR you had previously where you are explicitly re-setting on Zod schema fail

https://github.com/leather-io/mono/pull/550/

kyranjamie:
I think that we should remove the code you've linked here entirely. 

No schema validation during the reconciliation process. Or, if there is, the schemas should be version aware such that they can parse correctly between versions.

kyranjamie (approved):
Approving now if this fixes the issue short term

pete-watters:
Thanks @kyranjamie 👍  I'm going to merge this and get a potential fix in flight. I can amend / improve later as needed

PR #1689, opened 2025-10-14
feat: add stamps service and integrate to collectibles service

pete-watters:
> Try out Leather build 8442183 — [Extension build](https://github.com/leather-io/mono/actions/runs/18559662041), [Test report](https://leather-io.github.io/playwright-reports/feat/stamps-collectibles-full), [Storybook](https://feat/stamps-collectibles-full--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/stamps-collectibles-full)

This PR:
- adds a new stamps API call to the collectibles service
- integrates stamps to the mobile app

I am using the same API as we use in the extension - https://stampchain.io. This code works however often I am having problems with the API call timing out on subsequent renders. I can keep investigating improving that.

You can see a demo of a tree stamp in the mobile app here:

https://github.com/user-attachments/assets/af8b36d0-dc94-44c9-857f-2bc6a08081f7

### DIFF
--- a/apps/extension/src/app/services/use-account-addresses.ts
+++ b/apps/extension/src/app/services/use-account-addresses.ts
@@ -1,19 +1,49 @@
 import { useMemo } from 'react';
 
+import { deriveAddressIndexZeroFromAccount, getNativeSegwitPaymentFromAddressIndex, getTaprootPaymentFromAddressIndex } from '@leather.io/bitcoin';
 import { createAccountAddresses } from '@leather.io/utils';
 
 import { useBitcoinAccountXpubs } from '@app/store/accounts/blockchain/bitcoin/bitcoin.hooks';
+import { useNativeSegwitAccount } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useTaprootAccount } from '@app/store/accounts/blockchain/bitcoin/taproot-account.hooks';
 import { useStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
 
 export function useAccountAddresses(accountIndex: number) {
   const accountXpubs = useBitcoinAccountXpubs(accountIndex);
   const stxAccount = useStacksAccount(accountIndex);
+  const nativeSegwitAccount = useNativeSegwitAccount(accountIndex);
+  const taprootAccount = useTaprootAccount(accountIndex);
 
   return useMemo(() => {
-    return createAccountAddresses(
+    const baseAddresses = createAccountAddresses(
       { fingerprint: 'master', accountIndex },
       accountXpubs,
       stxAccount?.address
     );
-  }, [accountXpubs, stxAccount, accountIndex]);
+
+    if (!baseAddresses.bitcoin) return baseAddresses;
+
+    const nativeSegwitAddress = nativeSegwitAccount
+      ? getNativeSegwitPaymentFromAddressIndex(
+          deriveAddressIndexZeroFromAccount(nativeSegwitAccount.keychain),
+          nativeSegwitAccount.network
+        ).address
+      : '';
+
+    const taprootAddress = taprootAccount
+      ? getTaprootPaymentFromAddressIndex(
+          deriveAddressIndexZeroFromAccount(taprootAccount.keychain),
+          taprootAccount.network
+        ).address
+      : '';
+
+    return {
+      ...baseAddresses,
+      bitcoin: {
+        ...baseAddresses.bitcoin,
+        zeroIndexNativeSegwitPayerAddress: nativeSegwitAddress,
+        zeroIndexTaprootPayerAddress: taprootAddress,
+      },
+    };
+  }, [accountXpubs, stxAccount, accountIndex, nativeSegwitAccount, taprootAccount]);
 }
--- a/apps/mobile/src/hooks/use-account-addresses.ts
+++ b/apps/mobile/src/hooks/use-account-addresses.ts
@@ -3,7 +3,10 @@ import { useMemo } from 'react';
 import { Account } from '@/store/accounts/accounts';
 import { useAccounts, useAccountsByFingerprint } from '@/store/accounts/accounts.read';
 import { AccountStatus } from '@/store/accounts/utils';
-import { useBitcoinAccounts } from '@/store/keychains/bitcoin/bitcoin-keychains.read';
+import {
+  useBitcoinAccounts,
+  useBitcoinPayerAddressFromAccountIndex,
+} from '@/store/keychains/bitcoin/bitcoin-keychains.read';
 import {
   useStacksSignerAddressFromAccountIndex,
   useStacksSigners,
@@ -34,8 +37,8 @@ function deriveTotalAccountAddresses(
         );
         return uniqueArray(
           walletKeychains.map(key => extractAccountIndexFromPath(key.keyOrigin))
-        ).map(accountIndex =>
-          createAccountAddresses(
+        ).map(accountIndex => {
+          const baseAddresses = createAccountAddresses(
             {
               fingerprint: wallet.fingerprint,
               accountIndex,
@@ -47,8 +50,26 @@ function deriveTotalAccountAddresses(
             stacksSigners
               .fromAccountIndex(wallet.fingerprint, accountIndex)
               .map(signer => signer.address)[0]
-          )
-        );
+          );
+
+          if (!baseAddresses.bitcoin) return baseAddresses;
+
+          const { nativeSegwit, taproot } = bitcoinAccounts.accountIndexByPaymentType(
+            wallet.fingerprint,
+            accountIndex
+          );
+
+          return {
+            ...baseAddresses,
+            bitcoin: {
+              ...baseAddresses.bitcoin,
+              zeroIndexTaprootPayerAddress:
+                taproot?.derivePayer({ change: 0, addressIndex: 0 }).address ?? '',
+              zeroIndexNativeSegwitPayerAddress:
+                nativeSegwit?.derivePayer({ change: 0, addressIndex: 0 }).address ?? '',
+            },
+          };
+        });
       });
 }
 
@@ -62,8 +83,8 @@ function deriveWalletAccountAddresses(
     ? []
     : accountsByFingerprint.list
         .map(account => account.accountIndex)
-        .map(accountIndex =>
-          createAccountAddresses(
+        .map(accountIndex => {
+          const baseAddresses = createAccountAddresses(
             { fingerprint, accountIndex },
             bitcoinAccounts.list
               .filter(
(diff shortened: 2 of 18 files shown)


apps/extension/src/app/services/use-account-addresses.ts
```diff
@@ -1,19 +1,25 @@
...
   const accountXpubs = useBitcoinAccountXpubs(accountIndex);
   const stxAccount = useStacksAccount(accountIndex);
+  const currentNetwork = useCurrentNetwork();
 
   return useMemo(() => {
-    return createAccountAddresses(
-      { fingerprint: 'master', accountIndex },
-      accountXpubs,
-      stxAccount?.address
+    return addDerivedAddressesToAccountAddresses(
```
pete-watters: I'm also adding the derived addresses to accounts on the extension as I plan to use this service there also soon but I'm unsure if that's necessary there.

apps/mobile/src/hooks/use-account-addresses.ts
```diff
@@ -35,47 +41,54 @@ function deriveTotalAccountAddresses(
...
-              accountIndex,
-            },
-            walletKeychains
-              .filter(key => extractAccountIndexFromPath(key.keyOrigin) === accountIndex)
-              .map(getDescriptorFromKeychain)
-              .filter(isDefined),
-            stacksSigners
-              .fromAccountIndex(wallet.fingerprint, accountIndex)
-              .map(signer => signer.address)[0]
+          addDerivedAddressesToAccountAddresses(
```
pete-watters: The Stamps API needs to query using the segwit address so I am adding it to account addresses. 

Maybe there is a better way I could do this?

apps/extension/src/app/services/use-account-addresses.ts
alexp3y: Could use `useCurrentAccountNativeSegwitAddressIndexZero()` and similar for taproot for getting derived addresses here instead?

apps/mobile/src/hooks/use-account-addresses.ts
alexp3y: Makes sense and could be useful to have these available on `AccountAddresses` in general.

To populate the addresses I wonder if we really need to add new derivation utils when the wallets should be able to provide the fully derived addresses.

For example, `useBitcoinPayerAddressFromAccountIndex()` here in `mobile` should get you both taproot and segwit address. The wallet hooks may have better memoization built-in as well.

packages/models/src/account.model.ts
```diff
@@ -9,6 +9,8 @@ export const accountIdSchema = walletIdSchema.and(z.object({ accountIndex: z.num
 export const bitcoinAddressInfoSchema = z.object({
   taprootDescriptor: z.string(),
   nativeSegwitDescriptor: z.string(),
+  taprootAddress: z.string().optional(),
+  nativeSegwitAddress: z.string().optional(),
```
alexp3y: Maybe we can add a `zeroIndex` prefix here to make it super clear which addresses these are?

packages/services/src/collectibles/collectibles.service.ts
```diff
@@ -119,4 +141,25 @@ export class CollectiblesService {
...
+    address: string,
+    signal?: AbortSignal
+  ): Promise<{ asset: StampAsset; blockHeight: number }[]> {
+    try {
+      const response = await axios.get<{
+        data: { stamps: Stamp[] };
+      }>(`https://stampchain.io/api/v2/balance/${address}`, {
+        signal,
+        timeout: 10000,
+      });
```
alexp3y: It's only a single call but we should still follow our existing convention here and create a dedicated `StampchainApiClient`.

There you can also import and use the `CacheService`, which should help w/ re-render issues.

packages/services/src/collectibles/collectibles.service.ts
```diff
@@ -7,13 +8,24 @@ import {
...
+  createStampAsset,
+  mapBisInscriptionToCreateInscriptionData,
+  sortByBlockHeight,
+} from './collectibles.utils';
+
+interface Stamp {
+  stamp: number;
+  stamp_url: string;
+  block_index?: number;
+}
```
alexp3y: This type belongs to Stampchain and should be to flag that prefixed - `StampchainStamp` (ideally moved to a dedicated client).

packages/services/src/collectibles/collectibles.utils.ts
```diff
@@ -23,3 +29,13 @@ export function mapBisInscriptionToCreateInscriptionData(
 export function sortByBlockHeight(a: { blockHeight: number }, b: { blockHeight: number }) {
   return b.blockHeight - a.blockHeight;
 }
+
+export function createStampAsset(stampData: { stamp: number; stamp_url: string }): StampAsset {
```
alexp3y, line 33: 🦾

packages/services/src/collectibles/collectibles.service.ts
pete-watters: OK thanks! I considered that and thought it could be overkill but it makes sense and I'll refactor 👍

packages/services/src/collectibles/collectibles.service.ts
pete-watters: I'll make a client and move it, thanks!

apps/extension/src/app/services/use-account-addresses.ts
pete-watters: Thanks @alexp3y 👍  I'll check that out and see can I avoid doing this. I think `useCurrentAccountNativeSegwitAddressIndexZero` us deprecated though and it says to instead use `signer.address`

apps/mobile/src/hooks/use-account-addresses.ts
pete-watters: Thanks @alexp3y I should have used `useBitcoinPayerAddressFromAccountIndex` 🤦

apps/extension/src/app/services/use-account-addresses.ts
kyranjamie: I have difficult understanding what `addDerivedAddressesToAccountAddresses` means

packages/bitcoin/src/utils/bitcoin.descriptors.ts
```diff
@@ -12,3 +18,59 @@ export function getDescriptorFromKeychain<T extends { keyOrigin: string; xpub: s
...
+        getBtcSignerLibNetworkConfigByMode(network)
+      );
+      return payment.address;
+    }
+
+    return undefined;
+  } catch {
+    return undefined;
+  }
+}
```
kyranjamie: Can you help me understand why we need this? Is accessing the first native segwit address in the mobile codebase difficult currently?

The problem with this is that, at the package level, we're baking in assumptions that its only the first address that matters. I don't think we should do this, because that isn't how bitcoin works.

apps/mobile/src/hooks/use-account-addresses.ts
kyranjamie: I know it's hard because our wallet works in the model that we predominantly use a single address. But writing code that assumes we will only ever use those first addresses, will make any move away from that extremely painful, so I think we should be very careful,  and not assume that bitcoin accounts and bitcoin addresses are 1:1.

@pete-watters can you help me understand what the difficulty here is? It seems like many of these changes are to feed the native segwit address of the first account into the stamps service? Is there more to it than that?

apps/extension/src/app/services/use-account-addresses.ts
pete-watters: In our code for `createAccountAddresses` it seems that we add descriptors for taproot and segwit but not the actual addresses. 

For the stamps API I need to use the segwit address in the API call. It seemed to me that it would be good to add the first derived address from that descriptor to the account so that I can use this to query the API. 

That is essentially what `addDerivedAddressesToAccountAddresses` is doing. Perhaps it could be named better but also so could `createAccountAddresses` if it's actually creating BTC descriptors and a STX address. 

I am looking into what @alexp3y has suggested now to use the hooks:
- `useCurrentAccountNativeSegwitAddressIndexZero` - extension
- `useBitcoinPayerAddressFromAccountIndex` - mobile

Those are both client side hooks to so if I use those I will need to then pass the segwit address on to the collectibles service. I can change my code  if that's a better approach?

With the way I have implemented it now, we add the first derived address to our 'bitcoinAddressInfoSchema'  in the `accountAddressesSchema` so that the service has direct access without having to pass them around. 

I appreciate that there could be more addresses derived but I wonder how deep into the users addresses do we want to go to check for stamps? The API seems quite basic in that it only accepts a single address at time. Should I derive more addresses and loop through them to check more deeply? 

I am happy to re-work this code and can even scrap and re-write the whole thing if you think that's best. The goal here was to add basic support for Stamps in mobile so we could have it included in the collectibles service for use on both platforms

apps/mobile/src/hooks/use-account-addresses.ts
pete-watters: @kyranjamie I have added an explanation of what I did and why [here](https://github.com/leather-io/mono/pull/1689/files#r2435054408).

Perhaps this implementation is non optimal as it pollutes `AccountAddresses` with just the first derived address. It seemed better to me to extend `AccountAddresses` which we already pass to the collectibles service to give it the extra data it needs to query for stamps. 

Otherwise I would be obtaining the address in two different ways on `extension` and `mobile` to then feed into the service to pass on to the stamps API. 

I can also skip adding this to the service altogether and just use the query we already have? `createGetStampsByAddressQueryOptions` ? 

In the `extension` implementation we have now, we [query using the ](https://github.com/leather-io/mono/blob/dev/apps/extension/src/app/features/collectibles/components/bitcoin/stamps.tsx)`currentAccountBtcAddress` 

From my analysis of the code I think we also aren't yet adding `src20` tokens to the account balance service. That uses the same stampchain API so we should also consider integrating that to balances before we refactor it on the extension

pete-watters: It finally clicked and I am re-writing. 👍

packages/services/src/collectibles/collectibles.utils.ts
pete-watters: Thanks for the pity 💪  😂

packages/bitcoin/src/utils/bitcoin.descriptors.ts
pete-watters: I have refactored the code to pass in the address from mobile instead.

packages/models/src/account.model.ts
pete-watters: I've changed this

apps/extension/src/app/services/use-account-addresses.ts
pete-watters: I've changed this to use the existing hooks to add to the `accountAddressesSchema`. I initially misunderstood and thought I would need to pass them to the collectibles service directly. 

I've changed it to keep a similar approach but to use existing hooks instead

kyranjamie: > I appreciate that there could be more addresses derived but I wonder how deep into the users addresses do we want to go to check for stamps?

For stamps definitely the first address is fine.

My point is that we should ensure the rest of the codebase, unrelated to stamps, accurately models how bitcoin works, and doesn't bake in behaviours specific to the first address (which is a presentational concern of our wallet). Creating reusable structures to represent bitcoin accounts, that contains the first address only in it, counts as baking that assumption in.

IMO we should almost never read a pre-rendered bitcoin address string, and only ever read them as the result of a fn call to `account.derivePayer`. Bitcoin addresses as a concept should be absent the state structure.

Using these hooks is a better way to do it because they derive addresses when needed, rather than caching them upfront (hard to do with a 2³² addresses lol, thanks chat). 

The problem I see here is that, first, some things are named poorly. Something return bitcoin descriptors and stacks addresses shouldn't be called "addresses" as this would suggest addresses are returned. Second, there are some difficulties reading addresses with ease. 

I'm sure there are ways we can make it trivial to read addresses that don't involve larger changes to how we manage account state.

pete-watters:
@alexp3y @kyranjamie : Thanks for the feedback. I've re-worked this to hopefully improve it. If you notice anything else please let me know and I can update it.

alexp3y (approved):
Looks good!

apps/extension/src/app/services/use-account-addresses.ts
alexp3y: @kyranjamie Agreed `AccountAddresses` isn't the best name. What it is is a representation of an Account that includes pre-derived, chain-specific identifiers that are broadly useful, accurate and deterministic (but not necessarily comprehensive) and allows features and services to refer to an account by its various identifiers w/o having to directly derive them.

I think in this case, the first index Bitcoin address is often especially useful, so it probably doesn't hurt to include it here as long as it's clearly prefixed - i.e. `indexZeroNativeSegwitAddress`

pete-watters: Thanks both for the feedback, it has helped me learn a lot. I am going to merge this but we can consider a refactor of `AccountAddresses` soon after? I can do it right away if deemed urgent.

PR #1696, opened 2025-10-14
refactor(swap): simplify swap state organization

tigranpetrossian:
Refactoring in swap state to prevent it growing out of control.

* Organizes the growing dumping ground of helpers by domain. 
* Extracts most of the logic out of the main hook, making it more of an orchestrator.
* Removes asset visibility adapter, as it's going to be handled at service-level.

The PR effectively does nothing except moving things around, and is well tested, but feedback on existing code is still very welcome.


apps/mobile/src/features/swap/swap-state/swap.queries.ts
```diff
@@ -28,118 +29,142 @@ type CustomQueryOptions<TQueryFnData, TError = Error, TData = TQueryFnData> = Om
...
+  baseSwapAsset,
+  targetSwapAsset,
+  baseAmount,
+  strategy,
+  fairMarketRate,
+  slippage,
+  queryOptions,
+}: UseSwapQuotesQueryParams) {
+  const debounceDelay = 350;
+  const debouncedBaseAmount = useDebouncedValue(baseAmount, debounceDelay);
```
kyranjamie, line 98: Help me understand why we need custom debounce logic, rather than refetch/staleTime config of the query?

tigranpetrossian: Quote query stays up to date as long as its input is satisfied. Debounce is for the amount to avoid hammering the API as we type. <br>Unrelated to that, staleTime/refetchingInterval yet to be added and fine tuned.

apps/mobile/src/features/swap/swap-state/utils/secondary-amount.ts
```diff
@@ -0,0 +1,41 @@
+import { SecondaryAmount, SwapInternalState } from '@/features/swap/swap-state/swap-state.types';
+import { whenInputCurrencyMode } from '@/utils/when-currency-input-mode';
+
+import { Money } from '@leather.io/models';
+
+interface ComputeSecondaryAmountStateParams {
+  state: SwapInternalState;
+  queryStatus: 'pending' | 'success' | 'error';
+  isFetching: boolean;
+  derivedAmounts: { crypto: Money | null; quote: Money | null };
```
pete-watters, line 10: The `derivedAmounts` type is repeated a few times, maybe it can be refactored to be shared?

pete-watters (approved):
Nice work 👍

kyranjamie (approved):

apps/mobile/src/features/swap/swap-state/utils/secondary-amount.ts
tigranpetrossian: Definitely, this is getting out of hand. The name could probably be revised too.
I already have changes stacked on top of this, will do this for the next PR.

PR #1711, opened 2025-10-17
fix(mobile): scope compliance check to network to close loophole bypassing compliance

pete-watters:
This PR makes a small change to the `addressComplianceValidator` to add the `chain` and `shouldCheckCompliance` to the `cacheKey`. 

I ran `codex` on the codebase to try it out and it recommended this as a fix as otherwise an address could be used when in `Testnet` mode and cached as compliant before returning to `Mainnet` and bypassing the cache

### DIFF
--- a/apps/mobile/src/features/send/utils.spec.ts
+++ b/apps/mobile/src/features/send/utils.spec.ts
@@ -1,0 +1,78 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('@leather.io/query', () => ({
+  defaultStacksFees: { estimates: [] },
+  isAddressCompliant: vi.fn(),
+}));
+
+import { addressComplianceValidator } from '@/features/send/utils';
+import { isAddressCompliant } from '@leather.io/query';
+
+const memoizedValidator = addressComplianceValidator as typeof addressComplianceValidator & {
+  clear?: () => void;
+};
+
+describe('addressComplianceValidator', () => {
+  beforeEach(() => {
+    vi.mocked(isAddressCompliant).mockReset();
+    memoizedValidator.clear?.();
+  });
+
+  afterEach(() => {
+    memoizedValidator.clear?.();
+  });
+
+  it('memoizes identical compliance lookups', async () => {
+    vi.mocked(isAddressCompliant).mockResolvedValue(true);
+
+    await addressComplianceValidator({
+      address: 'addr',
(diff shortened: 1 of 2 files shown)


kyranjamie:
Not sure I follow the need for this. The address differs across networks, so seems redundant to also list network?

pete-watters:
> Not sure I follow the need for this. The address differs across networks, so seems redundant to also list network?

I unleashed `codex` to find some 'bug's but it seems man beat machine again? 

Based on their findings:
- in `complianceValidator` we return `true` if `network !== 'mainnet'` 
- if you use `testnet` to send to a banned address mainnet address  compliance will pass
- the compliance check is memoised based on address only 
- that means you can switch back to `mainnet` and send to the banned address without validation

I'm happy to scrap this if that's not correct

tigranpetrossian:
> > Not sure I follow the need for this. The address differs across networks, so seems redundant to also list network?
> 
> I unleashed `codex` to find some 'bug's but it seems man beat machine again?
> 
> Based on their findings:
> 
> * in `complianceValidator` we return `true` if `network !== 'mainnet'`
> * if you use `testnet` to send to a banned address mainnet address  compliance will pass
> * the compliance check is memoised based on address only
> * that means you can switch back to `mainnet` and send to the banned address without validation
> 
> I'm happy to scrap this if that's not correct

It's technically correct in a very narrow context. This isn't memoized throughout the lifecycle of the app, if someone is talented enough to somehow switch networks while this is still in memory, they would definitely be able to bypass the check in more easier and creative ways than exploiting a transient memoizer.

pete-watters:
> > > Not sure I follow the need for this. The address differs across networks, so seems redundant to also list network?
> > 
> > 
> > I unleashed `codex` to find some 'bug's but it seems man beat machine again?
> > Based on their findings:
> > 
> > * in `complianceValidator` we return `true` if `network !== 'mainnet'`
> > * if you use `testnet` to send to a banned address mainnet address  compliance will pass
> > * the compliance check is memoised based on address only
> > * that means you can switch back to `mainnet` and send to the banned address without validation
> > 
> > I'm happy to scrap this if that's not correct
> 
> It's technically correct in a very narrow context. This isn't memoized throughout the lifecycle of the app, if someone is talented enough to somehow switch networks while this is still in memory, they would definitely be able to bypass the check in more easier and creative ways than exploiting a transient memoizer.

Thanks @tigranpetrossian 👍  lets scratch this off as un-necessary so

PR #1719, opened 2025-10-21
feat: stacks transaction fees service

alexp3y:
> Try out Leather build 878ab96 — [Extension build](https://github.com/leather-io/mono/actions/runs/18713539447), [Test report](https://leather-io.github.io/playwright-reports/feat/transaction-fees-service), [Storybook](https://feat/transaction-fees-service--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/transaction-fees-service)

Introduces a new unified and extensible `TransactionFees` model.

Implements a `StacksTransactionFeesService` that returns tiered fees using the new model by porting existing Stacks fee logic into services in a simplified way, combining live fee rates from the Hiro Stacks API with defaults + min/max boundaries supplied via new app config.

A `BitcoinTransactionService` is also included as a stub and will be implemented next.

The new fees model and services are currently not used by apps but will be integrated by the new swaps implementation before being rolled out to other features (token sends, fee selection, etc.)

### DIFF
--- a/packages/models/src/fees/transaction-fees.model.ts
+++ b/packages/models/src/fees/transaction-fees.model.ts
@@ -1,0 +1,42 @@
+import { Money } from '../money.model';
+import { Blockchain } from '../types';
+
+export const transactionFeeTiers = ['low', 'standard', 'high'] as const;
+export type TransactionFeeTier = (typeof transactionFeeTiers)[number];
+
+export interface TransactionFees {
+  readonly chain: Blockchain;
+  readonly options: Record<TransactionFeeTier, TransactionFeeQuote>;
+}
+
+export const transactionFeeQuoteType = ['flat', 'feeRate', 'evm1559'] as const;
+export type TransactionFeeQuoteType = (typeof transactionFeeQuoteType)[number];
+
+export interface BaseTransactionFeeQuote {
+  readonly type: TransactionFeeQuoteType;
+  readonly value: Money;
+}
+
+export interface FlatTransactionFeeQuote extends BaseTransactionFeeQuote {
+  readonly type: 'flat';
+}
+
+export interface FeeRateTransactionFeeQuote extends BaseTransactionFeeQuote {
+  readonly type: 'feeRate';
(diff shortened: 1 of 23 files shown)


packages/models/src/fees/transaction-fees.model.ts
```diff
@@ -0,0 +1,41 @@
+import { Money } from '../money.model';
+
+export const transactionFeeTiers = ['low', 'standard', 'high'] as const;
+export type TransactionFeeTier = (typeof transactionFeeTiers)[number];
+
+export interface TransactionFees {
+  chain: 'stacks' | 'bitcoin';
```
kyranjamie:
```suggestion
  chain: SupportedBlockchains;
```

packages/models/src/fees/transaction-fees.model.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface FlatTransactionFeeQuote extends BaseTransactionFeeQuote {
+  type: 'flat';
+}
+
+export interface FeeRateTransactionFeeQuote extends BaseTransactionFeeQuote {
+  type: 'feeRate';
+  rate: number;
+  rateUnit: 'sats/vB' | 'µSTX/byte';
+  estimatedTxSize: number;
+  sizeUnit: 'vB' | 'byte';
```
kyranjamie: should these be added separately so they can be discriminating types?

kyranjamie (approved):

packages/models/src/fees/transaction-fees.model.ts
alexp3y: I could def see that for if/when we add any additional type-specific (btc vs stx) fields. At this point I think it's okay b/c the services will build these authoritatively as readonly and apps will just consume passively, which should guard against invalid field combinations (e.g. `{ rateUnit: 'sats/vB',  sizeUnit: 'byte', ... }`)

PR #1720, opened 2025-10-21
fix(mobile): prevent URL oscillation in browser deep linking

### LINKED ISSUE #1721: [Android] WebView deep linking loop: URL oscillates between default and target URLs
## Issue Description
When opening deep links on Android, the app sometimes oscillates between the Leather default URL and the deep link target URL.
In some cases, this oscillation prevents the target link from opening at all.
### Evidence
A video demonstrating the issue has been attached for reference: 
[Demo Video 1](https://drive.google.com/file/d/15ghWh_msL2jTzwvfmh6UxcIvYm06gSB0/view)
[Demo Video 2](https://www.loom.com/share/d6c3f8ff29c847b88875d305ef133daf)
### Expected Behavior
The deep link should open directly to its target URL without any oscillation.
### Actual Behavior
1. The WebView alternates rapidly between the default and deep link target URL.
2. Navigation never settles on the target page, it just opens the default url([leather.io](http://leather.io/)).
3. Occurs only on Android; iOS is unaffected.

nageenhaider:
## Problem Description
The in-app browser had an issue where URLs would oscillate between the Leather default URL and the deep link target URL, sometimes preventing the target URL from opening at all. This appears to happen only on Android.

### Possible Issue
The `goToUrl()` function was being called unconditionally on every navigation state change in the WebView. This meant that even during intermediate loading states and navigation transitions, the URL would be updated, causing the search bar and internal state to change prematurely.

### Proposed Solution
Added a conditional check in the `handleWebViewNavigationStateChange` function(used as the onNavigationStateChange callback for the WebView component) to only call `goToUrl()` if not in the middle of navigation and when the URL has actually changed.

### DIFF
--- a/apps/mobile/src/features/browser/browser/browser.tsx
+++ b/apps/mobile/src/features/browser/browser/browser.tsx
@@ -59,7 +59,9 @@ export function Browser({
   async function handleWebViewNavigationStateChange(newNavState: WebViewNavigation) {
     setNavState(newNavState);
     const url = new URL(newNavState.url);
-    goToUrl(newNavState.url);
+    if (!newNavState.loading && newNavState.url !== searchUrl) {
+      goToUrl(newNavState.url);
+    }
     if (newNavState.loading) return;
 
     const screenshotUri = await captureScreenshot(viewShotRef, url.hostname);


pete-watters (approved):
Thanks @nageenhaider

PR #1723, opened 2025-10-22
feat: rehaul approver flow

edgarkhanzadian:
Implement better UX for submitting transactions and displaying the status of transactions as per Figma designs [here](https://www.figma.com/design/NbCzd27zkHOuhOgWD0x0nw/%F0%9F%91%8D-Product-area--Approver-view?node-id=29404-62309&p=f&t=F7JtnCvJ6feFpBRD-0)

Also refactored the flow while i was at it.

This still would display a toast error when the transaction fails to broadcast, but that makes it easier for the user to retry the transaction in case if a nonce is wrong or network was unavailable at a time of the transaction broadcast. I suggest we raise it during the design office hours.

https://github.com/user-attachments/assets/3c8305d8-d7ca-4a23-85ea-3d41ff59941a

https://github.com/user-attachments/assets/8e1ec877-9cc3-403a-b7d4-48c7a3195428

NOTE: Still need to do a pass of my own over the PR to do a little clean up, but general logic is done


kyranjamie (changes requested):
What prompted you to work on these changes @edgarkhanzadian? I've pushed back on these designs before and think we should not proceed with them.

They look great, as does your implementation, but they change a behaviour that many apps work with: that the approver flow is finished when a tx is broadcast. In isolation they look fantastic, but in practice they won't work with some apps. These changes introduce an uncertainty (from the app's perspective) between where the first flow finishes and the next one starts. 

If the approver flow doesn't finish at the same time we send the txid response to an app, **the app doesn't know whether the user is in the context of our approver flow or their UX**. It could be either. Some apps trigger events, e.g. an animation or another approver flow when the the first has completed. But, with these changes, these actions happen in the background because the user might not have manually dismissed the window. 

What happens in this PR if an app triggers an approver event while the previous flow is still open?

apps/mobile/src/features/approver/components/approver-buttons.tsx
```diff
@@ -1,43 +1,145 @@
...
+      setApproverState('broadcasted_checkmark');
+      setTimeout(() => {
+        setApproverState('broadcasted_details');
+      }, 1000);
+    },
+    onBroadcastError() {
+      setApproverState('retry');
+    },
+  }));
+  function boop() {
```
kyranjamie: Can we call this something else?

apps/mobile/src/features/approver/components/approver-buttons.tsx
```diff
@@ -1,43 +1,145 @@
...
+    if (pressCount === 5) {
+      displayToast({ type: 'info', title: t`Alright, alright, I get it!` });
+    }
+    setPressCount(pressCount + 1);
```
kyranjamie: What is this? I don't think we should use language as casual as this in the product

apps/mobile/src/features/approver/components/approver-buttons.tsx
```diff
@@ -1,43 +1,145 @@
...
+import { assertUnreachable, match } from '@leather.io/utils';
+
+export interface ApproverMethods {
+  onBroadcast(): void;
+  onBroadcastSuccess(): void;
+  onBroadcastError(): void;
+}
 
-import { ApproverState } from '../utils';
+type ApproverButtonsType = 'in_house' | 'approver';
```
kyranjamie: How do the buttons vary between in house and approver?

apps/mobile/src/features/approver/layouts/base-stx-message-approver.layout.tsx
```diff
@@ -59,14 +65,25 @@ export function BaseStxMessageApproverLayout({
...
-          </Button>
-          <Button flex={1} onPress={onApprove}>
-            {t`Approve`}
-          </Button>
-        </Approver.Actions>
+        <ApproverButtons
+          ref={buttonsRef}
+          onBack={onCloseApprover}
+          onApprove={() => {
+            buttonsRef.current?.onBroadcast();
```
kyranjamie: I don't really follow why these events are handled this way? These could be triggered internally no? How is the imperative handled needed?

apps/mobile/src/features/approver/components/approver-buttons.tsx
edgarkhanzadian: there was a discussion regarding easter eggs, but removing this as indeed it's not needed

apps/mobile/src/features/approver/components/approver-buttons.tsx
edgarkhanzadian: Removed

apps/mobile/src/features/approver/components/approver-buttons.tsx
edgarkhanzadian: They varied by the titles, but removed for now

apps/mobile/src/features/approver/layouts/base-stx-message-approver.layout.tsx
edgarkhanzadian: ~Moved from imperative handler to state-keeping hook~ ah right everything can be done inside of the component, moving it in there

edgarkhanzadian:
Updated the PR s.t. approver flow would close as soon as the rpc request is finished, while the stx/btc/sip10 sends would wait for the user's input.

https://github.com/user-attachments/assets/87e44455-237d-4227-b240-7ea4227c9bb5

https://github.com/user-attachments/assets/a92f57c7-f014-4456-9885-27115dd8fc7d


https://github.com/user-attachments/assets/0792d614-9d14-4fb0-81b5-fcd3f8e52187

apps/mobile/src/features/approver/layouts/base-stx-tx-approver.layout.tsx
```diff
@@ -162,18 +175,37 @@ export function BaseStxTxApproverLayout({
...
-              {sendButtonTitle ?? t`Approve`}
-            </Button>
-          </Approver.Actions>
+          <ApproverButtons
+            onBack={onBack}
+            onClose={onCloseApprover}
+            onApprove={async () => {
+              const txid = await onApprove();
+              setBroadcastedTxid(txid);
+            }}
```
kyranjamie, line 188: Should state manipulating actions take place in the layout component?

apps/mobile/src/features/approver/layouts/types.ts
```diff
@@ -0,0 +1,4 @@
+export type OnApproveFuncSignTx = () => Promise<string>;
+
+export type OnApproveFuncSigningMessage = () => Promise<void>;
+export type OnApproveFuncGetAddresses = () => void;
```
kyranjamie: With use-case specific actions within the core approver feature, we're not really separating the responsibilities of instances of approver and the approver itself.

kyranjamie (approved):

apps/mobile/src/features/approver/layouts/base-stx-tx-approver.layout.tsx
edgarkhanzadian: i believe it's OK if manipulating state is used for displaying correct layout. In this case it is for 1) disabling pressable items 2) adding functionality to a clipboard button (which isn't a layout thing but it is a very simple functionality)

apps/mobile/src/features/approver/layouts/types.ts
edgarkhanzadian: Yea these types confuse more than help, i'll remove these

PR #1725, opened 2025-10-22
feat(state): adds state package

kyranjamie:
> Try out Leather build 4f710e2 — [Extension build](https://github.com/leather-io/mono/actions/runs/18838666793), [Test report](https://leather-io.github.io/playwright-reports/feat/state-package), [Storybook](https://feat/state-package--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/state-package)

This PR continues with the proposal in https://github.com/leather-io/mono/pull/1670, and is precursory work for multi-wallet support in extension.

I've added the `@leather.io/state` package and migrated across the wallet store, and flattened the state structure for keychains. The chain-specific nesting wasn't necessary as the keychains themselves are mostly agnostic of chain details. I've labelled the `chain` anyway in case we we need a discriminator. This requires a migration, which I've tested and demoed below. 

**Migration demo**

See version number increment and state change

https://github.com/user-attachments/assets/766c70a6-8e72-4ab5-8d24-1137daa55130


tigranpetrossian (approved):
Huge step towards consistency, thanks Kyran.

apps/mobile/src/store/migrations/migrate-0-1-flatten-keychains.ts
```diff
@@ -0,0 +1,30 @@
+import { EntityState } from '@reduxjs/toolkit';
```
pete-watters, line 1: Is this a short-lived file until we determine that all users are on `X` version and have been migrated? 

Maybe we can incentivise them to upgrade sooner?

pete-watters (approved):
Great work here @kyranjamie  👍

apps/mobile/src/store/migrations/migrate-0-1-flatten-keychains.ts
kyranjamie: I wouldn't say this is that short-lived. If a user returns to their wallet after a 12 months cycling around the world trip, their wallet should update correctly.

Migrations don't really hurt, it's fine to keep them here long term

PR #1726, opened 2025-10-22
feat: integrate onramper in mobile and extension

camerow:
> Try out Leather build a1fc814 — [Extension build](https://github.com/leather-io/mono/actions/runs/18908669882), [Test report](https://leather-io.github.io/playwright-reports/will/onramper-example), [Storybook](https://will/onramper-example--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=will/onramper-example)


apps/extension/src/app/pages/fund/fund.tsx
```diff
@@ -1,57 +1,104 @@
-import { Outlet, useParams } from 'react-router';
+import { useEffect, useRef } from 'react';
+import { Outlet } from 'react-router';
 
-import type { Blockchain, BtcBalance, CryptoCurrency, StxBalance } from '@leather.io/models';
+import crypto from 'crypto';
```
kyranjamie: We already have a lib for handling hashing operations https://github.com/paulmillr/noble-hashes

edgarkhanzadian:
A little bit skeptical about buy/sell buttons placements in a scrollview on mobile, but i'm not sure how we can fit more buttons on that row otherwise.

https://github.com/user-attachments/assets/794277c4-2c38-4aa9-bad4-dba5ae5d8fb4


https://github.com/user-attachments/assets/3f682215-3f14-4401-b549-b1e2f7b98ab6


https://github.com/user-attachments/assets/f0f5bd33-99fe-4b20-9971-ef18f0bedf7d

camerow:
@edgarkhanzadian I tend to agree but as an initial stop gap i think it's ok. We'll be forced to solve for it once swaps are out and can go consolidate / relocate later. @fabric-8 do you have any thoughts here?

fabric-8:
How about we only offer "buy" on home for now and additionally sell only within the respective token details view?

camerow:
@edgarkhanzadian let's go with that for now ^

edgarkhanzadian:
> @edgarkhanzadian let's go with that for now ^

i've setup feature flags for both, so we can enable only buy and not sell

tigranpetrossian (approved):

packages/ui/src/exports.web.ts
```diff
@@ -1,8 +1,22 @@
+export * from './components/approver/approver.web';
+export * from './components/avatar';
+export * from './components/collectibles/index.web';
+export * from './components/favicon/favicon.web';
+export * from './components/onramper/index.web';
+export * from './components/pressable/pressable.web';
+export * from './components/select';
+export * from './components/sheet';
```
kyranjamie, line 7: Not sure why we have `index.ts` files here. They exports should always use the `*.{web,native}.tsx` naming otherwise we'll mix files

kyranjamie (commented):
Thanks @camerow 

Here we've moved onramper into `ui`. `ui` is a domain-agnostic package, and I'm a little concerned to start mixing generic buttons with leather-specific feature. Seems like a dilution of a worthy, narrow responsibility. Though, it is great to see shared code cross-platform.

I wonder if we should make a separate domain-aware UI package. Propose name `features`. A package that depends on UI and layers on top domain-aware features. One that remains platform-agnostic, using the [render prop pattern](https://www.patterns.dev/react/render-props-pattern/) to do so.

cc/ @leather-io/leather-devs

pete-watters:
> Thanks @camerow
> 
> Here we've moved onramper into `ui`. `ui` is a domain-agnostic package, and I'm a little concerned to start mixing generic buttons with leather-specific feature. Seems like a dilution of a worthy, narrow responsibility. Though, it is great to see shared code cross-platform.
> 
> I wonder if we should make a separate domain-aware UI package. Propose name `features`. A package that depends on UI and layers on top domain-aware features. One that remains platform-agnostic, using the [render prop pattern](https://www.patterns.dev/react/render-props-pattern/) to do so.
> 
> cc/ @leather-io/leather-devs

This is a good idea and something @camerow and I were talking about last week. A `features` package would be helpful to share domain specific code e.g. `activity` / `collectibles` / `onramper` etc. shared code could live in there. It could have a similar setup to the `ui` package where it supports `.native` and `.web` and means we could share business logic and components without bloating the `ui` package.

Render prop seems pretty clean also so could be a better fit but

packages/ui/src/exports.web.ts
pete-watters: I agree this isn't perfect but we also have other instances of that in this file already:
```
export * from './components/select';
export * from './components/sheet';
export * from './components/toast';
export * from './components/tooltip';
``` 
etc. 

We can refactor that separately

pete-watters (approved):
Nice work @camerow & @edgarkhanzadian . 

I agree with @kyranjamie that `OnRamper` shouldn't live in the UI package but think we can merge this now and then migrate once we have setup a new `features` package. 

The features package would be a nice pre-requisite to our upcoming work on `OnBoarding` / `extension` revamp as it would allow us to add a lot of UI code there as we go

edgarkhanzadian:
@pete-watters @kyranjamie great thoughts! I agree it didn't feel right to put it in the ui package and having something like feature would be much more natural. I'll set it up as part of this PR, shouldn't be too hard

packages/features/.editorconfig
```diff
@@ -0,0 +1,3 @@
+# Windows files
+[*.bat]
+end_of_line = crlf
```
kyranjamie: do we need this if we have editorconfig at root?

edgarkhanzadian: Probably not, removing

packages/features/tsup.config.native.ts
```diff
@@ -0,0 +1,26 @@
...
+  minifyWhitespace: true,
+  splitting: true,
+  treeshake: true,
+  target: 'es2022',
+  loader: {
+    '.js': 'jsx',
+    '.otf': 'file',
+    '.png': 'file',
+    '.svg': 'jsx',
+    '.ttf': 'file',
```
kyranjamie: we don't need much of this config

edgarkhanzadian (approved):

mica000:
CleanShot 2025-11-17 at 18 01 33@2x


Moonpay seems to appear out of work for most of the currencies I try. Is anyone seeing the same

PR #1730, opened 2025-10-23
feat: add bitcoin transaction fees service

alexp3y:
> Try out Leather build 5885a6f — [Extension build](https://github.com/leather-io/mono/actions/runs/18754923297), [Test report](https://leather-io.github.io/playwright-reports/feat/btc-transaction-fees-service), [Storybook](https://feat/btc-transaction-fees-service--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/btc-transaction-fees-service)

Builds on recent work to add a `BitcoinTransactionFees` service returning the new unified `TransactionFees` model.

Starting to slowly shrink the circle around our coin-selection logic, which can now be mostly encapsulated in a `CoinSelectionService` and treated with some higher level abstractions. Both new services (btc transaction fees and coin-selection) can fetch live fee rates and combine with account available UTXOs, reducing requirements and complexity for implementing clients.

Renamed the "uneconomical" UTXOs and balances to "dust" for accuracy. Dust amounts can be removed from the available UTXO set and balance regardless of Tx context. Whereas any "uneconomical" determinations are inherently linked to specific transaction permutations and therefore can only be done during coin-selection.

Currently, neither service is used but they are available to support the net-new swaps implementation. 

Follow-on work will add address-type specific dust thresholds to the dust UTXO determination, as well as improvements to the coin-selection algorithm itself.

### DIFF
--- a/packages/services/src/coin-selection/bitcoin-coin-selection.service.ts
+++ b/packages/services/src/coin-selection/bitcoin-coin-selection.service.ts
@@ -1,0 +1,82 @@
+import { injectable } from 'inversify';
+
+import {
+  CalculateMaxSpendResponse,
+  CoinSelectionOutput,
+  CoinSelectionRecipient,
+  calculateMaxSpend,
+  determineUtxosForSpend,
+  determineUtxosForSpendAll,
+} from '@leather.io/bitcoin';
+import { Money, OwnedUtxo } from '@leather.io/models';
+
+import { LeatherApiClient } from '../infrastructure/api/leather/leather-api.client';
+import { AccountRequest } from '../types';
+import { UtxosService } from '../utxos/utxos.service';
+
+export interface CalculateMaxSpendRequest {
+  account: AccountRequest;
+  recipient: string;
+  feeRate?: number;
+}
+
+export interface CoinSelectionRequest {
+  account: AccountRequest;
+  recipients: CoinSelectionRecipient[];
+  feeRate?: number;
+  isMaxSpend?: boolean;
+}
+
+export interface CoinSelectionResult {
+  inputs: OwnedUtxo[];
+  outputs: CoinSelectionOutput[];
+  estimatedTxSize: number;
+  fee: Money;
+}
+
+@injectable()
+export class BitcoinCoinSelectionService {
+  constructor(
+    private readonly utxosService: UtxosService,
+    private readonly leatherApiClient: LeatherApiClient
+  ) {}
+
+  async performCoinSelection(
+    { account, recipients, feeRate, isMaxSpend }: CoinSelectionRequest,
+    signal?: AbortSignal
+  ): Promise<CoinSelectionResult> {
+    const { available: availableInputUtxos } = await this.utxosService.getAccountUtxos(
+      account,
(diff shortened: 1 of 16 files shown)


packages/services/src/coin-selection/bitcoin-coin-selection.service.ts
```diff
@@ -0,0 +1,82 @@
...
+    const { available: availableInputUtxos } = await this.utxosService.getAccountUtxos(
+      account,
+      signal
+    );
+    if (!feeRate) {
+      const feeRates = await this.leatherApiClient.fetchBitcoinFeeRates({ signal });
+      feeRate = feeRates.standard.rate;
+    }
+    const result = isMaxSpend
+      ? determineUtxosForSpendAll({ feeRate, recipients, utxos: availableInputUtxos })
```
pete-watters, line 57: Maybe we can later rename `determineUtxosForSpendAll` as `determineMaxSpendUtxos`? 

It could be just me but reading this was unclear. Then when I looked into it I see we still have duplicate code in the `extension` and `package` for it we can refactor soon

packages/services/src/utxos/utxos.service.ts
```diff
@@ -24,7 +24,7 @@ export interface UtxoTotals {
   inbound: OwnedUtxo[];
   outbound: OwnedUtxo[];
   protected: OwnedUtxo[];
-  uneconomical: OwnedUtxo[];
+  dust: OwnedUtxo[];
```
pete-watters, line 27: Good job with this rename 👍

packages/services/src/utxos/utxos.utils.ts
```diff
@@ -52,10 +52,11 @@ export function isUnconfirmedUtxo(utxo: Utxo) {
   return !utxo.height;
 }
 
-export const uneconomicalSatThreshold = 500;
+export const dustSatThreshold = 500;
 
-export function isUneconomicalUtxo(utxo: Utxo) {
-  return Number(utxo.value) < uneconomicalSatThreshold;
+// TODO: Add address-type specific dust thresholds
```
pete-watters, line 57: Do we **need** to do this?

pete-watters (approved):
Great stuff @alexp3y 👍

packages/services/src/utxos/utxos.utils.ts
alexp3y: Yeah, it should be a quick to-do but we need different thresholds for P2TR vs P2WPKH b/c we handle both UTXO types here and the thresholds are quite different (546 vs 330).

packages/services/src/coin-selection/bitcoin-coin-selection.service.ts
alexp3y: I agree that would be a nice improvement. There are several relative easy follow-on refactors we can do simultaneously here to improve (remove duplication, rename for clarity, improve size estimations) without needing to mess with the underlying coin-selection algorithm.

PR #1738, opened 2025-10-28
fix: subsequent approver requests + signMessage response

edgarkhanzadian:
> Try out Leather build 5162256 — [Extension build](https://github.com/leather-io/mono/actions/runs/18907489121), [Test report](https://leather-io.github.io/playwright-reports/fix/subsequent-approver-requests), [Storybook](https://fix/subsequent-approver-requests--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/subsequent-approver-requests)

- Set imperative handle on approver sheet
- Add 500ms delay before opening the sheet if the sheet is not fully closed yet
- Fix signMessage rpc response, it should also have "message" property in the response

https://github.com/user-attachments/assets/fcf2b1ae-d056-4152-94fa-85539e7dccca

### DIFF
--- a/apps/mobile/src/core/global-sheet-provider.tsx
+++ b/apps/mobile/src/core/global-sheet-provider.tsx
@@ -1,5 +1,9 @@
 import { createContext, useContext, useRef } from 'react';
 
+import {
+  ApproverSheetInstance,
+  type ApproverSheetRef,
+} from '@/features/browser/approver-sheet/approver-sheet';
 import { ReceiveSheetInstance, ReceiveSheetRef } from '@/features/receive/receive-sheet';
 import { SendSheetInstance, SendSheetRef } from '@/features/send/send-sheet';
 import {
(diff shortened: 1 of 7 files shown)


kyranjamie (approved):
Thanks @edgarkhanzadian. Any reason the type compiler didn't catch the response not containing all the values?

pete-watters (approved):

edgarkhanzadian:
@kyranjamie good point, first thing i've noticed is that for some reason we use `looseObject` in signMessage schema definition 🤔 Also, i've noticed that we don't have `message` property in the result of the rpc response

edgarkhanzadian:
@kyranjamie removed looseObject zod definitions from rpc responses.

kyranjamie (approved):
Thanks @edgarkhanzadian—yeah for the `result` objects I guess we don't ever want `looseObject` (we do more often for inbound payloads such that we don't break existing apps that might've been using an older api/prop)

PR #1740, opened 2025-10-29
fix(mobile): update token routes to use serialized assetIds

pete-watters:
> Try out Leather build 19af5f4 — [Extension build](https://github.com/leather-io/mono/actions/runs/18916956781), [Test report](https://leather-io.github.io/playwright-reports/fix/collectible-details-bugs), [Storybook](https://fix/collectible-details-bugs--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/collectible-details-bugs)

This PR updates the routing for SIP-009 other tokens to use a serialized `CryptoAssetId` as a route parameter. 

SIP-009 collectibles in the same collection have the same `assetId` so we need to use `tokenId` to help distinguish between them. 

This refactor also helps clean up the implementation of Token details

### DIFF
--- a/apps/mobile/src/components/home/home-with-account-screen.tsx
+++ b/apps/mobile/src/components/home/home-with-account-screen.tsx
@@ -21,8 +21,10 @@ import { useSettings } from '@/store/settings/settings';
 import { useAccountScaledBalanceAnalytics } from '@/utils/analytics-hooks';
 import { useRouter } from 'expo-router';
 
+import { btcAsset, stxAsset } from '@leather.io/constants';
 import { AccountId, CryptoAssetProtocols } from '@leather.io/models';
 import { SheetInstance } from '@leather.io/ui/native';
+import { getAssetId, serializeAssetId } from '@leather.io/utils';
 
 import { AccountScreenHeader } from './account-screen-header';
 import { AssetTabs } from './components/asset-tabs';
@@ -60,6 +62,7 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
   }
 
   const router = useRouter();
+
   function onOpenToken({ assetId, assetProtocol }: TokenDetailsProps) {
     router.navigate({
       pathname: '/(tabs)/(index)/[assetProtocol]/[assetId]',
@@ -101,7 +104,7 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
                     ? () =>
                         onOpenToken?.({
                           assetProtocol: CryptoAssetProtocols.nativeBtc,
-                          assetId: 'BTC',
+                          assetId: serializeAssetId(getAssetId(btcAsset)),
                         })
                     : undefined
                 }
@@ -114,7 +117,7 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
                     ? () =>
                         onOpenToken?.({
                           assetProtocol: CryptoAssetProtocols.nativeStx,
-                          assetId: 'STX',
+                          assetId: serializeAssetId(getAssetId(stxAsset)),
                         })
                     : undefined
                 }
--- a/apps/mobile/src/features/balances/assets/assets-list.tsx
+++ b/apps/mobile/src/features/balances/assets/assets-list.tsx
@@ -2,13 +2,13 @@ import { ReactElement, useMemo } from 'react';
 
 import { Screen } from '@/components/screen/screen';
 import { sortSip10Balances } from '@/features/balances/assets/utils/sort-sip10-balances';
-import { useRunesFlag } from '@/features/feature-flags';
 import { RefreshControl } from '@/features/refresh-control/refresh-control';
 import { TokenDetailsProps } from '@/features/token/types';
 import { useRunesAccountBalance } from '@/queries/balance/runes-balance.query';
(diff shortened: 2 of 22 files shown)


packages/utils/src/assets/asset-id.ts
```diff
@@ -68,13 +68,13 @@ export function getAssetId(asset: CryptoAsset): CryptoAssetId {
...
     case 'sip9':
-      return buildSip9AssetId(asset);
+      return createSip9AssetId(asset);
     default:
       assertUnreachable(asset);
   }
 }
 
-function buildSip9AssetId(asset: Sip9Asset): CryptoAssetId {
+export function createSip9AssetId(asset: Sip9Asset): CryptoAssetId {
```
pete-watters, line 85: I'm making this small change to a util that's only used in `mobile` but the `extension` Playwright tests are failing for Shard 3. 

I guess we need to test everything if we change packages but those tests can be flaky and flag non existent issues

apps/mobile/src/features/token/stacks/sip9-details.tsx
```diff
@@ -15,7 +16,9 @@ export function Sip9TokenDetails({ assetId, account }: Sip9TokenDetailsProps) {
   const { fingerprint, accountIndex } = account;
   const height = useCollectibleHeight();
 
-  const collectible = useAccountCollectibleByAssetId(fingerprint, accountIndex, assetId);
+  const [sip9AssetId, tokenId] = assetId.split('|');
```
alexp3y: Maybe we should add a utility for this - Parsing out the sub-fields of `Sip9AssetId.id`? The pipe-delimiter should ideally be a hidden implementation detail (could change in theory).

apps/mobile/src/features/token/stacks/sip9-details.tsx
```diff
@@ -15,7 +16,9 @@ export function Sip9TokenDetails({ assetId, account }: Sip9TokenDetailsProps) {
...
+  assertExistence(sip9AssetId, 'SIP-9 Asset ID is required');
+  const collectible = useAccountCollectibleByAssetId(fingerprint, accountIndex, sip9AssetId);
```
alexp3y: Having to use the underlying Sip9 field "sip9.assetId" for this call make it a bit confusing (lots of "assetIds" getting thrown around). 

It would be nice to completely hide `sip9.assetId` (and any other asset-specific identifier fields) going forward in favor of `AssetId`. 

Might be out of scope but using an `AssetId` object here in this hook (and then just using the `matchesAssetId` in the implementation) could clean things up a bit.

alexp3y (approved):
Couple of comments but might be out of scope here.

apps/mobile/src/components/home/home-with-account-screen.tsx
```diff
@@ -101,7 +104,7 @@ export function HomeScreenWithAccount({ currentAccount }: HomeScreenWithAccountP
                     ? () =>
                         onOpenToken?.({
                           assetProtocol: CryptoAssetProtocols.nativeBtc,
-                          assetId: 'BTC',
+                          assetId: serializeAssetId(getAssetId(btcAsset)),
```
pete-watters, line 107: @alexp3y : `BTC` / `STX` don't need `assetId` but I am doing it for route consistency.

Now that I am serialising the `assetId` and it contains the protocol, perhaps I should revisit the route structure and remove the protocol level? `'/(tabs)/(index)/[assetProtocol]/[assetId]'` => `'/(tabs)/(index)/[assetId]'`

Components can still be grouped in folders but maybe this will be a better approach to use in the extension also as we can have cleaner visible URLs?

pete-watters:
> Couple of comments but might be out of scope here.

Thanks for the feedback and for helping come up with a solution. I've refactored the code for both `Token` and `Collectible` details routing to use `SerializedCryptoAssetId` and this has fixed problems with `SIP-009` and also improved the current implementation. 

Let me know if you spot any other things I could improve. I'm considering flattening the routes now we are looking at this anyway

apps/mobile/src/features/balances/assets/assets-list.tsx
```diff
@@ -36,18 +36,11 @@ export function AssetsList({ sip10Data, runesData, header, onPressToken }: Asset
         renderAsset({
           item,
           onPress: onPressToken
-            ? () => {
-                if (item.asset.protocol === 'sip10')
-                  onPressToken?.({
-                    assetId: item.asset.assetId,
-                    assetProtocol: item.asset.protocol,
-                  });
-                if (runesFlag && item.asset.protocol === 'rune')
```
pete-watters, line 45: Pulling off the plaster of the `runesFlag` as it has been turned on for a while now

pete-watters:
@alexp3y : I'm just merging this based on previous approval so I can consume it in the next UI branch. When you have time please check and let me know if I can improve anything else here 🙏

apps/mobile/src/components/home/home-with-account-screen.tsx
alexp3y: Agree on staying consistent even if it's not technically necessary.

It seems like we could probably get rid of /{assetProtocol} now considering we branch to all protocol sub-components from a single route handler and we can get everything we need directly from the assetId.

PR #1745, opened 2025-10-30
fix(mobile): token routing refactor

pete-watters:
> Try out Leather build bd8e7dd — [Extension build](https://github.com/leather-io/mono/actions/runs/18935457652), [Test report](https://leather-io.github.io/playwright-reports/fix/token-routing), [Storybook](https://fix/token-routing--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/token-routing)

This PR: 
- fixes a bug introduced by [this PR](https://github.com/leather-io/mono/pull/1740/files) where SIP-009 token routes were not working 
- merges code for collectibles and tokens - a collectible is just another token
- flattens the collectibles route structure
    - now we pass a serialised `assetId` that contains the `protocol`
    - this means there is no need to have `protocol` as part of the route


apps/mobile/src/app/(tabs)/(index)/[assetId]/index.tsx
```diff
@@ -36,8 +35,10 @@ export default function AccountTokenScreen() {
       return <InscriptionDetails account={currentAccount} assetId={assetId} />;
     case CryptoAssetProtocols.stamp:
       return <StampDetails account={currentAccount} assetId={assetId} />;
-
     default:
-      assertUnreachable(assetProtocol);
+      if (!isSupportedAssetProtocol(assetProtocol)) {
```
Comment (pete-watters) [inline apps/mobile/src/app/(tabs)/(index)/[assetId]/index.tsx:39]: Some assets are not supported (`brc20`, `src20`, `stx20`). They won't appear in the UI yet so I added a guard for supported protocols

packages/models/src/assets/asset-id.model.ts
```diff
@@ -3,4 +3,5 @@ import { CryptoAssetProtocol } from './asset.model';
 export interface CryptoAssetId {
   protocol: CryptoAssetProtocol;
   id: string;
+  tokenId?: number;
```
pete-watters, line 6: I'm adding this as optional as it's needed for `SIP-009` to ensure unique assets are matched. That seemed cleaner than having a separate type and handling for that

tigranpetrossian (approved):

packages/models/src/assets/asset-id.model.ts
alexp3y: Adding this here overloads the protocol-agnostic model type with concerns specific to `Sip9Asset`. 

It shouldn't be necessary for comparisons though - The Sip9 `tokenId` value is already integrated into the `id` field for Sip9 assets via the `createSip9AssetId` helper so any equivalence helpers should already work as-is:
e.g.
```typescript
matchesAssetId({ protocol: 'sip9', assetId: 'asdf', tokenId: '123' } as Sip9Asset, { protocol: 'sip9', id: 'asdf|123' }) // -> true
```

packages/utils/src/assets/asset-id.ts
```diff
@@ -107,12 +98,14 @@ export function serializeAssetId(assetId: CryptoAssetId): SerializedCryptoAssetI
...
   if (!Object.keys(CryptoAssetProtocols).includes(protocol)) {
     throw new Error(`Unrecognized Asset Protocol: ${protocol}`);
   }
+
   return {
     protocol: protocol as CryptoAssetProtocol,
     id,
+    tokenId: tokenId ? parseInt(tokenId, 10) : undefined,
   };
 }
```
alexp3y, line 111: If the goal is to access the Sip9 `tokenId` of the deserialized assetId object, as an alternative to adding it to `CryptoAssetId`, we could introduce some additional protocol-specific helpers to construct / parse the multipart `id` field:

```typescript
export function createSip9AssetId(asset: Sip9Asset): CryptoAssetId {
  return {
    protocol: asset.protocol,
    id: formatSip9IdField(asset),
  };
}

export function formatSip9IdField(asset: Sip9Asset) {
  return `${asset.assetId}|${asset.tokenId}`;
}
export function parseSip9IdField(id: string) {
  const [assetId, tokenId] = id.split('|');
  return { assetId, tokenId: Number(tokenId) };

// elsewhere...
const { protocol, id } = deserializeAssetId(serializedAssetId);

if (protocol === 'sip9') {
  const { assetId, tokenId } = parseSip9IdField(id);
}


```

alexp3y:
@pete-watters Added a few comments before I saw you had merged.

pete-watters:
> @pete-watters Added a few comments before I saw you had merged.

Thanks @alexp3y! I'll check and apply improvements retrospectively, feedback is always welcome.

packages/utils/src/assets/asset-id.ts
pete-watters: Thanks @alexp3y 👍  I will refactor this to avoid bloating `CryptoAssetId`. I had tried to do as you suggested with a separate `SIP-9` parser but it seemed messier so I went for the quick win.  

I'll take a look and refactor it soon.

PR #1746, opened 2025-10-30
feat(mobile): add bns token details

pete-watters:
> Try out Leather build 7145a82 — [Extension build](https://github.com/leather-io/mono/actions/runs/18939868347), [Test report](https://leather-io.github.io/playwright-reports/feat/bns-collectible-details), [Storybook](https://feat/bns-collectible-details--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/bns-collectible-details)

This PR adds token details pages for BNS names and some refactors

DEMO:

https://github.com/user-attachments/assets/b2dfd962-a18b-4ae0-9f8b-3a9aeaaf6529


packages/services/src/bns/bns.utils.ts
```diff
@@ -44,6 +48,8 @@ export function buildBnsName(owner: string, name: string, namespace: string): Bn
     name,
     namespace,
     fullName: `${name}.${namespace}`,
+    renewalHeight: '',
```
pete-watters: Adding this as `''` as they get overwritten when mapping with `bnsV2`

packages/models/src/bns.model.ts
```diff
@@ -10,6 +10,8 @@ export interface BnsName {
   name: string;
   namespace: string;
   fullName: string;
+  renewalHeight: string;
```
alexp3y: Should `renewalHeight` be a `number`?

packages/models/src/bns.model.ts
```diff
@@ -10,6 +10,8 @@ export interface BnsName {
...
+  registeredAt: string;
```
alexp3y: Maybe`registeredAt` could be either a `Date` or a `number`/  timestamp and converted when mapping from the API value? That would make it more widely useful.

apps/mobile/src/shared/display-preference.ts
```diff
@@ -37,3 +42,31 @@ export function getAccountDisplayPreferencesKeyedByType(): Record<
     },
   };
 }
+export function getChainDisplayLabel(
+  chainOrPreference: 'bitcoin' | 'stacks' | AccountDisplayPreference
+): string {
+  const matchChain = match<'bitcoin' | 'stacks' | AccountDisplayPreference>();
+  return matchChain(chainOrPreference, {
```
tigranpetrossian, line 49: `match` strikes again. We should really remove this helper. This is the equivalent of the code below, except without two functions and a generic.

```ts
{
    bitcoin: t`Layer 1 • Bitcoin`,
    taproot: t`Layer 1 • Bitcoin`,
    'native-segwit': t`Layer 1 • Bitcoin`,
    stacks: t`Layer 2 • Stacks`,
    bns: t`Layer 2 • Stacks`,
  }[chainPreference]
```

tigranpetrossian (approved):
Thanks Pete. Happy to see this in the app.

packages/ui/src/components/collectibles/utils/is-bns.shared.ts
```diff
@@ -0,0 +1,3 @@
+export function isBns(name: string): boolean {
```
alexp3y: As a UI util that only has access to `name` this makes sense. But in cases where we have access to the full `Sip9Asset` object, a check that looks at the actual contract principal or assetId would be more robust.

alexp3y (approved):
Looks great 🤩

apps/mobile/src/features/token/components/token-details-card.tsx
```diff
@@ -1,17 +1,15 @@
...
+  title?: React.ReactNode;
 }
 export function TokenDetailsCard({ children, title }: TokenDetailsCardProps) {
   return (
     <Box backgroundColor="ink.background-primary" px="5" py="3">
-      {typeof title === 'string' ? (
+      {title && (
         <Text variant="label03" py="2">
-          {title}
+          {typeof title === 'string' ? title : <>{title}</>}
```
kyranjamie: This change looks odd to me, previously you could override, but now even when passing a ReactNode it goes within Text?

packages/ui/src/components/collectibles/utils/is-bns.shared.ts
kyranjamie: A slight copy change would break this fn

packages/models/src/bns.model.ts
pete-watters: `registeredAt` and `renewalHeight` are both block numbers so I have updated the model as such, thanks! 🙇

apps/mobile/src/features/token/components/token-details-card.tsx
pete-watters: Thanks 👍 this was a mistaken auto format, rolling back

packages/ui/src/components/collectibles/utils/is-bns.shared.ts
pete-watters: Thanks! This `name` is the SIP-9 collection name that we got back from the API so I deemed it safe originally. 

Now we are using `assetIdentifier` in the routes I have refactored to remove this function and to [assert if it's a BNS](14ce674b0f146581a81ac0b1617b7868879f1e15) based on the `assetName`

apps/mobile/src/shared/display-preference.ts
pete-watters: Thanks @tigranpetrossian 👍 I moved this code and copied it. 

I went ahead and got rid of `match` in [this PR](https://github.com/leather-io/mono/pull/1749) to get rid of it forever.

PR #1747, opened 2025-10-30
Some Swaps preparation work

tigranpetrossian:
This PR is mostly some early refactor/improvements preceding larger fee handling and form review PRs.

* Simplify swap amount validation
* Extract few things into hooks, place in correct folders
* Temporarily alias Stacks `availableBalance` to  `availableUnlockedBalance` to work around a [naming issue](https://trustmachines.slack.com/archives/C05LAP952E7/p1760601239815099), until resolved.
* Add DerivedAmounts type: ([comment from the past PR](https://github.com/leather-io/mono/pull/1696#discussion_r2432190455))


apps/mobile/src/features/swap/swap-state/validation/swap-validation.ts
```diff
@@ -82,7 +84,18 @@ export interface ValidationResult {
 
 function validateAmount(context: ValidationContext): BaseAmountIssue | undefined {
   const { state, derivedAmounts } = context;
-  const { baseAmount, baseSwapAsset, inputCurrencyMode } = state;
+  const { baseSwapAsset, baseAmount, inputCurrencyMode } = state;
+  const decimals = baseSwapAsset?.asset.decimals ?? Number.MAX_SAFE_INTEGER;
```
kyranjamie: Why huge number if asset has no decimals? Should null not assume 0?

apps/mobile/src/features/swap/swap-state/swap-state.types.ts
```diff
@@ -16,6 +16,36 @@ export type PresetPercentage = 0.25 | 0.5 | 0.75 | 1;
 
 export type SwapQuoteStrategy = 'best' | 'fastest' | 'cheapest';
 
+export interface DerivedAmounts {
+  crypto: Money | null;
```
kyranjamie, line 20: why crypto and quote and not more generic `base`? Stablecoins etc blur the gap a little bit re what consists "crypto".

apps/mobile/src/features/swap/swap-state/validation/swap-validation.ts
tigranpetrossian: This is a badly written handling of a case when base asset isn't set. The state doesn't prevent you from typing in an amount without selecting a base asset (although in reality asset is pretty much always pre-populated), and auto-corrects the decimal places when you pick an asset.

This should be written differently to communicate intent better:
```ts
const decimals = baseSwapAsset 
  ? baseSwapAsset.asset.decimals 
  : someDefaultArbitraryReasonablyLargeNumber
```

Will update.

apps/mobile/src/features/swap/swap-state/swap-state.types.ts
tigranpetrossian: True. We already have a bit of a naming conflict here which is likely going to cause lots of confusion. I'm planning to bring this up and try revisiting, but it will affect other parts of the app.

The gist of it is `quote` app-wide is what used to be `fiat`, because the denominated currency is no longer Fiat, we use BTC as a display currency option. That already messes with Swap 
"quote".
On top of that in Swap "base" is the payable in general: base/target asset, base/target amount.

Very much open to suggestions on how to solve both conundrums at once.

pete-watters (approved):
Nice work 👍

apps/mobile/src/features/swap/swap-state/validation/swap-validation.ts
tigranpetrossian: Updated. Rest assured, most amount field validation checks, except for balance-related, are just last resort safety gates. The logic itself has constraints, plus it's expected that the UI won't will gate illegal input.

PR #1753, opened 2025-10-31
Add network fees to swap state

tigranpetrossian:
Adds network fee support to the swap state.

The final result is a query constructed from the fee information coming from the service and user's selection. Supports fixed or tiered, and an optional custom fee, all of which can be configured per protocol.


apps/mobile/src/features/swap/swap-state/use-swap-state.ts
```diff
@@ -1,43 +1,72 @@
...
+} from './swap.queries';
 
 export interface UseSwapStateProps {
   accountRequest: { account: AccountAddresses };
   baseAsset?: SwappableFungibleCryptoAsset;
   targetAsset?: SwappableFungibleCryptoAsset;
-  marketDataService: MarketDataService;
-  swapService: SwapService;
   quoteCurrencyPreference: QuoteCurrency;
+  swapService: SwapService;
```
tigranpetrossian, line 46: @kyranjamie Per our past conversation, the core dependencies for this machine have grown quite a bit. I'm working on a standardized provider to be used in the extension and mobile as a better injection mechanism, will add it once the signing/broadcasting logic is in.

The current direct passing isn't yet terrible, but it will definitely get out of hand as the feature grows.

apps/mobile/src/features/swap/swap-state/strategies/execution-type/build-transaction/build-transaction/build-sbtc-bridge-transfer-tx.ts
```diff
@@ -0,0 +1,109 @@
...
+  network: NetworkConfiguration,
+  account: AccountAddresses,
+  payer: BitcoinNativeSegwitPayer
+): Promise<SbtcDeposit> {
+  const client = network.chain.bitcoin.mode === 'mainnet' ? clientMainnet : clientTestnet;
+
+  return buildSbtcDepositTx({
+    amountSats,
+    network: getSbtcNetworkConfig(network.chain.bitcoin.mode),
+    stacksAddress: account.stacks?.stxAddress ?? '', // TODO: ???
```
pete-watters: Do you need to do something here? `// TODO: ???` ? 🤔

pete-watters (approved):
Great work @tigranpetrossian 👏

apps/mobile/src/features/swap/swap-state/strategies/execution-type/build-transaction/build-transaction/build-sbtc-bridge-transfer-tx.ts
tigranpetrossian: Removed, should be fine as is. TODO was mostly my frustration with `stxAddress` being nullable. Kyran's work will hopefully tighten up those types.

PR #1756, opened 2025-10-31
feat: add thumbnails for videos

pete-watters:
This PR adds code to auto generate collectible video thumbnails based on the video content. 

It plays the video then takes a screenshot of it to use as a thumbnail in the collectibles list. 

This gives a more interactive feel and improves the display on the collectibles list page

**Demo**

https://github.com/user-attachments/assets/82e705e1-cf74-479d-a4df-2f741a1e47f6

### DIFF
--- a/packages/ui/src/components/collectibles/native/collectible-video.native.tsx
+++ b/packages/ui/src/components/collectibles/native/collectible-video.native.tsx
@@ -1,5 +1,7 @@
-import { useRef } from 'react';
-import { WebView } from 'react-native-webview';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import { WebView, type WebViewMessageEvent } from 'react-native-webview';
+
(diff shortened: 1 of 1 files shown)


alexp3y (approved):
Awesome. I did notice quite a bit of latency on first load in your demo. Is that due to this? Can we pre-load this somehow so the thumbnail creation logic executes asynchronously?

pete-watters:
> Awesome. I did notice quite a bit of latency on first load in your demo. Is that due to this? Can we pre-load this somehow so the thumbnail creation logic executes asynchronously?

I can take a look and see if pre-load is possible, thanks! We can give it a try of doing it on the homescreen. 

I think the way it works now is as we are using `FlashList` for the elements, it only lazy loads them when viewing collectibles. The first time you view a video one, it plays the video in the background to then take a screenshot of it. I will see what I can do to make it faster

PR #1762, opened 2025-11-03
feat(web): scaffold portfolio page

kyranjamie:
Scaffolds the portfolio page


pete-watters:
This code LGTM but I'm not sure if the feature is production ready yet / if we want to release it right now. 

I've added [this PR](https://github.com/leather-io/mono/pull/1826) to feature flag it so we can safely merge and get the code improvements and test it out on the live site. This will also mean we can take in the improvements to `activity` etc. and use them in the `extension`

pete-watters (approved):
This code LGTM as a first iteration gated behind a feature flag. It also includes other code refactors useful elsewhere. 

@kyranjamie what do you think about merging this? I've gated it behind a feature flag in [this PR](https://github.com/leather-io/mono/pull/1826) which also protects access to the route

kyranjamie:
I think we're good to merge this @pete-watters

PR #1767, opened 2025-11-04
chore: update use-enroll-transaction.ts

314159265359879:
updated the contract for enrollment for sBTC rewards.


314159265359879:
tested with ledger, I can sign the enroll transaction without issue.

pete-watters (approved):
Thanks @314159265359879

PR #1773, opened 2025-11-04
feat: defi integrations

alexp3y:
Service integrations with various Stacks DeFi protocols providing user position information.

- Bitflow
- Granite V1
- Stacking DAO
- Zest Borrow

PR includes demo hooks in `/apps/mobile/src/queries/defi` illustrating potential usage.


apps/mobile/src/queries/defi/granite.query.ts
```diff
@@ -0,0 +1,21 @@
...
+import { balanceQueryOptions } from '../balance/balance-query-options';
+
+export function useGranitePosition(address: string) {
+  return toFetchState(useGranitePositionQuery(address));
+}
+
+function useGranitePositionQuery(address: string) {
+  const { fiatCurrencyPreference } = useSettings();
+  return useQuery({
+    queryKey: ['granite-service-get-zest-position', address, fiatCurrencyPreference],
```
kyranjamie: why zest in namee?

packages/services/src/defi/zest/zest-borrow.utils.ts
```diff
@@ -0,0 +1,114 @@
...
+    throw new Error(cvToString(response.value));
+  }
+
+  throw new Error(
+    `Unexpected Zest get-user-borrow-balance Clarity Value type: ${getCVTypeString(response)}`
+  );
+}
+
+export function parseZestGetZTokenBalanceResponseCV(response: ClarityValue): bigint {
+  //
```
pete-watters: Rogue `//` here

packages/services/src/defi/zest/zest-borrow.utils.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (response.type === ClarityType.Tuple) {
+    const borrowed = response.value['assets-borrowed'];
+    const supplied = response.value['assets-supplied'];
+
+    if (borrowed.type !== ClarityType.List || supplied.type !== ClarityType.List) {
+      throw new Error('Unexpected Zest get-user-assets-read response: expected List values');
+    }
+
+    return {
+      borrowedAssetPrincipals: borrowed.value.map(cv => (cv as any).value as string),
```
pete-watters: Can we avoid this casting or if not add some mocks + tests?

packages/services/src/defi/stacking-dao/stacking-dao.service.ts
```diff
@@ -0,0 +1,250 @@
...
+      this.hiroStacksApiClient.getAddressFtBalances(address, {
+        signal,
+      }),
+      this.hiroStacksApiClient.getNftHoldings(address, {
+        signal,
+      }),
+      this.getSbtcReward(address, signal),
+    ]);
+
+    if (
```
pete-watters: This is a monster `if(`, maybe the clause can be extracted to const's

apps/mobile/src/queries/defi/bitflow.query.ts
```diff
@@ -0,0 +1,21 @@
+import { toFetchState } from '@/components/loading';
```
pete-watters: Thanks for adding these examples 👍   maybe they should go into the Web App though as we aren't using these on mobile and I'm not sure if we will have portfolio / positions there? (We should probably show these somewhere on mobile) 

I wonder should we move back towards having more queries shared from `packages/query` that accept network / currency as params? That could help us create a more unified presentation layer for the UI that can do more data manipulation in a shared way

pete-watters (approved):
Great work @alexp3y ! It's exciting to see this all come together

packages/services/src/defi/stacking-dao/stacking-dao.service.ts
alexp3y: This was a bit overprotective. Realized this isn't actually needed and removed it.

packages/services/src/defi/zest/zest-borrow.utils.ts
alexp3y: Added some shared Clarity type guards / utils that prevent this and make this code a lot more readable.

apps/mobile/src/queries/defi/bitflow.query.ts
alexp3y: Including in `mobile` was just aspirational :D Agreed - I will remove them.

Having centrally shared queries like you mentioned makes sense, but I think we should discuss the exact packaging strategy... IMO the existing `@leather.io/query` is better considered a legacy package that we attempt to wind-down (i.e. no new additions or usages if possible). There's so much (largely obsolete) business logic intertwined there that perhaps a net-new package consisting of pure query wrappers is the cleanest option moving forward.

pete-watters: Sounds good! I agree **@leather.io/query** got out of hand. What would be useful is a presentation intermediary layer that we can use to help us format service data centrally. 

It could be even data mapping utils that we use in client queries

PR #1780, opened 2025-11-05
fix: responsive portfolio

edgarkhanzadian:


pete-watters:
This works in making it responsive but it's adding a weird multi level background to the elements

https://github.com/user-attachments/assets/17c6f416-a532-4b95-b1ea-da50e7e4517e

apps/web/app/pages/portfolio/portfolio-table/portfolio-table.tsx
```diff
@@ -240,6 +240,16 @@ export function PortfolioTable({ assets, isLoading }: PortfolioTableProps) {
           </Table.Body>
         </Table.Table>
       </Table.Root>
+      <styled.div
```
pete-watters, line 243: Do we need this?

pete-watters (approved):
Happy to merge this to make it responsive then further polish later as we finish it off

apps/web/app/pages/portfolio/portfolio-table/portfolio-table.tsx
edgarkhanzadian: yeap that just gives a little overlay that at the bottom of the scrollview

edgarkhanzadian:
> This works in making it responsive but it's adding a weird multi level background to the elements

Hm i can't reproduce it locally in arc or chrome.

PR #1782, opened 2025-11-10
Fee Service updates

tigranpetrossian:
@alexp3y 
Need your feedback. I've made two related changes to the fee service to make the usage a little more straightforward:

1. Split `feeRate` type into Stacks and Bitcoin with specific metadata for each. This way the consuming side can just directly refer to calculation units without having to narrow it down.
2. Added `inputs` and `outputs` to `BitcoinTransactionFeeQuote`. I can't remember where we landed in our discussions after you initially suggested doing this, but having these seems to make a lot of sense on the usage site -- both spares the flow a UTXO round trip, and having to recalculate what's already technically in the memory.

### DIFF
--- a/packages/models/src/fees/transaction-fees.model.ts
+++ b/packages/models/src/fees/transaction-fees.model.ts
@@ -1,15 +1,26 @@
 import { Money } from '../money.model';
 import { Blockchain } from '../types';
+import { OwnedUtxo } from '../utxo.model';
 
 export const transactionFeeTiers = ['low', 'standard', 'high'] as const;
 export type TransactionFeeTier = (typeof transactionFeeTiers)[number];
+
+export interface CoinSelectionOutput {
+  value: bigint;
+  address?: string;
+}
 
 export interface TransactionFees {
   readonly chain: Blockchain;
   readonly options: Record<TransactionFeeTier, TransactionFeeQuote>;
 }
 
-export const transactionFeeQuoteType = ['flat', 'feeRate', 'evm1559'] as const;
+export const transactionFeeQuoteType = [
+  'flat',
+  'bitcoinFeeRate',
+  'stacksFeeRate',
+  'evm1559',
+] as const;
 export type TransactionFeeQuoteType = (typeof transactionFeeQuoteType)[number];
 
 export interface BaseTransactionFeeQuote {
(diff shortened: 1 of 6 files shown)


packages/models/src/fees/transaction-fees.model.ts
alexp3y, line 1: Couple of ideas here - 
1)  Once we introduce chain-specific `FeeQuote` types, we could implement a top-level union that enforces the correct / matching quote type by chain. Each `TransactionFees` type could then be returned by its specific service (e.g. `getBitcoinTransactionFees(): Promise<BitcoinTransactionfees>`

2) Perhaps we add an intermediate type for shared `FeeRate` fields and extend from that for the specific Stacks / Bitcoin quote implementations (and perhaps others in the future)?

```typescript

export interface BaseTransactionFees {
  readonly chain: Blockchain;
  readonly options: Record<TransactionFeeTier, TransactionFeeQuote>;
}

export interface StacksTransactionFees extends BaseTransactionFees {
  readonly chain: 'stacks';
  readonly options: Record<TransactionFeeTier, StacksFeeRateTransactionFeeQuote>;
}

export interface BitcoinTransactionFees extends BaseTransactionFees {
  readonly chain: 'bitcoin';
  readonly options: Record<TransactionFeeTier, BitcoinFeeRateTransactionFeeQuote>;
}

export interface EthereumTransactionFees extends BaseTransactionFees {
  readonly chain: 'ethereum';
  readonly options: Record<TransactionFeeTier, EvmTransactionFeeQuote>;
}

export type TransactionFees =
  | StacksTransactionFees
  | BitcoinTransactionFees
  | EthereumTransactionFees;

....

export interface BaseFeeRateTransactionFeeQuote extends BaseTransactionFeeQuote {
  readonly type: 'bitcoinFeeRate' | 'stacksFeeRate';
  readonly rate: number;
  readonly rateUnit: string;
  readonly estimatedTxSize: number;
  readonly sizeUnit: string;
}

export interface BitcoinFeeRateTransactionFeeQuote extends BaseFeeRateTransactionFeeQuote {
  readonly type: 'bitcoinFeeRate';
  readonly rateUnit: 'sats/vB';
  readonly sizeUnit: 'vB';
}

export interface StacksFeeRateTransactionFeeQuote extends BaseFeeRateTransactionFeeQuote {
  readonly type: 'stacksFeeRate';
  readonly rateUnit: 'µSTX/byte';
  readonly sizeUnit: 'byte';
}

export type TransactionFeeQuote =
  | FlatTransactionFeeQuote
  | BitcoinFeeRateTransactionFeeQuote
  | StacksFeeRateTransactionFeeQuote
  | EvmTransactionFeeQuote;
```

packages/models/src/fees/transaction-fees.model.ts
```diff
@@ -21,12 +32,22 @@ export interface FlatTransactionFeeQuote extends BaseTransactionFeeQuote {
...
-export interface FeeRateTransactionFeeQuote extends BaseTransactionFeeQuote {
-  readonly type: 'feeRate';
+export interface BitcoinTransactionFeeQuote extends BaseTransactionFeeQuote {
+  readonly type: 'bitcoinFeeRate';
+  readonly rate: number;
+  readonly rateUnit: 'sats/vB';
+  readonly estimatedTxSize: number;
+  readonly sizeUnit: 'vB';
+  readonly inputs: OwnedUtxo[];
+  readonly outputs: CoinSelectionOutput[];
```
alexp3y, line 42: I know this gives us coin-selection results for free, but now seeing this implemented and I think it does overload / expose implementation details on the Fees model too much.

The overhead of calling coin-selection twice (once for the fee estimate, again for tx construction) is minimal enough to be worth maintaining a clean separation, both on the model and in the execution flow, where I think coin-selection should be called pretty explicitly. WDYT?

alexp3y (commented):
Nice idea on the chain-specific quote types - Added a couple of suggestions to build on that. Not so crazy about adding the coin-selection results in though, some thoughts there.

packages/models/src/fees/transaction-fees.model.ts
tigranpetrossian: Thanks Alex. True, this felt like a crutch when adding, and feels even more so that now. I think I went this it part to avoid introducing UTXO knowledge into the flow, part thinking the coin selection not being exported from `services` is intentional (I assume it's not?).

Going to close this, because the PR is unmergeable regardless (botched commit order), and address both your comments in the next one.

alexp3y: Right, I didn't omit the coin-selection service from exports intentionally - Feel free to add it. In general, all services can be considered public.

PR #1783, opened 2025-11-10
refactor(mobile): refactor stacks broadcasting hook

tigranpetrossian:
Two changes to the existing stacks broadcasting hook on mobile:

1. Rename: `useBroadcastStxTransaction` → `useBroadcastStacksTransaction`
2. Extract the inner implementation into a separate `broadcastStacksTransaction` function for reuse.


tigranpetrossian:
Moved to #1784

PR #1784, opened 2025-11-10
Refactor Swaps global dependencies

tigranpetrossian:
This attempts to normalize Swaps global dependencies into a single organized `SwapDependencies` bag that `extension` and `mobile` can construct and inject, accounting for differences in implementations in both.

Also sneaking in few other small refactoring commits:
* Use single `AssetAvatar` component throughout the mobile Swap
* Refactor amount presets and increase tap target sizes.
* Rename: `useBroadcastStxTransaction` → `useBroadcastStacksTransaction`, extract the inner queryFn `broadcastStacksTransaction` function for reuse.


apps/mobile/src/features/swap/components/asset-avatar.tsx
```diff
@@ -0,0 +1,35 @@
+import {
```
pete-watters, line 1: Not now or for this PR but it could be good to share this kind of logic somewhere ( I don't know where yet). 

Mentioning as @camerow hit this issue implementing icons on Web portfolio. We often repeat the same logic in different places for different things.

Open discussion on how we can share things better cross platform. The `ui` package is getting a bit bloated. Maybe we need a feature specific package(s) / a presentation layer.

apps/mobile/src/features/swap/use-swap-dependencies.ts
```diff
@@ -0,0 +1,73 @@
...
+      bitcoinPayer,
+      signBitcoinPsbt: () => {
+        throw new Error('unimplemented');
+      },
+      broadcast: broadcastBitcoinTransaction,
+    },
+  };
+}
+
+// Forked from apps/mobile/src/queries/transaction/use-bitcoin-broadcast-transaction.ts
```
pete-watters, line 57: Should we re-write `transaction/use-bitcoin-broadcast-transaction.ts` to support this instead? (I'm confused here) 

Are we concerned with UTXO's here?

apps/mobile/src/features/swap/components/asset-avatar.tsx
tigranpetrossian: I was going to comment on exactly this in the PR and it slipped my mind. I definitely intend to make single global `AssetAvatar` out of this, but avoiding doing it right now because refactoring all the usage sites is time consuming.
Not yet sure where it will live either, maybe some sort of a package housing all domain-aware assets.

pete-watters (approved):
Great work @tigranpetrossian 👍  LGTM, just added some discussion points

apps/mobile/src/features/swap/components/asset-avatar.tsx
pete-watters: Sounds good! I mentioned it as we discussed it briefly but agreed you would have some great ideas here. Lets think about it in the back of our minds

apps/mobile/src/features/swap/use-swap-dependencies.ts
tigranpetrossian: Possibly, but not urgent.`use-bitcoin-broadcast-transaction.ts` has lots of baked in opinions that aren't always relevant. The entire signing/broadcasting story is actually in desperate need of normalization across apps because implementations are extremely different.

This particular broadcaster isn't even going to be used in Swaps at the moment to be honest, we only support BTC -> sBTC, which is broadcast via helper from the `sbtc` package.
I might intentionally replace this with a placeholder in the next PR, at least for now.

pete-watters: OK great, thanks for explaining 👍  Agreed we need to factor in normalisation and sharing of this logic

PR #1786, opened 2025-11-10
fix: use leather IPFS for collectibles, ref LEA-3258

pete-watters:
This PR updates the collectibles service so that we use our own Leather IPFS instance to serve IPFS hosted collectibles

### DIFF
--- a/packages/services/src/assets/sip9-asset.utils.ts
+++ b/packages/services/src/assets/sip9-asset.utils.ts
@@ -22,7 +22,7 @@ import {
   getContractPrincipalFromAssetIdentifier,
 } from './stacks-asset.utils';
 
-type GammaNftMetadata = z.infer<typeof gammaNftMetadataSchema>;
+export type GammaNftMetadata = z.infer<typeof gammaNftMetadataSchema>;
 
 function optionalize<A, R>(fn: (a: A) => R) {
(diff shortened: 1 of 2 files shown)


packages/services/src/assets/sip9-asset.utils.ts
```diff
@@ -59,6 +59,29 @@ export function getNonFungibleTokenId(hex: string): number {
   return clarityValue.type === 'uint' ? Number(clarityValue.value) : 0;
 }
 
+const leatherIpfsUrl = 'https://leather.quicknode-ipfs.com/ipfs/';
+
+function getLeatherIpfsUrl(url: string): string {
```
tigranpetrossian: Mostly nitpicks:
* something like `toLeatherIpfsUrl` or `rewriteToLeatherIpfsUrl` would be clearer, since this is technically a transformation.
* Comments are redundant, the code is pretty self-explanatory
* early return: `if (!url.includes('/ipfs/')) return url;`

tigranpetrossian (approved):

packages/services/src/assets/sip9-asset.utils.ts
pete-watters: Thanks @tigranpetrossian, I'll update

PR #1787, opened 2025-11-10
Add execution to swap state

tigranpetrossian:
Adds swap execution for both chains to the swap state machine.
Also splits the monstrously large test file into few parts, as it grew to a point where github was refusing to display it.


tigranpetrossian:
@alexp3y The commit from #1782 is now in this PR: https://github.com/leather-io/mono/pull/1787/commits/64ef3ec22ee3b785f0c3faca6b675124e0d323ed -- removed inputs & outputs from the fee response, using the coin selection service directly is much more intuitive. 
I didn't touch the fee model mostly due to time constraints, but on board with your suggestions. Let's catch up some time this week, I might have few more longer-term ideas there too.

apps/mobile/src/features/swap/swap-state/strategies/execution-type/transform-transaction-fees.ts
```diff
@@ -22,19 +17,18 @@ export function transformTransactionFees(
...
   const options = buildOptions(transactionFees);
   const selected = buildSelection(feeCapabilities, tier, customFee);
-  const value = calculateSelectedFeeValue(transactionFees, selected);
 
   return {
     mode: 'tiered',
     customFeeEnabled: feeCapabilities.customFeeEnabled,
     options,
-    value,
+    calculation: calculateSelectedFeeCalculation(transactionFees, selected),
```
tigranpetrossian: excellent naming

apps/mobile/src/features/swap/swap-state/strategies/execution-type/execution-type.ts
```diff
@@ -58,6 +61,19 @@ const stacksContractCallStrategy: ExecutionStrategy = {
...
+
+    const unsignedTx = await buildStacksTx(
+      executionData as StacksContractCallSwapExecutionData,
+      stacks.stacksNetwork,
+      stacks.stacksSigner,
+      fee.calculation.value,
+      nonce
+    );
+    const signed = await stacks.stacksSigner.sign(unsignedTx);
+    await stacks.broadcast({ tx: signed, stacksNetwork: stacks.stacksNetwork });
```
kyranjamie, line 75: We don't trigger approver here for the signing flow?

apps/mobile/src/features/swap/swap-state/strategies/execution-type/execution-type.ts
```diff
@@ -97,6 +114,68 @@ const sbtcBridgeTransferStrategy: ExecutionStrategy = {
...
+          bitcoin.bitcoinPayer.address,
+          BigInt(output.value),
+          networkMode
+        );
+      }
+    });
+
+    const signedDepositTx = await bitcoin.signBitcoinPsbt(deposit.transaction.toPSBT());
+    signedDepositTx.finalize();
+    await bitcoin.sbtcClient.broadcastTx(signedDepositTx);
```
kyranjamie, line 173: Ideally signing logic would be encapsulated, e.g. if we later add hardware or mcp signing later

apps/mobile/src/features/swap/swap-state/strategies/execution-type/transform-transaction-fees.ts
kyranjamie: calculate the calculation? infer? derive?

kyranjamie (approved):

apps/mobile/src/features/swap/swap-state/strategies/execution-type/transform-transaction-fees.ts
tigranpetrossian: image
More than one person dislikes this apparently, will revise.

apps/mobile/src/features/swap/swap-state/strategies/execution-type/execution-type.ts
tigranpetrossian: This was one of the options initially, but I think we had a lengthy discussion during the planning and decided against for multiple reasons:
1. Look & feel - Approver looks great in isolation for external transactions, but quite bulky for internal flows. Designing cohesive flows (a la Send in the extension) with shared transitions, and things like progressive disclosure are very challenging with the Approver.
2. Mobile approver flows in particular are quite inflexible for sharing live form state from the core flow. Mobile Send is a good example for this: a complete disconnect in terms of two-way sync, e.g. edit fees - go back to form, updates aren't reflected. This can be changed, though it's a considerable effort only to circle back to 1.

I think Approver is an easy shortcut, but for internal flows we should just have well-designed minimal review pages that users can glance over and confirm, or dig into advanced details via progressive disclosure as needed.

apps/mobile/src/features/swap/swap-state/strategies/execution-type/execution-type.ts
tigranpetrossian: I haven't looked closely at hardware signing yet, so I don't have a clear picture of how to best represent it. Will definitely revise once working on the extension version.

PR #1788, opened 2025-11-11
feat(ui): add useInterval and useCountdown hooks

tigranpetrossian:
Adds two general-purpose hooks to be used in Swaps UI for live estimates (upcoming PR). Also replaces the simpler `use-interval` in the extension.

### useInterval

Runs a callback (sync or async) at an interval, with some customization options. Returns lifecycle info:

```ts
// Simple
useInterval(() => console.log(...), 15000);

// Metadata
const { status, lastStartedAt, lastCompletedAt, nextRunTime  } = useInterval(fooAsync, 15000, { enabled: false });
```

### useCountdown
```ts
const remainingSeconds = useCountdown(15000)
console.log(remainingSeconds); // 15, 14, 13, ...
```

#### Optionally combine the two

```ts
function poll() {
  const interval = 15000;
  const { nextRunTime  } = useInterval(fooAsync, interval);
  const remainingSeconds = useCountdown(nextRunTime)
}
```


pete-watters (approved):
Nice work 👍 This LGTM but should it be in the `ui` package or in the `utils` package? 

I think we are eager to deprecate `utils` if I understand so if so this makes sense here

kyranjamie (approved):

kyranjamie:
All non-ui packages should be unaware of react imo, so think UI is the correct choice for this package

tigranpetrossian:
I'm occasionally contemplating making a `react` package, but then it's not like `useInterval` is React-specific either, just happens to be written in it.

PR #1789, opened 2025-11-11
feat: collectible type logo overlay

pete-watters:
This PR adds a collectible type logo on top of collectibles in the gallery view to indicate chain

Demo:

https://github.com/user-attachments/assets/e108c919-5f40-440e-ac75-ff95e7eea05c

Gamma seems to be having API issues right now so Stacks NFTs are not loading


packages/ui/src/components/collectibles/native/collectible-type-icon-overlay.native.tsx
```diff
@@ -0,0 +1,51 @@
...
+    default:
+      return null;
+  }
+}
+
+interface CollectibleTypeIconOverlayProps extends PropsWithChildren {
+  protocol: NonFungibleCryptoAsset['protocol'];
+}
+
+export function CollectibleTypeIconOverlay({
```
tigranpetrossian: Is there a reason why this component is in the `ui` and not the app? UI should stay domain-agnostic (this rule is already broken in few places, something we're planning to eliminate). 
Dependency on domain-specific code/packages like `models` is extremely problematic.

alexp3y (approved):
Looks good 👍

PR #1791, opened 2025-11-11
fix: protect against failing stampchain call

alexp3y:
The stampchain.io API is failing consistently with 500 errors that are received after a considerable (~3s delay). 

We don't cache the failure responses, so the request is repeated and latency experienced on each re-render.

This PR protects against this by wrapping the request in a try/catch and returning an empty array on failure. The relatively low cache time means the request will still be retried (in 30s), but not on re-renders.


pete-watters (approved):
Thanks @alexp3y 👍  I noticed this API being a bit flaky so hopefully this helps

kyranjamie (approved):

PR #1795, opened 2025-11-12
feat(web): add empty state

kyranjamie:
Adds empty state view to the portfolio


edgarkhanzadian (approved):

tigranpetrossian (approved):

apps/web/app/pages/portfolio/components/activity-list/index.tsx
```diff
@@ -16,19 +16,28 @@ export function ActivityList({ activity, isLoading }: ActivityListProps) {
...
-        >
+        <Flex p="space.06" textAlign="center" fontSize="24px" justifyContent="center" flexGrow={1}>
           <LoadingSpinner />
         </Flex>
       </Stack>
     );
   }
 
+  if (activity.length === 0) {
+    return (
```
pete-watters, line 27: We have a component for this in mobile that shows an image of a hammock - `ActivityEmpty`, it would be good to use a similar setup for that / ask for an empty design

apps/web/app/pages/portfolio/portfolio.page.tsx
```diff
@@ -52,11 +72,28 @@ export function PortfolioPage() {
...
+    );
+  }
+
+  const isLoading =
+    isConnected && (btcQuery.isLoading || sip10Query.isLoading || stxQuery.isLoading);
 
   return (
     <PortfolioPageLayout
-      overview={<PortfolioSummary />}
+      overview={<PortfolioSummary balance={isConnected ? totalBalance : dummyTotalBalance} />}
```
pete-watters: I don't think you need this check for `isConnected` as you are returning early above. 

Also you already have a const for `isConnected` so you can use that instead of`status !== 'connected'` on line 75?

apps/web/app/pages/portfolio/portfolio.page.tsx
```diff
@@ -21,12 +25,28 @@ function sortAssetsByValue(a: PortfolioAsset, b: PortfolioAsset) {
   return a.asset.symbol.localeCompare(b.asset.symbol);
 }
 
+export function PortfolioPageSkeleton() {
```
pete-watters, line 27: I guess we want to advertise the portfolio with mock data if the user hasn't connected yet? 

In that case I'd call this `MockPortfolio` or something like that as `Skeleton` is used elsewhere for loading states? 

Maybe we should call `dummy` `mock` too?

apps/web/app/pages/portfolio/portfolio.page.tsx
kyranjamie: thanks pete yes can remove

apps/web/app/pages/portfolio/portfolio.page.tsx
kyranjamie: This component really is just the skeleton, the dummy data uses the same components just with fake data in the feature's root

apps/web/public/images/extension-logo.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
pete-watters, line 1: We have this SVG in `packages/ui` - `leather-lettermark` . 

In future, it would be good to move the web app images over to the UI package and eliminate redundancy

kyranjamie: Noted, I think these are different images, but do agree we should try and colocate

PR #1797, opened 2025-11-12
feat: activity items better labeling and swap icons

camerow:


tigranpetrossian (approved):
Happy to see the swap type handled here. 

Debt is creeping up in this component. I have a stashed away version of this that refactors this using a simple per-activity type view model, but it still needs some work.
Will pick it up again once I have some room to breathe.

apps/web/app/pages/portfolio/components/activity-list/utils.ts
```diff
@@ -104,13 +118,25 @@ export function getActivityTitle(activity: OnChainActivity) {
     case 'swapAssets':
       return `Swap Assets`;
     case 'lockAsset':
-      return `Lock Asse`;
```
pete-watters, line 107: Haa nice typo fix 👍

apps/web/app/pages/portfolio/components/activity-list/utils.ts
```diff
@@ -104,13 +118,25 @@ export function getActivityTitle(activity: OnChainActivity) {
...
+      return `Lock Asset`;
     default:
       return `Unknown`;
   }
 }
 
 export function getBalancesText(activity: OnChainActivity) {
+  if (activity.type === 'swapAssets') {
```
pete-watters, line 128: Maybe you can just pass in `crypto` and `quote` instead of `activity` to clean it up?

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -15,45 +25,113 @@ interface StatusIndicatorProps {
...
+      if (type === 'executeSmartContract') {
+        return <FunctionActivityIcon width={16} height={16} />;
+      }
+      if (type === 'swapAssets') {
+        return <SwapIcon width={16} height={16} />;
+      }
+      if (type === 'receiveAsset') {
         return <ReceivedIcon width={16} height={16} />;
       }
+      // show hidden icon
```
pete-watters, line 46: Whats this all about? Received with `0` dimensions?

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -15,45 +25,113 @@ interface StatusIndicatorProps {
...
       assertUnreachable(status);
   }
 }
 
+interface AssetAvatarProps extends AvatarProps {
+  asset: CryptoAsset;
+  indicator?: ReactElement;
+}
+
+function AssetAvatar({ asset, indicator, size, ...rest }: AssetAvatarProps): ReactElement {
```
pete-watters, line 60: Maybe this could be in it's own file - `asset-avatar` as it's probably useful elsewhere we are repeating this

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -15,45 +25,113 @@ interface StatusIndicatorProps {
...
+          contractId={asset.contractId}
+          imageCanonicalUri={asset.imageCanonicalUri}
+          indicator={indicator}
+          name={asset.name}
+          size={size}
+          {...rest}
+        />
+      );
+    default:
+      return <Sip10AvatarIcon contractId="" imageCanonicalUri="" name="" size={size} {...rest} />;
```
pete-watters, line 78: Should we default to `Sip10`? We should probably:
- cover other protocols here (runes, ordinals etc.) 
- assert unreachable here for default

For now we can just default to`<Avatar` if that's easier. 

I added similar code [here](https://github.com/leather-io/mono/pull/1789/files#diff-5a34182a8b8b35c77c61efe64a1e5690aa2c879cbf604f3c9730ae385b7b992b) for the collectibles icon overlay lately. 

Just flagging as theres room for consolidation here. It would be especially great if we can figure out a cool way of sharing the logic across platforms (one switch to return web/mobile component based on context)

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -15,45 +25,113 @@ interface StatusIndicatorProps {
...
+    );
+  }
+
+  const asset = 'asset' in activity ? activity?.asset : undefined;
+  if (asset) {
+    return <AssetAvatar asset={asset} indicator={statusIndicator} />;
   }
+
+  return (
+    <Sip10AvatarIcon
```
pete-watters, line 130: Lets make this a generic `Avatar` also? `Sip10AvatarIcon` with no URI is just a fallback avatar

pete-watters (approved):
Great stuff @camerow 👍  it's great to have you in the code.

I mentioned some discussion points for collaboration but this is good to go as is

pete-watters:
Nice work @camerow . I solved the type issues by solving issues with [duplicate helpers](gate the new portfolio behind a LD flag). We have more of these helpers I didn't refactor yet:
- in the `extension` - `makeStacksAddressExplorerLink`, `makeStacksTxExplorerLink` etc.
- in `web` - `makeExplorerTxLink`

@kyranjamie @tigranpetrossian : this is a good example of why we were adding the `features/` package. That would give us a place to centralise this shared logic. 

I see that we have `makeActivityLink` in `packages/models`. Is that the correct place for that? I thought `models` should be purely for types etc. If that's a better place to share this kind of logic then we could use that instead of a `features` package. We still end up working across multiple areas to do one feature though and also need a better way to share data manipulation

PR #1798, opened 2025-11-13
fix: psbt testnet

edgarkhanzadian:
Fix PSBT signing on testnet

https://github.com/user-attachments/assets/b44d04c8-54b4-428b-a9b2-b92f3987364f

### DIFF
--- a/apps/mobile/src/shared/utils.ts
+++ b/apps/mobile/src/shared/utils.ts
@@ -1,0 +1,64 @@
+import { useSettings } from '@/store/settings/settings';
+import {
+  getStacksNetworkFromName,
+  useNetworkPreferenceStacksNetwork,
+} from '@/store/settings/settings.read';
+import { StacksNetwork, StacksNetworks } from '@stacks/network';
+
+import { NetworkConfiguration, defaultNetworksKeyedById } from '@leather.io/models';
+
+interface GetBtcNetworkFromRequestParams {
+  paramsNetwork: string | undefined;
+  defaultNetwork: NetworkConfiguration;
+}
+
+function getBtcNetworkFromRequestParams({
+  paramsNetwork,
+  defaultNetwork,
+}: GetBtcNetworkFromRequestParams) {
+  if (paramsNetwork) {
+    // default to testnet4 if rpc request specifies "testnet"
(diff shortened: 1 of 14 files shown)


kyranjamie:
I notice in the approver ux we're capitalising the bitcoin address. We shouldn't do this.

apps/mobile/src/shared/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (paramsNetwork === 'testnet') return defaultNetworksKeyedById.testnet4;
+    const foundNetwork = Object.values(defaultNetworksKeyedById).find(network => {
+      return network.chain.bitcoin.bitcoinNetwork === paramsNetwork;
+    });
+
+    if (foundNetwork) {
+      return foundNetwork;
+    }
+    throw new Error('Wrong network supplied in the rpc request params');
+  }
```
kyranjamie, line 30: Should this logic live somewhere more generic, imagine we'll need network selection logic globally

edgarkhanzadian: this does seems like a utility that we could share between applications, but i am unsure which package would suit best for this function

edgarkhanzadian: for now i can move it to apps/mobile/src/shared folder until we find a better place for it, wdyt?

apps/mobile/src/shared/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+    // default to testnet4 if rpc request specifies "testnet"
+    if (paramsNetwork === 'testnet') return defaultNetworksKeyedById.testnet4;
+    const foundNetwork = Object.values(defaultNetworksKeyedById).find(network => {
+      return network.chain.bitcoin.bitcoinNetwork === paramsNetwork;
+    });
+
+    if (foundNetwork) {
+      return foundNetwork;
+    }
+    throw new Error('Wrong network supplied in the rpc request params');
```
camerow: Possibly provide the value of `paramsNetwork` here for a slightly improved debugging experience - eg a typo could be easily scanned if provided.

camerow (commented):
Left a small comment but will leave this one up to @kyranjamie to ultimately approve.

kyranjamie (approved):

PR #1799, opened 2025-11-13
feat: add yield model types

alexp3y:
Adds an extensible Yield model and service for retrieving yield providers, products, and positions.

`YieldProvider` -> `YieldProduct` -> `YieldPosition`

Adjusts existing "DeFi" services to use and return these types.


alexp3y:
@kyranjamie I included some generated helpers in `yield.helpers.ts` that provide some type guards and other utils for working with the model. Not sure which / if any might be useful. We can get rid of any we don't need.

Haven't tested this extensively so should be considered WIP. But the types and data flow should all be in place.

The top-level `YieldService` has some utility methods for getting Providers / Products as well as Positions by category, provider, etc...

packages/models/src/yield/helpers/yield.helpers.ts
```diff
@@ -0,0 +1,308 @@
...
+export function filterPositionsByProtocol(
+  positions: YieldPosition[],
+  protocol: YieldProviderKey
+): YieldPosition[] {
+  return positions.filter(p => p.provider === protocol);
+}
+
+/**
+ * Filter positions by product
+ */
```
kyranjamie: Can we tone down some of these comments? In many places they don't add any details beyond the fn name

kyranjamie (approved):
Service and model structure LGTM, nice work @alexp3y

PR #1800, opened 2025-11-13
chore(web,mobile,extension,analytics): migrate to mixpanel analytics

camerow:
Remove segment and use mixpanel directly instead.


tigranpetrossian (approved):

apps/mobile/src/utils/analytics.ts
```diff
@@ -1,21 +1,46 @@
...
+  };
+}
+
+const mixpanelClient = process.env.EXPO_PUBLIC_MIXPANEL_TOKEN
+  ? new Mixpanel(process.env.EXPO_PUBLIC_MIXPANEL_TOKEN, false)
+  : getMockedMixpanel();
+
+void mixpanelClient.init().then(() => {
+  configureMixpanelInstance(mixpanelClient as Mixpanel);
+  mixpanelClient.registerSuperProperties({ platform: 'mobile' });
```
kyranjamie: Are these all the super properties? What about version, and other helpful user data?

apps/mobile/src/utils/analytics-plugins.ts
```diff
@@ -1,44 +1,17 @@
...
-      return {
-        ...event,
-        properties: { ...event.properties, platform: 'mobile', network: this.network },
-      };
-    }
-    return event;
-  }
+export function configureMixpanelInstance(instance: Mixpanel) {
+  mixpanelInstance = instance;
+}
```
kyranjamie: This is a little strange to me. Why are we using a fn to mutate a let in another file?

packages/analytics/src/types.ts
```diff
@@ -5,19 +5,25 @@ export interface DefaultProperties {
...
-  track(event: string, ...args: any[]): Promise<any>;
-  group(groupId: string, traits?: any, ...args: any[]): Promise<any>;
-  identify(...args: any[]): Promise<any>;
-  page?(category?: string, name?: string, ...args: any[]): Promise<any>;
-}
-
-export interface AnalyticsClientConfig<T extends AnalyticsClientInterface> {
-  client: T;
+export interface AnalyticsClientConfig {
+  client: {
```
kyranjamie: Why is there now a nesting for the `client` object?

apps/mobile/src/utils/analytics.ts
edgarkhanzadian: good point, let me see what i can add here

apps/mobile/src/utils/analytics-plugins.ts
edgarkhanzadian: good point, seems to be a leftover from development. removing

packages/analytics/src/types.ts
edgarkhanzadian: Sure, i'll move it to a separate interface. I mostly wanted to remove the generic type here as it wasn't that much helpful.

kyranjamie (approved):

PR #1802, opened 2025-11-13
feat: add LD to extension and web

edgarkhanzadian:
Add Launch Darkly to both extension and web.

example of flags in web react sdk:

As per @tigranpetrossian suggestion, it would be better to use:

```
import { useFlags as useLDFlags } from 'launchdarkly-react-client-sdk';

interface FeatureFlags {
    feature_flag: boolean
}

function useFlags() {
  return useLDFlags<FeatureFlags>()
}
```


tigranpetrossian (changes requested):
Thanks for doing this @edgarkhanzadian, it was long overdue.

There's some work to be done here, so I'm going to leave all my comments here:

**1. Initialization**
`withLDProvider` doesn't guarantee flags being ready when app is rendered, meaning users might briefly see disabled features flicker, or enabled features show up late ([docs](https://launchdarkly.com/docs/sdk/client-side/react/react-web#initialize-using-withldprovider)). I think this is a substandard experience. We should use the `asyncWithLDProvider` to guarantee flag correctness. 

**2. The usage:**
The nice things about `useFlags` on the React Web SDK is that we don't have to create hooks for each flag. Instead of doing:

```ts
export function useTestFlag() {
  const flags = useFlags<FeatureFlags>();
  return flags.test;
}
```

Define a custom global hook once with the generic baked in:
```ts
import { useFlags as useLDFlags } from 'launchdarkly-react-client-sdk';

function useFlags() {
  return useLDFlags<FeatureFlags>()
}
```

Usage:

```ts
const { tokenDetails } = useFlags() // free auto-complete, pseudo type-safety.
```

**3. Identification**
This is still an open question, but I'm _almost_ certain that if we don't explicitly identify context to LD, it will treat each session as brand new, meaning we won't be able to gradual rollouts/experiments/variations, and we'll get flawed analytics. 
I'm not 100% sure if LD doesn't automatically persist some sort of ID for anonymous sessions, but we need to be absolutely certain sessions are differentiated before adding this.

edgarkhanzadian:
@tigranpetrossian thank you for your feedback.

1. Didn't realize the withLDProvider wrapper let's app load without LD, will look into it. Good catch!

2. Yea i was unsure if we want to keep the same feature flag interface as on mobile or just use useFlags here. I'm good with both solutions and this PR doesn't include the feature flag usage in code anyway

3. I think LD does provide a persistent anonymous context, but will double check. We can also just generated uuid for each user

apps/extension/src/app/common/device-id/index.ts
```diff
@@ -0,0 +1,19 @@
+function generateDeviceId() {
+  return crypto.randomUUID();
+}
+
+function setDeviceId(id: string) {
```
tigranpetrossian: Worth future-proofing this a little bit:
1. Extract the key into  a constant and version it, e.g. `device-id-v1`. I also think device id as a name is a bit misleading, this isn't a device id per se and isn't deterministic.
2. All access to `localStorage` is unsafe and needs to be wrapped in try/catch. `localStorage` throws all the time: private mode, storage full, user-disabled, etc.

tigranpetrossian (approved):

PR #1806, opened 2025-11-14
fix: eslint warnings

edgarkhanzadian:


apps/extension/.eslintrc.cjs
```diff
@@ -68,8 +68,8 @@ module.exports = {
         message: avoidFetchMsg,
       },
     ],
-    '@typescript-eslint/no-floating-promises': ['warn'],
-    '@typescript-eslint/no-unnecessary-type-assertion': ['warn'],
+    '@typescript-eslint/no-floating-promises': ['error'],
```
pete-watters, line 71: Now that we have the extension in `mono` can we set it up to use `@leather.io/eslint-config` ? 

Then we can share the same config better and make these updates on all platforms

pete-watters (approved):

apps/extension/.eslintrc.cjs
edgarkhanzadian: for sure, we'll get there eventually. Removing these warnings would be a prerequisite for that work

PR #1807, opened 2025-11-14
fix(mobile): collectible reliability

pete-watters:
This PR addresses some bugs encountered when testing collectibles. 

- Adds better fallbacks when image sources are not found
- Fixes a bug going to stamps details
- Adds thumbnails properly for GLTF 
    - static on gallery page, interactive on details
- Uses `ord.io` to load some collectibles in the list view more reliably
    - `html` and `image` types sometimes don't render correctly on gallery view
    - sometimes they are so to load / sometimes the HTML content doesn't work

**Demo**

https://github.com/user-attachments/assets/89d24224-b11d-47af-acba-e8f719b803b0


packages/ui/src/components/collectibles/native/image-unavailable.native.tsx
```diff
@@ -1,22 +1,33 @@
...
 import { CollectibleCard } from './collectible-card.native';
 
-interface ImageUnavailableProps extends HasChildren {
+interface ImageUnavailableProps extends PropsWithChildren {
   height?: number;
+  message?: ReactNode;
 }
 
-export function ImageUnavailable({ children, height = 200 }: ImageUnavailableProps) {
+const defaultMessage = 'Image currently unavailable';
```
pete-watters, line 12: In the mobile app I pass in a translated value via `getImageUnavailableLabel` . This is just a fallback

packages/models/src/assets/asset.model.ts
```diff
@@ -120,6 +120,7 @@ export interface InscriptionAsset extends BaseNonFungibleCryptoAsset {
   readonly offset: string;
   readonly preview: string;
   readonly src: string;
+  readonly thumbnailSrc?: string;
```
pete-watters, line 123: @alexp3y I'm adding this field to distinguish between `preview` and `thumbnail`.

For HTML + image inscriptions I switched it to use `ord.io` to render images as it's more reliable and BIS wasn't as reliable

packages/ui/src/components/collectibles/native/image-unavailable.native.tsx
camerow: Why not wrap the fallback in a translation?

camerow (approved):

packages/ui/src/components/collectibles/native/image-unavailable.native.tsx
pete-watters: We don't have translations setup in the UI package and I'm not sure if we should? 

If we did it would simplify this implementation but also add increased scope here along with the issue of syncing translations. 

Maybe now we have `extension` in `mono` it's time to plan a translations package so we can share translations across all platforms? **OR** decide if we are even bothered translating at all

PR #1808, opened 2025-11-14
fix: Merge eslint configs

edgarkhanzadian:
1. Merge extension's eslint config into monorepo's one.
2. Use one single eslint command to lint them all


kyranjamie:
I noticed running eslint we're not using the flat file config. I was pretty sure @tigranpetrossian had migrated to this, yet now we still get the deprecation warning. 

```
(node:51274) ESLintRCWarning: You are using an eslintrc configuration file, which is deprecated and support will be removed in v10.0.0. Please migrate to an eslint.config.js file. See https://eslint.org/docs/latest/use/configure/migration-guide for details. An eslintrc configuration file is used because you have the ESLINT_USE_FLAT_CONFIG environment variable set to false. If you want to use an eslint.config.js file, remove the environment variable. If you want to find the location of the eslintrc configuration file, use the --debug flag.
```

Will we fix this in this PR?

edgarkhanzadian:
@kyranjamie yes, this one only uses flat config at the root of the monorepo

tigranpetrossian:
> I noticed running eslint we're not using the flat file config. I was pretty sure @tigranpetrossian had migrated to this, yet now we still get the deprecation warning. 
> 
> 
> 
> ```
> 
> (node:51274) ESLintRCWarning: You are using an eslintrc configuration file, which is deprecated and support will be removed in v10.0.0. Please migrate to an eslint.config.js file. See https://eslint.org/docs/latest/use/configure/migration-guide for details. An eslintrc configuration file is used because you have the ESLINT_USE_FLAT_CONFIG environment variable set to false. If you want to use an eslint.config.js file, remove the environment variable. If you want to find the location of the eslintrc configuration file, use the --debug flag.
> 
> ```
> 
> 
> 
> Will we fix this in this PR?

I think that's the intention. Extension eslint is still running off of its original config.

package.json
```diff
@@ -56,7 +56,7 @@
     "chalk": "5.3.0",
     "dependency-cruiser": "16.10.0",
     "dotenv": "16.4.5",
-    "eslint": "9.28",
+    "eslint": "9.34.0",
```
kyranjamie, line 59: why not latest version?

edgarkhanzadian: I did upgrade it to a higher version with a hope to fix an oom error (saw it somewhere in github discussions). It actually didn't help much but decided to keep it as a small version bump.

I can do an upgrade to the latest version in a separate PR

kyranjamie:
This is also 1minute slower than before 
2025-12-03-000570@2x

edgarkhanzadian:
@kyranjamie hmm it could be that i'm running more rules on apps/extension code using packages/eslint-config.
on `dev` we don't use monorepo's eslint config and extension has a separate lint rule list.

apps/extension/tests/mocks/mock-chrome.ts
```diff
@@ -1,18 +1,34 @@
...
   storage: {
     local: {
-      clear() {},
-      get() {},
-      getBytesInUse() {},
-      onChanged() {},
-      remove() {},
-      set() {},
+      clear() {
+        // Mock implementation
```
tigranpetrossian: I assume these are to silence the no empty function rule? Really a nitpick, but `clear: noop`, is slightly better than a comment here.

kyranjamie (approved):

pete-watters (approved):

tigranpetrossian (approved):

tigranpetrossian:
Unrelated to this PR, I find ` '@typescript-eslint/method-signature-style': ['error', 'method']` to be a very confusing rule given how we almost never use actual methods. @kyranjamie Can you remember the rationale behind this?

PR #1809, opened 2025-11-14
feat: enhance yield model

alexp3y:
This PR contains updates to the Yield model to favor more granular positions, supporting flexible UI designs.

Products are much more specific w/ multiple products per provider where possible - e.g. Stacking DAO `stSTX` and `stSTXbtc` are separate position objects.

Product services now return an array of yield positions instead of monolithic 1:1 position objects. 

Changes "Protocol" to "Provider" as the top-level entity to more naturally encompass non-protocol yield mechanism (e.g. native pooled stacking).


pete-watters (approved):
Great work 🚀

PR #1811, opened 2025-11-17
fix: client id uuid

edgarkhanzadian:
Addresses this Sentry issue https://github.com/leather-io/mono/pull/1802#issuecomment-3536397127


pete-watters (approved):

pete-watters:
Thanks @edgarkhanzadian 👍  I can see we also use it in the `provider` package and some mocks so maybe we can do a clean sweep and remove everywhere?

PR #1814, opened 2025-11-17
refactor(services): demo network as observable

kyranjamie:
Making this PR as a demo of conversation earlier with @alexp3y @edgarkhanzadian. 

This demos how a push-based approach could work. `network$` is a stream of events that updates. Consumers listen for changes, removing need to list deps array in useQuery to tell it to refetch.

This could equally be built with web standard `EventEmitter`s, but it lacks the tooling to easily compose together multiple streams, as comes with Rxjs (an implementation of a soon-to-be web standard)

### DIFF
--- a/apps/extension/src/app/query/stacks/balance/stx-balance.query.ts
+++ b/apps/extension/src/app/query/stacks/balance/stx-balance.query.ts
@@ -1,4 +1,7 @@
+import { useMemo, useSyncExternalStore } from 'react';
+
 import { type QueryFunctionContext, useQuery } from '@tanstack/react-query';
+import type { Observable } from 'rxjs';
 
 import { type AccountRequest, getStxBalancesService } from '@leather.io/services';
 
@@ -24,4 +27,29 @@ export function useGetStxAddressBalanceQuery(address: string) {
     enabled: !!address,
     ...balanceQueryOptions,
   });
+}
+
+export function useObservable<T>(observable: Observable<T>): T {
+  const store = useMemo(() => {
+    let currentValue: T;
+    const subscribe = (onChange: (value: any) => void) => {
+      const sub = observable.subscribe(value => {
+        currentValue = value;
+        onChange(value);
+      });
+      return () => sub.unsubscribe();
+    };
+    const getSnapshot = () => currentValue;
+    return { subscribe, getSnapshot };
+  }, [observable]);
+
+  return useSyncExternalStore(store.subscribe, store.getSnapshot);
+}
+
+export function useStxBalanceExperimentalStream(address: string) {
+  const stxBalance$ = useMemo(
+    () => getStxBalancesService().getStxAddressBalanceExperimentalStream(address),
+    [address]
+  );
+  return useObservable(stxBalance$);
 }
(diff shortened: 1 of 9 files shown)


apps/extension/src/app/services/extension-settings.service.ts
```diff
@@ -1,11 +1,18 @@
...
-import { store } from '@app/store';
+import { store, store$ } from '@app/store';
 import { selectTokenState } from '@app/store/manage-tokens/manage-tokens.slice';
 import { selectCurrentNetwork } from '@app/store/networks/networks.selectors';
 
 export class ExtensionSettingsService implements SettingsService {
+  network$ = store$.pipe(
+    map(state => selectCurrentNetwork(state)),
+    distinctUntilChanged()
+  );
```
kyranjamie, line 14: Network, which comes from Redux store, injected as `Observable` to service. No coupling to Redux, could come from any source.

packages/services/src/balances/stx-balances.service.ts
```diff
@@ -115,4 +116,44 @@ export class StxBalancesService {
...
+    address: string,
+    signal?: AbortSignal
+  ): Observable<AddressQuotedStxBalance> {
+    return combineLatest([
+      from(
+        Promise.all([
+          this.stacksTransactionsService.getPendingTransactions(address, signal),
+          this.marketDataService.getMarketData(stxAsset, signal),
+        ])
+      ),
```
kyranjamie: Wraps `Promise`s as `Observable`

packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -122,6 +124,41 @@ export class HiroStacksApiClient {
...
+        {
+          priority: hiroApiRequestsPriorityLevels.getAccountBalance,
+          signal,
+          throwOnTimeout: true,
+        }
+      );
+      return res.data;
+    };
+
+    return this.settings.network$.pipe(
```
kyranjamie, line 148: If `network$` emits change event, then refetch

apps/extension/src/app/query/stacks/balance/stx-balance.query.ts
```diff
@@ -25,3 +28,28 @@ export function useGetStxAddressBalanceQuery(address: string) {
...
+        onChange(value);
+      });
+      return () => sub.unsubscribe();
+    };
+    const getSnapshot = () => currentValue;
+    return { subscribe, getSnapshot };
+  }, [observable]);
+
+  return useSyncExternalStore(store.subscribe, store.getSnapshot);
+}
```
kyranjamie, line 47: To greatest extent possible, I'd want to avoid exposing the concept of `Observable`s to the react app, interfacing only with hooks. 

If services expose Observables, however, we would need some way to bind to them. Hence this hook.

This is the hard part imo, and where there is biggest scope for errors/bugs from stale closures etc.

pete-watters:
This looks like a good thing for us to try although complicated and there is a learning curve with RxJS. Perhaps it's something to come after finalisation of the `state` package? 

We are at a junction now where we need to delegate service data manipulation somewhere and using observables could help us with that goal, especially if it helps us move towards websockets. 

I just wonder is it a bit complex to do immediately now based on our goals. It could be a good idea to just do it and get it over with

kyranjamie:
Closing this PR as as we won't merge in this state, though concerns of the non-reactivity in the service layer remain

PR #1815, opened 2025-11-17
chore(mobile): refactor token destructure

pete-watters:
This PR:
- addresses some issues raised by @alexp3y  in [this PR](https://github.com/leather-io/mono/pull/1745)
- adds a fix to `mobile` where the **sBTC** avatar was not rendering properly

### DIFF
--- a/packages/models/src/assets/asset-id.model.ts
+++ b/packages/models/src/assets/asset-id.model.ts
@@ -3,5 +3,4 @@ import { CryptoAssetProtocol } from './asset.model';
 export interface CryptoAssetId {
   protocol: CryptoAssetProtocol;
   id: string;
-  tokenId?: number;
 }
--- a/packages/ui/src/components/avatar/sip10-avatar-icon.native.tsx
+++ b/packages/ui/src/components/avatar/sip10-avatar-icon.native.tsx
@@ -24,11 +24,13 @@ export function Sip10AvatarIcon({
 }: Sip10AvatarIconProps) {
   // TODO LEA-2551: use leather design system for more avatars
   if (name === 'sBTC') {
-    <Avatar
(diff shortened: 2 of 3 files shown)


packages/ui/src/components/avatar/sip10-avatar-icon.native.tsx
```diff
@@ -24,11 +24,13 @@ export function Sip10AvatarIcon({
 }: Sip10AvatarIconProps) {
   // TODO LEA-2551: use leather design system for more avatars
   if (name === 'sBTC') {
-    <Avatar
-      icon={<SbtcIcon width="100%" height="100%" />}
-      indicator={<StacksIcon width={16} height={16} />}
-      {...props}
-    />;
+    return (
```
pete-watters, line 27: I missed a `return` here so the Avatar was not showing on mobile at all

packages/models/src/assets/asset-id.model.ts
```diff
@@ -3,5 +3,4 @@ import { CryptoAssetProtocol } from './asset.model';
 export interface CryptoAssetId {
   protocol: CryptoAssetProtocol;
   id: string;
-  tokenId?: number;
```
pete-watters, line 6: @alexp3y this addresses concerns raised [here](https://github.com/leather-io/mono/pull/1745)

tigranpetrossian (approved):

packages/ui/src/components/avatar/sip10-avatar-icon.native.tsx
tigranpetrossian: I think I was the one who broke this. Strangely enough it was showing for me the whole time, I never realized it. Probably fell back to normal sip-10 avatar and was intermittently failing to load.

PR #1824, opened 2025-11-18
chore(extension): setup new Home component

pete-watters:
This PR sets the foundation for `extension` UI work to come:
- adds a feature flag `extension_revamp` that targets my clientId in LaunchDarkly
- duplicates the `Home` component to `HomeV2` (more to follow later)
- adds a switch to load the V2 home based on the feature flag


apps/extension/src/app/pages/home/home-v2.tsx
```diff
@@ -0,0 +1,112 @@
+import { Route } from 'react-router';
```
pete-watters, line 1: This component duplicates the old `Home` for now but will soon be the base for the extension refactor

apps/extension/src/app/routes/app-routes.tsx
```diff
@@ -88,11 +89,7 @@ function useAppRoutes() {
           >
             <Route
               path="/*"
-              element={
-                <AccountGate>
-                  <Home />
-                </AccountGate>
-              }
+              element={<AccountGate>{extension_revamp ? <HomeV2 /> : <Home />}</AccountGate>}
```
tigranpetrossian: Not a big deal, but in my experience you have less trouble doing the final cleanup if the flag routing is done inside `pages/home` vs. here.

`/pages/home.tsx` conditionally points to `pages/home-v1.tsx`
or `pages/home-v2.tsx`, later you just replace home.tsx contents without having to touch the routes.

apps/extension/src/app/routes/app-routes.tsx
```diff
@@ -71,7 +72,7 @@ export const homePageModalRoutes = (
 );
 
 function useAppRoutes() {
-  const { release_onramper_buy } = useFlags();
+  const { release_onramper_buy, extension_revamp } = useFlags();
```
tigranpetrossian: Unrelated, but LD should have a camel case setting somewhere to make these look less out of place.

tigranpetrossian (approved):
Happy to see the first legit branch by abstraction in the extension.

apps/extension/src/app/routes/app-routes.tsx
pete-watters: Thanks 👍 I'm just checking now and it seems all of our previous flags in `mobile` are already using snake case 😢 

We can easily change this setting `{ useCamelCaseFlagKeys:` for `web` and `extension` and change those as there are only 2 but then perhaps we should also update all of the mobile ones too?

I can open a task to refactor it if you think it's worth doing it?

apps/extension/src/app/routes/app-routes.tsx
pete-watters: Thanks for sharing your experience. I'll give a try to refactoring it to do it that way instead

apps/extension/src/app/routes/app-routes.tsx
tigranpetrossian: I think mobile can stay as is, because there's no `useFlags` there to do convert to camel case for us automagically, and we use custom hooks there anyway.

I should really get in touch with LD at some point and ask why they went with a different design for React Native SDK, because `useFlags` is much more convenient.

pete-watters: I'll merge this then add a fast follow to refactor `release_onramper_buy` & `extension_revamp` and make it use camel case CC @edgarkhanzadian

PR #1825, opened 2025-11-19
fix(extension): address speed issues with accounts switcher

pete-watters:


pete-watters:
Started work on some improvements here. I'll close for now and keep testing locally

PR #1826, opened 2025-11-19
chore(web): gate the new portfolio behind a LD flag

pete-watters:
This PR gates the new portfolio behind a LD flag so we can merge the branch safely and then release when it's ready.

This PR is going to `feat/portfolio` which will then enable us to merge [this PR](https://github.com/leather-io/mono/pull/1762) to `dev`

### DIFF
--- a/apps/web/app/features/feature-flags/index.ts
+++ b/apps/web/app/features/feature-flags/index.ts
@@ -1,4 +1,4 @@
-import { asyncWithLDProvider } from 'launchdarkly-react-client-sdk';
+import { asyncWithLDProvider, useFlags } from 'launchdarkly-react-client-sdk';
 import { VERSION } from '~/constants/constants';
 import { getClientId } from '~/utils/client-id';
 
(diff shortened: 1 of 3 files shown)


alexp3y (approved):
@pete-watters Should we also gate the route itself in `/web/app/rotues.ts` to prevent direct navigation?

pete-watters:
> /web/app/

Yes we probably should for safety but I was thinking it's easier to test if we don't do that. I can gate it anyway as a precaution, thanks!

pete-watters:
> @pete-watters Should we also gate the route itself in `/web/app/rotues.ts` to prevent direct navigation?

I gave this a try and it's tricky as React Router routes are defined at build time, but LaunchDarkly flags are resolved at runtime. So I can't use hooks in route configuration.

I can add an `ENV` variable to gate the route also but amn't sure if that's truly necessary. I went for a simpler implementation where the route exists but if the flag is not enabled and you manually navigate there you get re-directed to the 404 page

PR #1828, opened 2025-11-19
Add target fiat amount and balance to mobile swap form

tigranpetrossian:
Small change displaying target asset balance and the quote amount in fiat.

image


alexp3y (approved):
Looks great! 🔥

PR #1830, opened 2025-11-19
feat: add alex swap provider

alexp3y:
Adds an ALEX SDK swap provider implementation.

Also makes `SwapQuote` a discriminable Union Type to support more specific `providerQuoteData` implementations. Still requires some `unknown` types to avoid SDK type imports, but minimizes their necessity.


pete-watters (approved):
Great work @alexp3y

PR #1834, opened 2025-11-20
fix(features): react native imports used in shared file

kyranjamie:
`*.shared.ts` files are intended to contain platform-agnostic code shared between all platforms. RN specific code had been added to a shared file causing the web build to break.


pete-watters (approved):
Thanks @kyranjamie

PR #1837, opened 2025-11-20
feat(extension): refactor activity to features, update extension ui

pete-watters:
This PR adds to the feature flagged work for the new extension:
- adds a third tab for `NFTs` and re-jigs the old UI
- replaces `activity` with the new activity component from **portfolio**
- refactors **packages/features** to:
    -  be platform agnostic
    - share only feature specific utils and consts 
- moves shared `activity` logic to **packages/features/activity**
- refactors **packages/features/activity** to be platform agnostic and share
    - feature specific utils and consts
- moves shared `activity` query config to **packages/queries/activity**
    - standardises query config
    - standardises key generation using key factory
    - removed previous deprecated `totalActivity` query and renamed `account-activity` to `activity` 
- added `ActivityView` which is a UI ready view of activity
    - before we had the concept of `OnChainActivity` in our lists
    - `ActivityView` is shared across apps and pre-formats all UI data via RQ `select` 
    - now all source `Activity` gets transformed to `ActivityView` for standardised UI rendering

Here's a demo of activity on all platforms:

https://github.com/user-attachments/assets/08bd1848-8814-49ef-97b2-ddda03e9d537


apps/extension/src/app/features/activity-list/activity-list-v2.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ActivityList,
+  type OnChainActivity,
+  makeActivityLink,
+} from '@leather.io/features';
+
+import { openInNewTab } from '@app/common/utils/open-in-new-tab';
+import { useAccountActivity } from '@app/query/activity/account-activity.query';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+
+export function ActivityListV2() {
```
kyranjamie: Naming wise, it'd be more forward thinking to name this `ActivityList` and the old one `ActivityListLegacy` than version numbers.

packages/features/src/activity/index.shared.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export {
+  makeActivityLink,
+  getMempoolExplorerLink,
+  formatActivityCaption,
+  formatActivityStatusLabel,
+  getActivityTitle,
+  getBalancesText,
+  type BitcoinNetworkPreference,
+} from './activity.utils.shared';
```
kyranjamie: I think we should not use index files like this. It's annoying having to re-export everything, and for what? A slightly shorter import path?

packages/features/src/activity/utils.ts
```diff
@@ -144,13 +158,100 @@ export function getBalancesText(activity: OnChainActivity) {
...
+    addOperator(formatMoney(activity.value?.quote), getBalanceOperator(activity));
 
   return {
     formattedBalanceCrypto,
     formattedBalanceQuote,
   };
 }
+
+export type BitcoinNetworkPreference = 'mainnet' | 'testnet4' | 'signet';
+type StacksNetworkPreference = 'mainnet' | 'testnet';
```
kyranjamie: These types exist elsewhere, shouldn't be redefined here

pete-watters: Right now we have these types in `activity.utils` in the `models` package.  I moved them rather than re-defined. 

I agree they need to to be shared better though. We have a type in `DefaultNetworkConfigurations` in `network.model.ts` but this  `BitcoinNetworkPreference` has less supported networks. 

I think it dates back to initial mobile support - not covering all networks. 

Maybe we can align these better and curate `WalletDefaultNetworkConfigurationIds` to remove some legacy ones?

```
export enum WalletDefaultNetworkConfigurationIds {
  mainnet = 'mainnet',
  testnet = 'testnet',
  testnet4 = 'testnet4',
  signet = 'signet',
  sbtcTestnet = 'sbtcTestnet',
  sbtcDevenv = 'sbtcDevenv',
  devnet = 'devnet',
}
```

packages/features/src/activity/index.shared.ts
pete-watters: I can refactor but for these particular cases it makes it easier to subsequently export them from `exports.native.ts` and `exports.web.ts`

apps/extension/src/app/features/activity-list/activity-list-v2.tsx
pete-watters: I can change the naming from `V1` / `V2` to - `CurrentName` - `CurrentNameLegacy`. 

This revamp will take quite a while I image so I thought it safer to give versions

packages/models/src/network/network.model.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 import { Blockchain } from '../types';
 import { networkConfigurationSchema } from './network.schema';
 
+// TODO LEA-2285: Check if these should be here or in constants
```
pete-watters: We have a lot of constants in here. Is `packages/models` OK to use for constants? 

Should I instead move these to `packages/constants` ?

apps/extension/src/app/common/hooks/use-bitcoin-explorer-link.ts
```diff
@@ -1,6 +1,7 @@
 import { useCallback } from 'react';
 
-import { makeBitcoinTxExplorerLink } from '@app/common/utils';
+import { getMempoolExplorerLink } from '@leather.io/features';
```
tigranpetrossian: I think the old generic naming for the link creator was better: doesn't tie to a provider, but to a chain. Easier to change internally later.

packages/features/src/activity/utils.ts
```diff
@@ -1,14 +1,35 @@
...
+  NetworkConfiguration,
   type OnChainActivity,
   isFungibleAsset,
   isInscriptionAsset,
   isStampAsset,
 } from '@leather.io/models';
-import { minusSign } from '@leather.io/utils';
+import { FormatAmountOptions, createCurrencyFormatter, minusSign } from '@leather.io/utils';
+
+const currencyFormatter = createCurrencyFormatter({
```
tigranpetrossian: What's the reason for creating a new formatter here. Seems identical?

tigranpetrossian: Oh, didn't realize this was in `features`.  This reveals a critical flaw, apart from the duplication, different name, and omitting the tracing that's present in the apps, this will also be impossible to inject a locale into from the app because it's a brand new instance with a hardcoded locale.
The formatter should be injected into the functions using it, instead of recreating it in the `features` package.

packages/models/src/network/network.model.ts
alexp3y: I would say move them, but also our use of `packages/constants` has been pretty inconsistent up until now. It would be nice if that package was more organized instead of everything in a single grab bag `constants.ts` file.

If you do move them, maybe we can apply the same domain structure we've established elsewhere (e.g. `network.constants.ts`)? That would make it much more discoverable and promote proper usage / consolidation of constants going forward.

pete-watters: Thanks 👍 yes I think it's a good idea to move them and have them organised.

Similarly if we had some organisation like that in `packages/utils` it would negate the need for a `features` package at all and we could just share common feature utils from there

alexp3y: Agreed. We should reopen and prioritize the discussion on utilities strategy because we are super fragmented atm. We just need to decide on a convention / location, whether it's in `packages/models`, `packages/utils` or somewhere new and do a migration to start reducing duplication.
cc @tigranpetrossian

pete-watters:
Moving this back to draft as @tigranpetrossian had some suggested improvements and @alexp3y has added the new `queries` package. I'll re-work this to take both points on and re-open shortly

packages/models/src/network/network.model.ts
pete-watters: I'll do this in a follow up as this PR already got way too big

apps/extension/src/app/features/activity-list/activity-list.tsx
```diff
@@ -1,167 +1,96 @@
...
+import { ActivityItem } from './components/activity-item';
 
-function useTrBitcoinAddress() {
-  try {
-    return useZeroIndexTaprootAddress();
-  } catch (e) {
-    return '';
-  }
-}
+/*
```
pete-watters, line 15: I can get rid of this comment but testing `activity` and it only seems to go back to a certain point ( around August). Maybe we want to allow `loadMore` to load more data?

apps/extension/src/app/pages/home/home.tsx
```diff
@@ -1,9 +1,82 @@
...
+      width="100%"
+      bg="ink.1"
+      borderRadius="lg"
+      animation="fadein"
+      animationDuration="500ms"
+    >
+      <Box px={{ base: 'space.05', md: 0 }} pb={{ base: 'space.05', md: 0 }}>
+        <AccountCard
+          name={name}
+          availableBalance={
```
pete-watters: I'll refactor this to use the `Balance` component when working on centralising the balances UI

packages/queries/src/shared/query-options.ts
```diff
@@ -15,6 +15,15 @@ export const balanceQueryOptions = {
   gcTime: 5000,
 } satisfies Partial<UseQueryOptions>;
 
+export const activityQueryOptions = {
```
pete-watters, line 18: This is now centralised but we should consider how often we want activity to update. Is 5 seconds enough?

packages/features/src/activity/activity-links.ts
```diff
@@ -28,47 +29,32 @@ interface MakeActivityExplorerLinkArgs {
...
+
 function makeActivityExplorerLink({
   asset,
   txid,
   networkPreference,
 }: MakeActivityExplorerLinkArgs) {
   if (asset.chain === 'bitcoin') {
     return getMempoolExplorerLink({
-      networkPreference: networkPreference.chain.bitcoin.bitcoinNetwork as BitcoinNetworkPreference,
+      networkPreference: networkPreference.chain.bitcoin.bitcoinNetwork,
```
kyranjamie, line 40: I find `networkPreference` a strange choice of name. Is network a preference, or a matter of fact about a given transaction? This name conflates a user setting of a certain network with the unquestionable data of a transaction.

apps/extension/src/app/features/activity-list/components/activity-item.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function Item({ item, formatCurrency }: ItemProps) {
+  const { activityLink, title, caption, balances } = item;
+  const clickable = Boolean(activityLink);
+
+  const handleClick = useCallback(() => {
+    if (!activityLink) return;
+    openInNewTab(activityLink);
+  }, [activityLink]);
```
kyranjamie: It really isn't necessary to use a useCallback here. This is such a simple, cheap function, and there's a performance cost to adding the useCallback.

apps/extension/src/app/pages/home/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { useFlags } from '@app/features/feature-flags';
+
+import { Home as HomeCurrent } from './home';
+import { HomeLegacy } from './home-legacy';
+
+export function Home() {
+  const { extension_revamp } = useFlags();
+  return extension_revamp ? <HomeCurrent /> : <HomeLegacy />;
+}
```
kyranjamie: I'm not really clear on the benefit of flagging this?

apps/extension/src/app/pages/home/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { useFlags } from '@app/features/feature-flags';
```
kyranjamie: We don't really use index.ts files in extension and would prefer we avoided them

apps/mobile/src/features/activity/activity-item.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+function ActivityItemComponent({ item }: ActivityItemProps) {
+  const { balances, title, caption, activityLink } = item;
+  const { openURL } = useOpenURL();
+
+  const handlePress = useCallback(() => {
+    if (activityLink) {
+      openURL(activityLink);
+    }
+  }, [activityLink, openURL]);
```
kyranjamie: Same here

packages/features/src/activity/activity-balance.ts
```diff
@@ -0,0 +1,73 @@
+import type { Activity, Money } from '@leather.io/models';
+import { type FormatAmountOptions, minusSign } from '@leather.io/utils';
+
+import type { ActivityBalances } from './types';
+
+export type FormatMoney = (money: Money, options?: FormatAmountOptions) => string;
+
+function addOperator(balance: string, operator?: string) {
+  return operator ? `${operator} ${balance}` : balance;
+}
```
kyranjamie: this is repeated elsewhere

packages/queries/src/activity/activity.query-config.ts
```diff
@@ -0,0 +1,90 @@
...
+): UseQueryOptions<Activity[], Error, TData, QueryKey> {
+  const { queryKeyContext = [], select, ...queryOptions } = options;
+  return {
+    queryKey: [...createActivityByAssetQueryKey(account, asset, settings), ...queryKeyContext],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      getActivityService().getActivityByAsset(account, asset, signal),
+    select: select ?? (activity => defaultSelect(activity, settings)),
+    ...activityQueryOptions,
+    ...queryOptions,
+  } as UseQueryOptions<Activity[], Error, TData, QueryKey>;
```
kyranjamie: Can we use satisfies here, and in all other cases, to make sure we're not forcing the type?

packages/queries/src/activity/activity.query-config.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export function createActivityQueryKey(account: AccountAddresses, settings: UserSettings) {
+  return createServiceQueryKey(
+    'activity-service--get-activity',
+    createActivityKeyParams(account),
+    settings
+  );
+}
+
+export function defaultSelect(activity: Activity[], settings: UserSettings): ActivityView[] {
```
kyranjamie: unclear fn name

apps/extension/src/app/pages/home/home.tsx
```diff
@@ -1,9 +1,82 @@
...
+import { Assets } from '@app/pages/home/components/assets';
+import { homePageModalRoutes } from '@app/routes/app-routes';
+import { ModalBackgroundWrapper } from '@app/routes/components/modal-background-wrapper';
+import { AccountCard } from '@app/ui/components/account/account.card';
+
+import { AccountActions } from './components/account-actions';
+import { HomeTabs } from './components/home-tabs';
+import { useHomePageState } from './use-home-page-state';
+
+const emptyBalanceDisplay = '-.--';
```
kyranjamie: string repeated

apps/extension/src/app/pages/home/home.tsx
```diff
@@ -1,9 +1,82 @@
...
+    togglePrivateMode,
+    toggleSwitchAccount,
+  } = useHomePageState();
+
+  return (
+    <Stack
+      data-testid={HomePageSelectors.HomePageContainer}
+      px={{ base: 0, md: 'space.05' }}
+      py={{ base: 0, md: 'space.07' }}
+      gap={{ base: 0, md: 'space.06' }}
```
kyranjamie: Pretty please use array notation for responsive styles, we use this everywhere else in extension

A case why, beyond consistency:

1. I don't remember all the breakpoints by heart, and I don't want to. Is the next increment up xs or sm? With array notation, there is no ambiguity. The second breakpoint is in the second index of the array.
2. Ordering. Nothing stops someone mixing the order of these properties. Breakpoint styles are linear in their nature, smallest to biggest. When order matters, an array is usually the best choice.

apps/mobile/src/features/activity/activity-item.tsx
kyranjamie: Same here, useCallback not needed for this function

apps/mobile/src/features/activity/activity-item.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { useOpenURL } from '../browser/browser/use-open-url';
+
+interface ActivityItemProps {
+  item: ActivityView;
+}
+
+function ActivityItemComponent({ item }: ActivityItemProps) {
+  const { balances, title, caption, activityLink } = item;
+  const { openURL } = useOpenURL();
```
kyranjamie:
```suggestion
  const { openUrl } = useOpenUrl();
```

https://www.notion.so/trustmachines/Naming-Camelcase-acronyms-119d74b694c78080974aee5e3bc18bc9?source=copy_link

packages/ui/src/components/avatar/activity-avatar-icon.native.tsx
```diff
@@ -1,43 +1,46 @@
...
 import { Sip10AvatarIcon } from './sip10-avatar-icon.native';
 import { StxAvatarIcon } from './stx-avatar-icon.native';
 
-interface StatusIndicatorProps {
-  type: ActivityType;
-  status: OnChainActivityStatus;
-}
-function StatusIndicator({ type, status }: StatusIndicatorProps): ReactElement {
-  switch (status) {
+function renderStatusIndicator(indicator: ActivityStatusIndicatorId): ReactElement | null {
```
kyranjamie: Isn't a function that returns JSX just.... a component? What prompted this change?

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -106,30 +90,36 @@ function SwapAvatarIcon({ fromAsset, toAsset, indicator }: SwapAvatarIconProps):
...
 export function ActivityAvatarIcon({ activity }: ActivityIconProps) {
-  const statusIndicator = <StatusIndicator type={activity.type} status={activity.status} />;
+  const indicatorElement = renderStatusIndicator(activity.statusIndicator ?? 'hidden');
+  const avatarType =
+    activity.activityAvatar ??
+    (activity.fromAsset && activity.toAsset
+      ? 'swap'
+      : activity.asset || activity.fromAsset || activity.toAsset
+        ? 'asset'
+        : 'fallback');
```
kyranjamie: Repeated logic from packages/ui/src/components/avatar/activity-avatar-icon.native.tsx

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -106,30 +90,36 @@ function SwapAvatarIcon({ fromAsset, toAsset, indicator }: SwapAvatarIconProps):
...
-    return <AssetAvatar asset={asset} indicator={statusIndicator} />;
+  if (activity.asset) {
+    return <AssetAvatar asset={activity.asset} indicator={indicatorElement ?? undefined} />;
   }
 
   return (
     <Sip10AvatarIcon
-      contractId={activity.type}
-      indicator={statusIndicator}
+      contractId=""
```
kyranjamie: ?

apps/extension/src/app/features/activity-list/activity-list-legacy.tsx
```diff
@@ -0,0 +1,172 @@
...
+        )}
+      </ActivityListTabWrapper>
+      <Outlet />
+    </>
+  );
+}
+
+export function ActivityListLegacy() {
+  return <ActivityListLegacyInner />;
+}
```
kyranjamie: This component doesn't do much

kyranjamie (commented):
Hi @pete-watters great work. 

Can you rebase on dev so I can test out this branch. I get tsdown errors that've been fixed


```
@leather.io/queries:build: sh: tsdown: command not found
@leather.io/queries:build:  ELIFECYCLE  Command failed.
@leather.io/queries:build: ERROR: command finished with error: command (/Users/kyranjamie/dev/leather/mono-pull-requests/packages/queries) /Users/kyranjamie/Library/pnpm/pnpm run build exited (1)
@leather.io/ui:build:  ELIFECYCLE  Command failed.
@leather.io/queries#build: command (/Users/kyranjamie/dev/leather/mono-pull-requests/packages/queries) /Users/kyranjamie/Library/pnpm/pnpm run build exited (1)
```

tigranpetrossian:
I did a round of reviewing this, though still wrapping my head around things. `createActivityView` indeed improved things quite a bit, but I'm seeing lots of mixed responsibilities and cross-package concerns that seem to be a path to extreme entropy.

The most worrying aspect is that `createActivityQueryConfig` is the new `queries` package has knowledge about `ActivityView`. 
Meaning now we have `queries` -> `features` dependency. `queries` should have no business knowing about those VMs, those selects should happen at usage level vs. being baked in.

apps/extension/src/app/pages/home/index.tsx
pete-watters: Flagging this as it means I can stagger in new code to the extension without having one huge PR to overhaul everything. We have this [monster project](https://linear.app/leather-io/project/extension-mobile-onboarding-home-screen-2a5a69b712f7/overview) to overhaul the onboarding and all of the extension home. 

I thought it better to gradually get updates in then do extensive tests on the full new UI before releasing. 

If you want I can remove all of the old code now and just replace it as I go? If that was the case I could have started on that 2 weeks ago when it was first reviewed

packages/features/src/activity/activity-links.ts
pete-watters: I am open to changing this now but this is something baked into the code for a long time. I can change it to `network` instead? 

Explorer links are created based on the users settings. Should I change this instead to try and take the network from the transaction itself? I don't think we have that available in activity

packages/features/src/activity/activity-balance.ts
pete-watters: Thanks, I missed this. I'll move it to `features`

apps/mobile/src/features/activity/activity-item.tsx
pete-watters: I'll update this now but doing a quick search on `dev` and I get 55 results for `openURL` so it's not a new thing

packages/ui/src/components/avatar/activity-avatar-icon.native.tsx
pete-watters: Thanks Kyran, this was an automated mistake. I'll fix this and refactor the others to use JSX also

apps/extension/src/app/features/activity-list/components/activity-item.tsx
pete-watters: You're right, I thought it would help optimise this as it's rendered in a list but I will remove it

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
pete-watters: This is a mistake and a legacy bug I will fix, thanks! On web we use `Sip10AvatarIcon` with nothing passed as an empty fallback when we don't have an activity asset. 

I'll fix this properly now and make it clearer

apps/extension/src/app/pages/home/index.tsx
pete-watters: I can update this but this is a throwaway file for gating purposes. I had done things differently initially but changed it to us this index based on [this feedback](https://github.com/leather-io/mono/pull/1824#discussion_r2540883831) from @tigranpetrossian . 

The purpose of this file is to allow easy switching between both versions without touching the routes. 

If you think it's better I can just do away with this altogether and [get rid of the feature flag](https://github.com/leather-io/mono/pull/1837#discussion_r2588281971)?

apps/mobile/src/features/activity/activity-item.tsx
pete-watters: It seems this name comes from the `react-native` `Linking` package. I've refactored and renamed our own uses to match this convention

apps/extension/src/app/pages/home/home.tsx
pete-watters: I can change this now but I didn't add this right now, I am just moving this code we had already

packages/queries/src/activity/activity.query-config.ts
alexp3y, line 1: Agree with @tigranpetrossian 's comment about the dependency inversion here.

We should try to be pretty strict about `features` -> `queries` -> `services`

Ultimately the query config in `@leather.io/queries` should be a super thin wrapper w/o any feature-specific usage logic or types. 

Could be a pretty simple adjustment to allow `select` to be passed in from `features` via the `UseActivityQueryOptions` param.

alexp3y: Just refreshed and looks like this was just done almost exactly. @pete-watters as usual, way ahead of me :)

apps/extension/src/app/pages/home/home.tsx
pete-watters: I've updated this but essentially this whole file is duplicated with old and new home page

packages/queries/src/activity/activity.query-config.ts
pete-watters: I've refactored and updated this to apply the `select` in each consuming app

packages/queries/src/activity/activity.query-config.ts
pete-watters: Thanks 👍  this is dead code now I have refactored [as per](https://github.com/leather-io/mono/pull/1837#issuecomment-3611074884)

pete-watters:
> I did a round of reviewing this, though still wrapping my head around things. `createActivityView` indeed improved things quite a bit, but I'm seeing lots of mixed responsibilities and cross-package concerns that seem to be a path to extreme entropy.
> 
> The most worrying aspect is that `createActivityQueryConfig` is the new `queries` package has knowledge about `ActivityView`. Meaning now we have `queries` -> `features` dependency. `queries` should have no business knowing about those VMs, those selects should happen at usage level vs. being baked in.

Thanks for the feedback. I have refactored this to remove the dependancy on `features` in `queries` and do do the select and transform in each consuming app

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
pete-watters: I've refactored this in the latest commit to make the defaults more explicit. We need to do more work adding better support for more protocols and filling gaps with actual icons we can host ourself but hopefully improvements have been made now and that should be easier to do

pete-watters:
@tigranpetrossian @kyranjamie @alexp3y : Thanks for the reviews! I've went through a round of refactoring and applied almost all suggested changes and also some bug fixes. 

Outstanding items:
- should [I bother feature flagging](https://github.com/leather-io/mono/pull/1837#discussion_r2588095519) extension update or stagger new code in as its ready
- changing `networkPreference` [naming](https://github.com/leather-io/mono/pull/1837#discussion_r2588087161) 

I squashed commits up until the initial review and then added individual commits for each new work. 

I am investigating issues with some avatars being cut off but checking against `dev` and that is also the case there. I will start looking at replacing some of the fallback avatars with our own hosted avatars while I investigate that. 

**This branch**

Swap slightly cut off
Screenshot 2025-12-05 at 06 14 47

sBTC not filling space when indicator added 
Screenshot 2025-12-05 at 06 14 02

**dev branch**
Screenshot 2025-12-05 at 06 16 47
Screenshot 2025-12-05 at 06 16 20

alexp3y (approved):
Solid adjustments. I think the remaining items you flagged can likely be addressed elsewhere.

Overall this PR is a great step forward at adopting a solid addition (`@leather.io/features`) to a well-layered architecture.

tigranpetrossian (approved):
Thanks @pete-watters 
Re: avatars, I'll see if I can spend some time re-working the asset avatars once and for all, because we have few random problems across platforms this included.

apps/extension/src/app/common/hooks/use-bitcoin-explorer-link.ts
pete-watters: Good point, I can revert this and update `makeStacksTxExplorerLink` also

pete-watters:
> Thanks @pete-watters Re: avatars, I'll see if I can spend some time re-working the asset avatars once and for all, because we have few random problems across platforms this included.

Thanks @tigranpetrossian 👍 I've been speaking with @alexp3y about hosting some more of the SVGs ourselves so we can make improvements there and polish the UI. 

I'll check out about the `formatter` but I think that's outdated.

packages/features/src/activity/utils.ts
pete-watters: Thanks, I'll check this. It's not supposed to be in `features` and if so it was a mistake, I think I did that and this is stale now but I'll make sure.

kyranjamie:
I see this when I build myself 🤔 

image

pete-watters:
> I see this when I build myself 🤔
> 
> image

Good 👀  I didn't notice that, I'll check and fix!

The images is a known issue I mentioned [here](https://github.com/leather-io/mono/pull/1837#issuecomment-3615477269)

I will fix it ASAP. I was speaking with Alex about hosting more icons ourself via cloudflare so I will be giving this more attention soon and will address all UI issues then. As it's feature flagged for now it's safe to merge

packages/queries/src/activity/sip10-activity.query-config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { createServiceQueryKey } from '../shared/query-key.factory';
+import { activityQueryOptions } from '../shared/query-options';
+
+type Sip10ActivityQueryOptions<TData = Activity[]> = Omit<
+  UseQueryOptions<Activity[], Error, TData, QueryKey>,
+  'queryKey' | 'queryFn'
+> & {
+  queryKeyContext?: readonly unknown[];
+};
```
kyranjamie: We used to use a pattern of trying to make hooks configurable like this previously, [discussed in this comment](https://github.com/orgs/leather-io/discussions/61#:~:text=AppUseQueryConfig%20requires%20some%20typescript%2Dheavy%20boilerplate%20for%20each%20query%20definition.%20I%27m%20not%20a%20fan%20of%20the%20verbosity%2C%20nor%20the%20typing%20issues%20that%20come%20with%20this%20approach.%20Hooks%20in%20general%20are%20quite%20static%20and%20hard%20to%20configure%20with%20HOC%20style%20techniques.).

It gets ugly very quick, and queries are not really intended to be used this way. Suggest this is not something we adopt widely as a pattern.

apps/extension/src/app/features/activity-list/activity-list.tsx
```diff
@@ -1,167 +1,102 @@
...
+            justifyContent="center"
+            flexGrow={1}
+          >
+            <LoadingSpinner />
+          </Flex>
+        </Stack>
         <Outlet />
-      </ActivityListTabWrapper>
+      </>
     );
```
kyranjamie: Can we use a layout component? Here we're mixing all the UI and data fetching concerns in the same component.

apps/extension/src/app/pages/home/components/home-tabs.tsx
```diff
@@ -16,14 +16,22 @@ interface HomeTabsProps {
 export function HomeTabs({ children }: HomeTabsProps) {
   const navigate = useNavigate();
   const location = useLocation();
+  const activeTab = useMemo(() => {
+    if (location.pathname.startsWith(RouteUrls.Activity)) return RouteUrls.Activity;
+    if (location.pathname.startsWith(RouteUrls.Collectibles)) return RouteUrls.Collectibles;
+    return RouteUrls.Home;
+  }, [location.pathname]);
```
kyranjamie, line 40: Can you help me understand this change?

pete-watters: Highlighted tab was not updating properly when using browser navigation ( back links)

packages/queries/src/activity/sip10-activity.query-config.ts
pete-watters: OK I will investigate it 👍

apps/extension/src/app/pages/home/components/home-tabs.tsx
kyranjamie: Got it. Can it be dynamic? E.g. can't you read the values of `RouteUrls` directly from the `useLocation` hook?

apps/extension/src/app/features/activity-list/activity-list.tsx
pete-watters: I've updated this, thanks!

pete-watters:
> I see this when I build myself 🤔
> 
> image

I've re-tested this and it seems OK to me 🤔 


https://github.com/user-attachments/assets/9c97abd5-a07e-4b45-91ef-1e588f1c76cf

apps/extension/src/app/common/hooks/use-bitcoin-explorer-link.ts
pete-watters: I've updated this

packages/features/src/activity/utils.ts
pete-watters: Confirmed this is stale and not in `features`. Thanks for pointing out though

packages/queries/src/activity/sip10-activity.query-config.ts
pete-watters: I got to work updating this and also refactoring other queries but quickly it made this PR grow even more so I have split out that work into a fast follow up. I have a [draft PR here](https://github.com/leather-io/mono/pull/1895/).

Now the CI is unblocked I'll merge what has been reviewed and approved for this PR and add additional changes to this one

apps/extension/src/app/pages/home/components/home-tabs.tsx
pete-watters: I'll change this and more in a [follow up](https://github.com/leather-io/mono/pull/1895/)

PR #1839, opened 2025-11-21
fix(web): Miscellaneous test and CI improvements

kyranjamie:
Also using this PR to test how the tests run


apps/web/package.json
```diff
@@ -71,7 +71,7 @@
     "react-dom": "19.0.0",
     "react-hook-form": "7.62.0",
     "react-markdown": "10.1.0",
-    "react-router": "7.9.4",
+    "react-router": "7.9.6",
```
pete-watters, line 74: We should probably update the extension version too? Not here though

pete-watters (approved):
Nice work @kyranjamie 👍  good to get to the bottom of this.

If this shard is more reliable it saves a lot of dev time and CI resources

apps/web/package.json
kyranjamie: I think I added RR as a dependency that can diverge. Usage is pretty different between the two, client side routing vs SSR framework mode etc

PR #1840, opened 2025-11-21
fix: tsdown upgrade

edgarkhanzadian:
- Use tsdown instead of tsup
- Use project references
- Implement root `typecheck:tsc` to typecheck the whole project by just using the tsc with project references. Might come in handy for eslint work.
- Use `incremental: true` + `composite: true` on all ts packages
- Try out `isolatedDeclarations: true` on `constants` package. Package itself doesn't really change that much so i don't see any harm in using isolatedDeclarations just for this package for now (100ms builds instead of 800ms)


packages/ui/package.json
```diff
@@ -20,9 +20,9 @@
     "typecheck": "tsc --noEmit --project ./tsconfig.json"
   },
   "exports": {
-    ".": "./dist-web/exports.web.js",
+    ".": "./dist-web/exports.web.mjs",
```
kyranjamie: How come we have to use *.mjs files? I was under the impression most build systems let you choose whatever you want .js files to default to, esm nowadays

edgarkhanzadian: thanks for pointing that out, after some research found a tsdown option for outputting .js extensions

packages/ui/src/components/avatar/stamp-avatar-icon.web.tsx
```diff
@@ -1,8 +1,7 @@
-import StampAvatarIconSrc from '../../assets/images/stamps-avatar-icon.png';
 import { Avatar, type AvatarProps } from './avatar.web';
 
 const fallback = 'ST';
 
 export function StampAvatarIcon(props: AvatarProps) {
-  return <Avatar fallback={fallback} image={StampAvatarIconSrc} {...props} />;
+  return <Avatar fallback={fallback} image={undefined} {...props} />;
```
edgarkhanzadian: setting it to undefined here for now as we never had a `stamps-avatar-icon.png` file to begin with

pete-watters: We have one in `extension/public/assets/avatars/stamps-avatar-icon.png` I guess I forgot to move it into the `ui` package

apps/extension/tsconfig.json
```diff
@@ -40,5 +40,49 @@
   "exclude": ["test-app"],
   "parserOptions": {
     "extraFileExtensions": [".mdx"]
-  }
+  },
+  "references": [
```
pete-watters: Can we do this differently to share it from a central place? We already have the `tsconfig-config` package. 

It seems like we will need to keep updating these files?

eslint.config.js
```diff
@@ -40,10 +40,12 @@ export default tseslint.config(
       '**/dist-native/',
       '**/.expo/',
       '**/.turbo/',
-      '**/.tsup/',
+      '**/.tsdown/',
       '**/*.d.ts',
       '**/*.stories.{ts,tsx}',
-      '**/tsup.config*.ts',
+      '**/tsdown.config*.ts',
+      '**/panda.config*.ts',
```
pete-watters, line 47: Why do we need to ignore `panda.config` and `react-router.config`? 

On line 50 we also ignore `/apps/extension/`, maybe we can remove that now?

packages/test-config/README.md
```diff
@@ -0,0 +1 @@
+# @leather.io/test-config
```
pete-watters, line 1: How about just calling this `@leather.io/test` so we can use it for more things in future? 

 Maybe even shared test code itself, mocks and the extension test-app?

You should add some content to the README too

packages/ui/src/components/avatar/stamp-avatar-icon.web.tsx
pete-watters: We actually do have one in `packages/ui/src/assets-web/images/stamps-icon.png`, maybe we can just update the path here?

packages/ui/src/components/approver/components/approver-advanced.native.tsx
```diff
@@ -1,14 +1,15 @@
...
+import type { ReactNode } from 'react';
 
 import { ChevronDownIcon } from '../../../icons/chevron-down-icon.native';
 import { ChevronUpIcon } from '../../../icons/chevron-up-icon.native';
 import { Pressable } from '../../pressable/pressable.native';
 import { Text } from '../../text/text.native';
 import { useApproverContext, useRegisterApproverChild } from '../approver-context.shared';
 
-interface ApproverAdvancedProps extends HasChildren {
+interface ApproverAdvancedProps {
```
pete-watters: Is there a benefit to removing `extends HasChildren`? Should we avoid that?

pete-watters (approved):
Nice work @edgarkhanzadian 👍  I added a few Q's but overall it LGTM

apps/extension/tsconfig.json
edgarkhanzadian: nope, for each package we need to specify project's dependencies

eslint.config.js
edgarkhanzadian: if we do want to lint these we also need to include these in the tsconfig.json, and i don't think there is much benefit we can get from compiling or linting these files

packages/test-config/README.md
edgarkhanzadian: i think just having package `test` might be a little confusing, i'd suggest we leave it as test-config for now

packages/ui/src/components/approver/components/approver-advanced.native.tsx
edgarkhanzadian: good point! I was getting a tsc error here before but it seems it got resolved after a few iterations

packages/ui/src/components/avatar/stamp-avatar-icon.web.tsx
edgarkhanzadian: Good point, added

packages/test-config/README.md
pete-watters: OK 👍  maybe a readme would help anyway?

apps/extension/tsconfig.json
pete-watters: OK. It seems like a bit of a pain and something easily forgotten

packages/test-config/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "prettier": "3.5.1",
+    "tsdown": "0.16.5",
+    "typescript": "5.9.3"
+  },
+  "keywords": [
+    "leather",
+    "leather wallet",
+    "test"
+  ],
+  "main": ".prettierrc.js",
```
tigranpetrossian: Probably a copy-paste error, should be `./dist/index.js`?

packages/cms/tsconfig.json
```diff
@@ -5,5 +5,5 @@
     "outDir": "./dist/client"
   },
   "include": ["**/*.ts"],
-  "exclude": ["./dist/", "./src/studio/**"]
+  "exclude": ["./dist/"]
```
tigranpetrossian: I think `studio` is external code that should stay excluded.

packages/query/tsconfig.json
```diff
@@ -6,5 +6,25 @@
     "outDir": "./dist"
   },
   "include": ["**/*", ".*.ts", "config/*.json"],
-  "exclude": ["./dist/"]
+  "exclude": ["./dist/"],
```
tigranpetrossian: I think this is the best time to exclude `node_modules` from all packages, not excluding it is a mistake that contributes to TS compilation slowdowns considerably.

packages/constants/tsconfig.json
```diff
@@ -1,8 +1,14 @@
 {
   "extends": ["@leather.io/tsconfig-config/tsconfig.base.json"],
-  "include": ["**/*", ".*.ts"],
+  "include": ["src/**/*", "src/.*.ts"],
   "exclude": ["./dist/"],
   "compilerOptions": {
-    "outDir": "./dist"
-  }
+    "outDir": "./dist",
+    "isolatedDeclarations": true
```
tigranpetrossian: I don't think it's worth having a different config in a single package, especially if it requires disabling the relevant rule in the eslint config, which is the reason we decided against it to begin with when we were discussing it at length.

tigranpetrossian (changes requested):
Looks great @edgarkhanzadian! I haven't worked with `references` much, but it looks like a significant improvement from what I understand. 

Blocking to address the two important issues:
1. `src/studio` in `cms` should probably stay excluded, it's an external app not worth typechecking
2. We need to exclude `node_modules` in all package `tsconfig.json` files. Isn't an error made in this PR, but it's ]critical and this might be the best place to do it.

packages/test-config/package.json
edgarkhanzadian: yes! Good catch !

packages/query/tsconfig.json
edgarkhanzadian: i think it should be excluded by default

tigranpetrossian: Not if you have \`"include": \["\*\*/\*"\]\`, another option is not to the latter.

packages/constants/tsconfig.json
edgarkhanzadian: well yes but it is a package that is updated like once per month and the update itself in 90% of times isn't going to be disturbed by this setting. IMO enabling this setting specifically for this type of packages is quite convenient

packages/query/tsconfig.json
edgarkhanzadian: Looking through the ref: https://www.typescriptlang.org/tsconfig/#exclude

seems like `exclude` removes all filenames from whatever `include` found with its pattern match.
`exclude` does have an array of directories set by default, but if we explicitly set `exclude` to something else and don't specify node_modules, it does let ts work through node_modules directory 🤔 

So in this case, yeap we are going into node_modules because we set exclude to something else

edgarkhanzadian: nice catch!

tigranpetrossian: The easy way to find out is always `tsc --showConfig`

tigranpetrossian: P.S. Perhaps the best way is to specify this in the base config?<br>Edit: … although I don't know how configs are merged tbh

edgarkhanzadian: seems like setting it in base config does work, but if we for any reason decide to write `exclude` in the package's tsconfig too, it overrides whatever setting we have in the base config's `exclude` :/

tigranpetrossian: I had a hunch that was going to be the case 🙃

edgarkhanzadian: i think the best way would be to include only the files in the `src` directory

kyranjamie (changes requested):
My read of TS references is that each packages does not need to reference the other. This would be quite an unsustainable way of declaring refs. In these examples they describe once at mono root

Image


https://nx.dev/blog/typescript-project-references

https://moonrepo.dev/docs/guides/javascript/typescript-project-refs

edgarkhanzadian:
@kyranjamie i agree it is maintenance-heavy

In the example that you linked, they do mention they need references in both root of the monorepo and in each of the packages: https://nx.dev/blog/managing-ts-packages-in-monorepos#improving-performance-with-project-references

official typescript website also mentions that we should have references in both the root tsconfig and in the projects: https://www.typescriptlang.org/docs/handbook/project-references.html#structuring-for-relative-modules

That makes sure every package knows which package it depends on. Then, when we run tsc --build from the root of the monorepo, it knows the order in which it should build the packages.

Now we don't use tsc --build in monorepo root and we instead use turborepo for building the monorepo, but we do use tsc --build in mobile/ and web/ apps, and those commands need the whole chain of references to know the build order.

EDIT: Actually, we probably need to remove tsc --build completely or use only tsc --build s.t. we don't deal with outputs of both tsc and tsdown. Looking into that

edgarkhanzadian:
yea, removing references until we decide on using tsc --build in the future

edgarkhanzadian:
@kyranjamie removed references from projects and apps, only left those in the root tsconfig.json
@tigranpetrossian cleaned up tsconfig inputs to only search for files inside src/ folder.

Removed tsc --build that was causing outputs from multiple build scripts in the same dist directory.

We had *multiple* packages with no prettier setup, fixed those too

edgarkhanzadian:
eslint speeds are also drastically increased now. It only takes ~50 seconds to lint the whole monorepo with tslint type rules enabled ⭐

edgarkhanzadian:
still has room for improvement but much better than previous ~5-6 minutes with oom errors

apps/extension/scripts/mono-link-ui.js
```diff
@@ -35,15 +35,14 @@ async function modifyUiPackageJson() {
       '@leather.io/constants': 'file:../mono/packages/constants',
       '@leather.io/models': 'file:../mono/packages/models',
       '@leather.io/tokens': 'file:../mono/packages/tokens',
-      '@leather.io/utils': 'file:../mono/packages/utils'
+      '@leather.io/utils': 'file:../mono/packages/utils',
```
kyranjamie: we can delete this entire file

kyranjamie (approved):
Is this an upgrade or a downgrade?

apps/extension/scripts/mono-link-ui.js
edgarkhanzadian: sounds good. It does have a script in package.json but it seems to be unused too. Deleting both

tigranpetrossian: So happy to never have to do this again

tigranpetrossian (approved):

packages/cms/tsconfig.json
tigranpetrossian: Scratch this. Didn't realize this wasn't pointing to the actual sanity studio output. Shouldn't have been excluded to begin with.

PR #1843, opened 2025-11-24
refactor(extension): migrate utxo fetching to services, ref LEA-3262

alexp3y:
Migrates extension UTXO fetching to use the `UtxosService` and replaces coin-selection function with the `@leather.io/bitcoin`.


apps/extension/src/app/common/transactions/bitcoin/fees/bitcoin-fees.spec.ts
```diff
@@ -1,24 +1,21 @@
...
 
-import type { UtxoResponseItem } from '@leather.io/query';
+import type { OwnedUtxo } from '@leather.io/models';
 import { createMoney } from '@leather.io/utils';
 
 import { filterUneconomicalUtxos } from '../utils';
 import { calculateMaxBitcoinSpend } from './calculate-max-bitcoin-spend';
 
-function generateTxId(value: number): UtxoResponseItem {
+function generateTxId(value: number): OwnedUtxo {
```
pete-watters, line 10: Maybe we should move these bitcoin specific utils to the bitcoin package while we are working on them?

alexp3y: Agree we should migrate shared utils to/from `@leather.io/bitcoin` including the remaining coin-selection and max-spend utils that are still duplicated here.

There are some small differences in the two versions of some utilities that need to be reconciled, so I think we should do the bitcoin util migration as a standalone change to ensure that there are no regressions in extension.

pete-watters (approved):
Nice work!

PR #1847, opened 2025-11-25
refactor(extension): migrate taproot utxos to services

alexp3y:
Extends upon work from original `UtxosService` migration and targets taproot UTXOs and inscriptions.

Merging this back to the original refactor branch so that we can consolidate testing on a single build before releasing to dev.


pete-watters (approved):
Great work!

PR #1852, opened 2025-11-26
chore: update extension collectibles and refactor packages UI

pete-watters:


pete-watters:
Revisiting this in a clean branch

PR #1854, opened 2025-11-27
fix: collectibles service exports

pete-watters:
This PR fixes an issue introduced by [this PR](https://github.com/leather-io/mono/pull/1846/). This further exposes our vulnerability due to lack of automated UI tests for mobile. 

Collectibles were not loading and giving the error:
```
collectiblesState {"errorMessage": "Found unexpected missing metadata on type \"CollectiblesService\". \"CollectiblesService\" constructor requires at least 3 arguments, found 0 instead.
Are you using @inject, @multiInject or @unmanaged decorators in every non optional constructor argument?
```

Good auld `codex` saved the day and:
- Collectibles DI now works again by switching the service dependencies back to runtime imports, so Inversify can see the constructor
    metadata when resolving CollectiblesService (packages/services/src/collectibles/collectibles.service.ts:1-28).
  - Applied the same fix to the underlying services so their API-client dependencies are also available at runtime, preventing the container
    from tripping when it instantiates them (packages/services/src/collectibles/inscriptions.service.ts:1-44, packages/services/src/
    collectibles/stamps.service.ts:1-42, packages/services/src/collectibles/sip9s.service.ts:1-57).

### DIFF
--- a/packages/services/src/collectibles/collectibles.service.ts
+++ b/packages/services/src/collectibles/collectibles.service.ts
@@ -3,9 +3,9 @@ import { injectable } from 'inversify';
 import { NonFungibleCryptoAsset } from '@leather.io/models';
 
 import type { AccountRequest } from '../types';
-import type { InscriptionsService } from './inscriptions.service';
-import type { Sip9sService } from './sip9s.service';
-import type { StampsService } from './stamps.service';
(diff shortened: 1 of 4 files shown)


pete-watters:
@alexp3y also fixed this [here](https://github.com/leather-io/mono/pull/1855)

PR #1855, opened 2025-11-27
fix: collectibles services type imports

alexp3y:
Fixes bug preventing collectibles service from loading.

Removes type imports from the new decomposed collectibles services.


pete-watters (approved):
Thanks @alexp3y 👍  I'll close [this duplicate](https://github.com/leather-io/mono/pull/1854)

PR #1856, opened 2025-11-27
feat: init queries package

alexp3y:
Introduces the new `@leather.io/queries` as the canonical React Query adapter between apps and `@leather.io/services`.
```
Apps (extension/mobile/web)
↓
@leather.io/queries
↓
@leather.io/services (business logic)
```

Builds on pattern introduced in PR https://github.com/leather-io/mono/pull/1837 to separate out shared query config. Package is distinct from the legacy `@leather.io/query` package in that it contains exclusively RQ configuration and helpers, primarily intended for application-layer service calls.

* Exports fully typed RQ configs
* Implements a simple query key registry + query key factory that 
* Deterministically builds keys, auto-injecting required `UserSettings` dependencies

At the point of usage, apps provide query-specific params + global `UserSettings` object.

* Adds app-specific `useUserSettings()` adapters (extension/mobile/web) to compile local settings state into the shared `UserSettings` contract before calling query configs.
* Fully migrates market data query (extension/mobile/web)... (BTC / Sip10 balance query config also included as references) 

```typescript
import { useQuery } from '@tanstack/react-query';

import { createMarketDataQueryConfig } from '@leather.io/queries';

function useMarketData(asset: FungibleCryptoAsset) {
  const settings = useUserSettings(); // App-specific adapter -> UserSettings
  return useQuery(createMarketDataQueryConfig(asset, settings));
}
```


packages/queries/src/shared/query-key.factory.ts
```diff
@@ -0,0 +1,25 @@
+import type { UserSettings } from '@leather.io/services';
```
pete-watters, line 1: This is a good improvement and will help us have consistency across keys

pete-watters (approved):
Great work @alexp3y 👍  

I am happy to support this as a short / long term solution.

Having `@leather.io/queries` separate from `@leather.io/query` gives us a clear distinction and will help us deprecate `@leather.io/query`  sooner and in a cleaner manner.

kyranjamie (approved):
Commendable effort holding firm on continued RQ adoption. These changes improve things in the sense this is a more organised way to approach the same paradigm, so happy to approve.

But I think should be clear that this does not solve the dependency problem, contrary to what's written in the README. One can still forget to add `settings` to the query key, it's just in a different place now. The opaque disconnect between what injected properties a service call depends on, and those injected properties having to be explicitly defined in the query key remains.

We've had many issues in the past with over zealous queries causing 429s. This makes me skeptical of a solution that will cause _more query refreshes_. Consider that with this pattern of passing in settings to the query key: if service call A depends on `setting.a` only, when `setting.b` changes, we trigger a new request unnecessarily.

alexp3y:
> But I think should be clear that this does not solve the dependency problem, contrary to what's written in the README. 

Thanks @kyranjamie. Agree the wording on the Readme should be more clear on what exactly this solves.

To clarify, the query key registry (`query-key.registry.ts`) is a required central configuration point for all queries and their settings dependencies. When adding a new query, you _have to_ define its key prefix alongside the individual settings dependencies that it will be reactive to (i.e. [`setting.a`, `setting.c`]).

Those settings are then extracted by the query key factory from the full`UserSettings` object supplied by the app. So, there's **no ambiguity for developers using queries** in apps, and **no concern for over-fetching**.

We're not solving the underlying fact that you need to make the settings dependency declarations, **but** this does help highlight and focus the concern. Moving the declaration from individual points of usage to a 1-time registration event in a centralized file, next to all other queries that are declared this way, makes it basically impossible to overlook.

It's an attempt at an incremental improvement on what we already have. Hopefully it helps unblock migration work and gives us an organized pattern to iterate on.

PR #1859, opened 2025-11-27
chore: fix onboarding tests

edgarkhanzadian:
- Fix 1 onboarding test with sbtc mocks
- Make 3 more onboarding tests less flaky

### DIFF
--- a/apps/extension/tests/fixtures/fixtures.ts
+++ b/apps/extension/tests/fixtures/fixtures.ts
@@ -47,6 +47,16 @@ export const test = base.extend<TestFixtures>({
   page: async ({ context }, use) => {
     const pages = context.pages();
     const page = pages.length > 0 ? pages[0] : await context.newPage();
+    await page.addStyleTag({
+      content: `
+        *, *::before, *::after {
+          animation-duration: 0s !important;
+          animation-delay: 0s !important;
+          transition-duration: 0s !important;
+          transition-delay: 0s !important;
+        }
+      `,
+    });
     await use(page);
   },
   extensionId: async ({ context }, use) => {
--- a/apps/extension/tests/mocks/mock-sbtc.ts
+++ b/apps/extension/tests/mocks/mock-sbtc.ts
@@ -1,0 +1,15 @@
+import type { Page } from '@playwright/test';
+
+import { SBTC_EMILY_API_URL } from './constants';
+
+const sbtcEmilyUrl = `${SBTC_EMILY_API_URL}/deposit*`;
+
+export async function mockMainnetTestAccountSbtcDepositRequests(page: Page) {
+  await page.route(sbtcEmilyUrl, route =>
+    route.fulfill({
+      json: {
+        deposits: [],
+      },
+    })
+  );
+}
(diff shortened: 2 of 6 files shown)


apps/extension/tests/mocks/mock-sbtc.ts
```diff
@@ -0,0 +1,17 @@
+import type { Page } from '@playwright/test';
+
+import { SBTC_EMILY_API_URL } from './constants';
+
+const sbtcEmilyUrl = `${SBTC_EMILY_API_URL}/deposit*`;
+
+export async function mockMainnetTestAccountSbtcDepositRequests(page: Page) {
+  await Promise.all([
```
kyranjamie: Why promise all here? There's one promise

edgarkhanzadian: good catch, that's leftover

apps/extension/tests/specs/onboarding/onboarding.spec.ts
```diff
@@ -121,18 +114,63 @@ test.describe('Onboarding an existing user', () => {
...
 
       test('that the wallet generates the correct Taproot address', async ({ homePage }) => {
+        await homePage.page.addStyleTag({
+          content: `
+        *, *::before, *::after {
+          animation-duration: 0s !important;
+          animation-delay: 0s !important;
+          transition-duration: 0s !important;
+          transition-delay: 0s !important;
+        }
```
kyranjamie: If this does work, suggest adding it as a global test env rule rather than per test

edgarkhanzadian:
@kyranjamie still flaky but the probability of passing i'd say is much higher now. I suggest we merge this to help out the team while we look for a better solution here

apps/extension/tests/specs/onboarding/onboarding.spec.ts
```diff
@@ -121,18 +114,24 @@ test.describe('Onboarding an existing user', () => {
 
     test.describe('Bitcoin', () => {
       test('that the wallet generates the correct Native Segwit address', async ({ homePage }) => {
+        test.slow();
```
pete-watters, line 117: Do we still need all of these `test.slow()` even after adding the mocking?

apps/extension/tests/fixtures/fixtures.ts
```diff
@@ -47,6 +47,16 @@ export const test = base.extend<TestFixtures>({
   page: async ({ context }, use) => {
     const pages = context.pages();
     const page = pages.length > 0 ? pages[0] : await context.newPage();
+    await page.addStyleTag({
+      content: `
+        *, *::before, *::after {
```
pete-watters, line 52: Is waiting for the animations causing problems? It would be good to rebase the commits so each one explains the fixes applied as there are a few different things happening:
- mocks added
- animations disabled
- increased delays with `test.slow`

pete-watters (approved):
Nice work @edgarkhanzadian 👍

apps/extension/tests/specs/onboarding/onboarding.spec.ts
edgarkhanzadian: mocking is not connected to these tests. For some reason it is quite performance heavy to open radix dialogs, so i'm letting the test run a little longer to give it a chance to go through the flow

kyranjamie:
image

this is the problem with the flow approach

kyranjamie (approved):

PR #1860, opened 2025-11-28
chore(mobile): update ui native module path in metro config

tigranpetrossian:
Updates the `native.ts` path in metro config for dev mode. This is the fast refresh override, it was only crashing the dev build.


edgarkhanzadian (approved):
nice catch, thank you! @tigranpetrossian

PR #1862, opened 2025-12-01
Swap asset selector improvements

tigranpetrossian:
Few improvements to the Swap asset selector:

* Add missing error state
* Only display search when assets exists or loading (prevent layout shifts)
* Add basic transitions between empty state and the asset list

https://github.com/user-attachments/assets/cf6474c4-8465-44b5-a618-1ea0b76a9cc5


pete-watters (approved):
Nice work 👍

PR #1863, opened 2025-12-01
fix(extension): remove coinbase integration

kyranjamie:
We either need to merge this and kill the integration and launch Onramper, or fix it cc @camerow


pete-watters:
I'll close this, similarly to https://github.com/leather-io/mono/pull/1890 as I think it's not needed just yet

PR #1867, opened 2025-12-02
Mobile app tabs improvements

tigranpetrossian:
Adds couple of basic improvements to the global tab behavior:

1. Scroll up the current screen if the tab button pressed. 
2. Navigate back to the root screen from a sub-route when the tab is pressed.

Both mimic the behavior of most apps using bottom tabs.

https://github.com/user-attachments/assets/c02032bc-a3f0-475c-be0a-77655c2f295e


pete-watters (approved):

pete-watters (approved):
LGTM apart from type error

kyranjamie (approved):
Solid UX improvement

PR #1871, opened 2025-12-02
feat(extension): use dynamic signer to build psbts

alexp3y:
Transaction construction in `extension` uses a single signer by assuming index 0 addresses for all inputs.

Now with xpub-based UTXO lookups, inputs can come from non-zero address indices, causing signing failures when transactions mix UTXOs from different indices.

This PR creates a reusable `useBitcoinSignerFromInput()` hook that:
* Extracts the address index from each input's derivation path
* Returns the correct signer for that sepcific index
* Ensures each input uses its own script in the `witnessUtxo` field

Also, adds signing config support:
* Tx generation returns `BitcoinInputSigningConfig[]` holding correct derivation path info for each input
* Passes config to signing functions

### DIFF
--- a/apps/extension/src/app/pages/send/ordinal-inscription/hooks/use-generate-ordinal-tx.ts
+++ b/apps/extension/src/app/pages/send/ordinal-inscription/hooks/use-generate-ordinal-tx.ts
@@ -12,6 +12,7 @@ import { OrdinalSendFormValues } from '@shared/models/form.model';
 
 import { useCurrentNativeSegwitUtxos } from '@app/query/bitcoin/utxos/utxos.hooks';
 import { useBitcoinScureLibNetworkConfig } from '@app/store/accounts/blockchain/bitcoin/bitcoin-keychain';
+import { useBitcoinSignerFromInput } from '@app/store/accounts/blockchain/bitcoin/bitcoin-signer';
 import { useCurrentAccountNativeSegwitSigner } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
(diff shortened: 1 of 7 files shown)


apps/extension/src/app/pages/send/ordinal-inscription/hooks/use-generate-ordinal-tx.ts
```diff
@@ -146,13 +171,13 @@ export function useGenerateUnsignedOrdinalTx(inscriptionInput: UtxoWithDerivatio
         if (Number(output.value) === 0) return;
 
         if (!output.address) {
-          tx.addOutputAddress(nativeSegwitSigner.address, BigInt(output.value), networkMode);
+          tx.addOutputAddress(changeSigner.address, BigInt(output.value), networkMode);
```
kyranjamie, line 174: This naming could be a bit confusing, given that the second to last part of the derivation path is receive/change.

apps/extension/src/app/components/loaders/btc-balance-loader.tsx
```diff
@@ -19,6 +19,7 @@ export function BtcAssetItemBalanceLoader({
   children,
 }: BtcAssetItemBalanceLoaderProps) {
   const nativeSegwitBalance = useNativeSegwitBtcAccountBalance(accountIndex);
+  console.log('nativeSegwitBalance', JSON.stringify(nativeSegwitBalance, null, 2));
```
kyranjamie: log left over

kyranjamie (approved):
Thanks for doing this Alex. Hope the `signerConfig` stuff made sense. In hindsight, I think we could've used the `bip32Derivation` field and dropped the config entirely, but that's a bigger change.

pete-watters (approved):

PR #1873, opened 2025-12-03
Pressable updates

tigranpetrossian:
Two Pressable updates:

### Remove `legacyTouchablePressEffect`

This is now the baked in default `pressEffect` in `Pressable`. I've initially created this is a temporary crutch when moving away from `TouchableOpacity`, but apparently it's the default preference judging from all the usages in newly added features since then. 
The `pressEffect` functionality will be removed altogether once we're able to update to Reanimated V4 and use the new built-in CSS-like transitions for this.

### Add ReactElement support to Buttons (non-breaking)

The current implementation only allows passing our in-house icons as a **Component**:
```tsx
<Button iconStart={PlusIcon} />
// The button internally decides what size and color to apply to the button based on its variant.
```

This PR adds ability to specify arbitrary elements that will be rendered as is:
```tsx
<Button iconStart={<Box />} />

<Button iconStart={<SpinnerIcon />} />
```

### DIFF
--- a/apps/mobile/src/app/create-new-wallet.tsx
+++ b/apps/mobile/src/app/create-new-wallet.tsx
@@ -11,15 +11,7 @@ import { t } from '@lingui/core/macro';
 import { ImageBackground } from 'expo-image';
 
 import { generateMnemonic } from '@leather.io/crypto';
-import {
-  BlurView,
-  Box,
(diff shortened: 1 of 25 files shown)


pete-watters (approved):
Nice refactor 👍

PR #1875, opened 2025-12-03
ci: organise workflows 🧹

kyranjamie:
I've been touching some actions lately and running extension deploys, and they're a mess: mixed naming conventions and no structure for the part of the codebase they touch. This PR enforces a kebab case (aligned with github docs) for the filenames, workflow jobs, and job names.

### DIFF
--- a/.github/workflows/repo:workflow-checks.yml
+++ b/.github/workflows/repo:workflow-checks.yml
@@ -1,0 +1,48 @@
+name: repo:workflow-checks
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+  push:
+    branches:
+      - main
+      - dev
+    paths:
+      - '.github/workflows/**'
+
+permissions:
+  contents: read
+
+jobs:
+  workflow-names:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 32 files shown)


.github/workflows/repo:workflow-checks.yml
```diff
@@ -0,0 +1,48 @@
...
+    branches:
+      - main
+      - dev
+    paths:
+      - '.github/workflows/**'
+
+permissions:
+  contents: read
+
+jobs:
```
alexp3y, line 17: Maybe these naming convention checks could be consolidated into a single job?

edgarkhanzadian (approved):

alexp3y (approved):
This is awesome. Much needed clean-up. 🧼🦾  Love the naming convention and enforcement of matching filenames to workflows names. Always found it super annoying trying to trace what's running in the GH Actions tab.

.github/workflows/repo:workflow-checks.yml
kyranjamie: Thanks Alex. 

Already merged, but benefit of this is that if one breaks, it's more clear which one has broken in github ui. They're super fast anyway.

PR #1877, opened 2025-12-03
ci: lint broken workflow files

kyranjamie:
I broken the dev extension workflow, fixing here and adding a workflow linter as CI silently ignores broken workflows. I realise we can actually delete this dev build though, we just need the PR build job


tigranpetrossian (approved):
Courtesy approval, don't know how to read or write bash scripts.

PR #1878, opened 2025-12-03
fix(web): hide price fields from portfolio table

pete-watters:
This PR hides **price** and **price change** from Web Portfolio as we don't have price data yet

Screenshot 2025-12-03 at 15 28 16


kyranjamie (approved):
Thanks for taking this on Pete, LGTM

camerow (approved):

PR #1883, opened 2025-12-03
fix(extension): remove Windows-specific Ledger warning

tigranpetrossian:
Removing the Windows callout, since [it's no longer necessary](https://trustmachines.slack.com/archives/C05LAP952E7/p1764790911461339).

**Edit:**
Simply updated the description with instructions.


kyranjamie:
Many users could still be on outdated firmwares. Shouldn't we just change this message to say that, if it's broken, they should try updating?

kyranjamie:
@copilot rather than deleting this warning, instead we want to keep it but adjust the copy and remove the link. We want to explain to the user that, if they're experiencing problems connecting their Ledger they should try updating the firmware to the latest version in Ledger Live

edgarkhanzadian (approved):

tigranpetrossian:
@kyranjamie Copilot went on an overengineering frenzy, so I just updated it in place.

PR #1884, opened 2025-12-04
refactor(ui): update reanimated v4, @gorhom/bottom-sheet

tigranpetrossian:
- Re-work pressEffect to use new CSS Transitions
- Address breaking changes in the bottom sheet library
- Update types for lists

**Update:** This is ready for review, but I'll babysit it for about a week before merging, since it affects two of the most widely used components in the app.

I decided to rework and keep the `pressEffect` because it's still unreasonably cumbersome to implement pressed state transitions. Both the new API and the implementation are vastly simpler:

```tsx
// Simple transition
<Pressable pressEffect={{ opacity: { from: 1, to: 0.8 } }} />

// Custom config
<Pressable pressEffect={{ opacity: { from: 1, to: 0.8, duration: 150, delay: 100, timing: 'ease-out' } }} />

// Transforms are now supported
<Pressable pressEffect={{ transform: { from: [{ scale: 1, rotate: '0deg' }], to: [{ scale: 0.95, , rotate: '7deg' }] } }} />
```


kyranjamie (approved):

kyranjamie:
I see this is still in draft, so a premature ✅

tigranpetrossian:
> I see this is still in draft, so a premature ✅

Going to be for quite some time--many breaking changed to address. This is a non-trivial, but a much coveted upgrade.

edgarkhanzadian:
Looking forward to this!

kyranjamie (approved):

packages/ui/src/components/sheet/sheet-containers.native.tsx
```diff
@@ -32,7 +33,12 @@ function useSheetStyles() {
 
 export function SheetView({ style, ...props }: BottomSheetViewProps) {
   const styles = useSheetStyles();
-  return <BottomSheetView style={[styles.root, styles.contentContainer, style]} {...props} />;
+  return (
+    <BottomSheetView
+      style={StyleSheet.flatten([styles.root, styles.contentContainer, style])}
```
edgarkhanzadian, line 38: Do we need to flatten all of reanimated styles in a new version?

packages/ui/src/components/pressable/pressable.build-press-effect-styles.native.ts
```diff
@@ -0,0 +1,50 @@
...
+  for (const property of pressEffectProperties) {
+    const config = pressEffect[property];
+    if (!config) continue;
+
+    (style[property] as typeof config.to) = isPressed ? config.to : config.from;
+    transitionProperty.push(property);
+    transitionDuration.push(config.duration ?? defaultDuration);
+    transitionTimingFunction.push(config.timing ?? 'ease');
+    transitionDelay.push(config.delay ?? 0);
+  }
```
edgarkhanzadian, line 42: I'd suggest using remeda instead of for...loops:

```suggestion
  R.pipe(
    R.entries(pressEffect),
    R.map(([property, config]) => {
      (style[property] as typeof config.to) = isPressed ? config.to : config.from;
    transitionProperty.push(property);
    transitionDuration.push(config.duration ?? defaultDuration);
    transitionTimingFunction.push(config.timing ?? 'ease');
    transitionDelay.push(config.delay ?? 0);
    }),
  );
```

It also should be possible to skip mutating objects if we return values in the map here and then use fromEntries to reconstruct the object from the array, but that might be too much remeda :D

edgarkhanzadian (approved):
Great work! Added a suggestion for using remeda instead of a for...loop

packages/ui/src/components/sheet/sheet-containers.native.tsx
tigranpetrossian: No not at all. This is just addressing a breaking change in BottomSheetView & friends. It used to flatten internally, but stopped in one of the recent minor updates ¯\_(ツ)_/¯. Nothing else has to be done on our end.

packages/ui/src/components/pressable/pressable.build-press-effect-styles.native.ts
tigranpetrossian: I actually had a little more elaborate fluent expression here, as the initial draft supported many more properties I ended up removing. Remeda is great for declarative transformations, but this is one of the places where it improve much if at all. I wouldn't worry about small local mutations either--they're harmless.

tigranpetrossian:
I've been test driving this for few days now. It's ready to merge. Going to hold to do it after the current release just in case. cc @edgarkhanzadian

PR #1890, opened 2025-12-05
fix(extension): Coinbase onramp

pete-watters:
This PR aims to restore previous functionality we have where users could fund via Coinbase Pay. 

- [x] Recovered previous 'Fund' page code - restoring Coinbase Pay and others
- [x] Added On-ramper as a new option there  
- [x] updated the Coinbase pay library `@coinbase/cbpay-js` 
- [x] fixed error with `destinationWallets` – now we’re using addresses/assets per the latest cbpay-js definitions and docs
- [x] fixed invalid addresses error 
- [x] added possibility to pass a token - `COINBASE_SESSION_TOKEN`  

**Outstanding**
We still get and error for secure initialization as we need to either  supply a `sessionToken` from the BE or else disable `“Require secure initialization”` in our settings. 

**Demo**

Here's a demo of the UI showing both options. I'm not sure exactly what we want to show but we need to first urgently fix this issue. 

On-ramper is not working for me locally either which relates to [this issue I think](https://trustmachines.slack.com/archives/C05LAP952E7/p1764871901879639)

https://github.com/user-attachments/assets/fdea5adc-f33d-4993-a716-6865696e859e

### DIFF
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -49,7 +49,7 @@
     "@bitcoinerlab/secp256k1": "1.0.2",
     "@bitflowlabs/core-sdk": "2.4.0",
     "@blockstack/stacks-transactions": "0.7.0",
-    "@coinbase/cbpay-js": "2.1.0",
+    "@coinbase/cbpay-js": "2.4.0",
     "@fungible-systems/zone-file": "2.0.0",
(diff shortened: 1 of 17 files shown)


pete-watters:
I'll close this as it seems not as urgent as first thought. We can re-visit once Coinbase KYC is completed

PR #1893, opened 2025-12-06
fix(extension): remove deprecated sbtc reprocessing status

tigranpetrossian:
Removes deprecated reprocessing status and add rbf status addressing breaking Emily API changes.

The updated schema is [here](https://github.com/stacks-sbtc/sbtc/blob/7ab8ed3fcbfd3811b4799e8de493defabae6738f/emily/openapi-gen/generated-specs/public-emily-openapi-spec.json#L2090).

While at it, also tweaked the query retry logic. This should eliminate unnecessary retries for irrelevant errors like 4**.

### DIFF
--- a/apps/extension/src/app/components/sbtc-deposit-status-item/sbtc-deposit-status-item.tsx
+++ b/apps/extension/src/app/components/sbtc-deposit-status-item/sbtc-deposit-status-item.tsx
@@ -15,12 +15,13 @@ import { TransactionItemLayout } from '../transaction-item/transaction-item.layo
 function getDepositStatus(status: SbtcStatus) {
   switch (status) {
     case 'pending':
-    case 'reprocessing':
       return 'Pending deposit';
     case 'accepted':
       return 'Pending mint';
     case 'failed':
       return 'Failed';
+    case 'rbf':
+      return 'Replaced';
     case 'confirmed':
     default:
       return '';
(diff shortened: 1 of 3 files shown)


apps/extension/src/app/components/sbtc-deposit-status-item/sbtc-deposit-status-item.tsx
```diff
@@ -15,12 +15,13 @@ import { TransactionItemLayout } from '../transaction-item/transaction-item.layo
...
   switch (status) {
     case 'pending':
-    case 'reprocessing':
       return 'Pending deposit';
     case 'accepted':
       return 'Pending mint';
     case 'failed':
       return 'Failed';
+    case 'rbf':
+      return 'Replaced';
```
tigranpetrossian, line 24: "Replaced" or "Being replaced"?

kyranjamie (approved):

apps/extension/src/app/components/sbtc-deposit-status-item/sbtc-deposit-status-item.tsx
tigranpetrossian: @kyranjamie Does this copy look right to you? I'm on the fence because "Replaced" sounds like it has already taken place, and I'm not sure if that's the case when the API is responding with this. There isn't much documentation around this.

edgarkhanzadian (approved):

PR #1898, opened 2025-12-08
fix(extension): show submitted Stacks transactions immediately

fabric-8:
Shows submitted Stacks transactions immediately in the activity list instead of waiting for them to appear in the mempool.

- Display submitted transactions at top of activity list until confirmed
- Updated mempool/transaction queries to refetch properly when transactions exist
- Modified both extension and query packages

### DIFF
--- a/apps/extension/src/app/features/activity-list/activity-list.tsx
+++ b/apps/extension/src/app/features/activity-list/activity-list.tsx
@@ -1,16 +1,21 @@
-import { useCallback } from 'react';
+import { useCallback, useEffect, useMemo, useRef } from 'react';
 import { Outlet } from 'react-router';
 import { Virtuoso } from 'react-virtuoso';
 
 import { type ActivityView } from '@leather.io/features';
 
+import { safelyFormatHexTxid } from '@app/common/utils/safe-handle-txid';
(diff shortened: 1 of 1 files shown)


pete-watters:
Thanks @fabric-8 👍  This looks like a good start to me. I will do some tests now and can build upon it to help get it implemented

pete-watters (commented):
This looks good so far @fabric-8 👍  To get this mergable you will need to run prettier on the code - `pnpm format` from `mono` route should do the trick. 

I will need to do manual tests on this but I think a better way could be to have `SubmittedTransactionList` included in the main activity list. `SubmittedTransactionList` is one of the older components anyway so I'm not sure if it will match the new style. 

I'm not sure about manually changing the `refetchInterval` based on `hasSubmittedTxs`. We can probably do this a cleaner way by using `hasSubmittedTxs` or similar as part of the query key and having the `submittedTxs` as part of the normal activity feed but kept sorted at the top.

I'm not familiar with the code for submitted TXs but I can dig into it soon and integrate this along with other activity improvements

edgarkhanzadian:
We keeping this or should we close this?

pete-watters:
I took it into my larger PR, so we can close this off

PR #1900, opened 2025-12-08
feat(services): add swap quote execution constraints

alexp3y:
Adds an `isExecutable` flag to the `SwapQuote` type as an authoritative determination on whether a quote can be executed or not.

This allows for valid quotes that may not be executable due to constraint violations. (e.g. sBTC limit checks)

This PR Also:
* Adds an `executionConstraints[]` array to `SwapQuote`, describing any constraints affecting executability. Each constraint has a reason code and the underlying amount `threshold` that was violated.
* Checks the sBTC limits via the Emily API were implemented in the `SbtcBridgeSwapProvider` and influence the executability of its swap quotes.
* Enforces the `isExecutable` flag in `SwapService.getExecutionData`
* Makes all Swap Model type fields final (readonly)

### DIFF
--- a/packages/services/src/infrastructure/api/emily/emily-api.client.ts
+++ b/packages/services/src/infrastructure/api/emily/emily-api.client.ts
@@ -1,0 +1,35 @@
+/* eslint-disable func-style */
+import axios from 'axios';
+import { inject, injectable } from 'inversify';
+
+import { Types } from '../../../inversify.types';
+import type { HttpCacheService } from '../../cache/http-cache.service';
+import { selectBitcoinNetworkMode } from '../../settings/settings.selectors';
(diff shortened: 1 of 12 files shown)


packages/services/src/infrastructure/api/emily/emily-api.client.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable func-style */
```
pete-watters, line 1: Maybe you can refactor `fetchFn` to be a `function` component instead?

pete-watters (approved):

tigranpetrossian (approved):
Thanks so much Alex, I'll start working on a UI to present this.

PR #1903, opened 2025-12-09
feat(extension): update collectibles and add details screens                         

pete-watters:
This PR follows the path paved by `activity` to update collectibles in the `extension` following a similar pattern

✨ Features:

- [x] `CollectibleView` type introduced to have a shared interface for UI code
- [x] Collectible UI components refactored out of `packages/ui` and placed in their respective apps 
- [x] legacy support for `extension` maintained 
- [x] added token details for collectibles + tokens on extension
- [x] ported 'Send Inscription' over and simplified to not use route params  
- [x] added loading states for home screens in new extension  

🚧  More work:

- []  Pre-select `Send` token when going from token details page
- [] Improve `Receive` dialog. I think we should make this a non-routable dialog to avoid the bugs we have with background content for routable dialogs

**NOTE** - the gamma API has been giving 500 errors so we are showing fallback UI for SIP-9 tokens right now using hiro meta data. This needs to be investigated further

Here's a demo:

https://github.com/user-attachments/assets/66ea59b2-a522-4dc7-87a0-6b407eaeea30


package.json
```diff
@@ -74,7 +74,7 @@
     "yarn": "please-use-pnpm"
   },
   "keywords": [],
-  "packageManager": "pnpm@10.18.3+sha512.bbd16e6d7286fd7e01f6b3c0b3c932cda2965c06a908328f74663f10a9aea51f1129eea615134bf992831b009eabe167ecb7008b597f40ff9bc75946aadfb08d",
+  "packageManager": "pnpm@10.26.0+sha512.3b3f6c725ebe712506c0ab1ad4133cf86b1f4b687effce62a9b38b4d72e3954242e643190fc51fa1642949c735f403debd44f5cb0edd657abe63a8b6a7e1e402",
```
kyranjamie: Please update pnpm in another PR, will make it easier to identify if this changes behaviours

pete-watters: Thanks 👍  I've added the update to [this PR](https://github.com/leather-io/mono/pull/1965) instead

apps/extension/src/app/features/collectibles/components/bitcoin/inscription-html.tsx
```diff
@@ -0,0 +1,79 @@
...
+      title={title}
+    >
+      <Box
+        bg="ink.text-primary"
+        color="ink.background-secondary"
+        height="100%"
+        overflow="hidden"
+        p="space.04"
+        width="100%"
+        dangerouslySetInnerHTML={{ __html: sanitizeContent(query.data) }}
```
kyranjamie: Is this new code, or moved? I'm not comfortable with this. This is non-iframed HTML injected straight into the DOM. Sanitized yes but holes can be found in sanitization libraries.

pete-watters: This is new and you are right, it needs to go - good catch. 

This is still a working draft and next on my list to come back to so I will get rid of it. 

It comes from trying to use the new mobile code in the extension more quickly. I will change it so it is safer and uses an iframe

pete-watters: I've updated this to use the `<Iframe` component we already have instead. Thanks @kyranjamie !

apps/extension/src/app/features/send-inscription/send-inscription-dialog.tsx
```diff
@@ -0,0 +1,108 @@
+import { useState } from 'react';
```
pete-watters, line 1: @kyranjamie  I am changing this dialog in the new version to avoid passing route state. 

I think now is a good time to check - is it OK to remove other routable dialogs e.g. - `Receive`? Now we have more places to open it from its a real mess having it work from each path and still have the correct background location.

What do you think about dumbing it down so it's no longer routable?

kyranjamie:
image 🙃

apps/extension/src/app/features/send-inscription/send-inscription-dialog.tsx
kyranjamie: The idea of using route state for this is that you can basically call the component from anywhere.

So long as that's still the case, and there isn't more state juggling as a result, I'm fine with that. 


If we remove routing entirely, suggest use of react-call

apps/extension/src/app/features/activity-list/activity-list.tsx
```diff
@@ -1,16 +1,37 @@
...
 import { ActivityItem } from './components/activity-item';
 import { ActivityListLayout } from './components/activity-list.layout';
+import { createSubmittedActivityViews } from './submitted-activity-view';
+
+type ActivityListRow =
+  | ActivityView
+  | {
+      key: string;
+      kind: 'sbtc-deposit';
+      deposit: import('@app/query/sbtc/sbtc-deposits.query').SbtcDeposit;
```
kyranjamie, line 27: unneeded inline import

apps/extension/src/app/features/activity-list/activity-list.tsx
```diff
@@ -1,16 +1,37 @@
...
+    };
+
+function isSbtcDepositRow(
+  item: ActivityListRow
+): item is Extract<ActivityListRow, { kind: 'sbtc-deposit' }> {
+  return (item as any).kind === 'sbtc-deposit';
```
kyranjamie, line 33: do we have to cast here?

apps/extension/src/app/features/collectibles/components/bns.web.tsx
```diff
@@ -0,0 +1,42 @@
...
+    </Box>
+  );
+
+  if (onPress) {
+    return (
+      <CollectibleCard height={height}>
+        <button
+          type="button"
+          onClick={onPress}
+          style={{ border: 0, padding: 0, margin: 0, background: 'transparent', width: '100%' }}
```
kyranjamie, line 33: styled.button

apps/extension/src/app/features/collectibles/components/collectible-audio.web.tsx
```diff
@@ -0,0 +1,62 @@
...
+            </styled.span>
+          </Box>
+        </button>
+      </CollectibleCard>
+    );
+  }
+
+  return (
+    <CollectibleCard height={size}>
+      <audio
```
kyranjamie, line 50: styled.audio

apps/extension/src/app/features/collectibles/components/collectible-audio.web.tsx
```diff
@@ -0,0 +1,62 @@
...
+  alt: string;
+  size?: number;
+  onPress?(): void;
+}
+
+export function CollectibleAudio({ src, alt, size = 200, onPress }: CollectibleAudioProps) {
+  if (onPress) {
+    return (
+      <CollectibleCard height={size}>
+        <button
```
kyranjamie, line 18: styled.button

apps/extension/src/app/features/collectibles/components/collectible-html.web.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: CollectibleHtmlProps) {
+  const [hasError, setHasError] = useState(false);
+  const showFallback = hasError || !src;
+
+  if (showFallback) {
+    return <ImageUnavailable height={height} />;
+  }
+
+  const iframe = (
+    <iframe
```
kyranjamie, line 27: styled.iframe

apps/extension/src/app/features/collectibles/components/collectible-html.web.tsx
```diff
@@ -0,0 +1,55 @@
...
+        pointerEvents: onPress ? 'none' : undefined,
+      }}
+      onError={() => setHasError(true)}
+    />
+  );
+
+  if (onPress) {
+    return (
+      <CollectibleCard height={height}>
+        <button
```
kyranjamie, line 43: styled.button

kyranjamie:
In future it'd be great to have some smaller PRs. Token details for instance could've been done separately to the home page / introduction of tabs?

pete-watters:
> image 🙃

Ah yikes sorry about that 😭

pete-watters:
> In future it'd be great to have some smaller PRs. Token details for instance could've been done separately to the home page / introduction of tabs?

Yes I know and sorry about that, I just felt good momentum working on token details and wanted to try get all this done sooner. I can split it up into smaller chunks

pete-watters:
@kyranjamie - I'm going to work on splitting this up. If you are mid review please continue but otherwise I will open several smaller PRs. 

I struggle with the balance sometimes as if I have good momentum I like to keep going but I appreciate it's a nightmare to review for you then. 

It also seems tricky to get a review at all so sometimes it's easier to get one on a big PR than a few smaller ones

edgarkhanzadian:
Hello! I've started the review of the extension itself, would wait on the smaller PRs to review the code.

I've also included some issues connected to activity but it might not have been a part of this PR.

I'm getting an issue with the activity view here.

https://github.com/user-attachments/assets/753dd299-0ea4-418d-9d81-a5a60bc17f0f

Activity of token details have weird padding


https://github.com/user-attachments/assets/6860edcf-c696-47ab-8a60-864e86412dcf

Seems like some sip10s don't have activity? 


https://github.com/user-attachments/assets/7df67b0c-b81f-4fe1-b556-72f74eef20da

Seems like some nfts have an issue with rendering:

nft

on longer header titles we i'd suggest we center the header

centerHeader

Activity tab has a different padding compared to tokens tab


https://github.com/user-attachments/assets/49525e34-b989-446d-b4bc-7b425101de54


Details screen seems narrower:

https://github.com/user-attachments/assets/b98c5ad8-d9aa-4459-b4fb-8c719bb99222

kyranjamie (commented):
Nice @pete-watters, there's an impressive amount of work here.

I will say, this PR is currently very hard to review given how large it is, and there are a few too many obviously vibe-coded elements to it. Could txs details not have been done in a separate PR?

apps/extension/src/app/features/asset-list/bitcoin/btc-crypto-asset-item/btc-crypto-asset-item.tsx
```diff
@@ -10,15 +13,29 @@ interface BtcCryptoAssetItemProps {
...
   isLoading,
   onSelectAsset,
+  onOpenToken,
   isLoadingAdditionalData,
 }: BtcCryptoAssetItemProps) {
   const isPrivate = useIsPrivateMode();
 
+  function handleSelectAsset(symbol: string) {
+    if (onOpenToken) {
+      onOpenToken({
```
tigranpetrossian, line 29: What's the difference between `onOpenToken` and `onSelectAsset`? There's both a naming clash (token/asset) and an uncertainty of what happens upon selection.

apps/extension/src/app/features/asset-list/bitcoin/btc-crypto-asset-item/btc-crypto-asset-item.tsx
```diff
@@ -10,15 +13,29 @@ interface BtcCryptoAssetItemProps {
...
+        assetId: serializeAssetId(getAssetId(btcAsset)),
+      });
+      return;
+    }
+    onSelectAsset?.(symbol);
+  }
+
+  const onSelectAssetProp = onOpenToken || onSelectAsset ? handleSelectAsset : undefined;
```
tigranpetrossian, line 37: Doesn't this effectively make `handleSelectAsset` redundant? The first `if` clause of the handler will never run.

apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx
```diff
@@ -32,20 +36,29 @@ export function RunesAssetList({
   });
   const { isEnabled } = useManagedRunesTools(accountIndex);
 
+  const hasRunes = runes.state === 'success' && runes.value && runes.value.runes.length > 0;
+
   useEffect(() => {
-    if (runes.value && runes.value.runes.length > 0 && setHasManageableTokens) {
+    if (hasRunes && setHasManageableTokens) {
```
tigranpetrossian, line 42: Not part of this PR, since this is just a refactor, but this `useEffect` points at a design flaw, runes are defined internally and trigger an upper-level state change + an extra render. Instead, `hasManageableTokens` can be just a derived flag based on `hasRunes`.

apps/extension/src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-item.tsx
```diff
@@ -62,7 +76,7 @@ export function Sip10TokenAssetItem({
         titleLeft,
         fiatBalance: formatCurrency(balance.quote.availableBalance),
         dataTestId: assetId,
-        onSelectAsset,
+        onSelectAsset: onOpenToken || onSelectAsset ? handleSelectAsset : undefined,
```
tigranpetrossian, line 79: Same issue as with the other list. `handleSelectAsset` will never run if any of the props exist, yet it references these props. The fact that this logic is duplicated in multiple places indicates a design issue as well.

apps/extension/src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
```diff
@@ -37,13 +40,15 @@ export function Sip10TokenAssetList({
   });
   const { isEnabled } = useManagedSip10Tools(accountIndex);
 
+  const hasSip10s = sip10s.state === 'success' && sip10s.value && sip10s.value.sip10s.length > 0;
+
   useEffect(() => {
-    if (sip10s.value && sip10s.value.sip10s.length > 0 && setHasManageableTokens) {
+    if (hasSip10s && setHasManageableTokens) {
```
tigranpetrossian, line 46: Same comment as with `runes-asset-list.tsx`

apps/extension/src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.tsx
```diff
@@ -31,6 +36,18 @@ export function StxCryptoAssetItem({
...
+    if (onOpenToken) {
+      onOpenToken({
+        assetId: serializeAssetId(getAssetId(stxAsset)),
+      });
+      return;
+    }
+    onSelectAsset?.(symbol);
+  }
+
+  const onSelectAssetProp = onOpenToken || onSelectAsset ? handleSelectAsset : undefined;
```
tigranpetrossian, line 49: Same as https://github.com/leather-io/mono/pull/1903/changes#diff-458c2a10e45939568c57bbfa678f74612522c0abbd09e4ccc7303b8541dc3905R29

pete-watters:
> Nice @pete-watters, there's an impressive amount of work here.
> 
> I will say, this PR is currently very hard to review given how large it is, and there are a few too many obviously vibe-coded elements to it. Could txs details not have been done in a separate PR?

Thanks @kyranjamie , I am splitting it up as we speak and will clean up as I go. I was trying to move faster here so we can get the bulk of this onboarding rework over with sooner and move on with priorities. 

I'll close it off for now and re-open more PRs better grouped together and with improvements. Thanks for checking it 👍

pete-watters:
> **NOTE** - the gamma API has been giving 500 errors so we are showing fallback UI for SIP-9 tokens right now using hiro meta data. This needs to be investigated further

Thanks @edgarkhanzadian - I'll check out these issues now. 

For the STX collectibles I added this note in the description as Gamma has been giving me 500 errors - `NOTE - the gamma API has been giving 500 errors so we are showing fallback UI for SIP-9 tokens right now using hiro meta data. This needs to be investigated further` 

I think most of the activity related issues are solved by https://github.com/leather-io/mono/pull/1937 but I will recheck what you mention to make sure

apps/mobile/src/features/token/collectibles-list.tsx
```diff
@@ -4,40 +4,40 @@ import { useWindowDimensions } from 'react-native';
...
-  item: NonFungibleCryptoAsset;
+  item: CollectibleView;
   height: number;
   onPress?(tokenDetails: TokenDetailsProps): void;
 }
 function renderCollectible({ item, height, onPress }: RenderCollectibleProps) {
-  const collectible = (() => {
-    switch (item.protocol) {
+  const asset = item.asset;
+  const content = (() => {
```
tigranpetrossian, line 26: IIFE+switch seems like an overkill, a simple object would work just fine here.

packages/features/src/formatting.ts
```diff
@@ -0,0 +1,38 @@
...
+    day: 'numeric',
+  });
+}
+
+/**
+ * Formats a Unix timestamp (seconds) to a date-time string with UTC.
+ * @param timestamp - Unix timestamp in seconds
+ * @returns Formatted date-time string like "2024-01-15 14:30 UTC"
+ */
+export function formatTimestampWithTime(timestamp: number): string {
```
tigranpetrossian, line 30: Out of my own curiosity, is there a reason we want to display UTC time here instead of users time zone?
Regardless, since we already use dayjs, this could probably be expressed in a simpler manner (untested LLM pseudo-code):
```
dayjs.unix(timestamp).utc().format('YYYY-MM-DD HH:mm [UTC]')
```

tigranpetrossian (changes requested):
Hey @pete-watters,

I'm about half-way through reviewing this. The collectibles related code in `features` looks good: well-decoupled and covered, but the PR itself is extremely large (to the point  where Github is refusing to show more than one file), and contains cross-feature & cross-package changes making it really hard to reason about.
Seeing large changes to activity across 3 packages (along with the another pending activity PR), lots of minor unrelated changes in the test app, some market history queries, token details, etc.

I'd suggest breaking this apart into multiple smaller PRs we can review one by one. Especially with `features` being a new addition we're still assessing, a little bit of careful scrutiny here is given, so that we don't end up with more debt than we're able to handle.

pete-watters:
Thanks for the reviews @tigranpetrossian @kyranjamie @edgarkhanzadian . In hindsight I should have spent extra time splitting up the work first before asking for review but I got caught up in eagerness to get this done. 

I have split the work up into 6 smaller PRs now:

1. Shared utilities in features package: https://github.com/leather-io/mono/pull/1981 
2. Collectibles UI component refactor: https://github.com/leather-io/mono/pull/1982 
3. Token details screens: https://github.com/leather-io/mono/pull/1984
4. Send inscription dialog: https://github.com/leather-io/mono/pull/1985 
5. Activity & SBTC integration: https://github.com/leather-io/mono/pull/1986

I will get them reviewed and staggered in separately in the order above and will apply feedback/ fixes suggested here

PR #1904, opened 2025-12-09
chore: expose clientId for werner

pete-watters:
This PR:

- manually un-gates the new activity UI pending client ID setup
- exposes clientID in the UI settings so we can get it for launch darkly
  
Screenshot 2025-12-09 at 09 31 36

### DIFF
--- a/apps/extension/src/app/components/launchdarkly-client-id.tsx
+++ b/apps/extension/src/app/components/launchdarkly-client-id.tsx
@@ -1,0 +1,24 @@
+import { useMemo } from 'react';
+
+import { Box, styled } from 'leather-styles/jsx';
+
+import { getClientId } from '@app/common/client-id';
+
+export function LaunchDarklyClientId() {
+  const clientId = useMemo(() => getClientId(), []);
+
+  if (!clientId) return null;
+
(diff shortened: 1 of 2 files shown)


apps/extension/src/app/components/launchdarkly-client-id.tsx
```diff
@@ -0,0 +1,24 @@
+import { useMemo } from 'react';
+
+import { Box, styled } from 'leather-styles/jsx';
+
+import { getClientId } from '@app/common/client-id';
+
+export function LaunchDarklyClientId() {
+  const clientId = useMemo(() => getClientId(), []);
```
kyranjamie, line 8:
```suggestion
const clientId = getClientId();
export function LaunchDarklyClientId() {
```

useMemo and useCallback are a plague

kyranjamie (approved):

apps/extension/src/app/components/launchdarkly-client-id.tsx
pete-watters: Thanks. I don't plan to merge this, it's just a quick way to get the client Ids

PR #1907, opened 2025-12-09
chore(extension): enforce camel casing of feature flags

pete-watters:
This PR enforces `camelCasing` for LD feature flags in the `extension` as suggested by @tigranpetrossian  [here](https://github.com/leather-io/mono/pull/1824#discussion_r2540885666). 

I've:
- enforced it in the LD `leather-extension` settings
- added new camel cased versions of our 2 flags - `extensionRevamp` and `releaseOnramperBuy` 
- updated our app code
    - flags themselves
    -  `reactOptions: { useCamelCaseFlagKeys: true }` 

Screenshot 2025-12-09 at 11 49 50

Screenshot 2025-12-09 at 11 48 35

### DIFF
--- a/apps/extension/src/app/features/feature-flags/index.ts
+++ b/apps/extension/src/app/features/feature-flags/index.ts
@@ -15,12 +15,12 @@ export function createLDProvider() {
       kind: 'clientId',
       key: getClientId(),
     },
-    reactOptions: { useCamelCaseFlagKeys: false },
+    reactOptions: { useCamelCaseFlagKeys: true },
   });
 }
 
 interface FeatureFlags {
-  release_onramper_buy: boolean;
+  releaseOnramperBuy: boolean;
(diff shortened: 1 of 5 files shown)


tigranpetrossian:
I was under the impression that you could just do `useCamelCaseFlagKeys: true` without having to do anything in the LD itself. Was that not the case?

kyranjamie (approved):
Great tidying up Pete

edgarkhanzadian:
Yeap, as far as i know there is no need to change the flag itself to camel case. Switching useCamelCaseFlagKeys to true would return camel cased flags (even if the flag itself is_like_this)

pete-watters:
@tigranpetrossian @edgarkhanzadian - when I changed this setting my existing feature flag wasn't working anymore so I decided to just change it on LD and it worked when I did.

PR #1919, opened 2025-12-15
fix: svg logo

edgarkhanzadian:
SVGO removes viewBoxes as part of svg optimization by default, but that also breaks svgs in our setup.
This PR turns off that feature.
 
welcomeScreen
home

### DIFF
--- a/packages/ui/tsdown.config.web.ts
+++ b/packages/ui/tsdown.config.web.ts
@@ -5,7 +5,21 @@ import { defineConfig } from 'tsdown';
 export default defineConfig({
   entry: ['exports.web.ts'],
   plugins: [
-    svgr({ typescript: true, dimensions: false }),
+    svgr({
+      typescript: true,
+      svgoConfig: {
+        plugins: [
+          {
+            name: 'preset-default',
+            params: {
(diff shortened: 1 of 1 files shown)


tigranpetrossian:
Hey @edgarkhanzadian, are you sure the `viewBox` was the culprit? Just checked the extension with disappeared logo and the `viewBox` was present on it.

Perhaps the added preset fixed whatever the issue was (I suspect lack of `width` and `height`).
Worth figuring out what goes into the preset and normalizing it across platform configs.

image

edgarkhanzadian:
@tigranpetrossian with `dimensions: false` it strips out width and height, but leaves viewBox in.
with `dimensions: true` (default config behavior), it strips out viewBox but leaves width and height.
We need both for the svgs to work correctly, so `dimensions: true` with `removeViewBox: false` worked correctly for me.

The released version has `dimensions: false`

fabric-8 (approved):
Great stuff, thanks for the quick fix @edgarkhanzadian !

kyranjamie (approved):

pete-watters (approved):

PR #1920, opened 2025-12-15
Refactor/remove account ordering

kyranjamie:
This PR removes some unused ordering logic (discussed as unneeded behaviour), and simplifies the account state slice by deriving unset state, rather than defaulting to unset values and persisting them on the store. The result is that the account store only requires a single property per entry, and everything else is inferred if not set. 

The behaviour of deriving icon from the account id leads to a possible immediate repeat icon over the previous implementation, but there's a one-in-twenty chance of this so don't think it's real problem. Motivation for this change is just to have the raw account state as minimal as possible, and derive anything that can be, to shift complexity away from the store and keep it as easy to reason about as possible.


pete-watters (approved):
Great work and some nice code improvements along the way. 

I'm all for removing these un-necessary complications from adding too complex features before we are ready for them

apps/mobile/src/store/accounts/utils.ts
```diff
@@ -22,19 +22,34 @@ export const accountIcons = [
...
+
+export function deriveIconFromAccountId(accountId: string): AccountIcon {
+  let hash = 0;
+  for (let i = 0; i < accountId.length; i++) {
+    const char = accountId.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash = hash & hash;
+  }
+  const index = Math.abs(hash) % accountIcons.length;
+  const icon = accountIcons[index];
```
edgarkhanzadian, line 57: Maybe we can use @noble/hashes here?
```suggestion
  const index = new DataView(sha256(accountId).buffer).getUint32(0) % accountIcons.length;
  const icon = accountIcons[index];
```

edgarkhanzadian (approved):

apps/mobile/src/store/accounts/utils.ts
kyranjamie: I thought about this as well, is a cleaner implementation, it's way slower though

```
 LOG  Hashing 1000 strings took 2.1094159483909607ms
 LOG  SHA256 Hashing 1000 strings took 284.6787919998169ms
```

kyranjamie: blake3 is faster but still ~170ms

PR #1928, opened 2025-12-17
fix: adjust CI build process and fix Android build

pete-watters:
I've made some CI changes to fix the android build and adjust what we are building and when.


apps/mobile/package.json
```diff
@@ -9,7 +9,7 @@
     "android:release": "expo run:android --variant release",
     "build": "pnpm lingui",
     "check:cocoapods": "sh scripts/enforce-cocoapods-version.sh",
-    "eas-build-post-install": "cd ../.. && pnpm install && pnpm build",
+    "eas-build-post-install": "cd ../.. && pnpm install && pnpm build:mobile",
```
pete-watters, line 12: @camerow @edgarkhanzadian I got the mobile build working but it's unclear if I need to run a full `pnpm build` here or if `pnpm build:mobile` will do? 

It seems like we are building all packages via `build` which includes web + extension. Maybe I need to be granular here and just build mobile dependancies?

pete-watters: I checked this out and it seems that `pnpm build:mobile` is fine here as in `turbo.json`, the build task has `"dependsOn": ["^build"], meaning for any package you build, Turbo also runs build for its dependency graph.

I've confirmed this manually running and checking the command

camerow (approved):

PR #1929, opened 2025-12-17
Add execution constraints support to swap

tigranpetrossian:
Adds support for swap execution constraints introduced in #1900 

1. Ensure the selected quote is always executable
2. Attach constraints to the quote selection result if none of the quotes are executable
3. Add a new status `useLiveSwapEstimate` for handling this, reflect in the UI.

Also added some tests for `useLiveSwapEstimate` to cover possible permutations. I'm not particularly pleased with injecting query object fixtures into it for testing, but I'm planning on moving this hook into the swap state manager, since it's useful enough to re-use in the extension. When that's done, those fixtures will no longer be necessary.

### DIFF
--- a/apps/mobile/src/features/swap/components/quote-preview/quote-preview-constrained.tsx
+++ b/apps/mobile/src/features/swap/components/quote-preview/quote-preview-constrained.tsx
@@ -1,0 +1,61 @@
+import { formatCurrency } from '@/utils/currency-formatter';
+import { t } from '@lingui/core/macro';
+
+import { ExecutionConstraint, SwappableFungibleCryptoAsset } from '@leather.io/models';
+import { Box, Text } from '@leather.io/ui/native';
+import { assertUnreachable } from '@leather.io/utils';
+
+interface QuotePreviewConstrainedProps {
+  constraints: ExecutionConstraint[];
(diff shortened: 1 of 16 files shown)


tigranpetrossian:
Tests are failing but the PR is still reviewable. `pre-push` didn't run for me locally so I missed the fixture updates, on it.

kyranjamie (approved):

apps/mobile/src/features/swap/components/quote-preview/quote-preview-constrained.tsx
```diff
@@ -0,0 +1,61 @@
...
+      return {
+        title: t`Amount too large`,
+        description: t`${operation} ${baseAssetSymbol} to ${targetAssetSymbol} is limited to ${threshold}`,
+      };
+    default:
+      assertUnreachable(constraint.reason);
+  }
+}
+
+function getOperationLabel(base: string, target: string): string {
```
tigranpetrossian, line 57: @alexp3y Took a shortcut here to adjust some copy based on pair. This might be a good candidate for additional metadata on `SwapQuote`. I'm slowly making a list of those as I encounter them, so we can look through all at once at some point.

PR #1932, opened 2025-12-17
fix: use rq type narrowing

kyranjamie:
This PR demos exactly the same type narrowing functionality, without a wrapper function. 

It's identical except `state` becomes `status` and `value` becomes `data`.


edgarkhanzadian (commented):
definitely a +1 on this one

kyranjamie:
Closing for now, but hopefully we can take this on at some point soon, AI should be able to fairly reliably remove this cc @edgarkhanzadian

PR #1934, opened 2025-12-17
feat(mobile): improve target asset display ordering swap

tigranpetrossian:
Introduces a basic priority based ordering to make the asset selector a bit more presentable in zero states.

### DIFF
--- a/apps/mobile/src/features/swap/swap-state/utils/asset-selection.ts
+++ b/apps/mobile/src/features/swap/swap-state/utils/asset-selection.ts
@@ -2,13 +2,35 @@ import { filter, map, pipe, sortBy } from 'remeda';
 
 import { btcAsset, stxAsset } from '@leather.io/constants';
 import {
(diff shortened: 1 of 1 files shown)


edgarkhanzadian (approved):
I wonder if it would make sense in the future to have the priority list served from services config? Same way we are doing `defaultEnabled` assets: https://github.com/leather-io/workers/blob/3b2688f648cb4cab7dc3ddc0918c3b4d19b5a0d0/workers/api-gateway/src/app-config/app-config.ts#L78-L117

tigranpetrossian:
Definitely need to centralize this in some way, although use-cases slightly vary in different features, so we'll need to do a little thinking there or find a middle ground that works in most places. <br><br>@alexp3y [is looking into adding an asset listing service](https://linear.app/leather-io/issue/LEA-3319/asset-list-services), that might be the best place to do this sort of thing.

PR #1935, opened 2025-12-17
feat(extension): usdcx at top of asset list

kyranjamie:
This PR sets USDCx as a token fixed third in the list.

I'm not in love with these changes. Mocking the empty balance response value is pretty ugly, but not really sure how to otherwise satisfy the type the child component. Feedback welcome.

image


alexp3y:
@kyranjamie There's an existing service call that helps address your concern. `Sip10BalancesService.getSip10BalanceByAssetId` retrieves a specific token balance and will construct / return an empty balance using the user's quote currency setting.

Added a new commit that swaps in a USDCx balance hook to the loader. The hook uses the service call under-the-hood.

alexp3y (approved):

PR #1937, opened 2025-12-18
fix(mobile): correct activity UI regressions

pete-watters:
This PR:
- fixes some regressions to mobile activity 
- adds some improvements to icons and the loading screen
- fixes a type with the LD key `releaseOnramperBuy` -> `release_onramper_buy`

These issues were reported [here](https://trustmachines.slack.com/archives/C05LAP952E7/p1765359724391249)

1. The loading state momentarily shows the small (scrolled-up centered) title
2. There aren't enough skeletons rows in the loading state
3. The swap avatar is very obviously broken and misaligned
4. The empty state layout broken (this one might've already been there)
5. The cell structure overall doesn't match the [design here](https://www.figma.com/design/Zh8ljG9M0nB1m8r3eDxZkF/%F0%9F%93%A6-Project--Home?node-id=2729-38865&t=XSZ2Ea8zWgWnW0J7-4): title shows the token name instead of Sent,Received, dates aren't in "distance from now" format
6. Something funky happened with the avatars too, that affected Swaps asset selectors 

Here's a demo of it across extension + mobile APPS:

https://github.com/user-attachments/assets/5c59e173-fcfa-481d-ad00-6d514af33add


apps/mobile/src/features/activity/activity-loading.tsx
```diff
@@ -1,11 +1,26 @@
...
 
-export function ActivityLoading() {
+interface ActivityLoadingProps {
+  header?: ReactNode;
+  count?: number;
+}
+
+const estimatedRowHeight = 72;
+
+export function ActivityLoading({ header, count }: ActivityLoadingProps) {
```
pete-watters, line 14: This improves the skeleton loader by filling the screen height with loaders

apps/mobile/src/features/feature-flags/index.ts
```diff
@@ -97,7 +97,7 @@ export function useSwapFlag() {
 }
 
 export function useOnramperBuyFlag() {
-  return useBoolVariation('releaseOnramperBuy', false);
+  return useBoolVariation('release_onramper_buy', false);
```
pete-watters, line 100: This fixes a mistake I had made when refactoring the web feature flags casing

packages/ui/src/assets/icons/usdcx.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="320" height="320" viewBox="0 0 320 320" fill="none" xmlns="http://www.w3.org/2000/svg">
```
pete-watters, line 1: Adding this specifically as `usdcx.svg` so we can always show it correctly

packages/ui/src/components/avatar/asset-avatar-icon.web.tsx
```diff
@@ -20,12 +24,24 @@ export function AssetAvatarIcon({ asset, indicator, size, ...rest }: AssetAvatar
     case 'nativeBtc':
       return <BtcAvatarIcon indicator={indicator} size={size} {...rest} />;
     case 'sip10': {
+      if ('contractId' in asset && 'imageCanonicalUri' in asset && 'name' in asset) {
```
pete-watters, line 22: Adding these assertions as it helps us decouple from `'@leather.io/models'`. That dependancy will disappear from the UI package entirely soon

packages/ui/src/components/avatar/avatar.shared.ts
```diff
@@ -24,9 +24,6 @@ export function getSip10AvatarImage({
   contractId,
   name,
 }: GetSip10AvatarImageProps) {
-  // TODO: work is needed to give better SIP-10 support often the imageCanonicalUri is not set
```
pete-watters, line 27: I've uploaded a lot of SVGs to our image platform now so this work has been done

camerow:
@mica000 @fabric-8 @pete-watters 

I think these activity rows are missing some critical information. In Pete's video example the main focus of each row is the status (Sent / Received / etc.) and the relative date as well as the USD value and token amount. The token symbol is notably omitted from all of these rows, which I think severly hampers one's ability to scan for the token (save for the most obvious logos of BTC / STX)

Screenshot 2026-01-08 at 8 10 53 AM

I think these are hard to read and are missing some critical information, and each row has too much emphasis on the date / status. I would propose if we're redoing these we should do something more like Base seen here.

image

I think the main improvements here are:
1. ensuring we highlight the token symbol in the row
2. avoid repeating dates for days with lots of activity
3. save room for the truncated address which has some utility for verification - if a little imperfect

pete-watters:
> @mica000 @fabric-8 @pete-watters
> 
> I think these activity rows are missing some critical information. In Pete's video example the main focus of each row is the status (Sent / Received / etc.) and the relative date as well as the USD value and token amount. The token symbol is notably omitted from all of these rows, which I think severly hampers one's ability to scan for the token (save for the most obvious logos of BTC / STX)
> 
> Screenshot 2026-01-08 at 8 10 53 AM
> I think these are hard to read and are missing some critical information, and each row has too much emphasis on the date / status. I would propose if we're redoing these we should do something more like Base seen here.
> 
> image
> I think the main improvements here are:
> 
> 1. ensuring we highlight the token symbol in the row
> 2. avoid repeating dates for days with lots of activity
> 3. save room for the truncated address which has some utility for verification - if a little imperfect

I agree her @camerow  and think that would be a nice improvement. The good news is we have done great work now centralising activity cross platforms so future tweaks here will be easy. 

In the context of this particular PR I did it based on [this feedback](https://trustmachines.slack.com/archives/C05LAP952E7/p1765359724391249) I had introduced some regressions on `mobile` to block the release. The extension work here is gated behind a feature flag so I think we could safely review and merge the code as is and enhance activity further later? 

This is how the new activity looked before this PR change to make it more like `mobile`:
Screenshot 2026-01-09 at 11 41 00

💡 We could start an initiative to improve activity altogether and:
- use a table like we do for portfolio now 
- have better functions like pagination, search, filter etc. out of the box

As a user if I come here I am either:
- checking to see if my latest transaction has gone through
- looking for info on a specific thing 

It's not a great UX to just have a huge list going on forever, especially as some actions give multiple results

apps/extension/src/app/components/tx-asset-item.tsx
```diff
@@ -12,15 +12,23 @@ interface TxAssetItemProps extends HstackProps {
...
+            isStx
+              ? { protocol: 'nativeStx' }
+              : {
+                  protocol: 'sip10',
+                  contractId: iconString,
+                  imageCanonicalUri: imageCanonicalUri ?? '',
+                  name: ticker,
+                }
+          }
+          size="md"
```
kyranjamie, line 31: Not that it's introduced in this PR specifically, I see it was done this way before, but IMO it's the wrong approach to have chain- and chain-token specific details baked into this level of component. 

We need to minimise the number of logical forks, otherwise the codebase becomes a mess, wild `if (bitcion) edgeCase()` statements deep in the component tree. These details should either be passed in via props or a context.

apps/extension/src/app/features/activity-list/components/submitted-transaction-list/submitted-transaction-icon.tsx
```diff
@@ -1,14 +1,13 @@
...
 
 import { getTxSenderAddress } from '@app/common/transactions/stacks/transaction.utils';
-import { TransactionIconWrapper } from '@app/components/transaction/transaction-icon-wrapper';
 import { TransactionTypeIcon } from '@app/components/transaction/transaction-type-icon';
 
-interface SubmittedTransactionIconProps extends CircleProps {
+interface SubmittedTransactionIconProps extends BoxProps {
   transaction: StacksTransactionWire;
 }
 export function SubmittedTransactionIcon({ transaction, ...rest }: SubmittedTransactionIconProps) {
```
kyranjamie, line 13: Can we rename if this is stacks specific?

apps/extension/src/app/features/activity-list/components/activity-item.tsx
```diff
@@ -16,9 +16,39 @@ interface ItemProps {
...
+      break;
+    case 'failed':
+      swapStatusText = 'Swap failed';
+      break;
+    case 'swap':
+      swapStatusText = 'Swapped';
+      break;
+    default:
+      swapStatusText = '';
+  }
```
kyranjamie: this could be function outside the component, that then avoids the let

apps/mobile/src/components/loading/loading-item.tsx
```diff
@@ -1,13 +1,51 @@
...
+        alignItems="center"
+        justifyContent="center"
+        overflow="hidden"
+      >
+        {whenTheme({
+          light: (
+            <Image
+              style={{ width: 16, height: 16 }}
+              contentFit="cover"
+              source={require('@/assets/spinner-light.png')}
```
kyranjamie, line 29: This is an inline commonjs import. Should be non-inline esm import

apps/mobile/src/features/activity/activity-empty.tsx
```diff
@@ -11,23 +10,20 @@ const netSize = width * 0.65;
...
-      }
-    >
+    <Box flexGrow={1} px="5" alignItems="center" justifyContent="center" gap="4">
+      <Image
+        style={{
+          width: netSize,
+          height: netSize,
+        }}
+        contentFit="contain"
+        source={require('@/assets/stickers/net.png')}
```
kyranjamie, line 20: same here

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
```diff
@@ -15,6 +14,12 @@ import { AssetAvatarIcon } from './asset-avatar-icon.web';
 import { getAvatarUrl } from './avatar.shared';
 import { Avatar } from './avatar.web';
 
+type AssetForAvatar =
+  | { protocol: 'nativeStx' }
+  | { protocol: 'nativeBtc' }
+  | { protocol: 'sip10'; contractId: string; imageCanonicalUri: string; name: string }
+  | { protocol: string };
```
kyranjamie: repeated

packages/ui/src/components/avatar/asset-avatar-icon.native.tsx
```diff
@@ -1,15 +1,19 @@
...
 import { BtcAvatarIcon } from './btc-avatar-icon.native';
 import { Sip10AvatarIcon } from './sip10-avatar-icon.native';
 import { StxAvatarIcon } from './stx-avatar-icon.native';
 
+type AssetForAvatar =
+  | { protocol: 'nativeStx' }
+  | { protocol: 'nativeBtc' }
+  | { protocol: 'sip10'; contractId: string; imageCanonicalUri: string; name: string }
+  | { protocol: string };
+
```
kyranjamie: again

packages/ui/src/components/avatar/asset-avatar-icon.web.tsx
```diff
@@ -1,15 +1,19 @@
...
 import { BtcAvatarIcon } from './btc-avatar-icon.web';
 import { Sip10AvatarIcon } from './sip10-avatar-icon.web';
 import { StxAvatarIcon } from './stx-avatar-icon.web';
 
+type AssetForAvatar =
+  | { protocol: 'nativeStx' }
+  | { protocol: 'nativeBtc' }
+  | { protocol: 'sip10'; contractId: string; imageCanonicalUri: string; name: string }
+  | { protocol: string };
+
```
kyranjamie: and again

fabric-8:
I agree in regards of swapping date with truncated address, but I don't quite follow the symbol argumentation.

Considering the shared reference, I think we are doing more right now about highlighting the symbol as it's 1st row and larger where as in base activity it's muted + second row on the right.

So how about sth like [this](https://www.figma.com/design/fz3zc9YCYnRqXr60t9KvmQ/%F0%9F%A7%BE-Product-Area--Activity---Transaction-details?node-id=2160-107261&t=BC4IcSvLPT25xay6-1)?

image

I think it nice + consistent with amounts in other places where we pair the amount with the symbol, if we were to display the symbol for example additionally on the left we'd either have to display it twice in the row or remove it next to the actual amount which is worse result IMO.

apps/extension/src/app/features/activity-list/components/submitted-transaction-list/submitted-transaction-icon.tsx
pete-watters: I can rename it but this us legacy code I think it will be deleted soon anyway. Once we are happy with the new UI we will be getting through a lot of updates so I will add a note to this cleanup task if thats OK? [LEA-3340](https://linear.app/leather-io/issue/LEA-3340/cleanup-legacy-extension-code-refactor-tests)

apps/extension/src/app/components/tx-asset-item.tsx
pete-watters: I will make a note to refactor this away in [LEA-3340](https://linear.app/leather-io/issue/LEA-3340/cleanup-legacy-extension-code-refactor-tests)

apps/extension/src/app/features/activity-list/components/activity-item.tsx
pete-watters: Thanks 🙇  , what are your thoughts on banishing `let` entirely? I considered doing that in https://github.com/leather-io/mono/pull/1969 but thought it might be too harsh. 

I'd like to add a rule to banish let and force a more functional programming style

apps/mobile/src/components/loading/loading-item.tsx
pete-watters: Will I add a rule for this to https://github.com/leather-io/mono/pull/1969 also?

packages/ui/src/components/avatar/activity-avatar-icon.web.tsx
pete-watters: Thanks @kyranjamie 👍  that was sloppy work from me missing that

pete-watters:
Thanks @kyranjamie 👍  I fixed most of the issues I hope!

camerow:
I like it @fabric-8

apps/mobile/src/components/loading/loading-item.tsx
tigranpetrossian: This is the documented way for specifying `source` in `expo-image`. We have an exception in ESLint rules for this too. I suggest keeping as is.

pete-watters: Thanks @tigranpetrossian 👍  I knew we had it for good reason but forgot said reason. I'll revert my update to keep `require`

edgarkhanzadian (approved):

edgarkhanzadian (approved):
Thanks @pete-watters great work! Can you, please, squash before merge?

PR #1942, opened 2025-12-19
fix: givc usdcx priority in mobile and show in empty wallets

pete-watters:
This PR adds similar changes to https://github.com/leather-io/mono/pull/1935 but on mobile. 

- show USDCx at the top of the SIP-10 list
- show USDCx in an empty wallet state 
- don't allow it to be turned off in manage tokens


apps/mobile/src/features/balances/assets/assets-list.tsx
```diff
@@ -5,25 +5,64 @@ import { sortSip10Balances } from '@/features/balances/assets/utils/sort-sip10-b
...
+  header,
+  onPressToken,
+}: AssetsListProps) {
+  const { fingerprint, accountIndex } = account;
+  const { networkPreference } = useSettings();
+
+  const usdcxAssetId =
+    networkPreference.chain.bitcoin.mode === 'mainnet'
+      ? USDCX_ASSET_ID_MAINNET
+      : USDCX_ASSET_ID_TESTNET;
```
kyranjamie: Can we use `whenNetwork` operator here?

kyranjamie (approved):

PR #1943, opened 2025-12-19
chore: remove superflous query key prefix

pete-watters:
This PR fixes a mistake made in the queries package adding an extra un-needed prefix


alexp3y (approved):

kyranjamie (approved):
Nice, absolutely huge query keys now a one word smaller

kyranjamie:
I wonder if a memoized hash fingerprint of the key payload might be a better idea than the enormous keys we're using currently

PR #1946, opened 2025-12-22
refactor: upgrade panda css

kyranjamie:
Some festive spring cleaning 🎄


edgarkhanzadian (approved):

edgarkhanzadian:
Should we also enable lightningcss? It seems like panda supports that in this version, should make css generation faster

edgarkhanzadian:
https://panda-css.com/docs/docs/references/config#lightningcss

PR #1947, opened 2025-12-23
feat: onboarding action buttons [LEA-3297]

edgarkhanzadian:
https://github.com/user-attachments/assets/de0e1794-f987-416d-9b77-dde46fb85388

### DIFF
--- a/apps/extension/src/app/pages/home/components/account-actions-current/account-actions.tsx
+++ b/apps/extension/src/app/pages/home/components/account-actions-current/account-actions.tsx
@@ -1,0 +1,52 @@
+import { useNavigate } from 'react-router';
+
+import { HomePageSelectors } from '@tests/selectors/home.selectors';
+import { Flex } from 'leather-styles/jsx';
+
+import { RouteUrls } from '@shared/route-urls';
+import { replaceRouteParams } from '@shared/utils/replace-route-params';
+
+import { useConfigSwapsEnabled } from '@app/query/common/remote-config/remote-config.query';
+import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+import { BasicTooltip } from '@app/ui/components/tooltip/basic-tooltip';
+
+import { SwapsDisabledTooltipLabel } from '../swaps-disabled-tooltip-label';
+import { ActionButton } from './action-button';
+import { FundButtons } from './fund-buttons';
+import { TransferButtons } from './transfer-buttons';
+
+export function AccountActions() {
+  const navigate = useNavigate();
+  const stacksAccount = useCurrentStacksAccount();
+  const { isTestnet } = useCurrentNetworkState();
+
+  const swapsEnabled = useConfigSwapsEnabled();
+  const swapsBtnDisabled = !swapsEnabled || !stacksAccount || isTestnet;
+  function navigateToDefaultSwapRoute() {
+    return navigate(
+      replaceRouteParams(RouteUrls.Swap, {
+        base: 'STX',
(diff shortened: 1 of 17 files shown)


apps/mobile/package.json
```diff
@@ -172,7 +172,6 @@
     "@types/react": "19.0.12",
     "babel-preset-expo": "13.0.0",
     "concurrently": "8.2.2",
-    "firebase-tools": "13.30.0",
```
edgarkhanzadian: installing any dependency in the monorepo currently fails because we have this dependency in mobile repo. firebase-tools can't run with node v24 and we don't use this dependency at all.

apps/extension/src/app/pages/home/components/account-actions-current/account-actions.tsx
```diff
@@ -0,0 +1,52 @@
...
+    return navigate(
+      replaceRouteParams(RouteUrls.Swap, {
+        base: 'STX',
+        quote: '',
+      }).replace('{chain}', 'stacks')
+    );
+  }
+
+  return (
+    <Flex gap={{ base: 'space.01', md: 'space.04' }} py="space.04" overflowX="scroll">
```
kyranjamie:
```suggestion
    <Flex gap={['space.01', null, 'space.04']} py="space.04" overflowX="scroll">
```

plz use array notation for responsive styles

edgarkhanzadian: @kyranjamie updated to use the array, also it's probably to use the `sm` breakpoint here, changed that too

apps/extension/src/app/pages/home/components/account-actions-current/fund-buttons.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  function onBuy() {
+    return pageModeRoutingAction(RouteUrls.Fund);
+  }
+
+  function onSell() {
+    return pageModeRoutingAction(RouteUrls.Sell);
+  }
+
+  if (isAtLeastMd || !bothButtonsEnabled) {
```
pete-watters, line 52: I was going to say this seems odd but I guess the design/ a variation is like this? IMHO it would be better if we can avoid adding these code variations to change the UI based on width - just pick one set of buttons for both

apps/extension/src/app/pages/home/components/account-actions-current/fund-sheet.tsx
```diff
@@ -0,0 +1,31 @@
...
+        onClick={onBuy}
+        caption="Buy tokens securely with cash"
+        icon={<PaperPlaneIcon />}
+      />
+
+      <ActionDrawerButton
+        icon={<InboxIcon />}
+        onClick={onSell}
+        title="Sell"
+        caption="Sell tokens securely for cash "
```
pete-watters: It probably doesn't matter but there is an extra space after `cash "`. Just mentioning in case there is a UI space due to it

apps/mobile/package.json
pete-watters: Maybe we can remove this in a separate PR? I don't think we use firebase at all now and if we do / this breaks something we can investigate a better way to add it cc @camerow

pete-watters (approved):

apps/extension/src/app/pages/home/components/account-actions-current/fund-buttons.tsx
edgarkhanzadian: yea it might be tricky down the road, we can review this one more time before enabling the extension revamp on prod

mica000:
@edgarkhanzadian seems like we are also missing this logic for Asset's page
CleanShot 2026-01-06 at 10 58 07@2x

PR #1948, opened 2025-12-24
feat: home overview [LEA-3299]

edgarkhanzadian:
https://github.com/user-attachments/assets/8be5f0cf-b536-40b5-900e-9d95a8c7523e

### DIFF
--- a/apps/extension/src/app/components/layout/headers/header-grid-current.tsx
+++ b/apps/extension/src/app/components/layout/headers/header-grid-current.tsx
@@ -1,0 +1,33 @@
+import { Flex, Grid, GridItem, HStack } from 'leather-styles/jsx';
+
+import type { HasChildren } from '@app/common/has-children';
+
+import type { HeaderGridProps } from './header-grid';
+
+export function HeaderGrid({ leftCol, centerCol, rightCol, ...props }: HeaderGridProps) {
+  return (
+    <Grid
+      alignItems="center"
+      gridTemplateColumns={centerCol ? '2fr 4fr 2fr' : 'auto auto'}
+      gridAutoFlow="column"
+      width="100%"
+      {...props}
+    >
+      <GridItem justifySelf="start">
+        <Flex py={{ base: 0, md: 'space.01' }} ml="space.02">
+          {leftCol}
+        </Flex>
+      </GridItem>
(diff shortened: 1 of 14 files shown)


apps/extension/src/app/components/layout/headers/header-grid-current.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <Grid
+      alignItems="center"
+      gridTemplateColumns={centerCol ? '2fr 4fr 2fr' : 'auto auto'}
+      gridAutoFlow="column"
+      width="100%"
+      {...props}
+    >
+      <GridItem justifySelf="start">
+        <Flex py={{ base: 0, md: 'space.01' }} ml="space.02">
```
pete-watters, line 17: I'm happy to use object / array notation for these breakpoints but I know the general preference is to use array. 

Lets do a batch refactor separately to standardise this as array format and add a lint rule if we can?

apps/extension/src/app/features/settings/settings.tsx
```diff
@@ -50,24 +52,21 @@ import {
...
   canLockWallet?: boolean;
-  triggerButton: React.ReactNode;
   toggleSwitchAccount?(): void;
 }
-export function Settings({
-  canLockWallet = true,
-  triggerButton,
-  toggleSwitchAccount,
-}: SettingsProps) {
+export function Settings({ canLockWallet = true, toggleSwitchAccount }: SettingsProps) {
```
pete-watters, line 65: Maybe we can clean up this `canLockWallet = true` as I think there is more to it than that? They need a key etc. too

pete-watters (approved):
Great work @edgarkhanzadian ! It's coming together nicely

apps/extension/src/app/features/settings/settings.tsx
edgarkhanzadian: I don't want to mess up other screens accidentally once we enable this revamp. Trying to decrease the splash radius. We can revisit this place closer to the release.

apps/extension/src/app/components/layout/headers/header-grid-current.tsx
edgarkhanzadian: Yeap, sounds good to me

PR #1949, opened 2025-12-25
feat: updated settings [LEA-3295]

edgarkhanzadian:
https://github.com/user-attachments/assets/c75eed09-d9d6-4e74-ad8d-a1f90fcba830


apps/extension/src/app/components/icon-wrapper.tsx
```diff
@@ -18,3 +21,20 @@ export function IconWrapper({ children }: { children: React.ReactNode }) {
...
+export function NewIconWrapper({ children }: HasChildren) {
+  return (
+    <Box
+      className={css({
+        p: 'space.03',
+        borderRadius: 'round',
+        backgroundColor: 'ink.background-secondary',
+        borderColor: 'ink.component-background-hover',
+        borderWidth: 1,
+      })}
```
kyranjamie: These should be style props

apps/extension/src/app/features/settings/settings-current.tsx
```diff
@@ -0,0 +1,185 @@
...
+          <DropdownMenu.Content
+            align="end"
+            side="bottom"
+            sideOffset={8}
+            className={css({
+              width: 'settingsMenuWidth',
+              maxHeight: 'var(--radix-dropdown-menu-content-available-height)',
+              overflowY: 'scroll',
+              boxShadow: theme === 'dark' ? 'elevationDark' : 'elevationLight',
+            })}
```
kyranjamie, line 75: Dropdown not a panda element so css fine to use here

kyranjamie (approved):
Thanks for this @edgarkhanzadian, happy to have a settings page!

Did you test that the `AppVersion` component still works?

I wonder if some (power) users will complain that changing network is now more deeply nested in the settings, as it is sometimes changed often.

Suggest we get @fabric-8 and @mica000 eyes on this as these are significant visual changes

apps/extension/src/app/components/icon-wrapper.tsx
```diff
@@ -18,3 +21,20 @@ export function IconWrapper({ children }: { children: React.ReactNode }) {
     </Box>
   );
 }
+
+// TODO: seems like the design got updated for the icon wrapper. Should we update all of the instances of the icon wrapper?
```
pete-watters, line 24: Can you check this with design? Maybe you can help @fabric-8 ? 

Then we can tackle it now and get rid of the `TODO` handling it like the others - `ComponentCurrent` + `ComponentLegacy`

apps/extension/src/app/components/layout/headers/header-back-button.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
+
+import { ArrowLeftIcon } from '@leather.io/ui';
+
+import { HeaderActionButton } from './header-action-button';
+
+export function HeaderBackButton() {
+  const navigate = useNavigate();
+  return (
+    <HeaderActionButton
```
pete-watters, line 12: Maybe we can just add `HeaderActionButton` to settings? We only use it there

apps/extension/src/app/features/settings/settings-current.tsx
```diff
@@ -0,0 +1,185 @@
...
+          >
+            <DropdownMenu.Group>
+              {/** TODO: discuss this one with the team */}
```
pete-watters: What do you want to check here?

apps/extension/src/app/features/settings/settings-current.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const isPrivateMode = useIsPrivateMode();
+  const togglePrivateMode = useTogglePrivateMode();
+
+  const location = useLocation();
+
+  const { isPressed: showAdvancedMenuOptions } = useModifierKey('alt', 120);
+
+  const { theme } = useThemeSwitcher();
+
+  const bottomGroupItems = useMemo(
```
pete-watters: Maybe we can refactor this while we are here? I'm not sure if `useMemo` adds much and it could be simpler to just use `const`s and then include the components in the main JSX block? 🤔  

Then we could maybe even use the container/layout pattern here with this being the `.layout` and `settings.tsx` getting the data?

apps/extension/src/app/features/settings/settings-current.tsx
```diff
@@ -0,0 +1,185 @@
...
+                  </Flex>
+                </Flag>
+              </DropdownMenu.Item>
+              {bottomGroupItems.length > 0 && (
+                <>
+                  <Divider />
+                  <DropdownMenu.Group>{...bottomGroupItems}</DropdownMenu.Group>
+                </>
+              )}
+              {/* <AppVersion /> */}
```
pete-watters: Do we need `AppVersion`? 🤔

apps/extension/src/app/pages/settings/menu-buttons.tsx
```diff
@@ -0,0 +1,128 @@
...
+          setShowChangeNetwork(!showChangeNetwork);
+        }}
+        icon={<GlobeTiltedIcon />}
+      />
+
+      <SettingsButton
+        data-testid={SettingsSelectors.ToggleNotifications}
+        variant="switch"
+        title="Notification"
+        // TODO: update tooltip text here
```
pete-watters: Lets tackle the `TODO`s now? 
- `TODO: maybe need caption` 
- `TODO: update tooltip text here`
- `TODO: where should this navigate to?`

apps/extension/src/app/pages/settings/sticky-buttons.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <>
+      <Flex
+        gap="space.05"
+        pb="space.05"
+        pt="space.03"
+        background="ink.background-primary"
+        position="sticky"
+        bottom={0}
+      >
+        {hasDefaultInMemorySecretKey && hasKeys && walletType === 'software' && (
```
pete-watters, line 36: Maybe we can share this logic a bit better? We have `hasKeys && walletType === 'software'` a few times and a variation of this above in the settings page with `canLockWallet && hasKeys && walletType === 'software'`. 

Maybe we can make a `LockButton` component that has the logic it needs in one place?

packages/panda-preset/src/tokens.ts
```diff
@@ -12,4 +12,58 @@ export const tokens = defineTokens({
...
+        {
+          offsetX: 0,
+          offsetY: 0,
+          blur: 2,
+          spread: 0,
+          color: 'rgba(18, 16, 15, 0.12)',
+        },
+      ],
+    },
+    // TODO: elevationDark is autogenerated, discuss with designers if this shadow is good for dark mode.
```
pete-watters, line 41: 👍  can you share some example here so @fabric-8 / @mica000  can check?

packages/ui/src/components/dropdown-menu/dropdown-menu.web.tsx
```diff
@@ -65,8 +65,8 @@ const dropdownContentStyles = css({
   '--base-menu-padding': '0px',
   bg: 'ink.background-primary',
   borderRadius: 'sm',
-  boxShadow:
-    '0px 12px 24px 0px rgba(18, 16, 15, 0.08), 0px 4px 8px 0px rgba(18, 16, 15, 0.08), 0px 0px 2px 0px rgba(18, 16, 15, 0.08)',
+  // TODO: remove this boxShadow once we release home screen revamp
```
pete-watters, line 68: Can we pass this as a prop instead so we don't need the TODO?

pete-watters (approved):
Great work on this @edgarkhanzadian 👍  I mentioned a few things along the way but almost all are _nit_s / improvement suggestions

mica000:
Looking good @edgarkhanzadian! Thanks for picking this one up!

Wondering if we discarded moving "network" moving to a page as well to keep consistency or if we judge its better to keep in a modal. Designs [here](https://www.figma.com/design/Zh8ljG9M0nB1m8r3eDxZkF/%F0%9F%93%A6-Project--Home?node-id=2729-45771&t=HFLVBmekdq9gJAua-4)
CleanShot 2026-01-05 at 10 49 58@2x

mica000:
Tooltip copy is missing: _"Available for Bitcoin sends and receives"_
CleanShot 2026-01-05 at 11 40 00@2x

mica000:
- Hide button on the top right that gives access to settings while on the "Settings" page. 
- Network badge should remain visble

CleanShot 2026-01-05 at 11 42 00@2x

mica000:
The dropdown menu appears to have a div that adds unnecessary space. Could we remove it? But keep in mind that the line that divides the content also needs a vertical breather between the buttons "Privacy mode" and "lock". Designs [here](https://www.figma.com/design/Zh8ljG9M0nB1m8r3eDxZkF/%F0%9F%93%A6-Project--Home?node-id=2729-45742&t=HFLVBmekdq9gJAua-4)

CleanShot 2026-01-05 at 11 44 40@2x

mica000:
Could you fix the links redirects?
Help center: https://app.leather.io/help-center
Dev docs: https://leather.gitbook.io/developers/

mica000:
> Thanks for this @edgarkhanzadian, happy to have a settings page!
> 
> Did you test that the `AppVersion` component still works?
> 
> I wonder if some (power) users will complain that changing network is now more deeply nested in the settings, as it is sometimes changed often.
> 
> Suggest we get @fabric-8 and @mica000 eyes on this as these are significant visual changes

@kyranjamie @edgar, From your development perspective, is this behavior something you encounter frequently? I tried searching for data in Mixpanel, but we are not tracking anything in the settings area. Would you suggest we wait to see if the complaints increase, or should we reintroduce the surfacing network? In case we would like to surface this again, [here are the designs](https://www.figma.com/design/Zh8ljG9M0nB1m8r3eDxZkF/%F0%9F%93%A6-Project--Home?node-id=3573-43066&t=HFLVBmekdq9gJAua-4)

CleanShot 2026-01-05 at 12 28 54@2x

kyranjamie:
I'm sure in our overall analytics changing network does not happen that often. 

But Stacks ecosystem developers are a few loud voices that will do this frequently. If we add extra clicks to perform the same action as before, I anticipate complaints. If a small % of loud voices were to dislike this change, what would you suggest? Is there a simple solution to keep the network switcher easily accessible while not prominent for most users?

pete-watters:
> I'm sure in our overall analytics changing network does not happen that often.
> 
> But Stacks ecosystem developers are a few loud voices that will do this frequently. If we add extra clicks to perform the same action as before, I anticipate complaints. If a small % of loud voices were to dislike this change, what would you suggest? Is there a simple solution to keep the network switcher easily accessible while not prominent for most users?

@mica000 the [Radix UI dropdown](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) we use has a nested sub-menu. Maybe thats a nice compromise we can use for power users? 

We could show the network options in a submenu there and avoid too many clicks. That way users can quickly change network from the menu and it will still be an improvement as we won't need a dialog view for changing network. I suggest we keep `Add network` in the new settings page as that's a page view anyway.


Alternatively we could just use  an options group and show the networks listed there, similar to `People` in the radix example. 

Screenshot 2026-01-06 at 09 10 17

Now could be a good time to consider adding an 'Advanced mode' toggle setting in the 'Settings' page so we could get a good compromise between clean UX for regular users and also allow these power users to turn on visibility to quick access things like this. Most regular users will not change the network and seeing that option is even a bit confusing for noobs

apps/extension/src/app/components/layout/headers/header-back-button.tsx
edgarkhanzadian: I'm planning on reusing it in other header components in the future

apps/extension/src/app/features/settings/settings-current.tsx
edgarkhanzadian: Do we want to have the ledger menu item still in the dropdown or should we move it to the settings page?

cc @kyranjamie @mica000 @fabric-8

apps/extension/src/app/features/settings/settings-current.tsx
edgarkhanzadian: nope, thanks!

packages/ui/src/components/dropdown-menu/dropdown-menu.web.tsx
edgarkhanzadian: we would need to come back here and update this to theme dependent shadows once we get to releasing the extension revamp. Currently keeping it here to not affect the legacy ui

apps/extension/src/app/features/settings/settings-current.tsx
mica000: @edgarkhanzadian can you share a screenshot of what do you mean? Cant specifically localize this feature

mica000:
> > I'm sure in our overall analytics changing network does not happen that often.
> > But Stacks ecosystem developers are a few loud voices that will do this frequently. If we add extra clicks to perform the same action as before, I anticipate complaints. If a small % of loud voices were to dislike this change, what would you suggest? Is there a simple solution to keep the network switcher easily accessible while not prominent for most users?
> 
> @mica000 the [Radix UI dropdown](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) we use has a nested sub-menu. Maybe thats a nice compromise we can use for power users?
> 
> We could show the network options in a submenu there and avoid too many clicks. That way users can quickly change network from the menu and it will still be an improvement as we won't need a dialog view for changing network. I suggest we keep `Add network` in the new settings page as that's a page view anyway.
> 
> Alternatively we could just use an options group and show the networks listed there, similar to `People` in the radix example.
> 
> Screenshot 2026-01-06 at 09 10 17
> Now could be a good time to consider adding an 'Advanced mode' toggle setting in the 'Settings' page so we could get a good compromise between clean UX for regular users and also allow these power users to turn on visibility to quick access things like this. Most regular users will not change the network and seeing that option is even a bit confusing for noobs

Thanks @pete-watters! Ideally the solution should work the same way in all platforms, how would you imagine this working for mobile? 

@kyranjamie My intuition suggests that we should not implement these changes based on anticipated complaints, as we cannot be 100% certain they exist. I believe that keeping Networks inside Settings will also allow us more space to learn about any complaints and develop a better solution.

kyranjamie:
Great idea @pete-watters, would love to see this component used. I think it's fine that it isn't 100% like mobile. We shouldn't get caught up in "everything must be exactly the same". The general theme should be the same, the components can differ. Another option could be an "advanced" menu users access with right click or holding ctrl etc, or just enabling a global advanced mode.

@mica000 I think we can learn from experience working in the ecosystem. How do you do your job if you don't "anticipate" ways users will interact with our product in the most intuitive way? Your experience as a designer, and mine working in the ecosystem are both valid ways to "predict" how users will respond to product changes. We can be wrong, of course, but that doesn't mean we should completely ignore what we _think_ might happen.

Here's are more elaborate version of my concern, that I think we should avoid, and not just respond to: we release the changes, power users who change network often now have to click more to change network, Friedger or Alex Miller go to Stacks Signal chat and say "Hey Leather, why isn't network in dropdown, it takes way longer to change network now". Then, you have to go and design something new, and we have to go and add it back to the menu...

Happy to proceed as it is, and we can wait and see if I'm correct or not 😂

pete-watters:
> > > I'm sure in our overall analytics changing network does not happen that often.
> > > But Stacks ecosystem developers are a few loud voices that will do this frequently. If we add extra clicks to perform the same action as before, I anticipate complaints. If a small % of loud voices were to dislike this change, what would you suggest? Is there a simple solution to keep the network switcher easily accessible while not prominent for most users?
> > 
> > 
> > @mica000 the [Radix UI dropdown](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) we use has a nested sub-menu. Maybe thats a nice compromise we can use for power users?
> > We could show the network options in a submenu there and avoid too many clicks. That way users can quickly change network from the menu and it will still be an improvement as we won't need a dialog view for changing network. I suggest we keep `Add network` in the new settings page as that's a page view anyway.
> > Alternatively we could just use an options group and show the networks listed there, similar to `People` in the radix example.
> > Screenshot 2026-01-06 at 09 10 17
> > Now could be a good time to consider adding an 'Advanced mode' toggle setting in the 'Settings' page so we could get a good compromise between clean UX for regular users and also allow these power users to turn on visibility to quick access things like this. Most regular users will not change the network and seeing that option is even a bit confusing for noobs
> 
> Thanks @pete-watters! Ideally the solution should work the same way in all platforms, how would you imagine this working for mobile?
> 
> @kyranjamie My intuition suggests that we should not implement these changes based on anticipated complaints, as we cannot be 100% certain they exist. I believe that keeping Networks inside Settings will also allow us more space to learn about any complaints and develop a better solution.

The current version of mobile is working very differently than on the `extension` so I think that's OK? I imagine on `mobile` we would have more regular users than power users so I think it's OK to keep that as is. We don't have this dropdown menu on mobile at all as settings is already a page and I don't think we planned to change that?

pete-watters:
> Great idea @pete-watters, would love to see this component used. I think it's fine that it isn't 100% like mobile. We shouldn't get caught up in "everything must be exactly the same". The general theme should be the same, the components can differ. Another option could be an "advanced" menu users access with right click or holding ctrl etc, or just enabling a global advanced mode.
> 
> @mica000 I think we can learn from experience working in the ecosystem. How do you do your job if you don't "anticipate" ways users will interact with our product in the most intuitive way? Your experience as a designer, and mine working in the ecosystem are both valid ways to "predict" how users will respond to product changes. We can be wrong, of course, but that doesn't mean we should completely ignore what we _think_ might happen.
> 
> Here's are more elaborate version of my concern, that I think we should avoid, and not just respond to: we release the changes, power users who change network often now have to click more to change network, Friedger or Alex Miller go to Stacks Signal chat and say "Hey Leather, why isn't network in dropdown, it takes way longer to change network now". Then, you have to go and design something new, and we have to go and add it back to the menu...
> 
> Happy to proceed as it is, and we can wait and see if I'm correct or not 😂

@kyranjamie another idea I have is we could just remove it from here as initially planned but consider some kind of hot-key addition for these power users? Like `CTRL+N` just opens the network dialog we already have. I wonder if we can curate the list of networks also to remove some ones we don't need anymore? Or else show them in a more advanced mode

kyranjamie:
Yeah absolutely, or a cmd + k menu with access to all settings

apps/extension/src/app/features/settings/settings-current.tsx
edgarkhanzadian: discussed with the engineers in the slack thread, seems to be redundant: https://trustmachines.slack.com/archives/C05LRS7G44R/p1767775613461509?thread_ts=1767773103.899569&cid=C05LRS7G44R

PR #1952, opened 2026-01-05
feat: non-blocking inscription check

kyranjamie:
This PR removes the blocking runtime exception thrown when users try and spend inscriptions. While this was intended to protect user assets, it is too blocking and many users simply don't care about inscriptions. 

Here I've added a dialog that asks the user if they are happy to proceed with this situation. I've used the [`react-call`](https://github.com/desko27/react-call) library that helps call dialogs as promises. Encourage this pattern, as it offers a clean alternative to managing more component state using `useState` just to toggle a dialog's visibility on/off.

@fabric-8 I'm using our sheet component here, but as mentioned in slack the styles seem off. Ideal solution would be to fix the sheet component and adjust any typography you suggest.

**Demo**

https://github.com/user-attachments/assets/85934381-5fd9-4a13-88d6-9d950c83cd0b

Notice at the end the transaction fails for some unrelated reason 😢

### DIFF
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -151,6 +151,7 @@
     "react": "19.0.0",
     "react-animate-height": "3.1.1",
     "react-async-hook": "4.0.0",
+    "react-call": "1.8.1",
     "react-dom": "19.0.0",
     "react-dom-confetti": "0.2.0",
     "react-head": "3.4.2",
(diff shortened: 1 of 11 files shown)


pete-watters (approved):

edgarkhanzadian (approved):
let's get this merged and released together with change addresses fix

fabric-8:
@kyranjamie  Let's ommit the "x" in this sheet:
- it introduces ambigiuty (_"Does this do anything different from cancel?"_) 
- Adds more options to chose from although it's just doubling the cancel
- Doesn't look too great

kyranjamie:
@fabric-8 I'll follow up with these suggestions when doing the wider changes to the sheet, but merging for now as this is a blocking issue

PR #1953, opened 2026-01-05
fix(extension): restore stack fee guards on extension

pete-watters:
This PR makes changes described in [this PR](https://linear.app/leather-io/issue/LEA-3314/restore-fee-guards-on-extension) and on [this thread](https://trustmachines.slack.com/archives/C05LAP952E7/p1765287328998669) 

- [X] adds fee guards to the extension so a warning is shown if a fee about 5 STX is set
- [X] uses fee guard logic from `StacksTransactionFeesService` 
- [X] adds a fix account for memo/arg changes in fee calculation


pete-watters:
Closing this and will fix in https://github.com/leather-io/mono/pull/2002

PR #1954, opened 2026-01-05
fix(mobile): package advisories

kyranjamie:
Fixes package advisories, and a redirect loop issue with web app homepage.


apps/mobile/package.json
```diff
@@ -172,8 +172,8 @@
     "@types/react": "19.0.12",
     "babel-preset-expo": "13.0.0",
     "concurrently": "8.2.2",
-    "firebase-tools": "13.30.0",
-    "happy-dom": "18.0.1",
+    "firebase-tools": "15.1.0",
```
pete-watters: @edgarkhanzadian mentioned an issue with this package on [his PR](https://github.com/leather-io/mono/pull/1947#discussion_r2642774090)

Screenshot 2026-01-06 at 09 19 31

I'm not sure if it truly is blocking package installs but if so we should probably remove it. We don't really use firebase much now. I think we could potentially start to but I will need to catchup with @camerow on that

kyranjamie: Will remove here

pete-watters (approved):
Thanks for fixing this 🎯

apps/mobile/package.json
pete-watters: My hunch is it's for mobile notifications but we haven't finished that and I think it needs some BE work also so I think it's fine to remove for now then re-add if things break and we really do need it.

PR #1957, opened 2026-01-06
fix: change addresses support [LEA-3332]

edgarkhanzadian:
Add support for change addresses.
Tested using both software and hardware wallet on sbtc bridge

### DIFF
--- a/apps/extension/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
+++ b/apps/extension/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
@@ -1,28 +1,25 @@
 import { useCallback, useEffect } from 'react';
 
-import { P2TROut } from '@scure/btc-signer/payment';
 import { useInfiniteQuery } from '@tanstack/react-query';
 
-import type { BitcoinSigner } from '@leather.io/bitcoin';
-import { BitcoinAddress } from '@leather.io/models';
 import { BitcoinQueryPrefixes } from '@leather.io/query';
 import { createNumArrayOfRange } from '@leather.io/utils';
 
 import { useLeatherNetwork } from '@app/query/leather-query-provider';
 import { useBestInSlotApiRateLimiter } from '@app/query/rate-limiter/best-in-slot-limiter';
+import { useCurrentAccountNativeSegwitIndexZeroSigner } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useCurrentAccountTaprootSigner } from '@app/store/accounts/blockchain/bitcoin/taproot-account.hooks';
 
 import { useBitcoinClient } from '../../clients/bitcoin-client';
 
 const addressesSimultaneousFetchLimit = 3;
 const stopSearchAfterNumberAddressesWithoutBrc20Tokens = 3;
 
-export function useGetBrc20TokensQuery({
-  nativeSegwitAddress,
-  createTaprootSigner,
-}: {
-  nativeSegwitAddress: BitcoinAddress;
-  createTaprootSigner: ((addressIndex: number) => BitcoinSigner<P2TROut>) | undefined;
-}) {
+export function useGetBrc20TokensQuery() {
+  const createTaprootSigner = useCurrentAccountTaprootSigner();
+  const nativeSegwitSigner = useCurrentAccountNativeSegwitIndexZeroSigner();
+  const nativeSegwitAddress = nativeSegwitSigner.address;
+
   const network = useLeatherNetwork();
   const currentNsBitcoinAddress = nativeSegwitAddress;
   const client = useBitcoinClient();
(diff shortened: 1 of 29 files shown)


packages/bitcoin/src/utils/lookup-derivation-by-address.ts
```diff
@@ -26,19 +26,26 @@ export function lookupDerivationByAddress(args: LookUpDerivationByAddressArgs) {
 
   return (address: BitcoinAddress) => {
     const network = inferNetworkFromAddress(address);
+    // NOTE: this function is used for ordinals lookup. Should we lookup change addresses too?
+    const changeIndex = 0;
```
edgarkhanzadian: Should we lookup change addresses too here?

packages/bitcoin/src/utils/bitcoin.descriptors.ts
```diff
@@ -71,15 +71,28 @@ export function deriveAddressesFromDescriptor({
...
-  for (let i = 0; i < limit; i++) {
-    const address = whenSupportedPaymentType(paymentType)({
-      p2tr: getTaprootAddress({ index: i, keychain: accountKeychain, network }),
-      p2wpkh: getNativeSegwitAddress({ index: i, keychain: accountKeychain, network }),
-    });
-    results.push({
-      address,
-      path: derivationPathFn(network, accountKeychain.index - HARDENED_OFFSET, i),
-    });
+  // NOTE: Ask Alex if we should use change addresses for lookup
```
edgarkhanzadian: @alexp3y @kyranjamie i've added a change address support here too but let me know if it is not needed here

apps/extension/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
```diff
@@ -33,7 +30,8 @@ export function useGetBrc20TokensQuery({
   const getNextTaprootAddressBatch = useCallback(
     (fromIndex: number, toIndex: number) => {
       return createNumArrayOfRange(fromIndex, toIndex - 1).map(num => {
-        const address = createTaprootSigner(num).address;
+        // NOTE: safe to not check change address here?
+        const address = createTaprootSigner({ addressIndex: num, changeIndex: 0 }).address;
```
kyranjamie: This is for BRC20 tokens only so not particularly worried about defaulting to 0 here

packages/bitcoin/src/payments/p2tr-address-gen.ts
```diff
@@ -21,12 +21,20 @@ export function makeTaprootAccountDerivationPath(
...
-  return makeTaprootAccountDerivationPath(network, accountIndex) + `/0/${addressIndex}`;
+export function makeTaprootAddressIndexDerivationPath({
+  network,
+  accountIndex,
+  changeIndex,
+  addressIndex,
+}: {
+  network: BitcoinNetworkModes;
+  accountIndex: number;
+  changeIndex: number;
```
kyranjamie, line 32: Worth typing as a number literal? `0 | 1`?

packages/bitcoin/src/utils/bitcoin.descriptors.ts
kyranjamie: This fn is only used in one place in services for mempool look up. I would imagine we should add these here

packages/bitcoin/src/utils/lookup-derivation-by-address.ts
kyranjamie: I think it's fine assuming 0 here. This fn is already a bit of a hack to try and work backwards to find the matching address when we don't know the index.

packages/crypto/src/derivation-path-utils.ts
```diff
@@ -32,6 +32,14 @@ export const extractAddressIndexFromPath = extractSectionFromDerivationPath(
   DerivationPathDepth.AddressIndex
 );
 
+export function isTaprootDerivationPath(path: string) {
+  return extractPurposeFromPath(path) === 86;
+}
+
+export function isNativeSegwitDerivationPath(path: string) {
+  return extractPurposeFromPath(path) === 84;
+}
```
kyranjamie: Shouldn't this go in bitcoin pkg? Crypto is intended to be chain agnostic and bitcoin pkg builds on it.

kyranjamie (approved):
Thanks for taking on this refactor @edgarkhanzadian. It was my mistake (and against my own advice of writing code reflecting the protocol, rather than our use of it) not building it this way from the beginning.

packages/crypto/src/derivation-path-utils.ts
edgarkhanzadian: yeap good point will move it there

packages/bitcoin/src/payments/p2tr-address-gen.ts
edgarkhanzadian: i'd rather leave that as a separate PR tbh. I'd need to update `extractChangeIndexFromPath` and add some kind of error handling whenever we try to pass a number to change index.

PR #1959, opened 2026-01-06
test(extension): partial account wallet state broken

kyranjamie:
Glad I caught this, as it was only because I was using my Ledger to move funds I noticed. 

In Ledger mode, when accounts from only 1 chain are added, the activity tab breaks. Here I've added a test that breaks because of this bug. 

@pete-watters to take it from here. It might also be worth testing other functionality that might be impacted by the same issue. Based on the error, I think it has something to do with the activity service.


pete-watters:
Closing this as I have fixed the problem in https://github.com/leather-io/mono/pull/1960 and cherry picked this failing test to that PR

PR #1931, opened 2025-12-17
fix: use uniqueArray to useQueries and cache nft to avoid duplicate queries

pete-watters:
We are currently getting a flood of console errors in the extension about potential query duplication

image

I've resolved this by:
- making use of `uniqueArray` to sanitise inputs of `addresses` / `txid` / `blocks` / `xpubs` to `useQueries`
- creating a key map cache for [non-fungible-token-metadata.query.ts](https://github.com/leather-io/mono/compare/chore/console-warnings?expand=1#diff-3337f07374f134ea0c001647842ab4d3d1dd34264d57cd7cb222039fb52af694)

PR #1960, opened 2026-01-07
fix(extension): bug with activity crashing when ledger connected with Stacks only

pete-watters:
This PR fixes a bug @kyranjamie found when using ledger with Stacks only (Bitcoin not connected)

It was introduced by [this PR](https://github.com/leather-io/mono/pull/1931) which was an unrelated refactor to remove a lot of duplicate query keys we had and exposes a gap in our tests for Ledger activity. 

- In that PR I introduced `uniqueArray` to curate the address list we use for query keys
- In this scenario the address array for Bitcoin is `['','']` as its not connected so the addresses are both empty
- This means that `useGetBitcoinTransactionsByAddressListQuery` crashes as it has one input and not two 

You can see the fix here: 

https://github.com/user-attachments/assets/3b45ee84-c6f8-4157-99d5-5691cb2910c8

Unfortunately it means we will still start to see the duplicate query key warning again here on this state but that's correct in this case.  This code is soon to be deprecated and replaced by the new activity service so I think it's OK to accept this for now in the knowledge it will soon be removed entirely. 

Screenshot 2026-01-07 at 06 57 21


package.json
```diff
@@ -48,6 +48,8 @@
     "@tanstack/eslint-plugin-query": "5.62.1",
     "@testing-library/jest-dom": "6.6.3",
     "@testing-library/react": "16.3.0",
+    "@types/react": "19.0.12",
```
pete-watters: I added this un-related change as the `ui` package was failing to build for me when running `pnpm build`. I can drop it from here if its incorrect / adding unwanted noise 

The build was failing as the repo root pulls in `@types/react@19.2.7` via `@testing-library/react`, while `packages/ui` pins `@types/react@19.0.12`. 

I could likely get around this doing a local clean install but then it could probably happen again/to someone else so I think it's best to enforce the versions we want

kyranjamie (approved):
LGTM, but types shouldn't be in root pkg should they? Edit: missed comment

PR #1961, opened 2026-01-07
ci: fix flaky swap unit test [LEA-3336]

edgarkhanzadian:


apps/mobile/src/features/swap/swap-state/swap-state.reducer.ts
```diff
@@ -79,8 +79,8 @@ export function swapReducer(state: SwapInternalState, action: SwapActionObject):
         targetSwapAsset: null,
         baseAmount: '0',
         pairReconciliation: {
+          ...pairReconciliation,
           base: 'pending',
-          target: 'pending',
         },
```
edgarkhanzadian: @tigranpetrossian This is not related to the unit test but something i wanted to check just in case.
I've noticed that for `RECONCILE_BASE_WITH_PROVIDER` we reset the whole pairReconciliation object, but for `RECONCILE_TARGET_WITH_PROVIDER` we only reset the target. Wanted to double check if that is the expected behavior

apps/mobile/src/features/swap/swap-state/tests/use-swap-state.assets.spec.ts
```diff
@@ -367,8 +367,9 @@ describe('initial placholder base asset reconciliation', () => {
     await waitFor(() => {
       expect(result.current.baseAssetsQuery.status).toBe('success');
     });
-
-    expect(result.current.state.baseSwapAsset).toBeNull();
+    await waitFor(() => {
+      expect(result.current.state.baseSwapAsset).toBeNull();
+    });
```
edgarkhanzadian, line 372: Seems to be a race condition happening here. On rare occasions, baseAssetsQuery status is `success` but the `useSwapAssetReconciliation` hook doesn't have time to run. In that case, baseSwapAsset doesn't have enough time to get updated to null.
Just waiting for it to be null should be enough here.

pete-watters (approved):

apps/mobile/src/features/swap/swap-state/tests/use-swap-state.assets.spec.ts
kyranjamie: Doesn't this fix the the race condition in the test, and not address the same behaviour in the UI?

kyranjamie:
@edgarkhanzadian can this be merged soon 🙏🏼

apps/mobile/src/features/swap/swap-state/swap-state.reducer.ts
edgarkhanzadian: Discussed, this should not be changed

PR #1966, opened 2026-01-07
fix(extension): fix bug with tab navigation, ref LEA-3334

pete-watters:
This PR fixes a bug with navigation on the new activity tabs reported in LEA-3334. 

I've also added some `eslint` + `Claude.md` updates to discourage the use of ternary statements and suggest to use more functional coding patterns (e.g. avoid `let`)

https://github.com/user-attachments/assets/e3024956-f630-4d04-9823-06e1f23a8a19

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,6 +10,8 @@
 - Use arrow functions for callbacks.
 - Use object method shorthand syntax in objects and interfaces.
   constants.ts files.
+- Avoid nested ternary expressions; prefer clear branching or functional expressions.
+- Prefer `const` where possible; avoid `let` when it improves clarity.
 - Prefer constants over magic numbers or strings.
 
 ## Naming
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,12 +112,33 @@ const ids = items.map(x => x.id);
 - Our repo is a turbo monorepo which uses `pnpm` for package management and package.json scripts.
 - Many common actions can be found in respective pacakge and apps `scripts` in the package.json.
 
-### Updating pnpm version
+### Verify before completing
 
-When updating pnpm, you must update all three locations:
+Verify that no formatting, type, or lint errors are introduced:
 
-1. `.prototools` - version only (e.g., `pnpm = "10.26.0"`)
-2. `.github/actions/provision/action.yml` - version with SHA
-3. `package.json` - `packageManager` field with version and SHA
+```bash
+pnpm format:check
+pnpm lint
+pnpm typecheck
+```
 
-Use the helper script: `./scripts/update-pnpm-version.sh <version>`
+If `format:check` fails, run `pnpm format` to auto-fix formatting issues.
+
+For faster feedback when working in a specific package, use filtered commands:
+
+```bash
+pnpm --filter @leather.io/web lint
+pnpm --filter @leather.io/extension typecheck
+```
+
+Fix any errors before considering the task complete.
+
+### Prettier formatting
+
+Follow the project's Prettier configuration. Key rules:
+- Use single quotes for strings
+- Use trailing commas where valid in ES5 (objects, arrays)
+- 2-space indentation
+- 100 character print width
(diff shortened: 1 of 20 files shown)


.cursorrules
```diff
@@ -0,0 +1 @@
+CLAUDE.md
```
pete-watters: I have been using cursor agents a bit more as they seem to have improved. This is from a symlink and tells cursor agents to use the same rules we have in `Claude.md`

CLAUDE.md
```diff
@@ -109,3 +111,21 @@ const ids = items.map(x => x.id);
 
 - Our repo is a turbo monorepo which uses `pnpm` for package management and package.json scripts.
 - Many common actions can be found in respective pacakge and apps `scripts` in the package.json.
+
+### Verification before completing
```
pete-watters: Adding this so agents run these checks before giving their answers. A pre-commit hook could be better but that would also force it to run on each commit for everyone which could be slow and annoying most of the time. 

This way the generated code is at least pre-checked

apps/extension/src/app/pages/home/components/home-tabs.tsx
```diff
@@ -34,17 +34,22 @@ const homeTabs = [
 export function HomeTabs({ children }: HomeTabsProps) {
   const navigate = useNavigate();
   const location = useLocation();
-  const activeTab = useMemo(() => {
-    const matchingTab = homeTabs.find(tab => location.pathname.startsWith(tab.value));
-    return matchingTab?.value ?? RouteUrls.Home;
-  }, [location.pathname]);
+  const { pathname } = location;
```
pete-watters, line 37: This is the main fix to the bug LEA-3334 with the navigation not working properly

apps/mobile/src/utils/analytics-hooks.ts
```diff
@@ -128,12 +128,16 @@ export function useCollectiblesAnalytics({
...
-            ? collectible.content.contentType
-            : undefined;
+      function getContentType() {
+        if ('mimeType' in collectible && collectible.mimeType) {
+          return collectible.mimeType;
+        }
+        if ('content' in collectible && collectible.content?.contentType) {
+          return collectible.content.contentType;
+        }
+        return undefined;
```
edgarkhanzadian: nit:
```suggestion
        return;
```

eslint.config.js
```diff
@@ -66,6 +65,9 @@ export default defineConfig([
     files: ['{packages,apps}/**/*.{ts,tsx}'],
     extends: [baseConfig],
     rules: {
+      'no-nested-ternary': 'error',
+      'no-var': 'error',
+      'prefer-const': 'error',
```
edgarkhanzadian: Great addition of no-nested-ternary!

We already have no-var and prefer-const rules enabled in the base config, so we can remove it here.

Can you also add the no-nested-ternary to the base config in packages/eslint-config/index.js and remove it from the root eslint config?

.cursorrules
kyranjamie: Would rather this isn't committed. Project roots are already so crowded with configs and this is yet ide-specific one.

kyranjamie (approved):

CLAUDE.md
kyranjamie: > Verify before completing

> verify that no type or lint errors are introduced:

.cursorrules
pete-watters: OK can drop no problem 👍

eslint.config.js
pete-watters: Sure I can remove them. I actually wanted to enforce a rule of `no-let` and enforce more functional code but that could be a bit hardcore so thought I should check with the team first.

pete-watters:
Thanks for the feedback @kyranjamie  + @edgarkhanzadian 🙇  

I think I have addressed it all but if not let me know and I will make retrospective updates

edgarkhanzadian (approved):

PR #1967, opened 2026-01-08
feat(web): footer

fabric-8:
Introduce a shared web `StickyFooter` and update extension screens to place the footer outside padded content, so the footer/gradient (mirroring mobile aesthetics) run edge‑to‑edge while content retains consistent horizontal spacing.

CleanShot 2026-01-08 at 11 51 50@2x
CleanShot 2026-01-08 at 11 51 36@2x

### DIFF
--- a/packages/ui/src/components/sticky-footer/sticky-footer.web.tsx
+++ b/packages/ui/src/components/sticky-footer/sticky-footer.web.tsx
@@ -1,0 +1,30 @@
+import { Flex, type FlexProps, styled } from 'leather-styles/jsx';
+import { token } from 'leather-styles/tokens';
+
+import type { HasChildren } from '../../utils/has-children.shared';
+
+const fadeHeight = 24;
+
+export type StickyFooterProps = FlexProps & HasChildren;
+
+export function StickyFooter({ children, ...props }: StickyFooterProps) {
+  return (
+    <Flex position="sticky" bottom={0} background="ink.background-primary" width="100%" {...props}>
+      <styled.div
+        aria-hidden
+        style={{
+          position: 'absolute',
+          top: `-${fadeHeight}px`,
+          left: 0,
+          right: 0,
+          height: `${fadeHeight}px`,
+          pointerEvents: 'none',
+          background: `linear-gradient(180deg, transparent, ${token(
+            'colors.ink.background-primary'
+          )})`,
+        }}
+      />
+      {children}
+    </Flex>
+  );
+}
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/sticky-footer/sticky-footer.web.tsx
```diff
@@ -0,0 +1,30 @@
...
+const fadeHeight = 24;
+
+export type StickyFooterProps = FlexProps & HasChildren;
+
+export function StickyFooter({ children, ...props }: StickyFooterProps) {
+  return (
+    <Flex position="sticky" bottom={0} background="ink.background-primary" width="100%" {...props}>
+      <styled.div
+        aria-hidden
+        style={{
```
kyranjamie, line 15: These should be style props, as above, not all in a style object

kyranjamie:
Thanks @fabric-8 

What's the exact abstraction you're aiming for here? The combination of bottom positioning, and the fade style in one?

I'm uncertain about this because:
- In CSS lingo this is more an absolutely bottom positioned element. `position: sticky` is really meant for fixed elements as you scroll (e.g. the titles on changelog page). It can be used for this effect, but the name suggests something else. 
- A "footer" is a very general layout-focused concept. We have other footers that are bottom positioned: the web app's footer is a sticky footer for example.
- Positioning an element bottom can sometimes be surrounding element specific. E.g. here it only works because it is itself inside a Flex, so it's not an entirely encapsulated abstraction.

Perhaps this is more a naming concern than anything. Is this a more specific list footer?

I wonder if what is really needed is a list component that combines more layout-focused elements of the list (and basic styles like the fade).

```tsx
<List.Root>
  <List.Header>some title</List.Header>
  <List.Body>{listItems}</List.Body>
  <List.Footer>{actionsFixedToTheBottom}</List.Footer>
</List.Root>

fabric-8:
@kyranjamie Thanks for the feedback!

The intent here isn’t a global footer, it’s a bottom‑anchored action area with a subtle fade separation. So agree with your feedback that the name is too generic + ‘sticky’ misleading. How about we rename it to align with the existing `Approver.Actions` / action‑bar pattern (e.g. `ActionBar` / `BottomActionBar`)? It’s also true this assumes a vertical layout where the bar is the last child; Guess we could adjust the approach so the bar is always bottom‑anchored and the content gets the proper bottom padding (same pattern as `Approver.Actions`).  wdyt?

packages/ui/src/components/sticky-footer/sticky-footer.web.tsx
pete-watters: This keeps catching everyone out (me included) and is more of a code style preference than a problem. 

Anyway, it is creating churn so [I started work on adding a rule to enforce this](https://github.com/leather-io/mono/pull/1969) and also array notation for responsive styles.

I think having rules for stuff like that will just stop it happening and reduce churn. People will forget / AI will generate it without knowing better

fabric-8:
Closed in favour of #1972

PR #1971, opened 2026-01-08
fix(web): ensure btc is set as the rewardsToken and minimum commitment is core

camerow:
Ensure we show the correct minimum stacking amount and rewards token for StackingDAO. 

Ensure we provide a BTC awards address for stacking dao rewards.  
Screenshot 2026-01-08 at 12 59 16 PM
Screenshot 2026-01-08 at 12 59 27 PM

### DIFF
--- a/apps/web/app/data/data.ts
+++ b/apps/web/app/data/data.ts
@@ -1,5 +1,4 @@
 import { z } from 'zod';
-import { MIN_DELEGATED_STACKING_AMOUNT_USTX } from '~/constants/constants';
 
 // Providers are partner entities that offer yeild based services
 const providers = {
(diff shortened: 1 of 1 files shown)


apps/web/app/data/data.ts
kyranjamie: Shouldn't we update this value instead of hard coding?

pete-watters (approved):
This LGTM once you can fix the lint error

apps/web/app/data/data.ts
camerow: This seems to be a generic minimum stacking amount - not the minimum for that specific stacking pool. We should perhaps keep it for utility.

PR #1972, opened 2026-01-08
feat(extension): fade bottom action bar

fabric-8:
Adds a new bottomBarFade shadow token that creates a gradient fade effect for bottom action footers, allowing content to smoothly fade as it scrolls behind the buttons. This is an evolution of #1967

### DIFF
--- a/packages/panda-preset/src/semantic-tokens.ts
+++ b/packages/panda-preset/src/semantic-tokens.ts
@@ -4,4 +4,12 @@ import { semanticTokens as leatherSemanticTokens } from '@leather.io/tokens';
 
 export const semanticTokens = defineSemanticTokens({
   ...leatherSemanticTokens,
+  shadows: {
+    contentOverflowFade: {
+      value: {
+        base: '{shadows.contentOverflowFadeLight}',
+        _dark: '{shadows.contentOverflowFadeDark}',
+      },
+    },
+  },
 });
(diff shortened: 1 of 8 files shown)


packages/panda-preset/src/tokens.ts
```diff
@@ -65,5 +65,69 @@ export const tokens = defineTokens({
         },
       ],
     },
+    bottomBarFadeLight: {
```
kyranjamie: Is it not possible that this gradient might also be used elsewhere? 

I'd suggest omitting the "bar" from the core token name, as this suggests where it'll be used, making it inflexible for use elsewhere. `contentOverflowFadeLight`?

packages/panda-preset/src/semantic-tokens.ts
```diff
@@ -4,4 +4,12 @@ import { semanticTokens as leatherSemanticTokens } from '@leather.io/tokens';
...
 export const semanticTokens = defineSemanticTokens({
   ...leatherSemanticTokens,
+  shadows: {
+    bottomBarFade: {
+      value: {
+        base: '{shadows.bottomBarFadeLight}',
+        _dark: '{shadows.bottomBarFadeDark}',
+      },
+    },
+  },
```
edgarkhanzadian, line 14: nice! I'll need to add this to dropdown shadows too

edgarkhanzadian (approved):
Great work!

packages/panda-preset/src/tokens.ts
kyranjamie: thanks for updating!

kyranjamie (approved):

PR #1968, opened 2026-01-08
chore(web): update react router 7

kyranjamie:
Web app was failing to build for me, attempting update to fix

PR #1973, opened 2026-01-09
fix: react router audit

edgarkhanzadian:


.syncpackrc
```diff
@@ -1,12 +1,4 @@
 {
-  "versionGroups": [
-    {
-      "dependencies": [
-        "react-router"
-      ],
-      "isIgnored": true
-    }
-  ],
```
edgarkhanzadian, line 9: Any particular reason we wanted to keep react-router versions separate on web and extension?

kyranjamie: I added this. As one of our core deps on extension & web app, and their different uses in each app, I figured it's safe to keep them individual 

E.g. upgrading for web, and because of syncpack upgrading it for extension w/o manual testing seems like a bad idea

kyranjamie:
My PR already fixes this actually https://github.com/leather-io/mono/pull/1968

.syncpackrc
edgarkhanzadian: sure sounds good. I've synced the dependencies in this pr and after some testing it does seem like it is working correctly on both platforms

kyranjamie (approved):

PR #1975, opened 2026-01-09
fix(extension): make sure error boundary is centered, ref LEA-2563

### LINKED ISSUE #6247: The error boundary screen does not center the error
There is a small visual bug in the `ErrorBoundary` in that it does not center the content and it's left aligned:

Image

This can be replicated easily by adding a `throw new Error('test error');` to `pages/home`

pete-watters:
This is a small but impactful fix to center the error boundary. 

You can see it moving here:

https://github.com/user-attachments/assets/66a2f29a-0169-4584-b69c-1fc67ef6680e

### DIFF
--- a/apps/extension/src/app/features/errors/app-error-boundary.tsx
+++ b/apps/extension/src/app/features/errors/app-error-boundary.tsx
@@ -63,55 +63,56 @@ export function RouterErrorBoundary() {
   }
 
   return (
-    <Flex
-      alignItems="center"
-      justifyContent="center"
-      flexDirection="column"
-      px={['space.05', 'unset']}
-      py={['space.05', 'space.06']}
-      width="100%"
-      maxWidth="500px"
-    >
-      <Box mt="space.05">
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):

edgarkhanzadian:
Can we use sticky buttons here so user would always be able to see the buttons without scrolling? (it didn't work before too but i'd suggest we update that).
Also seems like the icon is hitting the ceiling now.

Now:
Screenshot 2026-01-09 at 20 31 07

Before:
Screenshot 2026-01-09 at 20 32 29

camerow (approved):

pete-watters:
> Can we use sticky buttons here so user would always be able to see the buttons without scrolling? (it didn't work before too but i'd suggest we update that).
> Also seems like the icon is hitting the ceiling now.

Good idea, I've done that while I am at it:

https://github.com/user-attachments/assets/e54c477a-133d-4d8a-937a-672442966b6c

PR #1977, opened 2026-01-09
feat(extension): make long function argument values expandable values rather than filling the screen

camerow:
https://github.com/user-attachments/assets/b98c3e4a-c09a-4dd0-90c5-e5dec6f8eda8

### DIFF
--- a/apps/extension/src/app/features/rpc-stacks-transaction-request/stacks/contract-call/function-arguments/function-argument-item.tsx
+++ b/apps/extension/src/app/features/rpc-stacks-transaction-request/stacks/contract-call/function-arguments/function-argument-item.tsx
@@ -1,20 +1,98 @@
+import { useState } from 'react';
+
 import { HStack, Stack, styled } from 'leather-styles/jsx';
 
 import { getStacksExplorerLink } from '@leather.io/features';
 import { ChainId } from '@leather.io/models';
-import { Link } from '@leather.io/ui';
+import { ChevronDownIcon, ChevronUpIcon, Link } from '@leather.io/ui';
 
 import { openInNewTab } from '@app/common/utils/open-in-new-tab';
 import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+
+const longValueThreshold = 512;
+const previewLength = 80;
+
+interface ExpandableValueProps {
+  value: string;
+}
+
+function FunctionArgumentValue({
+  value,
+  isPrincipal,
+  isLongValue,
+}: ExpandableValueProps & {
+  isPrincipal: boolean;
+  isLongValue: boolean;
+}) {
+  const { chain, isNakamotoTestnet } = useCurrentNetworkState();
+  const networkMode = chain.stacks.chainId === ChainId.Mainnet ? 'mainnet' : 'testnet';
+  if (isPrincipal) {
+    return (
+      <Link
+        size="sm"
+        variant="text"
+        wordBreak="break-all"
+        onClick={() =>
+          openInNewTab(
+            getStacksExplorerLink({
+              mode: networkMode,
+              type: 'address',
+              value,
+              isNakamoto: isNakamotoTestnet,
+            })
+          )
+        }
+      >
+        {value}
+      </Link>
+    );
+  }
+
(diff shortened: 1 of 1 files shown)


pete-watters (approved):
Thanks @camerow ! I had this on TODO's for too long, I appreciate the fix! 

Sometimes we are too caught up in big features we don't make enough time for small impactful changes like this. We need to improve in that respect

apps/extension/src/app/features/rpc-stacks-transaction-request/stacks/contract-call/function-arguments/function-argument-item.tsx
```diff
@@ -1,12 +1,57 @@
...
+        transition="height 0.2s ease-in-out"
+        visibility={isExpanded ? 'visible' : 'hidden'}
+      >
+        <styled.span
+          display="block"
+          textStyle="body.02"
+          wordBreak="break-all"
+          color="ink.text-subdued"
+        >
+          {value}
```
edgarkhanzadian: I'd personally prefer showing some value with the underscore with an ellipsis at the end. 
S.t. we at least show some part of the value. We could also use a hover instead of the button to display the full value. But that can be raised for a discussion with a design team later.
e.g.
smthlikethis

edgarkhanzadian (approved):

kyranjamie (changes requested):
I don't think we should, if too long, hide the entire contents. 

Rather, why not truncate it to a certain length with the same _show more_ button? This touches something rather sensitive, and provides an opportunity for malicious devs to hide payloads just by making the long enough to be auto hidden.

camerow:
@edgarkhanzadian / @kyranjamie good feedback - now we preview the first 80 characters if the limit is reached 

https://github.com/user-attachments/assets/2ee0be62-33a6-4f5c-ac0c-26d76e92f604

camerow:
cc @edgarkhanzadian / @kyranjamie requested change has been added

edgarkhanzadian:
I'd personally say it would be better to only enable this for values that are longer than 80chars so we don't show ellipsis/text decoration/chevron on those but i'm also ok to discuss it after with designers.

cc @fabric-8 @mica000

apps/extension/src/app/features/rpc-stacks-transaction-request/stacks/contract-call/function-arguments/function-argument-item.tsx
```diff
@@ -47,11 +70,35 @@ export function FunctionArgumentItem({ name, type, value }: FunctionArgumentItem
         >
           {value}
         </Link>
-      ) : (
-        <styled.span display="block" textStyle="body.02" wordBreak="break-all">
-          {value}
-        </styled.span>
-      )}
+      );
+    }
```
kyranjamie: This should be a variable with JSX, or a separate component.

A function that returns JSX is a component, and components shouldn't be defined inside other components https://react.dev/learn/your-first-component#:~:text=as%20you%20like.-,Pitfall,-Components%20can%20render

camerow: @kyranjamie updated

apps/extension/src/app/features/rpc-stacks-transaction-request/stacks/contract-call/function-arguments/function-argument-item.tsx
```diff
@@ -1,20 +1,98 @@
...
+function FunctionArgumentValue({
+  value,
+  isPrincipal,
+  isLongValue,
+}: ExpandableValueProps & {
+  isPrincipal: boolean;
+  isLongValue: boolean;
+}) {
+  const { chain, isNakamotoTestnet } = useCurrentNetworkState();
+  const networkMode = chain.stacks.chainId === ChainId.Mainnet ? 'mainnet' : 'testnet';
```
kyranjamie, line 28: We have a `whenNetwork` toggle you can use to avoid treating network as a boolean

kyranjamie (approved):

PR #1981, opened 2026-01-13
feat: add shared utilities to features package

pete-watters:
This PR lays the foundation of updating collectibles and adding token details to the extension and is the first in a series of PRs to break up work from https://github.com/leather-io/mono/pull/1903 

- Centralise collectibles rendering logic via a view in the `features` package 
    - Refactor collectible code to better share consts
    - Introduce `CollectibleView` to share logic across platfroms
    - Add tests for data mapping knowledge via `getInscriptionInfo` / `getSip9Info`
- Move shared type guards from `mobile` to features
    - improve to avoid usage of `X is Y` where possible
- Improve the `activity` features:
    - Add `sbtc` activity code to `features/activity`
    - Add shared filtering logic to use in client side queries 
- Fix a bug with how we handle encoded `ipfs` paths and handle `ipfs://` and `https://`

### DIFF
--- a/packages/features/src/activity/activity-status.ts
+++ b/packages/features/src/activity/activity-status.ts
@@ -89,6 +89,22 @@ export function formatActivityStatusLabel(activity: Activity) {
       return contractName ?? 'Unknown';
     }
     case 'swapAssets':
+      if ('sbtcBridgeStatus' in activity && activity.sbtcBridgeStatus) {
+        switch (activity.sbtcBridgeStatus) {
+          case 'pending':
+            return 'Pending deposit';
+          case 'accepted':
+            return 'Pending mint';
+          case 'failed':
+            return 'Failed';
+          case 'rbf':
+            return 'Replaced';
+          case 'confirmed':
+            return 'Confirmed';
+          default:
+            return undefined;
+        }
+      }
       if (isFungibleAsset(activity.fromAsset) && isFungibleAsset(activity.toAsset)) {
         return `${activity.fromAsset.symbol} → ${activity.toAsset.symbol}`;
       } else if (isStampAsset(activity.fromAsset) && isStampAsset(activity.toAsset)) {
--- a/packages/features/src/activity/activity-view.ts
+++ b/packages/features/src/activity/activity-view.ts
@@ -1,3 +1,4 @@
+import { SBTC_RECLAIM_URL } from '@leather.io/constants';
 import type { Activity, NetworkConfiguration } from '@leather.io/models';
 
 import { getActivityBalances } from './activity-balance';
@@ -17,6 +18,29 @@ function getActivityKey(activity: Activity): string {
   return `${activity.type}-${activity.timestamp}`;
 }
 
+function hasSbtcBridgeFailed(activity: Activity): boolean {
+  return (
+    activity.type === 'swapAssets' &&
+    'sbtcBridgeStatus' in activity &&
+    activity.sbtcBridgeStatus === 'failed'
+  );
+}
+
+function getActivityLink(
+  activity: Activity,
+  asset: ReturnType<typeof getActivityAsset>,
+  networkPreference: NetworkConfiguration
+) {
+  const showSbtcReclaimUrl = hasSbtcBridgeFailed(activity) && 'txid' in activity && activity.txid;
+  if (showSbtcReclaimUrl) {
+    return `${SBTC_RECLAIM_URL}${activity.txid}`;
+  }
+  if (hasTxDetails(activity) && asset) {
+    return makeActivityLink({ txid: activity.txid, networkPreference, asset });
+  }
+  return null;
+}
+
 export function createActivityView(
   activity: Activity,
   networkPreference: NetworkConfiguration
@@ -32,10 +56,8 @@ export function createActivityView(
   const statusIndicator = getActivityStatusIndicatorId(activity);
   const fromAsset = activity.type === 'swapAssets' ? activity.fromAsset : undefined;
   const toAsset = activity.type === 'swapAssets' ? activity.toAsset : undefined;
-  const activityLink =
-    hasTxDetails(activity) && asset
-      ? makeActivityLink({ txid: activity.txid, networkPreference, asset })
-      : null;
+
+  const activityLink = getActivityLink(activity, asset, networkPreference);
 
   return {
     key: getActivityKey(activity),
@@ -44,7 +66,7 @@ export function createActivityView(
     toAsset,
     title,
     caption: combinedCaption,
-    statusLabel,
+    statusLabel: statusLabel ?? null,
     balances,
     activityLink,
     activityAvatar,
--- a/packages/features/src/balance/balance-types.ts
+++ b/packages/features/src/balance/balance-types.ts
@@ -1,0 +1,29 @@
+import { CryptoAssetProtocols } from '@leather.io/models';
+import {
+  AccountQuotedBtcBalance,
+  AddressQuotedStxBalance,
+  RuneBalance,
+  Sip10Balance,
+} from '@leather.io/services';
+
+export type TokenBalance =
+  | AccountQuotedBtcBalance
+  | AddressQuotedStxBalance
+  | Sip10Balance
+  | RuneBalance;
+
+export function isAccountQuotedBtcBalance(value: TokenBalance): value is AccountQuotedBtcBalance {
+  return 'btc' in value;
+}
+
+export function isAddressQuotedStxBalance(value: TokenBalance): value is AddressQuotedStxBalance {
+  return 'stx' in value;
+}
+
+export function isSip10Balance(value: TokenBalance): value is Sip10Balance {
+  return 'asset' in value && value.asset.protocol === CryptoAssetProtocols.sip10;
+}
+
+export function isRuneBalance(value: TokenBalance): value is RuneBalance {
+  return 'asset' in value && value.asset.protocol === CryptoAssetProtocols.rune;
+}
--- a/packages/features/src/collectibles/collectible-details.spec.ts
+++ b/packages/features/src/collectibles/collectible-details.spec.ts
@@ -1,0 +1,189 @@
+import { describe, expect, it } from 'vitest';
+
+import { GAMMA_URL, HIRO_EXPLORER_URL, ORD_IO_URL } from '@leather.io/constants';
+import type { InscriptionAsset, Sip9Asset, Sip9Attribute, StampAsset } from '@leather.io/models';
+
+import {
+  DESCRIPTION_TRUNCATE_LENGTH,
+  filterSip9Attributes,
+  formatAttributeValue,
+  getGammaCollectionUrl,
+  getHiroExplorerContractUrl,
+  getInscriptionInfo,
+  getOrdExplorerUrl,
+  getSip9Info,
+  getStampInfo,
+  truncateDescription,
+} from './collectible-details';
+
+describe('getOrdExplorerUrl', () => {
+  it('returns undefined when no number provided', () => {
+    expect(getOrdExplorerUrl()).toBeUndefined();
+  });
+
+  it('returns ord.io URL with inscription number', () => {
+    expect(getOrdExplorerUrl(12345)).toBe(`${ORD_IO_URL}/12345`);
+  });
+});
+
+describe('getGammaCollectionUrl', () => {
+  it('returns undefined when no URL provided', () => {
+    expect(getGammaCollectionUrl()).toBeUndefined();
+    expect(getGammaCollectionUrl(null)).toBeUndefined();
+  });
+
+  it('returns gamma URL with path', () => {
+    expect(getGammaCollectionUrl('/collection/test')).toBe(`${GAMMA_URL}/collection/test`);
+  });
+});
+
+describe('getHiroExplorerContractUrl', () => {
+  it('returns undefined when no contract provided', () => {
+    expect(getHiroExplorerContractUrl()).toBeUndefined();
+    expect(getHiroExplorerContractUrl(null)).toBeUndefined();
+  });
+
+  it('returns Hiro explorer URL', () => {
+    expect(getHiroExplorerContractUrl('SP123.test')).toBe(
+      `${HIRO_EXPLORER_URL}/address/SP123.test`
+    );
+  });
+});
+
+describe('filterSip9Attributes', () => {
+  it('returns empty array for undefined', () => {
+    expect(filterSip9Attributes(undefined)).toEqual([]);
+  });
+
+  it('filters out invalid attributes', () => {
+    const attrs: Sip9Attribute[] = [
+      { traitType: 'Color', value: 'Red' },
+      { traitType: '', value: 'Blue' },
+      { traitType: 'Size', value: '' },
+      { traitType: 'Style', value: 'None' },
+      { traitType: 'Valid', value: 'Yes' },
+    ];
+
+    const result = filterSip9Attributes(attrs);
+    expect(result).toHaveLength(2);
+    expect(result[0].traitType).toBe('Color');
+    expect(result[1].traitType).toBe('Valid');
+  });
+});
+
+describe('formatAttributeValue', () => {
+  it('returns value without rarity when not present', () => {
+    expect(formatAttributeValue({ traitType: 'Color', value: 'Red' })).toBe('Red');
+  });
+
+  it('includes rarity percentage when present', () => {
+    expect(formatAttributeValue({ traitType: 'Color', value: 'Red', rarityPercent: 5.5 })).toBe(
+      'Red (5.5%)'
+    );
+  });
+});
+
+describe('truncateDescription', () => {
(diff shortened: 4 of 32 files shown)


packages/features/src/activity/activity-status.ts
```diff
@@ -89,6 +89,21 @@ export function formatActivityStatusLabel(activity: Activity) {
...
+            return 'Pending deposit';
+          case 'accepted':
+            return 'Pending mint';
+          case 'failed':
+            return 'Failed';
+          case 'rbf':
+            return 'Replaced';
+          case 'confirmed':
+          default:
+            return '';
```
edgarkhanzadian: let's do `null` or `undefined` instead of the empty string here. We do falsy value check in activity-view.ts file anyway, but it would better to tell tsc that the value can be empty.

packages/features/src/activity/activity-view.ts
```diff
@@ -32,10 +34,19 @@ export function createActivityView(
...
+      'sbtcBridgeStatus' in activity &&
+      activity.sbtcBridgeStatus === 'failed'
+    ) {
+      return `${sbtcReclaimUrl}${activity.txid}`;
+    }
+    if (hasTxDetails(activity) && asset) {
+      return makeActivityLink({ txid: activity.txid, networkPreference, asset });
+    }
+    return null;
+  })();
```
edgarkhanzadian: let's separate this to a function called `getActivityLink` or smth. I don't think there is a need for an IIFE here.

packages/features/src/balance/balance-types.ts
```diff
@@ -0,0 +1,62 @@
...
+export function isRuneBalance(value: TokenBalance): value is RuneBalance {
+  return (
+    value &&
+    typeof value === 'object' &&
+    'asset' in value &&
+    'quote' in value &&
+    'crypto' in value &&
+    'availableBalance' in value.crypto
+  );
+}
```
edgarkhanzadian, line 29: Usage of `X is Y` is quite dangerous if done incorrectly. Both functions would pass with the same TokenBalance object but tsc would claim that the same object is RuneBalance or Sip10Balance based on which of these functions it passed.
In this case you need to use `token.asset.protocol` which either would have a value of `CryptoAssetProtocols.sip10` or `CryptoAssetProtocols.rune`

We also have these functions implemented here: https://github.com/leather-io/mono/blob/dev/apps/mobile/src/features/balances/assets/render-assets.tsx#L10-L16

E.g.

```
// Returns type Sip10Balance successfully
isSip10Balance({
  value: {},
  asset: {} as RuneAsset,
  quote: {} as CryptoAssetBalance,
  crypto: {} as CryptoAssetBalance,
} as TokenBalance);

// Returns type RuneBalance successfully
isRuneBalance({
  value: {},
  asset: {} as RuneAsset,
  quote: {} as CryptoAssetBalance,
  crypto: {} as CryptoAssetBalance,
} as TokenBalance);
```

packages/features/src/balance/balance-types.ts
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    value &&
+    typeof value === 'object' &&
+    'stx' in value &&
+    'quote' in value &&
+    value.stx !== undefined &&
+    value.quote !== undefined &&
+    'availableUnlockedBalance' in value.stx
+  );
+}
```
edgarkhanzadian: Same goes for these. We should either remove these or make a much simpler boolean expression that clearly identifies AccountQuotedBtcBalance or AddressQuotedStxBalance. I'd suggest we cut down on `X is Y` type usage.

packages/features/src/collectibles/collectible-details.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+
+describe('getHiroExplorerContractUrl', () => {
+  it('returns undefined when no contract provided', () => {
+    expect(getHiroExplorerContractUrl()).toBeUndefined();
+    expect(getHiroExplorerContractUrl(null)).toBeUndefined();
+  });
+
+  it('returns Hiro explorer URL', () => {
+    expect(getHiroExplorerContractUrl('SP123.test')).toBe(
+      'https://explorer.hiro.so/address/SP123.test'
```
edgarkhanzadian: Can we use `HIRO_EXPLORER_URL` const here?

packages/features/src/collectibles/collectible-details.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+  truncateDescription,
+} from './collectible-details';
+
+describe('getOrdExplorerUrl', () => {
+  it('returns undefined when no number provided', () => {
+    expect(getOrdExplorerUrl()).toBeUndefined();
+  });
+
+  it('returns ord.io URL with inscription number', () => {
+    expect(getOrdExplorerUrl(12345)).toBe('https://ord.io/12345');
```
edgarkhanzadian: `ORD_IO_URL`

packages/features/src/collectibles/collectible-details.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+});
+
+describe('getGammaCollectionUrl', () => {
+  it('returns undefined when no URL provided', () => {
+    expect(getGammaCollectionUrl()).toBeUndefined();
+    expect(getGammaCollectionUrl(null)).toBeUndefined();
+  });
+
+  it('returns gamma URL with path', () => {
+    expect(getGammaCollectionUrl('/collection/test')).toBe('https://gamma.io/collection/test');
```
edgarkhanzadian: `GAMMA_API_URL` (not in constants package but probably a nice moment to add it there.)

packages/features/src/collectibles/collectible-details.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export function getGammaCollectionUrl(collectionExplorerUrl?: string | null): string | undefined {
+  if (!collectionExplorerUrl) return undefined;
+  return `https://gamma.io${collectionExplorerUrl}`;
+}
+
+export function getHiroExplorerContractUrl(contractId?: string | null): string | undefined {
+  if (!contractId) return undefined;
+  return `https://explorer.hiro.so/address/${contractId}`;
```
edgarkhanzadian: HIRO_EXPLORER_URL

packages/features/src/collectibles/collectible-details.ts
```diff
@@ -0,0 +1,133 @@
...
+import { getBitcoinExplorerLink } from '../activity/activity-links';
+
+export function getOrdExplorerUrl(inscriptionNumber?: number): string | undefined {
+  if (!inscriptionNumber) return undefined;
+  return `${ORD_IO_URL}/${inscriptionNumber}`;
+}
+
+export function getGammaCollectionUrl(collectionExplorerUrl?: string | null): string | undefined {
+  if (!collectionExplorerUrl) return undefined;
+  return `https://gamma.io${collectionExplorerUrl}`;
```
edgarkhanzadian: GAMMA_API_URL

packages/features/src/collectibles/collectible-details.ts
```diff
@@ -0,0 +1,133 @@
...
+    collectionName: collection?.name,
+    collectionUrl: getGammaCollectionUrl(collection?.collectionExplorerUrl),
+    creator: asset.creator,
+    rarityRank: asset.rarityRank,
+    totalItems: collection?.totalItems,
+    contractUrl: getHiroExplorerContractUrl(asset.contractId),
+    contentType: asset.content?.contentType,
+    attributes: filterSip9Attributes(asset.attributes),
+  };
+}
```
edgarkhanzadian, line 87: Why do we need Sip9Info if it pretty much matches Sip9Asset data? 🤔 We can call the functions right before we need to use those

packages/features/src/collectibles/collectible-details.ts
```diff
@@ -0,0 +1,133 @@
...
+    blockExplorerUrl: asset.blockHeight
+      ? getBitcoinExplorerLink({
+          id: asset.blockHeight.toString(),
+          type: 'block',
+          networkPreference: bitcoinNetwork,
+        })
+      : undefined,
+    blockHeight: asset.blockHeight,
+  };
+}
```
edgarkhanzadian, line 121: Same for stamp info

packages/features/src/collectibles/collectible-view.ts
```diff
@@ -0,0 +1,55 @@
...
+        title: asset.name || assetName || 'Unknown collectible',
+        subtitle: asset.collection?.name ?? 'Stacks collectible',
+        asset,
+        isBns,
+      };
+    }
+    default:
+      return assertUnreachable(asset);
+  }
+}
```
edgarkhanzadian, line 51: tbh i'd suggest keeping this closer to the function component where we are going to use this function. i'm not sure if moving this function further from the point of use is helping or distracting, could be wrong here.

packages/features/src/collectibles/sip9-media.ts
```diff
@@ -0,0 +1,82 @@
...
+  'audio/wav',
+  'audio/ogg',
+  'audio/aac',
+  'audio/flac',
+  'audio/webm',
+  'text/plain',
+  'application/octet-stream',
+  'model/gltf+json',
+  'model/gltf-binary',
+  '',
```
edgarkhanzadian: please let's not use empty string as a type 🙏 if we are unsure on the content type, maybe we could use null/undefined or `unknown`

packages/features/src/formatting.ts
```diff
@@ -0,0 +1,22 @@
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc';
+
+dayjs.extend(utc);
```
edgarkhanzadian: i wonder if using `dayjs.extend` outside of a function would create a side effect of affecting other files with dayjs usage. Can we move this into `formatTimestampWithTime` function?

packages/services/src/assets/sip9-asset.utils.spec.ts
```diff
@@ -136,6 +136,28 @@ describe('createSip9Asset', () => {
     });
   });
 
+  it('should transform ipfs:// URLs to the Leather IPFS gateway', () => {
+    const gammaIpfs = JSON.parse(JSON.stringify(mockGammaMetadata)) as GammaNftMetadata;
+    gammaIpfs.item.asset_content.content_url = 'ipfs://QmTestCid/some/image with spaces.png';
+
+    const asset = createSip9Asset(assetIdentifier, tokenId, undefined, gammaIpfs);
+    expect(asset.content.contentUrl).toContain(
+      'https://leather.quicknode-ipfs.com/ipfs/QmTestCid/'
```
edgarkhanzadian: let's get this url into a constant in constants package

pnpm-lock.yaml
```diff
@@ -1787,6 +1787,9 @@ importers:
       '@leather.io/services':
         specifier: workspace:*
         version: link:../services
+      '@leather.io/stacks':
+        specifier: workspace:*
+        version: link:../stacks
```
edgarkhanzadian, line 1792: Technically this should be the only change needed in the pnpm-lock based on the changes in features/package.json. Unsure why it comes with version updates on other packages. Can you try stashing only the pnpm-lock file (reverting it to dev branch) and trying to run pnpm i?

packages/features/src/collectibles/collectible-details.ts
```diff
@@ -0,0 +1,133 @@
...
+          type: 'tx',
+          networkPreference: bitcoinNetwork,
+        })
+      : undefined,
+    genesisTimestamp: asset.genesisTimestamp,
+    genesisBlockHeight: asset.genesisBlockHeight,
+    mimeType: asset.mimeType,
+    outputValue: asset.value,
+  };
+}
```
edgarkhanzadian, line 56: Same for inscription info

packages/features/src/balance/balance-types.ts
pete-watters: OK thanks I didn't know about those duplicate functions. I'll use those instead

pete-watters:
Thanks @edgarkhanzadian 👍 I'll update and let you know when its ready to re-check

packages/features/src/activity/activity-status.ts
pete-watters: Thanks! This was based on some older code we have:
```
function getDepositStatus(status: SbtcStatus) {
  switch (status) {
    case 'pending':
      return 'Pending deposit';
    case 'accepted':
      return 'Pending mint';
    case 'failed':
      return 'Failed';
    case 'rbf':
      return 'Replaced';
    case 'confirmed':
    default:
      return '';
  }
}
```

That is used in the legacy code we will get rid of soon
```
<TransactionListItem>
   <SbtcDepositTransactionItem>
``` 

I've changed it to be undefined and to revert to `''` in `activity-view.ts`

packages/features/src/balance/balance-types.ts
pete-watters: Thanks @edgarkhanzadian . I wasn't aware of how it can be unsafe so I appreciate the explanation 🙇

packages/features/src/collectibles/collectible-details.ts
pete-watters: Adding these abstractions means we can add unit tests to cover the transformation logic in isolation and helps give a clearer view model pattern. 

`getSip9Info` is:
- flattening nested props 
- transforming URLs
- filtering attributes 


I can change to call those functions inline but then we lack test coverage in that area. I am doing the same for `getInscriptionInfo` and `getStampInfo`. 

I can revert if you think thats best? Sometimes it's nice to have an abstraction and mocked unit tests to help document the intentions

packages/features/src/collectibles/collectible-view.ts
pete-watters: I am adding `createCollectibleView` here so that we can better share this view creation across platforms. 

It gets used in query selection local to both `extension` and `mobile` in `useAccountCollectiblesQuery`. We use it in the query select in both - `(collectibles => collectibles.map(createCollectibleView))` 

Having it in `features` means we can centralise that logic here and also cover it in tests but still allow `extension`/`mobile` to over-ride with a local select if they want

packages/features/src/collectibles/sip9-media.ts
pete-watters: Thanks @edgarkhanzadian 👍 🧑‍🏫

packages/features/src/formatting.ts
pete-watters: Thanks! This was leftover junk code, apologies

packages/features/src/collectibles/collectible-details.ts
pete-watters: I added these based on [these reasons](https://github.com/leather-io/mono/pull/1981#discussion_r2689187087) but can easily update if you think thats best

packages/features/src/collectibles/collectible-details.ts
pete-watters: I added these based on [these reasons](https://github.com/leather-io/mono/pull/1981#discussion_r2689187087) but can easily update if you think thats best

pete-watters:
Thanks for all the feedback @edgarkhanzadian  and your patience with me 🙏  You helped me fix a lot of issues with my code but also helped us add some nice improvements to exisiting code along the way

I've made almost all of the suggested changes apart from: 
- I didn't get rid of `getSip9Info` yet as I think thats [a useful abstraction](https://github.com/leather-io/mono/pull/1981#discussion_r2689187087). 
- I kept `createCollectibleView ` where it is [due to](https://github.com/leather-io/mono/pull/1981#discussion_r2689199264)

I could be wrong though? 🤔  (it wouldn't be the first time) .

Please let me know if you notice any other areas to improve

pnpm-lock.yaml
pete-watters: Thanks! I think this comes from the original awful monster PR and issues splitting it up

packages/features/src/activity/activity-view.ts
```diff
@@ -44,7 +66,7 @@ export function createActivityView(
     toAsset,
     title,
     caption: combinedCaption,
-    statusLabel,
+    statusLabel: statusLabel ?? '',
```
edgarkhanzadian: nullable statusLabel should be handled in the functional component where it is used, e.g. in ActivityItemComponent.

packages/features/src/collectibles/collectible-details.ts
edgarkhanzadian: sure, what i mean is that we can move the function and the tests to a place where the function is going to be called. Unless we plan to use these functions in mobile as well.

edgarkhanzadian (approved):

packages/features/src/collectibles/collectible-details.ts
pete-watters: Gotcha. They will be used on mobile too but I need to refactor it a bit to do that.  I will do that immediately and open a follow up PR

PR #1983, opened 2026-01-13
fix(extension): show validation errors on send form submit

pete-watters:
This PR addes some fixes for these issues: 
- Including feeguards from [this PR](https://github.com/leather-io/mono/pull/1953) makes changes described in [this PR](https://linear.app/leather-io/issue/LEA-3314/restore-fee-guards-on-extension) and on [this thread](https://trustmachines.slack.com/archives/C05LAP952E7/p1765287328998669) 
- LEA-3343: fixes a race condition bug loading nonce which required a double click. The button is now disabled until nonce is loaded
- LEA-3342: I added some mapping of error codes to messages (we may need to review the copy)

Here's a demo if it working for LEA-3343

https://github.com/user-attachments/assets/21fad845-227f-4e55-9c10-501c5ae410d5

### DIFF
--- a/apps/extension/src/app/pages/send/send-crypto-asset-form/family/stacks/use-stacks-common-send-form.tsx
+++ b/apps/extension/src/app/pages/send/send-crypto-asset-form/family/stacks/use-stacks-common-send-form.tsx
@@ -57,7 +57,7 @@ export function useStacksCommonSendForm({
         (acc, key) => ({ ...acc, [key]: true }),
         {}
       );
-      formikHelpers.setTouched(touchedFields);
+      void formikHelpers.setTouched(touchedFields);
       return false;
     }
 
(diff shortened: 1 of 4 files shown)


apps/extension/src/app/pages/send/send-crypto-asset-form/family/stacks/use-stacks-common-send-form.tsx
```diff
@@ -50,6 +51,16 @@ export function useStacksCommonSendForm({
   ) {
     const formErrors = await formikHelpers.validateForm();
 
+    // If there are validation errors, touch all error fields to display them
```
pete-watters, line 54: This helps prevent silent submission fails which I noticed when trying to connect to Zest with Ledger and not having STX to cover fees

kyranjamie:
How is LEA-3342 fixed here? This issue is that we need to calculate a higher fee when a memo is set, and there are no fee changes in this PR

pete-watters:
> LEA-3342

I missed the fee calculation part 🤦  I'll re-work this and add that

PR #1987, opened 2026-01-13
fix(mobile): store account name and status during creation

tigranpetrossian:
This partially rolls back [a previous change](https://github.com/leather-io/mono/commit/4a6effee18144b24f38922dc70cb036ca5474f85) that made default `status` and `name` fields for the account derived vs. stored. Ultimately derivation here is useful but not really critical, and it unfortunately caused a silent rehydration bug.

The `t` macro in initializeAccount was being called during redux-persist rehydration before lingui was initialized, causing wallet data to be silently wiped on dev reload.
Didn't seem to be causing issues in production, although I have seen related Sentry errors.

The PR can be good to go, but I'm currently figuring out why was this failing silently, and adding some relevant tests.

### DIFF
--- a/apps/mobile/src/store/accounts/accounts.write.ts
+++ b/apps/mobile/src/store/accounts/accounts.write.ts
@@ -1,3 +1,4 @@
+import { t } from '@lingui/core/macro';
 import { createAction, createEntityAdapter, createSlice } from '@reduxjs/toolkit';
 
 import { makeAccountIdentifer } from '@leather.io/crypto';
@@ -6,6 +7,7 @@ import { entitySchema, handleAppResetWithState, handleEntityActionWith } from '@
 import { userAddsAccount } from '@leather.io/state/keychains';
 import { userAddsWallet, userRemovesWallet } from '@leather.io/state/wallet';
 
+import { getWalletAccountsByAccountId } from '../utils';
 import { AccountIcon, AccountStatus, AccountStore, accountStoreSchema } from './utils';
 
 export const accountsAdapter = createEntityAdapter<AccountStore, string>({
@@ -26,7 +28,11 @@ export const accountsSlice = createSlice({
         const firstAccountIndex = 0;
         const id = makeAccountIdentifer(action.payload.wallet.fingerprint, firstAccountIndex);
 
-        accountsAdapter.addOne(state, { id });
+        accountsAdapter.addOne(state, {
+          id,
+          name: t`Account 1`,
+          status: 'active',
+        });
       })
 
       .addCase(userRemovesWallet, (state, action) => {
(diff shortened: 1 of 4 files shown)


apps/mobile/src/store/accounts/accounts.write.ts
```diff
@@ -26,7 +28,11 @@ export const accountsSlice = createSlice({
         const firstAccountIndex = 0;
         const id = makeAccountIdentifer(action.payload.wallet.fingerprint, firstAccountIndex);
 
-        accountsAdapter.addOne(state, { id });
+        accountsAdapter.addOne(state, {
+          id,
+          name: t`Account 1`,
```
edgarkhanzadian, line 33: This is interesting tho, so if i create the account when the app is in Spanish, and then revert the app to English, the name of the account would stay in Spanish right? 🤔 I wonder if that is something we need to think about

edgarkhanzadian (approved):

apps/mobile/src/store/accounts/accounts.write.ts
tigranpetrossian: You're right, I was completely blindsided by the existing setup, this won't work. Will revisit.

kyranjamie: This is solved by... deriving the default account names, no?

kyranjamie:
> and it unfortunately caused a silent rehydration bug 

Can you elaborate what this rehydration bug is? Why can't that be fixed?

I made this change as a way of simplifying the eventual migration of this over to the extension. I believe the PR to do this refactor yielded the same behaviour with a much more simple implementation.

tigranpetrossian:
I mentioned this part in the description. The rehydration runs before `i18n` has a change to initialize, part of the rehydration subroutine contains a call to `t` which fails causing the entire slice not to rehydrate.<br><br>Regardless, translation-wise this fix even more faultier than what it fixes, I completely overlooked that this wasn't properly interpolated to begin with (neither as derived, nor stored), and needs a revision.<br><br>I'll do another pass trying to extract the `t` from the store logic altogether.

apps/mobile/src/store/accounts/accounts.write.ts
tigranpetrossian: @kyranjamie Partially, but the current implementation caused a potentially critical bug due the order of bootstrapping (explained in the other comment). This should be properly solved by ideally removing `t` references from the store logic altogether. Working on that.

tigranpetrossian: ...partially, because the t`Account 1` never correctly allowed translation, that's the part I overlooked.

PR #1989, opened 2026-01-13
fix(mobile): wallet state not rehydrating after dev reload

tigranpetrossian:
A much simpler fix compared to #1987, turns out the root cause wasn't the read as I incorrectly assumed before, but the the rehydration logic itself was using the `initializeAccount` helper which in turn called `t`.
In other words: the consumption of the changes in the commit I was trying to rollback was causing the issue, not the commit itself.

### DIFF
--- a/apps/mobile/src/store/settings/settings-rehydration.ts
+++ b/apps/mobile/src/store/settings/settings-rehydration.ts
@@ -3,7 +3,7 @@ import { RootState } from '@/store';
 import { Action } from '@reduxjs/toolkit';
 import { REHYDRATE } from 'redux-persist';
 
-import { initializeAccount } from '../accounts/accounts';
+import { deserializeAccountId } from '../accounts/accounts';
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):

edgarkhanzadian (approved):
Nice catch! I do remember using the full account data in the currentAccount setting before moving to just fingerprint and index, seems like i've left this part behind during the refactor

tigranpetrossian:
Just a series of unfortunate coincidences ¯\\\_(ツ)\_/¯. Including me completely misinterpreting the whole thing.

camerow (approved):
Glad this got fixed, but perhaps some tests are in order to ensure it doesn't break in the future?

tigranpetrossian:
Agreed, this and few other parts of the store desperately need a better coverage. I'll log an issue and work on this in parallel with Swaps during the week.

PR #1995, opened 2026-01-14
chore(extension): revamp playwright tests

pete-watters:
This is some rough work from me investigating how best to update tests to account for the onboarding overhaul. I didn't spend a long time on this so its a bit rough and ready but just sharing to articulate ideas. 

It includes having a setup where we can also test stuff behind flags but I'm not sure we will do that?

There are some claude notes in [this file](https://github.com/leather-io/mono/pull/1995/changes#diff-6518121084874494bb0845208be53b42838f2d84ba0bf4afaf6f860356df18d5)

The aim is to:
- extract business logic from existing tests
- come up with easy to read BDD style `.feature` files
- share them from the `test-config` package and try and share across platform if we can 

For **mobile** we struggle to add maestro tests as we need to run a build for those which is expensive. The idea behind having these `.feature` files is that it gives us a clear picture of what we are testing and what we aren't. 

As we are refactoring the `extension` to use the same services as mobile it means that generally the behaviours should be quite similar. So if the `extension` covers similar business logic and uses the same services hopefully those tests can help block introduced issues in mobile too. 

Ideally in the future we can get `maestro` running more often in EAS using cached builds. I think taking these steps now will even help us get there as the `.feature` files are similar to maestro commands


apps/extension/tests/features/settings/settings-menu.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Settings Menu
+  As a wallet user
+  I want to access and modify wallet settings
+  So that I can customize my wallet experience
```
kyranjamie, line 4: Introducing BDD is a pretty big change to how we do things currently? Is it necessary? What value does it bring over the cost of introducing a new DSL?

kyranjamie (commented):
Thanks for sharing @pete-watters. I recognise this is a draft serves as part of a proposal, but the code changes make it real enough to cause me some alarm.

Firstly, this would be a massive PR.... there's a huge scope proposed here.

1. Introducing BDD Playwright
2. Migrating to MSW API mocks
3. LD handling
4. Testing revamp feature

These are all pretty big tasks that could easily have be their own effort/PR. There are many changes in this PR that warrant discussion. I'm open to BDD and taking this approach, if you really advocate for it. But let's discuss as a team, and get started with it separately from an already big testing effort.

Why do you propose our entire test infrastructure adapting because of this "revamp"? The improvements to the homepage are not some long-lasting permanent work. We're doing it once and it's done. It's ephemeral. So why are we forking our tests, and modifying the integration test structure? They should have no awareness this work ever took place.

I'd suggest our tests have no idea of LD or feature flags. We should only ever be testing the latest proposed version.

pete-watters:
> Thanks for sharing @pete-watters. I recognise this is a draft serves as part of a proposal, but the code changes make it real enough to cause me some alarm.
> 
> Firstly, this would be a massive PR.... there's a huge scope proposed here.
> 
> 1. Introducing BDD Playwright
> 2. Migrating to MSW API mocks
> 3. LD handling
> 4. Testing revamp feature
> 
> These are all pretty big tasks that could easily have be their own effort/PR. There are many changes in this PR that warrant discussion. I'm open to BDD and taking this approach, if you really advocate for it. But let's discuss as a team, and get started with it separately from an already big testing effort.
> 
> Why do you propose our entire test infrastructure adapting because of this "revamp"? The improvements to the homepage are not some long-lasting permanent work. We're doing it once and it's done. It's ephemeral. So why are we forking our tests, and modifying the integration test structure? They should have no awareness this work ever took place.
> 
> I'd suggest our tests have no idea of LD or feature flags. We should only ever be testing the latest proposed version.

Thanks @kyranjamie - this is honestly just a brain dump of ideas PR to help start discussions on what to do. I will close it off to avoid noise in the queue. 

I think switching the BDD will help document our user flows and help us to align tests more with mobile which we need to get  up and running. Maestro uses `.yaml` files so its similar to BDD - https://github.com/leather-io/mono/blob/dev/apps/mobile/maestro/legacy-flows/accounts-drawer.yaml 

The main idea here is to improve our existing tests as we go and come up with a pattern that we can use sustainably going forward. 

A better approach could be to instead migrate our current tests to use MSW as we go and tackle updating them as we un-flag items. 

We are all only too aware of issues with Shard 3 so I think its good to discuss strategies as we go

apps/extension/tests/features/settings/settings-menu.feature
pete-watters: I think it will help us better:
- document the feature requirements through repo code
- show us more clearly what areas are covered and which aren't
- let us to hopefully generate more tests based on text requirements
- align more to mobile

I agree that the value brought by this is debatable though as we have other priorities now. At the same time we spend countless hours waiting on certain shards to run so if we can improve that as we go here it would be great

PR #1991, opened 2026-01-14
fix(extension): bleeding shadows

fabric-8:
Ensures our gradient fade (set up as a shadow) doesn't bleed outside its surrounding/parent container.

PR #1998, opened 2026-01-15
fix(mobile): skip unrelated prepare scripts during EAS build

pete-watters:
This PR fixes a problem we introduced to the build process in PR #1928 (Dec 17) which was masked by EAS build caching. When `pnpm-lock.yaml` changed (react-router upgrade, etc.), the cache was invalidated, exposing the issue.

In PR #1928 I missed adding `--ignore-scripts` during `pnpm install` in `eas-build-post-install` to prevent unrelated app prepare scripts from running during mobile EAS builds. 

Below is more analysis from `claude` explaining the fix further

## Problem
During EAS mobile builds, `pnpm install` was running `prepare` scripts for ALL workspace packages, including `apps/web` which runs `panda codegen`. This failed because:

1. `panda codegen` loads `@leather.io/panda-preset/config` (TypeScript source)
2. That source imports from `@leather.io/tokens` (built dist)
3. During fresh install, tokens dist doesn't exist yet → build fails

The web/extension prepare scripts are completely irrelevant to mobile builds.

## Solution
Skip all prepare scripts during install, let turbo handle build orchestration:

```json
"eas-build-post-install": "cd ../.. && pnpm install --ignore-scripts && pnpm build:mobile"
```

**Why this works:**
- `pnpm build:mobile` uses turbo with `dependsOn: ["^build"]`
- Turbo automatically builds dependencies in the correct order
- Prepare scripts become redundant since turbo handles everything

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -9,7 +9,7 @@
     "android:release": "expo run:android --variant release",
     "build": "pnpm lingui",
     "check:cocoapods": "sh scripts/enforce-cocoapods-version.sh",
-    "eas-build-post-install": "cd ../.. && pnpm install && pnpm build:mobile",
+    "eas-build-post-install": "cd ../.. && pnpm install --ignore-scripts && pnpm build:mobile",
     "eas-build-pre-install": "corepack enable",
     "firebase:refresh": "firebase login && firebase apps:sdkconfig ANDROID --project leather-b5081 > google-services.json && firebase apps:sdkconfig IOS --project leather-b5081 > GoogleService-Info.plist",
     "format": "prettier . --write --ignore-path ../../.prettierignore",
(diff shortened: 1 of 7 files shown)


edgarkhanzadian:
Would it be possible to not run the web application's prepare script during the EAS build? I don't think it does anything useful for the mobile build.

pete-watters:
> Would it be possible to not run the web application's prepare script during the EAS build? I don't think it does anything useful for the mobile build.

Yes let me try that. I am working on this now as his PR https://github.com/leather-io/mono/pull/1991 made the EAS build fail. 

I did a successful build a few days ago but I think that was using a cache and this problem comes from this fix I added previously - https://github.com/leather-io/mono/pull/1928/changes .

I will see if we can avoid it entirely. I thought this should prevent that https://github.com/leather-io/mono/pull/1928/changes#r2627491844 but I will spend a bit longer on it now

pete-watters:
I managed to fix this in a much simpler way, thanks for the suggestion @edgarkhanzadian 👍

kyranjamie:
Using prepare is how panda recommend but I wish we hadn't ever used it. It's since been used elsewhere exacerbating the problem.

Learning here I think is not to use prepare/life cycle hooks at all, decoupling build requirements from package management. If web/mobile require x, their respective build systems should trigger build.

kyranjamie (approved):

PR #1999, opened 2026-01-15
fix: more updates for settings revamp

edgarkhanzadian:
Last touches for settings ui update

In the next PR would remove the feature flags, fix tests and enable this settings part of the revamp for prod
 
https://github.com/user-attachments/assets/c2e7cbdb-6627-4719-9b29-74026e855f1d


packages/ui/src/components/item-layout/item-layout.web.tsx
```diff
@@ -58,24 +56,16 @@ export function ItemLayout({
...
-          {isValidElement(titleRight) ? (
-            titleRight
-          ) : (
-            <styled.span textStyle="label.02">{titleRight}</styled.span>
-          )}
-          {isValidElement(captionRight) ? (
-            captionRight
-          ) : (
+          {titleRight && <styled.span textStyle="label.02">{titleRight}</styled.span>}
+          {captionRight && (
```
edgarkhanzadian: I assume this change broke the tests, looking into it

edgarkhanzadian: Yeah that was it. For context, we need to not render a span here if the component is nullable, s.t. we don't screw the alignment of the row
alignment

apps/extension/src/app/features/settings/settings-current.tsx
```diff
@@ -52,6 +53,76 @@ export function Settings({ canLockWallet = true }: SettingsProps) {
...
   const showLockWalletItem = canLockWallet && hasKeys && walletType === 'software';
   const showSignOutItem = hasKeys;
+  const hasDefaultInMemorySecretKey = useHasDefaultInMemoryWalletSecretKey();
+  const isWalletUnlocked = hasDefaultInMemorySecretKey || walletType === 'ledger';
+
+  const settingsItem = (
+    <DropdownMenu.Item
+      data-testid={SettingsSelectors.SettingsMenuItem}
+      onSelect={() => {
+        analytics.track('click_settings_menu_item');
```
tigranpetrossian, line 63: This is a hard-to-enforce rule, but the preference for event naming is `settings_menu_item_clicked`, to follow the object action framework.

apps/extension/src/app/pages/onboarding/set-password/set-password-current.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const { setPassword } = useKeyActions();
+  const finishSignIn = useFinishAuthRequest();
+  const navigate = useNavigate();
+  const { decodedAuthRequest } = useOnboardingState();
+
+  useEffect(() => {
+    analytics.page('view', '/set-password');
+  }, []);
+
+  const submit = useCallback(
```
tigranpetrossian, line 54: I have nothing against `useCallback` in shared contexts/state machines with many updates happening in a short window, but in an isolated page like this it's likely redundant.

apps/extension/src/app/pages/onboarding/sign-in/enter-mnemonic-current.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  DescriptionColumn,
+  TwoColumnLayout,
+} from '@app/components/layout/layouts/two-column-current.layout';
+import { MnemonicForm } from '@app/pages/onboarding/sign-in/mnemonic-form';
+
+import type { EnterMnemonicProps } from './enter-mnemonic';
+
+export function EnterMnemonic({ title, description }: EnterMnemonicProps) {
+  const [twentyFourWordMode, setTwentyFourWordMode] = useState(true);
```
tigranpetrossian, line 19: Not a dealbreaker, but a better design here would be `mode: 12 | 24` (or `mnemonicType`?`. Numbers like `12 | 24`  may or may not be the perfect discriminator, but certainly an improvement over a boolean flag tied to a particular "mode".

apps/extension/src/app/features/settings/settings-current.tsx
```diff
@@ -72,75 +143,17 @@ export function Settings({ canLockWallet = true }: SettingsProps) {
...
-                        textStyle="label.02"
-                        width="100%"
-                      >
-                        Sign out
-                      </Flag>
-                    </DropdownMenu.Item>
-                  )}
-                </>
-              )}
+              {isWalletUnlocked && settingsItem}
```
tigranpetrossian, line 146: For the current number of items this implementation is perfectly fine, but as this list grows it would be nice to have more of a config based item setup, where all conditions are declaratively defined in one place outside the render.

tigranpetrossian (approved):
@edgarkhanzadian Looks good, hopefully I haven't commented on any of the now legacy code. 
Left some minor feedback, mainly about telemetry events naming format.

apps/extension/src/app/features/settings/settings-current.tsx
edgarkhanzadian: Good point! That's a legacy event though and i'll leave it out of scope of the extension revamp. we need to go through a lot of analytics events and tweak those to a correct standard

apps/extension/src/app/pages/onboarding/set-password/set-password-current.tsx
edgarkhanzadian: unfortunately, legacy too. I mostly did this for updating TwoColumnLayout here. I definitely could make less changes in this PR and only feature flag parts of the one file instead of creating 3 files, but i am also planning on cleaning up these and other settings related ui from extensionRevamp feature flag in the next PR😅

apps/extension/src/app/pages/onboarding/sign-in/enter-mnemonic-current.tsx
edgarkhanzadian: Yeap, legacy. Only updated TwoColumnLayout here.

apps/extension/src/app/features/settings/settings-current.tsx
pete-watters: We have [a task for analytics ](https://linear.app/leather-io/issue/LEA-3095/kpis-event-tracking)in the Onboarding project. We can add a note there to audit them once we remove the old dead code.

We can try and add a custom `eslint` rule to enforce it too

apps/extension/src/app/features/settings/settings-current.tsx
pete-watters: As part of this overhaul we are specifically trying to reduce the number of items here and limit them so I think it's fine to leave as is

edgarkhanzadian: Yeap good point, i don't like the boolean mess here either. I will do another pass here when enabling this for prod

pete-watters (approved):
Great work @edgarkhanzadian 👍  I guess the next step here is removing the feature flag and deprecated legacy code and having a go at the tests?

I started vibing some test changes on [this branch](https://github.com/leather-io/mono/pull/1995) I'll add some notes there in case it helps.

apps/extension/src/app/features/settings/settings-current.tsx
camerow: @edgarkhanzadian if it's easy I would do it now as these types of things tend to get lost as we move on, the name change has the added effect of delineating the old menu uses and the new menu uses

PR #2000, opened 2026-01-15
feat: enable settings revamp

edgarkhanzadian:
First commit - Remove legacy code and a feature flag
Second commit - Update integration tests
So it can be reviewed on a commit-by-commit basis

https://github.com/user-attachments/assets/ec190d60-d02e-4bde-94f8-62e0603a6029


apps/extension/src/app/components/app-version.tsx
```diff
@@ -1,9 +1,67 @@
...
+    >
+      {children}
+    </styled.p>
+  );
+}
+
+function useAppVersion() {
+  const { pullRequestLink, isLatestBuild } = useIsLatestPullRequestBuild();
+
+  const version = useMemo(() => {
```
pete-watters: Maybe we don't need `useMemo` here? There are no dependancies and its not difficult to do this

apps/extension/src/app/features/settings/settings.tsx
```diff
@@ -1,12 +1,164 @@
-import { useFlags } from '../feature-flags';
-import { Settings as CurrentSettings } from './settings-current';
-import { Settings as LegacySettings } from './settings-legacy';
+import { useState } from 'react';
```
pete-watters, line 1: It could be good to have a `settings.tsx` file and a `settings.layout.tsx` file now?

apps/extension/src/app/pages/view-secret-key/view-secret-key.tsx
```diff
@@ -1,9 +1,66 @@
...
+
+  const header = (
+    <Header px="space.04">
+      <HeaderGrid leftCol={<HeaderBackButton />} rightCol={null} />
+    </Header>
+  );
+
+  if (showSecretKey) {
+    return (
+      <>
```
pete-watters: Maybe this part can be it's own component?

apps/extension/tests/specs/settings/settings.spec.ts
```diff
@@ -60,18 +61,14 @@ test.describe('Settings menu', () => {
...
-    const currentNetwork = await page.getByTestId(SettingsSelectors.CurrentNetwork).innerText();
-    test.expect(currentNetwork).toContain('mainnet');
-
-    await page.getByTestId(SettingsSelectors.ChangeNetworkAction).click();
-    await page.waitForTimeout(1000);
-    const networkListItems = await page.getByTestId(SettingsSelectors.NetworkListItem).all();
-    test
-      .expect(networkListItems)
-      .toHaveLength(Object.keys(WalletDefaultNetworkConfigurationIds).length);
+  test('that menu item allows changing networks', async ({ networkPage, settingsPage, page }) => {
```
pete-watters, line 64: What do you think about trying to mock the networks with MSW? 

We could have one real test of actually changing and if that passes then in the other tests we could use mocked networks? 

We use this pattern of changing networks a lot in tests so maybe we could improve it. Some tests might need actual funds but for a lot of UI stuff up until a certain point they won't

pete-watters (approved):
Great work here @edgarkhanzadian 👍  excited to see so much code disappear extensionRevamp

apps/extension/src/app/features/settings/settings.tsx
edgarkhanzadian: well it's all just one dropdown component, i don't think there is much need for that

apps/extension/tests/specs/settings/settings.spec.ts
edgarkhanzadian: i don't think this uses networks tho, it pretty much only changes the redux network state of the application

apps/extension/src/app/pages/view-secret-key/view-secret-key.tsx
edgarkhanzadian: Yeap, done!

apps/extension/tests/specs/settings/settings.spec.ts
pete-watters: I thought for a lot of tests it changes it to `Testnet` and uses the test wallet testnet funds for opening the send form?

PR #2001, opened 2026-01-15
fix(mobile): BTX/STX loading states

pete-watters:
This PR:
- fixes a bug found testing the mobile release candidate where the wrong loading state was shown for BTC/STX token details
    - an early return `activity.state === 'loading'` meant it was showing the activity loading screen instead
- adds a fix to activity key to ensure uniqueness by also using account fingerprint
    - if there are sub wallet account transfers the previous key could have duplicates
    - now we also include the account fingerprint to address this 
    - `receive` assets can have multiple operations to the same account e.g. a transaction with multiple outputs to different addresses in the same account
    
- Fixes this UI issue with activity lists https://linear.app/leather-io/issue/LEA-3362/mobile-activity-page-and-scrolling-are-broken

#### Bug

Wrong loader 

https://github.com/user-attachments/assets/14ee6834-e620-400b-b471-66c192209622

#### Fix

Here's the correct loading state + the key warning 

https://github.com/user-attachments/assets/1db67bfc-05ca-424d-b63e-b53b611215aa

### DIFF
--- a/apps/mobile/src/features/activity/activity-list.tsx
+++ b/apps/mobile/src/features/activity/activity-list.tsx
@@ -29,7 +29,7 @@ export function ActivityList({ data, header }: ActivityListProps) {
 
   if (data.state === 'error') {
     return (
-      <Screen.FlashList
+      <Screen.List
         data={[]}
         keyExtractor={keyExtractor}
         renderItem={renderItem}
@@ -40,7 +40,7 @@ export function ActivityList({ data, header }: ActivityListProps) {
   }
 
   return (
-    <Screen.FlashList
+    <Screen.List
       data={data.value}
       renderItem={renderItem}
       keyExtractor={keyExtractor}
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):

apps/mobile/src/features/activity/activity-list.tsx
```diff
@@ -29,7 +29,7 @@ export function ActivityList({ data, header }: ActivityListProps) {
 
   if (data.state === 'error') {
     return (
-      <Screen.FlashList
+      <Screen.List
```
edgarkhanzadian, line 32: Why remove FlashList here tho?

pete-watters: For some reason we started seeing these [weird spacing issues in the activity list](https://trustmachines.slack.com/archives/C05LAP952E7/p1768480135672339) [LEA-3362](https://linear.app/leather-io/issue/LEA-3362/mobile-activity-page-and-scrolling-are-broken )

I took a look at this and tried a couple of fixes (using a consistent estimated item height) but no luck.

The activity list on token details pages is appearing correctly and that is using `Screen.List` instead of `Screen.FlashList` so I switched this temporarily to fix that UI bug in the short term. 

`Screen.FlashList` is better for performance I think but we are trying to get a release over the line so I think it's OK to sacrifice performance for now to unblock release? Let me know what you think.

pete-watters:
@edgarkhanzadian - I'm going to merge this so I can prep a mobile release candidate. 

See [comment here](https://github.com/leather-io/mono/pull/2001#discussion_r2697962186) RE `FlashList`. 

I'll happily re-add it soon if desired but in the interim I want to have a good release candidate for testing and a quicker fix like this is favourable for now based on other priorities

PR #2003, opened 2026-01-15
feat: add account collectibles query

pete-watters:
This PR adds a collectibles query config to the features package


packages/queries/src/collectibles/account-collectibles.query-config.ts
```diff
@@ -0,0 +1,45 @@
...
+function createAccountCollectiblesKeyParams(request: AccountRequest) {
+  const { id, bitcoin, stacks } = request.account;
+  return [
+    id.fingerprint,
+    id.accountIndex,
+    bitcoin?.taprootDescriptor ?? null,
+    bitcoin?.nativeSegwitDescriptor ?? null,
+    bitcoin?.zeroIndexNativeSegwitPayerAddress ?? null,
+    bitcoin?.zeroIndexTaprootPayerAddress ?? null,
+    stacks?.stxAddress ?? null,
```
kyranjamie: why is all this data going in the key?

pete-watters: Thats a mistake, sorry! I'll fix it

packages/queries/src/collectibles/account-collectibles.query-config.ts
```diff
@@ -0,0 +1,37 @@
...
+  getCollectiblesService,
+} from '@leather.io/services';
+
+import { createServiceQueryKey } from '../shared/query-key.factory';
+import { collectiblesQueryOptions } from '../shared/query-options';
+
+export function createAccountCollectiblesQueryKey(request: AccountRequest, settings: UserSettings) {
+  return createServiceQueryKey(
+    'collectibles-service--get-account-collectibles',
+    [
```
pete-watters: @kyranjamie I've changed this but I'm not sure if this is still optimal. Can you advise on this?  Should I just use `[request]` as the key?

packages/queries/src/shared/query-options.ts
```diff
@@ -45,3 +45,12 @@ export const marketHistoryQueryOptions = {
   staleTime: 30000,
   gcTime: 30000,
 } satisfies Partial<UseQueryOptions>;
+
+export const collectiblesQueryOptions = {
+  refetchOnReconnect: false,
+  refetchOnWindowFocus: false,
+  refetchOnMount: true,
+  retryOnMount: false,
+  staleTime: minutesInMs(5),
```
pete-watters, line 54: Here I'm not sure if we should use:
- `oneMinInMs * 5`
- `minutesInMs(5)`

packages/queries/src/collectibles/account-collectibles.query-config.ts
alexp3y: @pete-watters You should just use `[request]`. Whatever is included in the query key factory signature should be included in the dependency array. Otherwise, any specific field combination becomes hidden hidden configuration, i.e. The assumption should be that if I'm passing a value into the query key factory, any changes to it will trigger a refetch.

packages/queries/src/shared/query-options.ts
alexp3y: IMO `minutesInMs(x)` is a more readable convention and we have helper functions for each time unit (seconds, minutes, weeks, etc...). 

The multiplication can get cumbersome as there's no good rule on how / when to combine different constants. You can easily end up with lengthy calculations. E.g. "2 months" becomes represented by something like `oneMinInMs * 60 * 24 * 30 * 2`

packages/queries/src/collectibles/account-collectibles.query-config.ts
pete-watters: OK thanks @alexp3y 👍

packages/queries/src/shared/query-options.ts
pete-watters: thanks!

edgarkhanzadian (approved):

PR #2006, opened 2026-01-15
POC: support ledger derivation paths

kyranjamie:
Following the [conversation with King yesterday on Slack](https://trustmachines.slack.com/archives/C05114YP8CV/p1768410439035709), @mica000's proposed callout to explain the situation to users , and the several other occasions of this limitation being noted as a pretty damning blocker, I've POCd the base functionality we need to toggle between classic Stacks derivation paths and the Ledger ones.

cc/ @fabric-8 @mica000 would like to jam on how we can integrate this. Inclined to suggest we have a toggle within the settings menu. Only visible when Ledger connected. For new users, we'll always pull both keys. For existing users, we'll have to prompt the user to add the keys with the modal connection flow. I think I would not try to advertise this feature too much. It's there if needed, and for users that explore settings, but think I would not highlight too much.

There are some annoying consequences to this allowing this toggle. For example, users already connected to dApps, that then change to Ledger addresses. Future RPC requests will use the Ledger path keys but expect Stacks path keys. I suggest we also clear existing sign-ins when toggling (so we should also communicate that). I think this should have it's own page within settings. It is not something users should do regularly, so intentional friction is a good idea. 

https://github.com/user-attachments/assets/143dcad6-86fd-4377-bd96-07b651c7a000

### DIFF
--- a/apps/extension/src/app/debug.ts
+++ b/apps/extension/src/app/debug.ts
@@ -72,6 +72,9 @@ const debug = {
   resetInscriptionState() {
     store.dispatch(settingsSlice.actions.resetInscriptionState());
   },
+  toggleStacksDerivationPathPreference() {
+    store.dispatch(settingsSlice.actions.toggleStacksAccountDerivationPreference());
+  },
 };
 
 export function setDebugOnGlobal() {
--- a/apps/extension/src/app/features/ledger/flows/request-stacks-keys/request-stacks-keys.utils.ts
+++ b/apps/extension/src/app/features/ledger/flows/request-stacks-keys/request-stacks-keys.utils.ts
@@ -2,7 +2,10 @@ import { bytesToHex } from '@noble/hashes/utils';
 import * as secp from '@noble/secp256k1';
 import StacksApp from '@zondax/ledger-stacks';
 
-import { makeStxDerivationPath } from '@leather.io/stacks';
+import {
+  makeStacksAccountDerivationPath,
+  makeStacksAccountLedgerCompatibleDerivationPath,
+} from '@leather.io/stacks';
 import { delay } from '@leather.io/utils';
 
 import { getIdentityDerivationPath } from '@shared/crypto/stacks/stacks-address-gen';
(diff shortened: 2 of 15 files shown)


edgarkhanzadian:
i'd suggest we only pull either ledger path keys or leather path keys, by asking users in the beginning of the setup. IMO that would simplify unexpected scenarios like connected dApps

kyranjamie:
> i'd suggest we only pull either ledger path keys or leather path keys, by asking users in the beginning of the setup. IMO that would simplify unexpected scenarios like connected dApps

Mmm, yeah this would make it a one-time thing rather than something configurable

mica000:
@kyranjamie @edgarkhanzadian What do you think of this solution? This would only be visible for Ledger user's

https://www.figma.com/design/VQPOPkkrfVcf0BEd9kz3HZ/%F0%9F%A6%BE-Ledger--Onboarding---Transaction-signing-2024?node-id=3327-55978&t=MFg7mBWoRw3fpydg-4

The idea is to have two educational callouts: one for Ledger users and one for Stacks users. Each callout would link to Account derivation path in Settings, where users can choose between Stacks (the protocol default) and Ledger.

For confirmation, we could show a toaster after selection, though the selected state alone might be sufficient.

In the Setting's page, there could be an entry point as well, called "Account derivation path" to allow users to access the page, instead of only via the call out.

As for the article: we can redirect them to this one: https://app.leather.io/support/guide/why-your-ledger-account-might-not-show-up-in-leather

mica000:
Added to the designs:
- Step at the onboarding to pick the Preferred address standard
- On the Account switcher, removed the call out to something less intrusive entry point (Tooltip up hovering "Select account")
- Added confirmation step upon switching account derivation path

Please take a look and let me know your thoughts :slightly_smiling_face:

Designs here: https://www.figma.com/design/VQPOPkkrfVcf0BEd9kz3HZ/%F0%9F%A6%BE-Ledger--Onboarding---Transaction-signing-2024?node-id=3327-5992 (

camerow:
After today's discussion we landed on supporting this from the onboarding create wallet flow as an initial choice. We'll add a section to the account screen as well informing users how to switch between the options by restoring their wallets.

mica000:
Final Designs here: 
- Onboarding Ledger modal: https://www.figma.com/design/VQPOPkkrfVcf0BEd9kz3HZ/%F0%9F%A6%BE-Ledger--Onboarding---Transaction-signing-2024?node-id=3392-107515&t=MFg7mBWoRw3fpydg-4
- Tooltip for Account switcher: https://www.figma.com/design/VQPOPkkrfVcf0BEd9kz3HZ/%F0%9F%A6%BE-Ledger--Onboarding---Transaction-signing-2024?node-id=3327-55978&t=MFg7mBWoRw3fpydg-4
- Article: Im updating the post to explain how to switch between the options by restoring their wallets.

PR #2013, opened 2026-01-16
refactor: remove bits of jotai part 4

edgarkhanzadian:
This one should be last for extension


apps/extension/src/app/store/ui/ui.hooks.ts
```diff
@@ -1,7 +1,36 @@
...
+  return {
+    isLoading: loadingState.value === 'loading',
+    setIsLoading() {
+      dispatch(uiActions.setIsLoading(key));
+    },
+    setIsIdle() {
+      dispatch(uiActions.setIsIdle());
+    },
+  };
+}
```
edgarkhanzadian, line 23: Tried keeping the key value here as well not to deviate too much from the original implementation. This would allow a developer to still see what created the loading state

kyranjamie (approved):

PR #2017, opened 2026-01-19
refactor(mobile): move the swap state machine into state package

tigranpetrossian:
This PR is mostly a noop refactor moving the Swap state into the package with no changes besides imports.

Moving did uncover some circular dependency issues though. Currently looking into eliminating those and figuring out why didn't these light up in `mobile`.


packages/state/package.json
```diff
@@ -25,20 +25,41 @@
   "exports": {
     ".": "./dist/index.js",
     "./wallet": "./dist/wallet/index.js",
-    "./keychains": "./dist/keychains/index.js"
+    "./keychains": "./dist/keychains/index.js",
+    "./swap": "./dist/swap/index.js"
```
pete-watters, line 29: I'm curious if `swap` would be netter shared via the `features` package? 

Of course the state is perfect being shared from here but perhaps the utilities could go into the `features` package.

pete-watters (approved):
Great work here @tigranpetrossian this is looking brilliant

packages/state/package.json
tigranpetrossian: I frankly don't see a reason for scattering it over multiple packages. What would be the benefit of that?

pete-watters: Just to follow the convention we started with `features` and mirror the other areas like `activity`, `balance`, `collectibles` etc. 

That way we keep `state` and pure utility functions seperate

tigranpetrossian: I'm not sure that convention applies here, but maybe I'm missing something. What's the practical benefit for doing that?

pete-watters: Lets forget it. I was just asking as perhaps some of the utilities may be useful elsewhere outside of `state`

PR #2018, opened 2026-01-19
fix(ci): mobile build

pete-watters:
This PR builds on from https://github.com/leather-io/mono/pull/1998 which did in fact not actually resolve the issue of EAS build breaking due to it trying to build for other platforms. 

In that PR ` --ignore-scripts` was introduced but that's needed by _lingui_ to compile translations and the `mobile` `postinstall` would not run

This PR implements @kyranjamie 's [feedback around avoiding lifecycle hooks](https://github.com/leather-io/mono/pull/1998#issuecomment-3754033017) AKA `prepare` and cleans up our CI

The change here:
- [x] Removes lifecycle hooks
    - banishes the use of `prepare` and relabels as `prepare:build`
    - `prepare` was causing problems as it runs automatically during install
    - this creates tight coupling between package installation and build requirements. 
 - [x] Adds explicit build scripts 
    - The use of `prepare:build` script allows us avoid automatic run in the NPM lifcycle
- [x] Optimizes mobile build 
    - `packages/ui/build-conditional.sh ` is introduced to conditionally builds based on `BUILD_TARGET`
    -  `BUILD_TARGET=mobile` only builds native UI (skips web/panda)
    -  Turbo's dependsOn: `["^build"]` handles the build order 
- [x] Adds a new CI job that tests the mobile build will work in GIT actions 
    - this saves us wasted effort and build resources in EAS 
    - it also lets us know earlier if potential issues are introduced


docs/BUILD_TARGET.md
```diff
@@ -0,0 +1,395 @@
+# Build System Quick Reference
```
pete-watters, line 1: This is a Claude generated doc explaining the problem with our old setup. I can keep it for reference or remove it

packages/ui/build-conditional.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
pete-watters, line 1: This script helps us intentionally only build what we need from the `ui` package

scripts/test-mobile-build.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
pete-watters, line 1: I'm adding this as a faster feedback loop to know if something has happened to break the mobile build in EAS. Right now we don't have an idea about this until we do a release. 

My hunch is that these failures will be less likely after this PR but I think it's still nice to have some PR feedback on issues that are breaking the mobile build.

camerow (approved):

PR #2019, opened 2026-01-19
chore: update app-error-boundary.tsx

kyranjamie:
update link in crash report

thanks @314159265359879


edgarkhanzadian (approved):

pete-watters:
@kyranjamie FYI just merging this as its a small but important fix. I rebased it with the fix to add an override for tar yesterday

PR #2020, opened 2026-01-19
fix: tar cve

kyranjamie:
Resolves issue with tar package


pete-watters (approved):
Thanks @kyranjamie 👍  I had started fixing this before checking in

PR #2021, opened 2026-01-19
fix(mobile): adjust swap activity icon alignment, simplify styling

tigranpetrossian:
image


pete-watters (approved):
Thanks for fixing this 👍

PR #2024, opened 2026-01-20
fix(mobile): remove token details flag

edgarkhanzadian:
Remove token details feature flag on mobile.


tigranpetrossian (approved):

pete-watters:
@edgarkhanzadian @tigranpetrossian I was thinking it better to keep this flag until we do the release? That way if we notice a lot of issues we could quickly turn it off again without re-releasing?

What are your thoughts here @camerow ?

camerow:
Token details is low impact if there are issues - not something I think we should worry to much about the need to do a "rollback" rather we should aim to quickly fix any issues found.

PR #2025, opened 2026-01-20
fix(crypto): leading zero loss in fingerprint calc

kyranjamie:
Extracting this from multi-wallet work. [Reported on Slack.](https://trustmachines.slack.com/archives/C05LAP952E7/p1768926469721129)

The first commit I've written myself, the second I've used Claude and it isn't ready. I'm currently trying to get a migration handled where we:
1. Migrate the redux store to use padded fingerprint values
2. Migrate the key by fingerprint encrypted store

One challenging I'm facing here is not having the tooling set up to debug this easily. On extension, I have some debug buttons to flash store to a past version, and reload, such that I can write some migration code, test, make changes, try again. I haven't gotten this far yet. 

@edgarkhanzadian I would love your help here if possible. I'm not attached to any of the code here so you can start again if needed. The main migration of the store is working, but I believe the remaining piece of this is the secure store.

**Before – `dev`**

2026-01-20-000618@2x

**After – `fix/fingerprint`**

image

### DIFF
--- a/apps/mobile/src/store/key-store.ts
+++ b/apps/mobile/src/store/key-store.ts
@@ -10,12 +10,12 @@ import {
   getMnemonicRootKeyFingerprint,
   makeAccountIdentifer,
 } from '@leather.io/crypto';
-import { stacksRootKeychainToAccountDescriptor } from '@leather.io/stacks';
+import { stacksRootKeychainToAccountDescriptorV2 } from '@leather.io/stacks';
 import { findHighestAccountIndexOfFingerprint, userAddsAccount } from '@leather.io/state/keychains';
 
 import { userTogglesHideAccount } from './accounts/accounts.write';
 import { useBitcoinAccounts } from './keychains/bitcoin/bitcoin-keychains.read';
-import { mnemonicStore } from './storage-persistors';
+import { mnemonicStore } from './secure-store/mnemonic-store';
 import { useAppDispatch } from './utils';
 import { useWallets } from './wallets/wallets.read';
 
@@ -54,6 +54,7 @@ export function useKeyStore() {
 
     createNewSoftwareWallet() {
       const mnemonic = generateMnemonic();
+      // FIXME: biometrics should not be automatically set to true here. We need to check the security preference first.
       return this.restoreWalletFromMnemonic({ mnemonic, biometrics: true });
     },
 
@@ -145,7 +146,7 @@ export function useKeyStore() {
 
       const stacksKeychainDescriptors = [
         {
-          descriptor: stacksRootKeychainToAccountDescriptor(rootKeychain, nextAccountIndex),
+          descriptor: stacksRootKeychainToAccountDescriptorV2(rootKeychain, nextAccountIndex),
           chain: 'stacks' as const,
         },
       ];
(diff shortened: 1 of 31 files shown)


packages/crypto/src/keychain.ts
```diff
@@ -45,17 +56,36 @@ export async function deriveChildKeychainFromMnemnonic(
...
 /**
  * Gets keychain fingerprint directly from mnemonic. This is useful for
  * referencing a mnemonic safely by an identifier.
  */
-export async function getMnemonicRootKeyFingerprint(mnemonic: string, passphrase?: string) {
-  const keychain = deriveRootBip32Keychain(await deriveBip39SeedFromMnemonic(mnemonic, passphrase));
+export function getMnemonicRootKeyFingerprint(mnemonic: string, passphrase?: string) {
+  const keychain = deriveRootBip32Keychain(deriveBip39SeedFromMnemonicSync(mnemonic, passphrase));
+  return fingerprintAsNumberToHex(keychain.fingerprint);
+}
```
edgarkhanzadian: why shouldn't we continue using async function here?

edgarkhanzadian (commented):
Thank you for the fix! I'll continue on the migrations from here.

packages/crypto/src/keychain.ts
edgarkhanzadian: considering this is a hot fix, i'd say we should keep this change as simple as possible and maybe introduce synchronous version in later updates?

kyranjamie: Appreciate it @edgarkhanzadian, thanks. 

We use this fn in redux selectors that aren't async, was main motivation for this.

apps/mobile/src/store/key-store.ts
```diff
@@ -54,6 +54,7 @@ export function useKeyStore() {
 
     createNewSoftwareWallet() {
       const mnemonic = generateMnemonic();
+      // FIXME: biometrics should not be automatically set to true here. We need to check the security preference first.
```
edgarkhanzadian, line 57: Should be done separately as not part of migration

edgarkhanzadian:
- Added migrations for both redux persist and expo secure store
- Added tests for migrations and for secure store
- Manually tested on both simulator and physical device

apps/mobile/src/store/secure-store/mnemonic-store.ts
```diff
@@ -0,0 +1,102 @@
...
+      getSecureStoreConfig(biometrics)
+    );
+  }
+
+  async function getMnemonic() {
+    const { mnemonic, passphrase, storeKeys } = await searchForMnemonic(secureStoreKeys);
+    if (!storeKeys || !mnemonic) {
+      // we need to error out here somehow
+      throw new Error('No mnemonic was found');
+    }
```
kyranjamie: Not just return null/undefined if not found?

kyranjamie (commented):
LGTM ✅, appreciate your help finishing this @edgarkhanzadian 

Will remove the v2 function alongside multi-wallet work

edgarkhanzadian (approved):

apps/mobile/src/store/secure-store/mnemonic-store.ts
edgarkhanzadian: hmm good point, should we send back a nullable mnemonic value on getMnemonic function or error it out inside getMnemonic. Currently, we do error out anyway because we pass mnemonic and passphrase through the `mnemonicSchema`. I made it that way s.t. we won't need to check if mnemonic is nullable or not when we deal with signing of transactions.

I've removed this piece of code from getMnemonic but i have left zod schema in the next commit: https://github.com/leather-io/mono/pull/2025/commits/5bd43e6c44185bea524d4be8dd45a880b1b598d0
We can discuss error handling later and implement it in a separate PR.

PR #2058, opened 2026-01-27
Will/maestro smoke tests

camerow:
This PR ensures we can run smoke tests more consistently - this seems to have fingerprinting working correctly as well, which will reduce much of the cost and build times. 

The core idea here is to use fingerprints to target specific builds, push js bundles to those builds via the `cicd` channel. That flow should take < 9 minutes and will only run when there are changes to `apps/mobile/` or `packages/` so extension specific work and web specific work is safe. 

In the case a full build must run, it should only ever need to be run once, and subsequent builds will use the `eas update` command to update the cached bundle unless changes are made to the mobile configs/packages.

Here is a fresh build, with changes to some app configs needed to support deep linking to expo builds on ios in maestro tests:

### DIFF
--- a/apps/mobile/.eas/workflows/maestro-smoke-test.yml
+++ b/apps/mobile/.eas/workflows/maestro-smoke-test.yml
@@ -1,0 +1,104 @@
+name: Maestro Smoke Test
+
+on:
+  pull_request:
+    branches:
+      - dev
+    paths:
+      - apps/mobile/**
+      - packages/**
+
+defaults:
+  tools:
+    node: 22.15.0
+    pnpm: 10.10.0
+
+jobs:
+  fingerprint:
+    name: Fingerprint
+    type: fingerprint
+
+  run_eas_update:
+    name: EAS Update
+    type: update
+    params:
+      platform: all
+      channel: cicd
+
+  get_android_build:
+    name: Check Android Build
+    needs: [fingerprint]
+    type: get-build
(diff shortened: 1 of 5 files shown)


pete-watters (approved):
Nice work here @camerow 👍  This change looks good to me and if the fingerprint is working it's a huge leap forward in helping us move towards more continuous delivery with less release anxiety. 

This new code looks good but if this is to be the new smoke tests we should probably retire the old ones while we are at it? See:
- `apps/mobile/.eas/workflows/ios-maestro-test.yml`
- `apps/mobile/.eas/workflows/android-maestro-test.yml`
- `maestro/flows/smoke-tests.yaml` 

Those were added previously back in [this PR](https://github.com/leather-io/mono/pull/1410) and are triggered by the PR tag - `e2e-tests`. In that PR I simplified our tests  and deprecated ones that had become out-dated. 

They were outdated as we were allowing code changes without running the tests. With this latest change to fingerprint we should be able to run the tests in CI more often and prevent them from going stale again and also catch many more regressions before it's too late. 

I have been working on a side-project to update those tests I hope to complete soon. That will enable us to also run the tests we have under `maestro/flows/legacy-flows`. 

The slow part of running the tests has been waiting for the builds which I believe this fingerprint fix should solve. 

We can assess speed of enabling the remaining tests and focus on extending the tests to core flows

PR #2067, opened 2026-01-28
feat(extension): go live with new extension home tab layout

pete-watters:
This PR goes live with changes made to the extension home page tabs:
- introducing a third tab for collectibles - while keeping existing `AssetList`, `Collectibles` + `Activity`
- fixes a long standing bug with the `Receive` dialog not opening properly from the activity tab
- refactors the `Send Inscription` flow so it can be opened properly from a tab

**NOTEs**
- the PR is perhaps slightly too big. I tried to separate both commits into their own PR but adding the collectibles tab is a breaking change for the existing send inscription flow
- I attempted to change the `Send Inscription` flow to deprecate route state and allow passing inscription ID as a route param however this would require further changes and the improvement gained is debatable

**DEMO**

https://github.com/user-attachments/assets/0071c2cf-3f28-49b0-8bb0-f617aa6ac57d

- `Send Inscription` took a long time to complete so I cut off the video
- `Activity` is also taking a long time to load 

It did work though:
Screenshot 2026-01-28 at 11 08 12
Screenshot 2026-01-28 at 11 06 38


apps/mobile/src/components/home/components/asset-tabs.tsx
```diff
@@ -27,7 +27,7 @@ export function AssetTabs({ listTab, setListTab }: AssetTabsProps) {
         />
         <TabButton
           isActive={listTab === 'collectibles'}
-          title={t`NFTs`}
+          title={t`Collectibles`}
```
pete-watters, line 30: This is an intended change to copy on `mobile`

apps/extension/src/app/routes/hooks/use-background-location.ts
```diff
@@ -0,0 +1,25 @@
+import { useLocation } from 'react-router';
```
pete-watters, line 1: Introducing this to help store the correct page background for this routable modal. 

We already have another similar hook `useBackgroundLocationRedirect`. 

I spent some time investigating a way of allowing 'Send Inscription' to work from anywhere by passing the inscription ID as a route param. 

This could help us move away from using route state but it involves making changes to working code that is un-necessary so thought it best to just leave it as is.

pete-watters (commented):
This PR is ready for review. @mica000 @fabric-8 @314159265359879  - if you have some time to test the build for issues / visual defects please do.

I just saw the design says `Tokens` now instead of `Assets`, I am working on the `Assets` updates now so can fix that with those.

edgarkhanzadian (approved):

pete-watters:
> This PR is ready for review. @mica000 @fabric-8 @314159265359879 - if you have some time to test the build for issues / visual defects please do.
> 
> I just saw the design says `Tokens` now instead of `Assets`, I am working on the `Assets` updates now so can fix that with those.

I removed the label and will merge this. I have a queue of PRs coming so I will get a design review on more things once ready

PR #2072, opened 2026-01-29
feat(extension): update assets design + fix bug with fallback avatars

pete-watters:
This PR updates the homepage assets design + fixes a bug with fallback avatars.

✨ Features:

- [x] Update design of new Asset list on homepage
- 
- [x] Add better E2E test coverage around the asset list 
- [x] Fix a bug with the fallback avatar not correctly showing a coloured avatar

💻 Demo: 

Here's a demo of the new UI beside the old (new on the right)

https://github.com/user-attachments/assets/f98db2d5-f23e-4047-852d-0e637259552a


apps/extension/src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
```diff
@@ -31,12 +32,19 @@ export function Sip10TokenAssetList({
...
   assetRightElementVariant,
   setHasManageableTokens,
+  setIsLoading,
 }: Sip10TokenAssetListProps) {
   const sip10s = useSip10AccountBalance(accountIndex, {
     includeHiddenAssets: assetFilter === 'all',
   });
   const { isEnabled } = useManagedSip10Tools(accountIndex);
 
+  useEffect(() => {
```
pete-watters: Adding this as sometimes when opening `Manage tokens` it flashes up 'No tokens' while its loading

packages/ui/src/components/avatar/sip10-avatar-icon.native.tsx
```diff
@@ -35,6 +40,7 @@ export function Sip10AvatarIcon({
     <Avatar
       image={getSip10AvatarImage({ imageCanonicalUri, contractId, name })}
       imageAlt={name}
+      fallback={getAvatarFallbackText(name)}
```
pete-watters, line 43: This fallback was mistakenly removed which meant for some avatars like - `BOB` we were showing a white avatar. 

Image

This fix makes sure it's using the fallback generated avatar:

Image

I reached out to @314159265359879 who will help me add some more of these missing avatars to the new image hosting solution we have. Then we can replace these without code changes

apps/extension/src/app/pages/home/components/assets-switch.tsx
```diff
@@ -1,9 +1,5 @@
-import { useFlags } from '@app/features/feature-flags';
-
 import { Assets } from './assets';
-import { AssetsLegacy } from './assets-legacy';
 
 export function AssetsSwitch() {
-  const { assetsRevamp } = useFlags();
-  return assetsRevamp ? <Assets /> : <AssetsLegacy />;
+  return <Assets />;
```
edgarkhanzadian: let's remove the AssetsSwitch component, wrapper seems unnecessary

apps/extension/src/app/pages/home/components/home-tabs.tsx
```diff
@@ -71,7 +71,7 @@ export function HomeTabs({ children, showCollectibles = true }: HomeTabsProps) {
         </Tabs.List>
       </Tabs.Root>
       <Suspense fallback={<LoadingSpinner pb="72px" />}>
-        <Box px={{ base: 'space.05', md: 0 }} width="100%">
+        <Box px={['space.05', 0]} width="100%">
```
edgarkhanzadian: tagging this just in case the intention was to just convert obj to array. In that case it should be `{['space.05', null, 0]}` here

apps/extension/src/app/pages/home/components/tab-content-header.tsx
```diff
@@ -0,0 +1,17 @@
...
+  rightElement?: ReactNode;
+}
+
+export function TabContentHeader({ leftElement, rightElement }: TabContentHeaderProps) {
+  return (
+    <Flex justifyContent="space-between" alignItems="flex-start">
+      {leftElement}
+      {rightElement}
+    </Flex>
+  );
```
edgarkhanzadian: I'd say imo creating wrapper (that is used once) for just a Flex container with 2 children seems quite unnecessary, can we remove it and write Flex component in AssetsTabHeader?

apps/extension/src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
edgarkhanzadian: It does seem like it is possible to move this code (might be wrong but flagging this):

```
          {isLoading && (
            <Stack h="100%" justify="center" align="center">
              <Spinner size="md" />
            </Stack>
          )}

          {!isLoading && !hasManageableTokens && (
            <Stack h="100%" justify="center" align="center">
              <Caption>No tokens found</Caption>
            </Stack>
          )}
```
To Sip10TokenAssetList and then move isLoading state into Sip10TokenAssetList. That way we'd ditch the props drilling with isLoading state

apps/extension/src/app/pages/home/components/assets-switch.tsx
pete-watters: Ah yes good catch, thanks!

apps/extension/src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
pete-watters: I will get rid of this altogether as it seems to be OK when moving to a page in the next PR  - https://github.com/leather-io/mono/pull/2073

apps/extension/src/app/pages/home/components/tab-content-header.tsx
pete-watters: I was planning to also use it for the collectibles section but you are right, its probably overkill having a specific component for it 

Screenshot 2026-01-29 at 16 51 52

apps/extension/src/app/pages/home/components/home-tabs.tsx
pete-watters: it was, thanks!

pete-watters:
Thanks @edgarkhanzadian ! Some great feedback there and things I missed. 

I have pushed two commits:
- [this one addresses your feedback](https://github.com/leather-io/mono/pull/2072/changes/021c7d0c26be812e4ff5137424dc2bc9448a5021)
- when re-reviewing the code again I realise that we are changing the name to `Token` in the UI so I took this chance to rename everything `Asset` - `Token` in [this commit](https://github.com/leather-io/mono/pull/2072/changes/592ff7e2551c70d222e954bb0fe7114245fc9228)

The second commit could be adding too much noise here so I can remove it from here if you want and add it in a follow up

alexp3y (approved):
LGTM 🚀

PR #2075, opened 2026-01-29
fix: remove pnpm extensions

edgarkhanzadian:
We don't need any of package extensions and most of the overrides anymore + those were actually breaking the dev expo build on ios.


tigranpetrossian (approved):

pete-watters (approved):
Nice @edgarkhanzadian

kyranjamie:
nice!

PR #2082, opened 2026-02-03
feat(extension): implement token details

pete-watters:
This PR adds the token details feature to the extension. It's big but the changes are focused and related:                                                                     
                                                                                               
  1. Token details page - The core feature                                                     
  2. 404/error handling - Necessary UX for when tokens don't exist or routes are invalid
  3. Refactor to improve routable modals app wide

- Add support for token details pages on `extension` 
- Add new UI for handling 404 pages in general and add some bug fixes
    - I added a 404 for unknown tokens
    - I fixed a bug to show a similar UI for unknown pages on the extension Home 
- Allow ability to view zero balance token details pages in case users bookmark URLs (not on mobile) 
- Global change to handling of background URLs for routable modals now we are showing from different source pages

**Demo: Empty Wallet**

https://github.com/user-attachments/assets/35c540f0-0b31-46b1-99b9-49cf3e609d69

**Demo: Populated Wallet**

https://github.com/user-attachments/assets/4855b3c2-e335-415b-bfc6-29e0805e8522

**Demo: Receive Modal**

https://github.com/user-attachments/assets/0ecc186e-26df-4637-86a7-58bdc17bfc26

**Demo: 404 Page**

https://github.com/user-attachments/assets/96e3a284-98b9-4011-b8cd-b9b903bc9cae


apps/extension/src/app/features/token/token-details.layout.tsx
```diff
@@ -0,0 +1,121 @@
...
+  layer: string;
+  contractDetails?: string;
+  descriptionText?: string;
+  balancesContent: ReactNode;
+  activity: ActivityView[];
+  isBuyEnabled?: boolean;
+  isSwapEnabled?: boolean;
+}
+
+export function TokenDetailsLayout({
```
pete-watters, line 42: I considered refactoring this to use composition and avoid prop drilling however that would mean there is more code repetition across token-specific layouts. 

While not ideal to drill these props I believe its a reasonable balance that reduces code duplication and applies consistent formatting

apps/extension/src/app/pages/home/home.tsx
```diff
@@ -20,6 +21,11 @@ import { AccountCard } from './components/account-card';
 import { HomeTabs } from './components/home-tabs';
 import { Tokens } from './components/tokens';
 
+function HomeNotFound() {
```
pete-watters, line 24: The extension uses React Router with a `/*` catch-all pattern on the Home route to support the modal overlay pattern (e.g., receive dialog opening over the home page). 

This means a traditional 404 route at the root level would never be reached for authenticated users.

Instead of fighting the routing architecture, we:

1.  Show the "Lost in the stack" content within the home page tabs when users navigate to
unknown routes
2. Keep the home header, balance, and action buttons visible so users aren't completely lost
3. Provide a "Go home" button that navigates back to the tokens tab

This gives users clear feedback that the route doesn't exist while maintaining navigation context, rather than showing an empty page or breaking the modal system.

**BEFORE**

Image

**AFTER**

Image

apps/extension/src/app/routes/components/global-modal-wrapper.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <>
+      <Routes location={backgroundLocation}>
+        <Route
+          path={RouteUrls.TokenDetails}
+          element={
+            <AccountGate>
+              <TokenDetails />
+            </AccountGate>
+          }
+        />
```
edgarkhanzadian, line 30: Why do we need to copy all of these routes in GlobalModalWrapper? And just generally why do we need GlobalModalWrapper?

apps/extension/src/app/features/asset-list/bitcoin/btc-crypto-asset-item/btc-crypto-asset-item.tsx
```diff
@@ -10,15 +13,21 @@ interface BtcCryptoAssetItemProps {
   isLoading: boolean;
   isLoadingAdditionalData?: boolean;
   onSelectAsset?(symbol: string): void;
+  onOpenToken?(details: TokenDetailsProps): void;
```
edgarkhanzadian: Why do we need both onSelectAsset and onOpenToken here? Can we refactor this to have only one `on<Something>` here?

apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx
```diff
@@ -65,6 +78,7 @@ export function RunesAssetList({
           titleLeft,
           fiatBalance: formatCurrency(rune.quote.totalBalance),
           dataTestId: rune.asset.runeName,
+          onSelectAsset: onOpenToken ? () => handleSelectRune() : undefined,
```
edgarkhanzadian: This checks for onOpenToken existence for the second time (first time is done in handleSelectRune) which seems unnecessary.

apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx
```diff
@@ -32,20 +36,29 @@ export function RunesAssetList({
...
   useEffect(() => {
-    if (runes.value && runes.value.runes.length > 0 && setHasManageableTokens) {
+    if (hasRunes && setHasManageableTokens) {
       setHasManageableTokens(true);
     }
-  }, [runes, setHasManageableTokens]);
+  }, [hasRunes, setHasManageableTokens]);
 
-  if (runes.state !== 'success' && !runes.value) return null;
+  if (!hasRunes) return null;
```
edgarkhanzadian: Generally it might be dangerous to add a derived variable to useEffect's dependency list. As this is a boolean it probably won't rerun useEffect much, but i'd still suggest moving hasRunes to useEffect.

apps/extension/src/app/features/asset-list/stacks/sip10-token-list/sip10-token-item.tsx
```diff
@@ -13,11 +15,13 @@ interface Sip10TokenItemProps {
   isEnabled: boolean;
   assetRightElementVariant?: AssetRightElementVariant;
   onSelectAsset?(symbol: string, contractId?: string): void;
+  onOpenToken?(details: TokenDetailsProps): void;
```
edgarkhanzadian: Same here, please let's use one `on<Something>` here.

apps/extension/src/app/features/asset-list/stacks/sip10-token-list/sip10-token-list.tsx
```diff
@@ -22,13 +23,15 @@ interface Sip10TokenListProps {
   assetFilter?: AssetFilter;
   assetRightElementVariant?: AssetRightElementVariant;
   onSelectAsset?(symbol: string, contractId?: string): void;
+  onOpenToken?(tokenDetails: TokenDetailsProps): void;
```
edgarkhanzadian: ⬆️

apps/extension/src/app/features/asset-list/stacks/sip10-token-list/sip10-token-list.tsx
```diff
@@ -37,13 +40,15 @@ export function Sip10TokenList({
...
   useEffect(() => {
-    if (sip10s.value && sip10s.value.sip10s.length > 0 && setHasManageableTokens) {
+    if (hasSip10s && setHasManageableTokens) {
       setHasManageableTokens(true);
     }
-  }, [sip10s, setHasManageableTokens]);
+  }, [hasSip10s, setHasManageableTokens]);
 
-  if (sip10s.state !== 'success' && !sip10s.value) return null;
+  if (!hasSip10s) return null;
```
edgarkhanzadian: same here regarding the derived boolean

apps/extension/src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.tsx
```diff
@@ -11,13 +14,15 @@ interface StxCryptoAssetItemProps {
...
   onSelectAsset?(symbol: string): void;
+  onOpenToken?(details: TokenDetailsProps): void;
 }
 
 export function StxCryptoAssetItem({
   balance,
   isLoading,
   isPrivate,
   onSelectAsset,
+  onOpenToken,
```
edgarkhanzadian: ⬆️

apps/extension/src/app/features/asset-list/token-list.tsx
```diff
@@ -40,11 +41,13 @@ interface TokenListProps {
   assetRightElementVariant?: AssetRightElementVariant;
   showUnmanageableTokens?: boolean;
   onSelectAsset?(symbol: string, contractId?: string): void;
+  onOpenToken?(tokenDetails: TokenDetailsProps): void;
```
edgarkhanzadian: ⬆️

apps/extension/src/app/features/token/hooks/use-token-market-info.ts
```diff
@@ -0,0 +1,48 @@
...
+  const description = useAssetDescription(asset);
+  const priceChange = usePriceChangePercentage(asset);
+
+  const isLoading = marketData.state === 'loading';
+  const hasError = marketData.state === 'error';
+
+  const price = marketData.state === 'success' ? marketData.value.price : undefined;
+  const changePercent = priceChange.state === 'success' ? priceChange.value : 0;
+  const descriptionText =
+    description.state === 'success' ? (description.value.description ?? '') : '';
```
edgarkhanzadian, line 33:
```suggestion
  const descriptionText = description?.value?.description ?? ''
```

apps/extension/src/app/features/token/rune-token-details.tsx
```diff
@@ -0,0 +1,66 @@
...
+  quote: { availableBalance: Money };
+  account: AccountAddresses;
+}
+
+function RuneTokenDetailsContent({ asset, crypto, quote, account }: RuneTokenDetailsContentProps) {
+  const marketInfo = useTokenMarketInfo(asset);
+  const activityQuery = useActivityByAsset(account, asset);
+
+  const name = asset.spacedRuneName ?? asset.runeName;
+  const symbol = asset.symbol ?? '';
```
edgarkhanzadian: Let's not provide an empty string symbol to RuneTokenDetailsLayout. Can we pass null/undefined?

apps/extension/src/app/features/token/token-details.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const { assetId, parsedAssetId } = useMemo(() => {
+    if (!assetPath) return { assetId: null, parsedAssetId: null };
+    try {
+      const serialized = urlPathToAssetId(assetPath);
+      return { assetId: serialized, parsedAssetId: deserializeAssetId(serialized) };
+    } catch {
+      return { assetId: null, parsedAssetId: null };
+    }
+  }, [assetPath]);
```
edgarkhanzadian: Can we simplify this?

```suggestion
  const { '*': assetPath } = useParams();

  const assetId = assetPath ? urlPathToAssetId(assetPath) : null;
  const parsedAssetId = assetId ? deserializeAssetId(assetId) : null; 
```

packages/features/src/token/token-details-utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!hasChangePercent) return '—';
+
+  const prefix = getChangePrefix(changePercent);
+
+  if (priceChangeDelta) {
+    return `${prefix}${changePercent.toFixed(2)}% (${priceChangeDelta})`;
+  }
+  return `${prefix}${changePercent.toFixed(2)}%`;
+}
```
edgarkhanzadian, line 31: Summoning @tigranpetrossian to check the formatting function

apps/extension/src/app/routes/components/global-modal-wrapper.tsx
pete-watters: Good catch with this duplication - it slipped past my review. I've re-implemented it without
  route duplication.                                                                           
   
  **The problem**: 

When opening a routable modal (like receive) from different pages (Home vs Token Details), we need the correct page to render as the background behind the modal. React Router uses `backgroundLocation` state to achieve this - it tells the router "render this location as the background while showing the modal on top."

On production now this works but we only use that modal on the homepage:
Screenshot 2026-02-11 at 08 35 23

  However, because receive routes are defined at the top level (not nested under Home or Token
  Details), the router doesn't know which background page to render. Without intervention, it
  would either show nothing or the wrong page.

  **The solution**: 

`ReceiveModalWrapper` wraps the receive routes and checks for `backgroundLocation`
  state. When present, it renders the appropriate background page (Home or Token Details) based
   on that location, then renders the modal via `<Outlet />`. This keeps the receive routes
  registered only once at the top level - no duplication needed.

  The mistake I made here which you correctly flagged here was me duplicating the receive
  routes under both Home and Token Details, which is worse than this approach.



**Demo of correct background pages**: 

https://github.com/user-attachments/assets/2687cc69-c685-4231-b9f8-d4131fecd467

apps/extension/src/app/features/token/hooks/use-token-market-info.ts
pete-watters: The current code checks `description.state === 'success'` before accessing the value, which is
  the pattern we use with our query results. 
  
  The suggested optional chaining would skip that state check. I can simplify it slightly to:
  ```
  const descriptionText = description.state === 'success' ? description.value.description ?? ''
   : '';
   ```

apps/extension/src/app/features/token/rune-token-details.tsx
pete-watters: Makes sense - I'll update the layout to accept `symbol?: string` and handle the missing case in the UI, rather than passing an empty string from the data layer.

apps/extension/src/app/features/token/token-details.tsx
pete-watters: The `useMemo` with try-catch is there because `deserializeAssetId `can throw for malformed asset
  IDs (e.g., if someone manually enters a bad URL). Without the try-catch, the whole page would crash instead of showing the error state.

I could extract it to a helper that returns null on error instead, which would make this cleaner:
```
  const assetId = assetPath ? urlPathToAssetId(assetPath) : null;
  const parsedAssetId = assetId ? safeDeserializeAssetId(assetId) : null;
  ```

File: apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx (hunk 5 above)
Comment (pete-watters) [reply apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx]: Good catch, thanks, I'll remove the redundant check inside `handleSelectRune`.

File: apps/extension/src/app/features/asset-list/bitcoin/btc-crypto-asset-item/btc-crypto-asset-item.tsx (hunk 4 above)
Comment (pete-watters) [reply apps/extension/src/app/features/asset-list/bitcoin/btc-crypto-asset-item/btc-crypto-asset-item.tsx]: `onSelectAsset` is used by the send flow (ChooseCryptoAsset) with signature `(symbol,  contractId?) => void` to navigate to the send form. onOpenToken is the new callback for token details with `signature ({ assetId }) => void`. 

  They have different signatures because the send flow needs symbol/contractId for URL construction, while token details uses the serialized assetId format. The `handleOpenToken ??  onSelectAsset` pattern lets the component work in both contexts.
  
We can probably refactor `onSelectAsset` to use the serialized assetId format but I didn't want to add extra noise to this PR. I can add it in a follow up?

File: apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx (hunk 6 above)
Comment (pete-watters) [reply apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx]: Good catch - I'll move the check inside the `useEffect`

File: apps/extension/src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx [hunk 16]
```diff
@@ -17,13 +19,15 @@ interface RunesAssetListProps {
   accountIndex: number;
   filter?: AssetFilter;
   assetRightElementVariant?: AssetRightElementVariant;
+  onOpenToken?(details: TokenDetailsProps): void;
```
kyranjamie: props naming odd here, given that the type is used for arguments (not component props, suggest a name that describes the data, which could just be TokenDetails?

apps/extension/src/app/features/asset-list/stacks/sip10-token-list/sip10-token-list.tsx
kyranjamie: Also find the double methods here hard to follow. The reason explained in earlier PR comment won't be clear to someone reading this code. 

If the two methods have different signatures, why not overload the function to support both signatures?

apps/extension/src/app/features/token/components/token-details-actions.tsx
```diff
@@ -0,0 +1,119 @@
...
+  symbol,
+  isBuyEnabled = true,
+  isSwapEnabled = true,
+}: TokenDetailsActionsRowProps) {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const { releaseOnramperBuy } = useFlags();
+
+  const receivePath = symbol === 'BTC' ? `/${RouteUrls.ReceiveBtc}` : `/${RouteUrls.ReceiveStx}`;
+  const swapChain = symbol === 'BTC' ? 'bitcoin' : 'stacks';
```
kyranjamie: Baking in conditional logic like this isn't a pattern I recommend. 

We should be passing in chain-specific config to use this component. Code should be designed in such a way that, if a third chain comes along, we don't need to dive through the entire component tree to find conditionals like this.

Use props or a context

packages/features/src/token/token-details-utils.ts
pete-watters: This is extracting/consolidating existing code. `getPriceChangeColor` already existed locally in both `apps/mobile/src/components/balance/price-change.tsx` and `apps/web/app/pages/portfolio/portfolio-table/portfolio-table-cells.tsx`.  I moved it to the shared `@leather.io/features` package to DRY it up.

apps/extension/src/app/features/asset-list/bitcoin/btc-crypto-asset-item/btc-crypto-asset-item.tsx
pete-watters: I've refactored this here now to just get it done, let me know if it looks OK

pete-watters:
Thanks for the feedback @kyranjamie + @edgarkhanzadian. I think I've addressed most of it and have done some refactoring to add improvements along the way. Let me know if you spot any other issues and I'd be happy to address them

apps/extension/src/app/features/token/components/token-details-balance-item.tsx
```diff
@@ -0,0 +1,99 @@
...
+          {truncateMiddle(address, 6)}
+        </styled.button>
+      );
+    }
+    return (
+      <styled.span textStyle="caption.02" color="ink.text-subdued">
+        {truncateMiddle(address, 6)}
+      </styled.span>
+    );
+  }
```
kyranjamie: as far as I understan, this is better a useMemo, as currently this is a component defined within a component, vs jsx block rerendered on dep change

apps/extension/src/app/features/token/components/token-overview.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export function TokenOverview({
+  icon,
+  amount,
+  amountSuffix,
+  fiatAmount,
+  actions,
+}: TokenOverviewProps) {
+  const displayAmount = amountSuffix ? `${amount} ` : amount;
```
kyranjamie: can we use the currency formatter here?

kyranjamie (approved):

apps/extension/src/app/features/token/components/token-details-balance-item.tsx
pete-watters: Good catch, I've changed it to `useMemo`. Thanks!

pete-watters:
@mica000 gave some design feedback here so I will apply that tomorrow and do another round of QA before merging.

packages/features/src/token/token-details-utils.ts
tigranpetrossian: This is fine. We do have a formatPercentage helper that could potentially do this, but it needs to be slightly amended, and checked whether the existing polyfills on mobile support `signDisplay`.
I can take care of this myself at some point.

pete-watters:
This PR seems fine and good to go but I can't seem to beat `shard-3` even when I try this approach https://github.com/leather-io/mono/pull/2108

PR #2086, opened 2026-02-04
refactor(extension): multi-wallet supporting state structure 

kyranjamie:
This PR prepares the codebase for multi-wallet support.

**Project**: https://linear.app/leather-io/project/extension-multi-wallet-support-3d983f214bfa

This is a reasonably large PR. I would love to have broken it into smaller chunks, but given the broad scope of changes across the entire extension codebase, I don’t think it would’ve been possible. It isn’t too massive, though: many of the ~2500 line additions belong to the migration, supporting tests and their dummy state payloads, and developer tooling that I will remove before merge.

I'll explain here in as much detail as possible all the changes. I’m not in love with all the changes, and invite criticism of all choices made.

I aim to get all the behind the scenes changes completed, merged, and released before starting the actual UI changes .

### Todos

- [x] Fix restore wallet flow
- [x] Migration analytics 
- [x] Update all integration tests 
- [x] Revert changes to switch account list
- [x] Remove duplicate sign out / reset wallet action
- [x] Force Ledger app versions upgrade
- [x] Make design improvements to modal
- [x] Remove feature dev tooling

## Account references

`number` → `{ fingerprint: string, accountIndex: number }`

Previously we used an account index number to reference accounts. In a single account model, that’s all we needed. As in mobile, we now use the `AccountId` type to reference accounts, introducing the `fingerprint`. This, and all corresponding state changes/migrations, make up the largest conceptual change introduced here.

Coincidentally, credit to @alexp3y for adding the descriptor to state for the bitcoin/stacks address notifications. Having the fingerprint available in state tree enabled most of the state updates to happen at migration-time, rather than having to do a two-step migration (also migrating state after the user’s signed in and the keys are available).

## Migration

The updates to the state follow the approach taken in mobile, reusing some of the shared state package slices.

The migration is written using `any`. I don’t like it, but see no other option. To type more accurately, we’d need to retain a previous version of `RootState`. This would have to be described explicitly, rather than derived from the reducer (unless we were also to keep a copy of the old reducers too). Using `immer`, during the migration the state is in a half-way state, further complicating any efforts to strongly type. 

Full unit test coverage in there in lieu of strong typing.

### Edge case: Stacks-only Ledger keys

In the case of software or Ledger wallets with bitcoin keys, we already know the master key fingerprint. If a user only has Stacks Ledger keys, we don’t, given a limitation in the stacks app for Ledger.

Originally we used the key “default” to key our only wallet by. Now that we expect a hex fingerprint, it now defaults to `00000000`.

Recently released is https://github.com/Zondax/ledger-stacks/releases/tag/v0.26.0. This release does include the `getMasterkeyFingerprint` method.

A decision needs to be made how to handle this. Not having the fingerprint makes it impossible to tie the corresponding stacks accounts to the shared device. I am erring on the side of forcing users to update to the latest app version, so we can get fingerprint. This will require a warning step when connecting devices. cc/ @fabric-8 @mica000.

### Edge case: Earlier application state, versions 1 & 2 migrations

I would've forgotten this if it weren't for the integration test we have. Suppose a very infrequent Leather user returns to their wallet after 3 years. They have an extremely old wallet version, that doesn't have the fingerprint in the state. The migration run. In this scenario we default to `00000000`. But, the state tree will expect to find the actual key fingerprint, which won't be there. This would completely break the wallet.

We need a sign-in time migration that, in the condition of a single software wallet with the zero fingerprint, read the fingerprint and update all the store values.

## Missing keychain migration

Not completed here is the migration to the pattern of persisting account keychains for Ledger and software wallets. The selectors are composed in such a way that we follow a similar flow of data, first deriving the keychain, then our account abstraction. 

I am inclined to take this on next, still within this PR. I haven’t done it yet in an attempt to try and limit the scope of breaking things, but now I have a mostly complete migration in place, don’t think it should be too much of a lift to further migrate to this way of managing keychains.

## Memoizing account initialisation

I’ve ran into render issues in this work. Some selectors follow a pattern of returning a function with signature

```tsx
(accountId: AccountId) => StacksAccount | undefined
```

This helps to create a uniform API with which to get accounts of any origin. Within these calls we first check if we have a matching in memory key, then for persisted Ledger keys, before returning `undefined`. The problem is that, without some funky memoization logic (see `memoizeAccountGenerator`), a new reference is returned each time, causing many parts of the codebase passing account as an effect dependency to trigger render loops.

Perhaps this approach of returning a function doesn’t suit a selector based state structure? I’d like to avoid this custom cache logic if possible. Open to feedback and still thinking on better ways to do this.

## Decoupling signing from payer details

Throughout the codebase we have a Bitcoin "signer" abstraction that has payment information (address, payment script, pub key etc) _and_ signing methods `sign` and `signAtIndex`. 

The problem is that only software wallets can arbitrarily call these signing methods. Ledger wallets must sign through it's own flow. These methods are only used in two places as well. By referring to "payers" (as already done in mobile), we have an abstraction that works for both wallet type's public key info.

Software signing methods are now pulled in specifically when needed. Thus, we now refer to _payers_ rather than _signers_ in the majority of use cases.


apps/extension/src/app/store/index.ts
```diff
@@ -63,17 +74,18 @@ const appReducer = combineReducers({
     bitcoin: bitcoinKeysSlice.reducer,
     stacks: stacksKeysSlice.reducer,
   }),
+  wallets: walletSlice.reducer,
+  keychains: keychainSlice.reducer,
```
edgarkhanzadian: If we are adding keychains here, can we remove `chains` slice?

apps/extension/src/app/store/chains/stx-chain.slice.ts
```diff
@@ -1,56 +1,69 @@
...
+      if (!state[fingerprint]) {
+        state[fingerprint] = {
+          highestAccountIndex: action.payload.accountIndex,
+          currentAccountStacksDescriptor: '',
+        };
+      } else {
+        state[fingerprint].highestAccountIndex = action.payload.accountIndex;
+      }
     },
   },
```
edgarkhanzadian, line 45: Doesn't this ignore the fingerprint that we are getting from the action payload?

apps/extension/src/app/store/software-keys/software-key.slice.ts
```diff
@@ -1,34 +1,48 @@
...
-      keyAdapter.removeOne(state as any, defaultWalletKeyId);
+    addNewWallet(state, action: PayloadAction<SoftwareKeyConfig>) {
+      keyAdapter.addOne(state, action.payload);
     },
 
-    debugKillStacks() {
-      // if (state.entities.default?.type !== 'ledger') return;
-      // state.entities.default.publicKeys = [];
+    signOut(state) {
+      if (state.salt) delete state.salt;
```
edgarkhanzadian: can we use immutable states here instead? And just generally in all of the redux reducers

edgarkhanzadian (commented):
Making baby steps in reviewing this PR. I'm going to add more review comments as i go through the PR

apps/extension/src/app/store/software-keys/software-key.slice.ts
kyranjamie: RTK uses immer internally so this is still immutable https://redux-toolkit.js.org/usage/immer-reducers#benefits-of-immer

apps/extension/src/app/store/index.ts
kyranjamie: Mmm, yes keychain shouldn't be here (yet), though that's coming next

apps/extension/src/app/store/software-keys/software-key.slice.ts
edgarkhanzadian: Nice, didn't know that!

apps/extension/src/app/store/active/active.slice.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { fingerprintMigration } from '@shared/storage/redux-persist';
+
+import { inMemoryKeySlice } from '../in-memory-key/in-memory-key.slice';
+
+export const userSwitchesAccount = createAction<AccountId>('active/userSwitchesAccount');
+
+interface ActiveState {
+  account: AccountId | null;
+}
```
edgarkhanzadian, line 16: What is the difference between active account and current account?

packages/bitcoin/src/signer/bitcoin-signer.ts
```diff
@@ -47,8 +47,6 @@ export interface BitcoinSigner<Payment> {
   derivationPath: string;
   address: BitcoinAddress;
   publicKey: Uint8Array;
-  sign(tx: btc.Transaction): void;
-  signIndex(tx: btc.Transaction, index: number, allowedSighash?: AllowedSighashTypes[]): void;
 }
```
edgarkhanzadian: Why remove sign and signIndex here?

apps/extension/tests/specs/store-migrations/store-migrations.spec.ts
```diff
@@ -1,50 +1,137 @@
...
+      createdOn: null,
+    });
+    test.expect(result.wallets.ids).toEqual(['ef7bfd66']);
+
+    test.expect(result.chains.stx.ef7bfd66).toBeDefined();
+    test.expect(result.chains.stx.default).toBeUndefined();
+
+    test.expect(result.onboarding).toBeUndefined();
+    test.expect(result.ordinals).toBeUndefined();
   });
```
edgarkhanzadian, line 140: Should we consider a deepEqual here? Same way that was done for v1->v2 state migration on mobile?

apps/extension/src/shared/storage/redux-persist.ts
```diff
@@ -1,199 +1,40 @@
...
     'analytics',
     'chains',
-    'ordinals',
     'softwareKeys',
     'appPermissions',
     'ledger',
     'networks',
-    'onboarding',
     'settings',
+    'wallets',
```
edgarkhanzadian, line 32: We might want to add `keychains` here too

apps/extension/src/shared/storage/migrations/migrate-2-3.ts
```diff
@@ -0,0 +1,270 @@
...
+        logger.error('Error extracting fingerprint from Bitcoin policy:', error, policy);
+      }
+    }
+  }
+
+  // Default only for Stacks-only wallets
+  return assumedZeroFingerprint;
+}
+
+export function migrateMultiWalletSupport(state: RootState) {
```
edgarkhanzadian: When the migration is needed, this state is not RootState right (as RootState corresponds to the latest version of the state)? I'd suggest we create interfaces for each store version and make the migrations as type safe as possible.

apps/extension/src/shared/storage/migrations/migrate-1-2.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      return acc;
+    },
+    { ids: [], entities: {} }
+  );
+
+  return newStacksKeys;
+}
+
+export function migrateToRenameKeysStoreModule(state: any): any {
```
edgarkhanzadian, line 28: same thing regarding store interfaces

apps/extension/src/app/store/utils/vault-reducer-migration.ts
```diff
@@ -16,11 +17,11 @@ export function migrateVaultReducerStoreToNewStateStructure(initialState: typeof
...
-      ids: ['default'],
+      ids: [assumedZeroFingerprint],
       entities: {
-        default: {
+        [assumedZeroFingerprint]: {
           type: 'software',
-          id: 'default',
+          id: assumedZeroFingerprint,
           encryptedSecretKey,
           salt,
```
edgarkhanzadian, line 26: Just curious why do we need migrateVaultReducerStoreToNewStateStructure at all? I though redux would handle all of the migrations

apps/extension/src/app/store/software-keys/software-key.selectors.ts
```diff
@@ -4,35 +4,50 @@ import { createSelector } from '@reduxjs/toolkit';
...
 );
 
-export const selectHasSecretKey = createSelector(
-  selectDefaultSoftwareKey,
-  softwareKey => !!softwareKey?.encryptedSecretKey
+export const selectWalletSalt = createSelector(
+  selectKeysSlice,
+  // State v3 migrates salt to softwareKeys root
+  state =>
+    state.salt ?? ((state.entities[assumedZeroFingerprint] as any)?.salt as string | undefined)
```
edgarkhanzadian: any chance we could do this a little safer?

apps/extension/src/app/store/ledger/stacks/stacks-key.slice.ts
```diff
@@ -7,11 +7,13 @@ function selectStacksKeysSlice(state: RootState) {
   return state.ledger.stacks;
 }
 
-// ts-unused-exports:disable-next-line
-export const { slice: stacksKeysSlice, adapter } = generateLedgerChainKeyStorageSlice<
-  StacksAppKeysResponseItem & { id: string; targetId: string }
+const { slice: stacksKeysSlice, adapter } = generateLedgerChainKeyStorageSlice<
+  StacksAppKeysResponseItem & { id: string; fingerprint: string }
```
edgarkhanzadian: Do we need targetId at all anymore?

apps/extension/src/app/store/ledger/ledger.selectors.ts
```diff
@@ -14,21 +17,29 @@ const selectNumberOfLedgerKeysPersisted = createSelector(selectLedger, ledger =>
...
+const selectHasLedgerKeys = createSelector(selectNumberOfLedgerKeysPersisted, numOfKeys =>
   numOfKeys.isGreaterThan(0)
 );
 
 const selectHasLedgerBitcoinKeys = createSelector(
-  selectNumberOfLedgerBitcoinKeysPersisted,
-  numOfKeys => numOfKeys.isGreaterThan(0)
+  [selectBitcoinKeychainEntities, selectCurrentAccount],
+  (bitcoinKeychainEntities, currentAccount) => {
+    const bitcoinKeysForCurrentWallet = Object.values(bitcoinKeychainEntities || {}).filter(
```
edgarkhanzadian: if i understand correctly, bitcoinKeychainEntities is always an object and we don't need `|| {}` here

apps/extension/src/app/store/common/wallet-type.selectors.ts
```diff
@@ -2,29 +2,114 @@ import { useSelector } from 'react-redux';
...
-    }
-    // TODO: better error handling
-    return undefined;
+// This renders a tree of wallets and their child accounts as an array. This
+// should be used to render virtualised lists, where list item components take
+// care of account look up themselves, as a way to lazily derive account details
+// when needed.
+const selectWalletAccountRefTree = createSelector(
+  [
+    selectWalletEntities,
```
edgarkhanzadian: walletEntities technically should always be an object, so no reason to fallback to `|| {}`

apps/extension/src/app/store/active/active.slice.ts
kyranjamie: I'm trying to diminish this concept a bit. We need to persist in state what the active account is (previously was in `chains.stx.default`), but also when an app connects we want to default to the account that it connected with, not the active one, overriding it. What I'm calling current. 

I will try and make it clear that generally we should not be consuming the "active" wallet.

apps/extension/src/app/store/common/wallet-type.selectors.ts
```diff
@@ -2,29 +2,104 @@ import { useSelector } from 'react-redux';
...
+      // Collect account indices from software Stacks accounts
+      const stxChainState = stxChain?.[wallet.fingerprint];
+      if (stxChainState) {
+        for (let i = 0; i <= stxChainState.highestAccountIndex; i++) {
+          allAccountIndices.push(i);
+        }
+      }
+
+      // Collect account indices from Ledger Stacks keys
+      ledgerStacksKeys
```
edgarkhanzadian: Do we need to also add ledger bitcoin keys?

apps/extension/src/app/store/common/wallet-type.selectors.ts
```diff
@@ -2,29 +2,104 @@ import { useSelector } from 'react-redux';
...
+  type: WalletType;
+  accounts: AccountId[];
+  createdOn: string | null;
+}
+
+// This renders a tree of wallets and their child accounts as an array. This
+// should be used to render virtualised lists, where list item components take
+// care of account look up themselves, as a way to lazily derive account details
+// when needed.
+const selectWalletAccountRefTree = createSelector(
```
edgarkhanzadian, line 46: I'm a little unsure why we need this function. Do i understand correctly that it tries to find the highest account for each wallet and then create an AccountId[] for each wallet?

apps/extension/src/app/store/chains/stx-chain.slice.ts
```diff
@@ -1,56 +1,69 @@
...
-    });
-  },
+  extraReducers: builder =>
+    builder
+      .addCase(keySlice.actions.signOut, () => ({}))
+
+      .addCase(userAddsWallet, (state, action) => {
+        state[action.payload.wallet.fingerprint] = {
+          highestAccountIndex: 0,
+          currentAccountStacksDescriptor: '',
```
edgarkhanzadian, line 54: Why use an empty string for a stacks descriptor here? I'd suggest either populating this with a correct value or setting this to null

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.selectors.ts
```diff
@@ -1,115 +1,204 @@
...
+
+function memoizeAccountGenerator<
+  TAccount extends SoftwareStacksAccount | HardwareStacksAccount | undefined,
+>(
+  fn: (accountId: AccountId, network: NetworkModes) => TAccount
+): (accountId: AccountId, network: NetworkModes) => TAccount {
+  const cache = new Map<string, TAccount>();
+
+  return (accountId: AccountId, network: NetworkModes): TAccount => {
+    const cacheKey = `${accountId.fingerprint}:${accountId.accountIndex}:${network}`;
```
edgarkhanzadian, line 160: Should we also add wallet type (software/hardware) to this cacheKey as well? What would happen if i add a mnemonic as a software wallet and then try to connect ledger with the same mnemonic to the leather wallet?

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.models.ts
```diff
@@ -1,20 +1,22 @@
...
+  };
 
-export interface HardwareStacksAccount {
+export interface HardwareStacksAccount extends AccountId {
   type: 'ledger';
-  index: number;
   address: string;
   stxPublicKey: string;
   dataPublicKey: string;
+  index: number;
```
edgarkhanzadian, line 20: What does index stand for? Considering this interface already extends AccountId that has fingerprint and accountIndex

apps/extension/src/app/services/accounts/use-account-addresses.ts
```diff
@@ -1,44 +1,39 @@
...
 
     const nativeSegwitAddress = nativeSegwitAccount
-      ? getNativeSegwitPaymentFromAddressIndex(
-          deriveAddressIndexZeroFromAccount(nativeSegwitAccount.keychain),
-          nativeSegwitAccount.network
-        ).address
+      ? deriveBitcoinPayerFromAccount(
+          nativeSegwitAccount.descriptor,
+          network.chain.bitcoin.mode
+        )({ change: 0, addressIndex: 0 }).address
```
edgarkhanzadian, line 29: This is awesome!

apps/extension/src/app/routes/app-routes.tsx
```diff
@@ -147,6 +148,16 @@ function useAppRoutes() {
...
+            path={RouteUrls.MultiWalletTest}
+            element={
+              <AccountGate>
+                <MultiWalletTest />
+              </AccountGate>
+            }
+          >
+            {requestBitcoinKeysRoutes}
+          </Route>
           <Route
```
edgarkhanzadian: Do we need to guard this using a feature flag or something like `__DEV__`?

apps/extension/src/app/store/common/wallet-type.selectors.ts
```diff
@@ -2,29 +2,104 @@ import { useSelector } from 'react-redux';
...
-      return WalletType.Ledger;
-    }
-    // TODO: better error handling
-    return undefined;
+interface WalletAccountRefTree {
+  fingerprint: string;
+  name: string;
+  type: WalletType;
+  accounts: AccountId[];
+  createdOn: string | null;
```
edgarkhanzadian, line 39: Can we keep createdOn as milliseconds instead of an ISO string? That would save us some time with parsing to Date object + it is simpler to use milliseconds for sorting in selectWalletAccountRefTree function

apps/extension/src/app/common/hooks/use-key-actions.ts
```diff
@@ -40,31 +45,35 @@ export function useKeyActions() {
...
-        return dispatch(createNewAccount());
+        if (!activeAccount) throw new Error('No active account');
+        return dispatch(createNewAccount(activeAccount.fingerprint));
       },
 
       async signOut() {
         await clearWalletSession();
         dispatch(networksSlice.actions.changeNetwork('mainnet'));
         dispatch(keyActions.signOut());
+        dispatch(resetWallet());
```
edgarkhanzadian, line 70: Do i understand correctly that `resetWallet` action only resets the active account slice?

edgarkhanzadian:
I also noticed we have an empty file `apps/extension/src/app/store/keychains/keychain.slice.ts` here. Do we need to keep a placeholder there or we can just remove it?

packages/bitcoin/src/signer/bitcoin-signer.ts
kyranjamie: Thanks for reminder here, I need to rename this/merge it with the other type `BitcoinPayer` (last section of PR body explains)

apps/extension/src/app/routes/app-routes.tsx
kyranjamie: I will be removing this

apps/extension/src/app/store/utils/vault-reducer-migration.ts
kyranjamie: Note this is a 3 year old file, written to handle an earlier refactor. The extension used to write the key directly to `localStorage`. When we later changed to a single state tree design, this migration pulls the values into the original state tree.

Out of an abundance of caution, I think we should leave this here, as very old user returning to the wallet from pre-2023 would require this migration for their wallet to work.

apps/extension/src/app/store/common/wallet-type.selectors.ts
kyranjamie: I would argue for ISO strings. IMO Human readability > micro performance optimisation. `+Date.parse(obj.createdOn)`

apps/extension/src/app/store/utils/vault-reducer-migration.ts
kyranjamie: It would've been better to have this be a migration as this is actually pretty bad, synchronous reads on every load 🤔  Maybe we can kill it

apps/extension/src/app/store/ledger/stacks/stacks-key.slice.ts
kyranjamie: I'm pretty sure not all Ledger's even return this anymore, think it's safe to axe

apps/extension/src/app/store/common/wallet-type.selectors.ts
kyranjamie: Consider the switch account list with 3x wallets with 100x accounts. An extreme but possible scenario. We need a virtual list to render these, as otherwise we have to pre-derive 300 accounts and it'll block the main thread causing super laggy list/ui.

What do we pass to the virtual list? We need a data structure that contains references to the accounts, but not the derived accounts themselves, so we can get each list item to lazy derive the account. That's why we have this tree of `AccountId[]`s

apps/extension/src/app/store/chains/stx-chain.slice.ts
kyranjamie: TDB on this, keychain changes should resolve

apps/extension/src/shared/storage/migrations/migrate-2-3.ts
kyranjamie: Good spot that this type is wrong. 

I tried to [justify these decisions here](https://github.com/leather-io/mono/pull/2086#:~:text=The%20migration%20is,of%20strong%20typing). Maintaining henceforth a full copy of the old state tree would quite a challenging/verbose effort. Not only, given the approach of using immer, the object can be somewhere between the two states during the migration. So, even if we did have 2 state tree interfaces, how do you satisfy the compiler half-way through the migration?

I did try to type this but it just became such a challenge that I instead focused on testing. We have unit test and integration coverage of the migration step, in place of this typing.

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.selectors.ts
kyranjamie: Were this to happen, it shouldn't really matter as the account info would be the same?

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.models.ts
kyranjamie: This is the account index (address for stacks), and remains here I think owing to some interopability with stacks.js, but let me double check

kyranjamie: I think this will be easier to resolve when taking on keychain changes so will hang fire on this

apps/extension/tests/specs/store-migrations/store-migrations.spec.ts
kyranjamie: How do you mean? Structured clone? I could but would be more verbose, this does test mostly just what's changed. We have both integration and unit tests for the migrate step

apps/extension/src/app/common/authentication/use-legacy-auth-bitcoin-addresses.ts
```diff
@@ -1,39 +1,60 @@
...
+      const p2wpkhTestnet = getAccountDetails('p2wpkh', 'testnet');
+      const p2trMainnet = getAccountDetails('p2tr', 'mainnet');
+      const p2trTestnet = getAccountDetails('p2tr', 'testnet');
+
+      return {
+        btcAddress: {
+          p2tr: {
+            mainnet: p2trMainnet?.address,
+            testnet: p2trTestnet?.address,
+            regtest: getAccountDetails('p2tr', 'regtest')?.address,
```
pete-watters: Maybe just set a const for this like the others? `const p2trRegtest`

apps/extension/src/app/features/asset-list/_components/connect-ledger-asset-button.tsx
```diff
@@ -21,7 +21,7 @@ export function ConnectLedgerButton({ chain }: ConnectLedgerButtonProps) {
     void navigate(`${chain}/connect-your-ledger`, {
       replace: true,
       state: {
-        [immediatelyAttemptLedgerConnection]: true,
+        [immediatelyAttemptLedgerConnection]: false,
```
pete-watters, line 24: Double checking this change is intentional

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
```diff
@@ -0,0 +1,166 @@
...
+    setEditedName(name);
+    setIsEditing(false);
+  }
+
+  return (
+    <Stack gap="space.02">
+      <HStack justifyContent="space-between" alignItems="center" py="space.02">
+        <HStack gap="space.02" alignItems="center" flex="1">
+          {isEditing ? (
+            <HStack gap="space.02" flex="1">
```
pete-watters: This could be a component

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
```diff
@@ -0,0 +1,166 @@
...
+              <input
+                type="text"
+                value={editedName}
+                onChange={e => setEditedName(e.target.value)}
+                onKeyDown={e => {
+                  if (e.key === 'Enter') handleSaveName();
+                  if (e.key === 'Escape') handleCancelEdit();
+                }}
+                autoFocus
+                style={{
```
pete-watters: Style objects?? 🫨

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
```diff
@@ -0,0 +1,166 @@
...
+              </Box>
+            ))
+          )}
+          {type === 'software' && (
+            <Box py="space.01">
+              <Button
+                onClick={() => onCreateAccount(fingerprint)}
+                size="sm"
+                variant="ghost"
+                style={{ fontSize: '12px', padding: '4px 8px' }}
```
pete-watters: Style objects could be props

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
kyranjamie: i know what a hypocrite I am lol (all of the developer-utilities are not going to be committed, will delete before merge)

apps/extension/src/app/features/asset-list/_components/connect-ledger-asset-button.tsx
kyranjamie: yep intended

apps/extension/src/app/features/ledger/flows/request-stacks-keys/ledger-request-stacks-keys.tsx
```diff
@@ -55,15 +82,30 @@ function LedgerRequestStacksKeys() {
...
-              targetId: latestDeviceResponse?.targetId || '',
-            }))
-          )
-        );
+
+        const keychains = resp.publicKeys.map(keys => {
+          const keyOrigin = createKeyOriginPath(fingerprint, keys.path);
+          const descriptor = createDescriptor(keyOrigin, keys.stxPublicKey);
+          return {
+            chain: 'stacks' as const,
```
pete-watters, line 90: Can we avoid this casting to `const`?

apps/extension/src/app/features/ledger/flows/stacks-tx-signing/ledger-sign-stacks-tx-container.tsx
```diff
@@ -58,91 +58,86 @@ function LedgerSignStacksTxContainer() {
...
+          });
+          await delay(400);
+          return false;
+        }
+
         return true;
-      }
+      },
+      async signTransactionWithDevice(stacksApp) {
+        // TODO: need better handling
```
pete-watters: Can you handle this now / log a ticket or add more details in the comment?

apps/extension/src/app/features/ledger/generic-steps/outdated-stacks-app-warning/outdated-stacks-app-warning-base.tsx
```diff
@@ -0,0 +1,169 @@
...
+              border="1px solid"
+              borderColor="ink.border-default"
+              flexShrink={0}
+            >
+              <styled.span textStyle="label.02">1</styled.span>
+            </Flex>
+          }
+        >
+          <styled.span textStyle="body.01">
+            Open Ledger Wallet → My Ledger and{' '}
```
pete-watters, line 87: Maybe you can use an interpolated string instead of `{' '}`

apps/extension/src/app/features/ledger/flows/request-stacks-keys/ledger-request-stacks-keys.tsx
kyranjamie: nope, defaults to string otherwise and 'stacks' literal is expected

kyranjamie: image

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
kyranjamie: will be deleted before merge

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
kyranjamie: to delete before merge

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-keychain.ts
```diff
@@ -1,60 +1,202 @@
...
+// each time, while not defaulting to only working with the existing network
+type RequirementsWithoutNetwork = Omit<BitcoinAccountDerivationRequirements, 'network'>;
+export const selectCurrentNetworkBitcoinAccountLookup = createSelector(
+  selectBitcoinAccountLookup,
+  selectCurrentNetwork,
+  (accountLookup, network) => (fingerprint: string) => (args: RequirementsWithoutNetwork) =>
+    accountLookup(fingerprint)({ ...args, network: network.chain.bitcoin.mode })
+);
+
+// TODO: very bad performance to use this selector to generate all keychains in
```
pete-watters: Can we do this right away in a fast follow up?

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin.hooks.ts
```diff
@@ -60,10 +60,10 @@ export function useHasCurrentBitcoinAccount() {
 }
 
 // Temporary - remove with privacy mode
```
pete-watters: Do you understand the relevance of this `// Temporary..` comment now given the context? maybe we can delete it? It's been temporary for 2 years

Image

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin.ledger.ts
```diff
@@ -1,34 +1,40 @@
 import { useSelector } from 'react-redux';
 
 import { bitcoinNetworkModeToCoreNetworkMode, inferNetworkFromPath } from '@leather.io/bitcoin';
+import { extractDerivationPathFromDescriptor } from '@leather.io/crypto';
 
-import { selectDefaultWalletBitcoinKeys } from '@app/store/ledger/bitcoin/bitcoin-key.slice';
+import { selectBitcoinKeychains } from '@app/store/keychains/keychain.selectors';
 import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 
 // TODO: Asset refactor: remove if determined unnecessary
```
pete-watters: If this is still unused lets get rid of it `// ts-unused-exports:disable-next-line useHasBitcoinLedgerKeychain`

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.models.ts
```diff
@@ -1,20 +1,25 @@
 import type { Account } from '@stacks/wallet-sdk';
 
+import type { AccountId } from '@leather.io/models';
+
+// TODO @kyranjamie
```
pete-watters: Are you planning to do this soon?

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.selectors.ts
```diff
@@ -1,115 +1,240 @@
...
-  const stxPrivateKey = deriveStxPrivateKey({ rootNode: rootKeychain, index } as any);
-  const pubKey = privateKeyToPublic(stxPrivateKey) as string;
+function initalizeSoftwareStacksAccount(
+  rootKeychain: HDKey,
+  accountId: AccountId,
+  network: NetworkModes
+): SoftwareStacksAccount {
+  const stxPrivateKey = deriveStxPrivateKey({
+    // Internal lib type mismatch
+    rootNode: rootKeychain as any,
```
pete-watters, line 41: I guess the `as any` is needed due to the mismatch? Maybe we can handle it better than this

apps/extension/src/app/store/accounts/blockchain/utils.ts
```diff
@@ -2,14 +2,15 @@ import { useCallback } from 'react';
 
 import type { Blockchain } from '@leather.io/models';
 
+import { useHasLedgerStacksKeys } from '@app/store/ledger/ledger.selectors';
+
 import { useHasCurrentBitcoinAccount } from './bitcoin/bitcoin.hooks';
-import { useHasStacksLedgerKeychain } from './stacks/stacks.hooks';
 
 // TODO: Asset refactor: remove if determined unnecessary
```
pete-watters, line 9: Lets also get rid of this if it's not used 
```
// ts-unused-exports:disable-next-line
export function useCheckLedgerBlockchainAvailable() 
```

.gitignore
```diff
@@ -19,3 +19,4 @@ tmp
 **/.claude/settings.local.json
 .cursorrules
 test-results
+copilot-instructions.md
```
pete-watters, line 22: 🫨  have you been cheating?!?

pete-watters (approved):
Great work @kyranjamie!

A monster PR, `+5,893 -2,519`, touching 220 files! And 5 drafts on the board 🫨  some would say too much AI, too fast! 

Joking aside, I gave the PR a review and added a few comments, most of them unimportant so happy to approve and get this merged. (I know how it feels to have a huge PR hanging over you). 

Lets get this merged and then iterate to improve it over time. 

As part of the review I got Claude to assess the code, here are some questions and concerns it mentioned in case it helps with the next iterations:

**Concerns and questions**:

1. **The 00000000 fingerprint default for Stacks-only Ledger keys** — this feels like it could become a subtle source of bugs. If two Stacks-only Ledger wallets both default to 00000000, their AccountIds could collide. Is there a plan to enforce the Ledger Stacks v0.26.0+ upgrade path more aggressively, or is this intended as a temporary shim? Worth a comment or runtime guard here.
2. **13.37% patch coverage** — The PR description acknowledges migration tests, but 2,604 uncovered lines is significant. Even acconyunting for migration boilerplate, the new hooks and selectors (like selectStacksAccountById, useBitcoinAccountLookup, deriveBitcoinPayerFromAccount) are load-bearing logic that would benefit from test coverage. The "Update integration tests" TODO suggests this is known, but merging with this coverage gap feels risky.
3. **any types in migrations** — Understandable, but these tend to linger. Would it be worth adding a // TODO: type narrowing or even a lint exception with an explanation, so they don't get normalized? Alternatively, defining lightweight interfaces for the "before" state shapes (even partial ones) could catch migration bugs at compile time.
4. **Memoization to prevent render loops** — This is mentioned in the description but I'd want to verify the memoization is using stable references. If AccountId objects are being created inline (e.g., { fingerprint, accountIndex } in render), they'll break shallow equality checks. Would be worth confirming that AccountId objects are memoized or structurally compared.
5. **The outstanding TODO list is long** — restore wallet flow, migration analytics, integration tests, Ledger version enforcement, design improvements, and dev tooling cleanup. Several of these (restore wallet, integration tests) feel like they should be pre-merge rather than follow-up. Is there a plan to gate the feature flag until these are resolved?
6. **Immer addition** — Adding immer as a new dependency is fine for immutable state updates, but worth confirming it's used consistently across the new slices rather than mixed with manual spread patterns.
7. **Removing pnpm typecheck from CLAUDE.md** — Is this intentional because it's currently broken from the migration, or a permanent change? If temporary, it should probably be restored before merge.

kyranjamie:
Thanks for feedback @pete-watters, is appreciated. At least 1.5k lines are the developer utilities and config json I'll remove before merge.

> Is there a plan to enforce the Ledger Stacks v0.26.0+ upgrade path more aggressively

Yes, there is. I've already built in a forced upgrade path to handle this exact scenario.

I'll review where I can improve test coverage, especially relating to pure state manip. functions.

Immer is already used internally in RTK. It is pretty crucial for the migration code, otherwise it would be an unreadable mess. It already is to be honest, that is why I have focused most unit testing effort there (also use of any).

I didn't remove typecheck from Claude.md, I removed the build step. This conflicts with an active dev session while using claude, which is super annoying, and isn't critical as we are checking the types.

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-keychain.ts
kyranjamie: https://linear.app/leather-io/issue/LEA-3494/perf-remove-use-of-usestacksaccounts

apps/extension/src/app/store/accounts/blockchain/stacks/stacks-account.selectors.ts
kyranjamie: update with better comment

pete-watters:
> Thanks for feedback @pete-watters, is appreciated. At least 1.5k lines are the developer utilities and config json I'll remove before merge.
> 
> > Is there a plan to enforce the Ledger Stacks v0.26.0+ upgrade path more aggressively
> 
> Yes, there is. I've already built in a forced upgrade path to handle this exact scenario.
> 
> I'll review where I can improve test coverage, especially relating to pure state manip. functions.
> 
> Immer is already used internally in RTK. It is pretty crucial for the migration code, otherwise it would be an unreadable mess. It already is to be honest, that is why I have focused most unit testing effort there (also use of any).
> 
> I didn't remove typecheck from Claude.md, I removed the build step. This conflicts with an active dev session while using claude, which is super annoying, and isn't critical as we are checking the types.

This is amazing PR overall and a huge step forward for the product, nice work 💪

apps/extension/src/app/features/developer-utilities/components/wallet-tree-section.tsx
pete-watters: It's fine 😂  realistically stuff like this matters less and less now when `prettier` has morphed into `claude` formatting this stuff. It's basically the same thing anyway style props / style objects. It just adds un-necessary cognitive load bothering to care about it

PR #2088, opened 2026-02-05
chore: update claude.md

edgarkhanzadian:


.claude/settings.json
```diff
@@ -0,0 +1,17 @@
...
+      "Bash(git status *)",
+      "Bash(git diff *)",
+      "Bash(git log *)",
+      "Bash(git branch *)",
+      "Bash(git rev-parse *)",
+      "Bash(gh *)",
+      "Bash(ls *)",
+      "Bash(xargs *)",
+      "Bash(grep *)",
+      "WebFetch(domain:raw.githubusercontent.com)"
```
kyranjamie: IMO config like this should personal, not committed. Plus, just use `--dangerously-skip-permissions` 😂

CLAUDE.md
```diff
@@ -1,185 +1,140 @@
...
 
-- **Pure business logic** → `@leather.io/domain` (organized by feature: `activity/`, `balances/`, `fees/`, etc.)
-- **Orchestration** (API calls, DI, caching) → `@leather.io/services`
-- **Protocol utilities** → `@leather.io/bitcoin` or `@leather.io/stacks`
-- **Generic utilities** → `@leather.io/utils`
-- **UI components** → `@leather.io/ui`
-
-## Code style
-
-- Don't use enums.
```
pete-watters, line 46: It seems like you are adding some great stuff but also deleting a lot of stuff, maybe for good reason?

I added a lot of these instructions to improve the quality of the output to conform to our standards. Often it was coming out with code that wasn't as we want it so I tried to improve that gradually over time 

Are we able to enforce these patterns a better way you are introducing here?

.claude/settings.json
camerow: Yea agreed, perhaps instead we should update `.gitignore` to ignore claude settings so these don't get accidentally committed?

CLAUDE.md
edgarkhanzadian: i think claude didn't remove this part, it just moved it inside the file. All of these rules are still enforced in claude.md

.claude/settings.json
edgarkhanzadian: i deleted this file, claude does create settings.local.json that should handle this part

tigranpetrossian (approved):

PR #2089, opened 2026-02-09
feat(mobile): first token [LEA-3352]

edgarkhanzadian:
https://github.com/user-attachments/assets/79219c03-51c0-4dda-8493-69b628a6d19e


apps/mobile/src/components/home/components/first-token-banner.tsx
```diff
@@ -0,0 +1,24 @@
...
+        <Text variant="label01">{t`Get your first token`}</Text>
+        <Text variant="caption01">
+          {t`Fund your wallet by buying tokens or transferring from another account.`}
+        </Text>
+      </Box>
+      <Box>
+        <Image
+          style={{ height: 48, width: 80 }}
+          contentFit="cover"
+          source={require('@/assets/exchange-icons.png')}
```
pete-watters, line 19: I think we will need this same image when implementing in the `extension` so maybe its best to share it from somewhere?

pete-watters (approved):
Nice work 👍 Maybe you can add the `extension` next and we can come up with a nice way to share the logic of when to show this screen.

Activity success and no activity seems OK but I wonder if it's 100% reliable? 🤔 Like maybe its the second account of a wallet that does have activity?

edgarkhanzadian:
yea i agree, this is going to display "buy your first token" for every new account, but imo it is fine to have that by design. We can also discuss this during design office hours

apps/mobile/src/components/home/components/first-token-banner.tsx
edgarkhanzadian: Good point. I'll add this to some shared place once i get this implemented in extension

camerow:
@edgarkhanzadian I agree that it's fine to show for every first account. It's a much nicer 0 state no matter what. We can improve if people complain but it looks quite nice man, good work!

PR #2091, opened 2026-02-09
chore(mobile): add comprehensive maestro E2E tests

pete-watters:
## Summary

Adds comprehensive Maestro E2E test coverage for the mobile app, with CI infrastructure to run tests reliably via EAS Update.

### Maestro E2E Test Suite

**Full test suite** (`full-suite-ci.yaml`) runs all tests in sequence:
- **Wallet Creation** — via developer console (bottom sheets don't present reliably in EAS Update CI)
- **Settings Navigation** — all settings screens (Display, Security, Networks, Help)
- **Send Flow** — verify send sheet and asset selection
- **Receive Flow** — verify receive sheet and address display
- **Network Switching** — Mainnet/Testnet4 round-trip
- **Wallet Removal** — cleanup via Settings
- **Wallet Restore** — restore from test mnemonic

**Shared reusable flows**: `create-wallet.yaml`, `create-wallet-dev-console.yaml`, `restore-wallet.yaml`, `remove-wallet.yaml`, `clean-up.yaml`

### Developer Console in EAS Update CI

`__DEV__` is `false` in EAS Update JS bundles (production-mode JS loaded into a dev client). To make the developer console available for Maestro tests:
- Switched dev console visibility from `__DEV__` to `!isProduction()` (checks `EXPO_PUBLIC_NODE_ENV`)
- Added `environment: development` to the EAS Update workflow step so the right env vars are baked into the JS bundle

### CI Improvements

- Added `code-quality-gate` job to `mobile:validate-build-pre-eas.yml` — fast-fails if repo code checks fail, avoids duplicating lint/typecheck work
- Narrowed PR path filter to only the 16 runtime dependency packages mobile uses — extension-only PRs no longer trigger mobile build validation
- Non-PR triggers (release, workflow_dispatch) skip the gate and proceed directly

### Bug Fixes

- Fixed Maestro network tap selectors — iOS accessibility tree exposes title+caption as single label, so selectors use combined patterns (`Testnet4.*Disabled`)
- Fixed network switch-back selector — after switching to Testnet4, Mainnet shows as `Disabled` not `Enabled`
- Reverted `.gitignore` change (`/android/` → `android/`) that caused EAS fingerprint drift and unnecessary fresh native builds
- Removed unused `TestId` enum members (`developerToolsWalletManagementButton`, `homeAccountCard`, `homeActivityButton`, `homeAddWalletButton`)
- Removed phantom `WalletList` import from developer console

## Test plan

- [ ] Verify Maestro E2E tests pass on both iOS and Android in EAS CI
- [ ] Confirm dev console button is visible in EAS Update CI environment
- [ ] Confirm dev console button is NOT visible in production builds
- [ ] Verify EAS builds use cached native builds (no fingerprint drift)
- [ ] Verify `mobile:validate-build-pre-eas` no longer triggers on extension-only PRs


kyranjamie:
bullish on this PR

packages/ui/src/components/sheet/sheet.native.tsx
```diff
@@ -63,6 +63,7 @@ export function Sheet({
 
   return (
     <BottomSheetModal
+      accessible={Platform.OS === 'ios' ? false : undefined}
```
pete-watters: `Create Wallet` passes on Android but not iOS and I tried to get it to work with this hack but it didn't yet.

My research said: iOS Bottom Sheet Accessibility Issue

  

> Problem: Maestro E2E tests pass on Android but fail on iOS — Maestro can't find any elements
>   (testID or text) inside @gorhom/bottom-sheet modals. The sheet is visually present but completely
>    invisible to the test framework.
> 
>   Root cause: In @gorhom/bottom-sheet v4.6.0+
>   (https://github.com/gorhom/react-native-bottom-sheet/pull/1288), the library started setting
>   accessible={true} and accessibilityRole="adjustable" by default on the BottomSheet component. On
>   iOS, when a parent view has accessible={true}, UIKit groups all its children into a single
>   accessibility element — meaning VoiceOver (and by extension Maestro's XCUITest driver) sees the
>   entire sheet as one opaque element labelled "Bottom Sheet" rather than exposing individual
>   buttons, text, etc. inside it. This doesn't affect Android because TalkBack handles
>   accessible={true} differently.
> 
>   Fix: Set accessible={false} on iOS for BottomSheetModal:
>   <BottomSheetModal
>     accessible={Platform.OS === 'ios' ? false : undefined}
>     // ...
>   />
> 
>   This is the documented solution from both the
>   https://github.com/mobile-dev-inc/maestro/issues/1493 and the
>   https://github.com/gorhom/react-native-bottom-sheet/issues/712. It restores per-element
>   accessibility on iOS while leaving Android behaviour unchanged.
> 
>   Note: This also improves real VoiceOver accessibility — previously iOS VoiceOver users couldn't
>   navigate individual elements inside sheets either.

pete-watters: @edgarkhanzadian @camerow do you have any ideas here? What about just skipping the iOS suite for now and getting the value added from comprehensive tests on Android? 

Ideally we could fix the bottom sheet issue but in the interim we could increase our E2E coverage on one platform at least. 

The tests seem to run pretty fast now too

pete-watters:
@kyranjamie @camerow @edgarkhanzadian this is a draft which needs some finesse but opening for review to discuss about how we can solve the issue of the bottom sheet tests failing for iOS only. 

Perhaps we just disable the iOS suite for now and accept the value of having a better set of tests on one platform that trying to get them on both?

https://github.com/leather-io/mono/pull/2091#discussion_r2804240176

packages/ui/src/components/sheet/sheet.native.tsx
camerow: Couldn't we just take the recommended fix here and set `accessible={false}` for ios tests?

pete-watters: I had tried that via the use of `Platform.OS`. I've hardcoded it with `accessible={false}` to see how that goes. 

If it still fails I will try something else or we could consider not running the tests on iOS. 

I could come up with a different hidden way of creating wallets just for tests

.github/workflows/mobile:validate-build-pre-eas.yml
```diff
@@ -7,52 +7,73 @@ on:
...
       - 'packages/**'
       - 'pnpm-lock.yaml'
       - 'turbo.json'
-  
+
   # Allow manual trigger
   workflow_dispatch:
 
 jobs:
+  code-quality-gate:
```
pete-watters, line 39: This makes sure we don't run the build too often by making sure the other standard checks have passed

apps/mobile/maestro/flows/android/010-create-wallet-sheet.yaml
```diff
@@ -0,0 +1,17 @@
+appId: io.leather.mobilewallet
```
pete-watters, line 1: This adds a specific test to make sure we can restore wallets from the bottom sheet. It's for Android only as iOS tests cannot use buttons in the sheet

packages/ui/src/components/sheet/sheet.native.tsx
pete-watters: I decided to fix this by:
- adding a new Android only smoke test for wallet restoration from the bottom sheet
- re-introducing the developer tools so we can create sheets from there for other tests
- developer tools were previously deleted in [this PR](https://github.com/leather-io/mono/commit/587cd6a54321487011a34c736b00cbe1b1a3fec1)

apps/mobile/src/components/screen/screen-header/components/header-actions.tsx
```diff
@@ -17,6 +24,14 @@ export function HeaderActions() {
 
   return (
     <Box alignItems="center" flexDirection="row" justifyContent="center">
+      {__DEV__ && (
```
pete-watters: I'm re-adding this to help with the iOS tests. It was removed in [this PR ](https://github.com/leather-io/mono/pull/2091#discussion_r2833891790). 

Re-adding this only in `DEV` means that the tests can use it to create wallets on both platforms. 

We then have a separate smoke test for wallet creation from the sheet on Android only. That means we can make sure that works without blocking other iOS tests

apps/mobile/maestro/flows/full-suite-ci.yaml
```diff
@@ -142,8 +142,7 @@ name: Full Test Suite CI
     id: 'homeSettingsButton'
 - tapOn:
     id: 'settingsNetworkButton'
-- tapOn:
-    text: 'Testnet4'
+- tapOn: 'Testnet4.*Disabled'
```
pete-watters, line 157: This uses a REGEX match to make sure it can find this button

.github/workflows/mobile:validate-build-pre-eas.yml
camerow: That's a nice improvement!

pete-watters: Yes it’s good to help not be wasteful with the builds

PR #2096, opened 2026-02-10
fix(extension): show fiat balance for zero balance tokens

pete-watters:
## Summary
- Fixes a bug where tokens with zero balance (like USDCx) were not showing their fiat value ($0.00)
- The condition `availableBalance.amount.toNumber() > 0` was hiding the fiat balance for zero balance tokens
- Always display the fiat balance regardless of the crypto balance amount

## Before

Screenshot 2026-02-10 at 11 01 45

## After

Screenshot 2026-02-10 at 11 01 05

### DIFF
--- a/apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
+++ b/apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
@@ -62,7 +62,7 @@ export function CryptoAssetItemLayout({
             data-state={isLoadingAdditionalData ? 'loading' : undefined}
             className={shimmerStyles}
           >
-            {availableBalance.amount.toNumber() > 0 ? fiatBalance : null}
+            {fiatBalance}
           </PrivateTextLayout>
           {titleRightBulletInfo}
         </BulletSeparator>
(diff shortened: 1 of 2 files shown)


apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -62,7 +62,7 @@ export function CryptoAssetItemLayout({
             data-state={isLoadingAdditionalData ? 'loading' : undefined}
             className={shimmerStyles}
           >
-            {availableBalance.amount.toNumber() > 0 ? fiatBalance : null}
```
pete-watters, line 65: I'm not sure if `availableBalance.amount.toNumber() > 0` is intentional but I think it looks weird not showing the 0 balance in fiat

edgarkhanzadian (approved):

apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
kyranjamie: this is weird code anyway,  why convert to a number to do math on a BigNumber, when BigNumber has math methods on it...

kyranjamie (approved):

apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
kyranjamie: this could well have been to prevent negative balances

PR #2097, opened 2026-02-10
chore(web): use default name set in environment

camerow:


apps/web/app/utils/analytics/analytics.ts
```diff
@@ -4,7 +4,7 @@ import { VERSION } from '~/constants/constants';
 import { configureAnalyticsClient } from '@leather.io/analytics';
 import { noop } from '@leather.io/utils';
 
-const writeKey = import.meta.env.LEATHER_MIXPANEL_TOKEN;
+const writeKey = import.meta.env.MIXPANEL_TOKEN;
```
edgarkhanzadian, line 7: you would also need to update this in github action yaml

PR #2098, opened 2026-02-11
refactor(extension): remove BitcoinSigner entirely

kyranjamie:
This PR address some feedback from @edgarkhanzadian's review. 

19f1dacb90109d704f227fbfd2521cfcd49fafa0

This commit finishes a refactor that had already partially been complete relating to `BitcoinSigner` / `BitcoinPayer`. In decoupling the signing logic, such to create a more forgiving abstraction, we can now remove the concept of `BitcoinSigner` entirely.

### DIFF
--- a/apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-payer.ts
+++ b/apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-payer.ts
@@ -2,21 +2,36 @@ import { useCallback } from 'react';
 
 import { HDKey, Versions } from '@scure/bip32';
 import { SigHash } from '@scure/btc-signer';
+import type { P2Ret, P2TROut } from '@scure/btc-signer/payment';
 
 import {
-  BitcoinSigner,
+  BitcoinNativeSegwitPayer,
+  BitcoinPayer,
+  BitcoinTaprootPayer,
   deriveAddressIndexKeychainFromAccount,
   isNativeSegwitDerivationPath,
   isTaprootDerivationPath,
-  makeNativeSegwitAddressIndexDerivationPath,
-  makeTaprootAddressIndexDerivationPath,
-  whenPaymentType,
 } from '@leather.io/bitcoin';
-import { extractAddressIndexFromPath, extractChangeIndexFromPath } from '@leather.io/crypto';
+import {
+  appendAddressIndexToPath,
+  extractAddressIndexFromPath,
+  extractChangeIndexFromPath,
+} from '@leather.io/crypto';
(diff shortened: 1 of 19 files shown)


apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-payer.ts
```diff
@@ -41,16 +56,20 @@ export const allSighashTypes = [
 interface MakeBitcoinPayerArgs {
   keychain: HDKey;
   network: BitcoinNetworkModes;
-  derivationPath: string;
+  keyOrigin: string;
+  masterKeyFingerprint: string;
+  paymentType: 'p2wpkh' | 'p2tr';
   paymentFn(keychain: HDKey, network: BitcoinNetworkModes): any;
```
edgarkhanzadian: Any chance we could drop `any` here? Is there no way for us to know the response of paymentFn? Maybe we could at least use `unknown`?

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-payer.ts
```diff
@@ -41,16 +56,20 @@ export const allSighashTypes = [
...
 }
 function makeBitcoinPayer<T extends MakeBitcoinPayerArgs>(args: T) {
```
edgarkhanzadian, line 66: Would it be possible to run the response of this function through something like `satisfies PaymentToPayer<X>` instead of using `as unknown as PaymentToPayer<ExtractPaymentReturn<TPaymentFn>>` below? E.g. maybe in that case we could create type `X` from something that we get from generic type `T`.
I haven't spent too much time on this part of the code so i understand that something like this might not be feasible or worth implementing, but trying to reduce the amount of `any/unknown` being introduced into the codebase.

edgarkhanzadian (approved):
Great work!

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-payer.ts
kyranjamie: I can type as either of the return values here yes

apps/extension/src/app/store/accounts/blockchain/bitcoin/bitcoin-payer.ts
kyranjamie: When using generic types like this, I don't think you can, because the compiler can't satisfy it, we're relying on the inferred type. 

payer = BitcoinAccount, and we need to narrow it based on the argument provided, which I think can you only do with casting.

PR #2099, opened 2026-02-11
feat: asset-list service

alexp3y:
Adds `AssetListService` as a composable orchestrator for fetching filterable, sortable, paginated lists of `FungibleCryptoAsset` with opt-in data hydration
  - Supports all fungible asset protocols (`nativeBtc`, `nativeStx`, `sip10`, `rune`) with per-asset hydration of market data, market stats, token analytics, and user balances
  - Adds RQ config in `@leather.io/queries` for consumption by apps
 - Adds an integration test file `asset-list-test.mjs` and script that runs through and validates various usage scenarios (requires replacement of address / xpub placeholders).

`AssetListService.getAssetList(request)` accepts a single `AssetListRequest` that declaratively describes what the caller needs:

  - `filters`: protocol subset, chain, threshold filters (minMarketCap, minTrustScore, minTrendingScore, minDistributionScore), hasBalance, includeHidden
  - `includes`: opt-in hydration flags (`marketData`, `marketStats`, `analytics`, `balance`) -> only fetches what's requested
  - `sort`: ordered array of { field, direction } with 12 sort fields (name, marketCap, price, quoteTotalBalance, quoteAvailableBalance, change1d/1w/1m, trustScore, trendingScore, distributionScore, holderCount) and
   multi-field tiebreaking
  - `pagination`: offset/limit with accurate meta.total reflecting post-filter count
  - `accountContext`: optional account info for balance hydration

### DIFF
--- a/packages/services/src/asset-list/asset-list.service.ts
+++ b/packages/services/src/asset-list/asset-list.service.ts
@@ -1,0 +1,367 @@
+import { injectable } from 'inversify';
+import { entries } from 'remeda';
+
+import { btcAsset, stxAsset } from '@leather.io/constants';
+import type {
+  FungibleCryptoAsset,
+  FungibleCryptoAssetProtocol,
+  MarketStats,
+  TokenAnalytics,
+} from '@leather.io/models';
+import { type SerializedCryptoAssetId, getAssetId, serializeAssetId } from '@leather.io/utils';
+
+import { FungibleAssetVisibilityService } from '../assets/fungible-asset-visibility.service';
+import { createRuneAsset } from '../assets/rune-asset.utils';
+import { createSip10Asset } from '../assets/stacks-asset.utils';
+import { BtcBalancesService } from '../balances/btc-balances.service';
+import { RunesBalancesService } from '../balances/runes-balances.service';
+import { Sip10BalancesService } from '../balances/sip10-balances.service';
+import { StxBalancesService } from '../balances/stx-balances.service';
+import { LeatherApiClient } from '../infrastructure/api/leather/leather-api.client';
+import { MarketDataService } from '../market/market-data.service';
+import { MarketStatsService } from '../market/market-stats.service';
+import { TokenAnalyticsService } from '../token-analytics/token-analytics.service';
+import type {
+  AssetListItem,
+  AssetListItemBalance,
+  AssetListRequest,
+  AssetListResponse,
+  AssetListSortField,
+} from './asset-list.types';
+import {
+  filterAssetsByChain,
+  filterAssetsByProtocol,
+  filterByDistributionScore,
+  filterByHasBalance,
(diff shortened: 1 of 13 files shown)


packages/services/src/asset-list/asset-list.utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    return marketCap !== undefined && marketCap >= minMarketCap;
+  });
+}
+
+export function filterByTrustScore(items: AssetListItem[], minTrustScore: number): AssetListItem[] {
+  return items.filter(item => {
+    const trustScore = item.analytics?.trustScore;
+    return trustScore !== undefined && trustScore >= minTrustScore;
+  });
+}
```
edgarkhanzadian: I'd suggest creating a function that `items.filter` would be able to consume.
Like this:
```
items.filter(filterByTrustScore(minTrustScore))
```

Applies to all of the other function utils in this file.
```suggestion
export function filterByTrustScore(minTrustScore: number): {
  return function(item: AssetListItem) {
    const trustScore = item.analytics?.trustScore;
    return trustScore !== undefined && trustScore >= minTrustScore;
  }
}
```

packages/services/src/asset-list/asset-list.service.ts
```diff
@@ -0,0 +1,367 @@
...
+
+    let items: AssetListItem[] = assets.map(asset => {
+      const id = serializeAssetId(getAssetId(asset));
+      return {
+        id,
+        asset,
+        marketStats: marketStatsMap.get(id) ?? undefined,
+        analytics: analyticsMap.get(id) ?? undefined,
+      };
+    });
```
edgarkhanzadian, line 115: Just to avoid type casting
```suggestion
    let items = assets.map(asset => {
      const id = serializeAssetId(getAssetId(asset));
      return {
        id,
        asset,
        marketStats: marketStatsMap.get(id) ?? undefined,
        analytics: analyticsMap.get(id) ?? undefined,
      } satisfies AssetListItem;
    });
```

edgarkhanzadian (approved):
Great work!

pete-watters (approved):
This looks great! 🚀

packages/services/src/asset-list/asset-list.service.ts
alexp3y: Hmm, don't know if I agree that this constitutes type casting. It's actually just standard typing `items` as an `AssetListItem` array. We're explicitly static typing the var at its initialization rather than validating an inferred type via "satisfies", which imo is superior pattern.

edgarkhanzadian: Ah yeap you're right, for some reason i thought that doing `let items: AssetListItem[]` is equivalent to `let items = <something> as AssetListItem[]`. My bad

PR #2100, opened 2026-02-11
feat(mobile): trending tokens [LEA-3351]

edgarkhanzadian:
https://github.com/user-attachments/assets/f22f623e-4607-4bc4-82e7-be8ebbff604c

### DIFF
--- a/apps/mobile/src/queries/asset-list/trending-tokens.query.ts
+++ b/apps/mobile/src/queries/asset-list/trending-tokens.query.ts
@@ -1,0 +1,17 @@
+import { useUserSettings } from '@/hooks/use-user-settings';
+import { useQuery } from '@tanstack/react-query';
+
+import { createAssetListQueryConfig } from '@leather.io/queries';
+import { AssetListRequest } from '@leather.io/services';
+
+const trendingTokensRequest: AssetListRequest = {
+  filters: { protocols: ['sip10'], minTrendingScore: 8 },
+  includes: { marketStats: true },
+  sort: [{ field: 'trendingScore', direction: 'desc' }],
+  pagination: { limit: 15, offset: 0 },
+};
+
+export function useTrendingTokensQuery() {
+  const settings = useUserSettings();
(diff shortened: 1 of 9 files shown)


apps/mobile/src/queries/asset-list/trending-tokens.query.ts
```diff
@@ -0,0 +1,17 @@
+import { useUserSettings } from '@/hooks/use-user-settings';
+import { useQuery } from '@tanstack/react-query';
+
+import { createAssetListQueryConfig } from '@leather.io/queries';
+import { AssetListRequest } from '@leather.io/services';
+
+const trendingTokensRequest: AssetListRequest = {
+  filters: { protocols: ['sip10'], minTrendingScore: 1 },
```
alexp3y: We might want to add a minimum trust score here... I've found that somewhere in the 8-10 range is a decent baseline for quality.

apps/mobile/src/queries/asset-list/trending-tokens.query.ts
```diff
@@ -0,0 +1,17 @@
...
+  includes: { marketStats: true },
+  sort: [{ field: 'trendingScore', direction: 'desc' }],
```
alexp3y, line 10: FYI The trending score can rank by either positive or negative trend (i.e. it extends both positive and negative, e.g. -3, -2, -1, 0, 1, 2, 3). 

I think we only had positive trend in design but food for thought we could also incorporate trending by negative change by switching the sort direction. cc @mica000

alexp3y (approved):
Good stuff 🦾

apps/mobile/src/queries/asset-list/trending-tokens.query.ts
edgarkhanzadian: Thank you for the fine tune! I've noticed when setting minTrendingScore to 8, we mostly get 0 "1d" changes, is that expected?
trending

PR #2102, opened 2026-02-11
fix(mobile): help header

edgarkhanzadian:
help


pete-watters (approved):
Thanks @edgarkhanzadian ! I noticed this broken when working on the maestro tests

PR #2105, opened 2026-02-12
feat: taproot sends [LEA-3443]

edgarkhanzadian:
Description of fixes and improvements:
- use single btc fee size estimator from @leather.io/bitcoin package
- update btc fee size estimator to count fee for taproot+ns mixed input transactions.
- use taproot+ns utxos instead of just ns utxos when doing calculations
- do not use only native segwit balance. Always use both taproot and ns balance combined

Manually tested:
- sending using only taproot utxos, only ns utxos, taproot+ns
- transaction rbf
- sending ordinals

Note:

fee estimation is going to be replaced by a service in a subsequent PR.

Potential things to do in subsequent PRs:
- remove parts of code responsible for retrieving taproot funds
- Try to use as less `new btc.Transaction()` as possible (maybe at least merge formTaprootOrdinalTx and formNativeSegwitOrdinalTx into one function)


packages/bitcoin/src/utils/bitcoin.utils.ts
```diff
@@ -388,3 +390,9 @@ export function isTaprootDerivationPath(path: string) {
 export function isNativeSegwitDerivationPath(path: string) {
   return extractPurposeFromPath(path) === 84;
 }
+
+export function isP2TROut(
+  signer: BitcoinSigner<P2Ret | P2TROut> | null
```
kyranjamie, line 395: BitcoinSigner is gone in my PR, now we only refer to "payers"

apps/extension/src/app/query/bitcoin/utxos/utxos.hooks.ts
```diff
@@ -1,16 +1,16 @@
...
 import { useTaprootAccountRequest } from '@app/services/accounts/use-taproot-account-request';
 import { toFetchState } from '@app/services/fetch-state';
 
 import { useNativeSegwitAccountRequest } from '../../../services/accounts/use-native-segwit-account-request';
 import { useGetAccountUtxosQuery } from './utxos.query';
 
-export function useCurrentNativeSegwitUtxos() {
-  const accountRequest = useNativeSegwitAccountRequest();
-  const query = useGetAccountUtxosQuery(accountRequest);
+function useBaseCurrentUtxos(request: AccountRequest) {
```
kyranjamie: what does base mean?

apps/extension/src/app/pages/send/ordinal-inscription/hooks/use-generate-ordinal-tx.ts
```diff
@@ -151,6 +158,9 @@ export function useGenerateUnsignedOrdinalTx(inscriptionInput: UtxoWithDerivatio
       // Fee-covering Native Segwit inputs
       inputs.forEach(input => {
         const signer = getSignerForInput(input);
+        const tapInternalKey = isP2TROut(signer)
+          ? { tapInternalKey: signer.payment.tapInternalKey }
+          : {};
```
kyranjamie, line 163: 👍🏼

apps/extension/src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
```diff
@@ -69,18 +73,28 @@ export function useGenerateUnsignedNativeSegwitTx({
...
-              script: p2wpkh.script,
-              amount: BigInt(input.value),
-            },
-          });
+          if (isTaprootDerivationPath(input.path)) {
+            tx.addInput({
+              txid: input.txid,
+              index: input.vout,
+              sequence: 0,
+              tapInternalKey: (inputSigner.payment as P2TROut).tapInternalKey,
```
kyranjamie: can we narrow the type on the `.payment` so the any isn't needed?

apps/extension/src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
```diff
@@ -69,18 +73,28 @@ export function useGenerateUnsignedNativeSegwitTx({
...
+            tx.addInput({
+              txid: input.txid,
+              index: input.vout,
+              sequence: 0,
+              witnessUtxo: {
+                script: inputSigner.payment.script,
+                amount: BigInt(input.value),
+              },
+            });
+          }
```
kyranjamie: Personal pref would be a whenPayment type or something to avoid else block

kyranjamie (approved):
Thanks @edgarkhanzadian LGTM. This will conflict with my work most notably because of the reframing of "signer" -> "payer".

apps/extension/src/app/query/bitcoin/utxos/utxos.hooks.ts
edgarkhanzadian: Yea not a good name. this is an internal function that should be used by useCurrentUtxos and useCurrentNativeSegwitUtxos. I'll rename it to just useUtxos

apps/extension/src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
edgarkhanzadian: ah yes forgot to reuse isP2TROut here

apps/extension/src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
edgarkhanzadian: using isP2TROut would avoid if/else here

PR #2106, opened 2026-02-13
fix(services): collectibles api improvements

pete-watters:
This PR adds some improvements to the collectibles API to address some bugs and un-desirable behaviour

1. [fix(services): prevent ipfs url double-encoding](https://github.com/leather-io/mono/pull/2106/changes/25de907c7ad6988264b80f971938542000234fc4) : fixes a bug where we were incorrectly encoding paths that were already encoded 
2. [fix(services): use promise.allsettled for resilient nft metadata fetc…](https://github.com/leather-io/mono/commit/e4a580836e8622b3ddec68ceb94567d4bf6088fd) makes a change to use `Promise.allSettled` for NFT metadata fetching so one API failure doesn't break the entire request
3. [fix(services): handle api errors gracefully in gamma and bns services](https://github.com/leather-io/mono/commit/e83ce78a31afc0d47c0c79fee17a4c1371824be8) Adds error handling to Gamma and BNS services (return null/empty instead of  throwing)
4. [fix(services): filter lp tokens from collectibles](https://github.com/leather-io/mono/commit/0ff010cbb65000567c9d3e52f89c39b500926e64) adds a filter to exclude LP (liquidity provider) tokens from collectibles calls. These are not real collectibles and cause 500 errors with the gamma API as there is not metadata for it because it's a DeFi position token. See [slack discussion here](https://trustmachines.slack.com/archives/C05LAP952E7/p1770744310876949)
5. [fix(services): add pagination for nft holdings endpoint](https://github.com/leather-io/mono/commit/d73442849bb000890b0adb2fd7d4d62f5061764e) adds pagination support to the collectibles holding endpoint. That allows it to fetch all the NFTs rather than only the first 100.

### DIFF
--- a/packages/services/src/infrastructure/api/gamma/gamma-api.client.ts
+++ b/packages/services/src/infrastructure/api/gamma/gamma-api.client.ts
@@ -20,16 +20,20 @@ export class GammaApiClient {
     contractPrincipal: string,
     tokenId: number,
     { signal, skipCache }: ApiRequestOptions = {}
-  ): Promise<GammaNftMetadata> {
+  ): Promise<GammaNftMetadata | null> {
     async function fetchFn() {
-      const res = await axios.get(
-        `${GAMMA_API_URL}/get-stacks-nft?id=${contractPrincipal}_${tokenId}`,
-        {
-          signal,
-        }
-      );
+      try {
+        const res = await axios.get(
+          `${GAMMA_API_URL}/get-stacks-nft?id=${contractPrincipal}_${tokenId}`,
+          {
+            signal,
+          }
+        );
 
-      return gammaNftMetadataSchema.parse(res.data);
+        return gammaNftMetadataSchema.parse(res.data);
+      } catch {
+        return null;
+      }
     }
     return skipCache
       ? await fetchFn()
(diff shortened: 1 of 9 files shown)


packages/services/src/infrastructure/api/hiro/hiro-stacks-api.client.ts
```diff
@@ -402,8 +400,16 @@ export class HiroStacksApiClient {
...
       );
-      return res.data.results;
-    };
+      return res.data;
+    }.bind(this);
+
+    async function fetchFn() {
+      return fetchHiroPages(fetchPage, {
+        limit,
+        pagesRequest: { allPages: true },
```
alexp3y, line 409: Maybe we want to put a limit on the pagesRequest here (`stopAfter: x`) in cases where users have hundreds / thousands of NFTs?

packages/services/src/infrastructure/api/gamma/gamma-api.client.ts
```diff
@@ -42,15 +46,19 @@ export class GammaApiClient {
...
-      return gammaCollectionMetadataSchema.parse(res.data);
+      try {
+        const res = await axios.get(
+          `${GAMMA_API_URL}/get-stacks-collection?contract_id_or_slug=${contractPrincipal}`,
+          {
+            signal,
+          }
+        );
+        return gammaCollectionMetadataSchema.parse(res.data);
+      } catch {
```
alexp3y, line 59: Might want to consider being slightly less restrictive here - For example, checking for certain HTTP response codes (4xx, even 5xx) that correspond to a `null` return value and allowing other error types to still propagate.

alexp3y (approved):
Some nice defensive additions here 🥷

pete-watters:
> Some nice defensive additions here 🥷

Thanks @alexp3y 👍 I'm going to merge this now and add improvements in a follow up as these changes are needed for other PRs. [I opened a reminder here](https://linear.app/leather-io/issue/LEA-3444/collectible-service-improvements). I'll get them added before we go live in the `extension` with this

PR #2108, opened 2026-02-13
feat(extension): make receive dialog tabs route-driven

pete-watters:
This PR makes a change to the receive modal to make it more route driven. I am trying this to see if its a way we can get around problems with flakiness in `shard-3`. 

Often these tests seem to fail and time out trying to onboard a user and verify a taproot address is generated. To do this now we:
- open the 'Receive' modal 
- click the collectibles tab 
- click taproot and assert for address

This PR makes a change that lets us bypass that and go directly to a route for '/receive/collectibles` 

https://github.com/user-attachments/assets/c00ed17f-eec6-4ad6-8c80-d32ac85037ec


pete-watters:
I opened this to try and improve shard 3 flakiness. It’s hard to tell if it is making a difference or not for sure but it could be a useful improvement nonetheless

pete-watters:
I'll close this off as its unclear if this is helpful or not

PR #2113, opened 2026-02-16
refactor(state, mobile): add a `SwapProvider` to swap state

tigranpetrossian:
Adds a provider on top of swap state to reduce potential prop drilling. This was less relevant on mobile, but makes more sense with react-router based setups on the extension and web.

Plus one small commit replacing leftover `@leather.io/ui/native` imports with `@leather.io/ui` for couple of shared hooks used in the state machine.

### DIFF
--- a/apps/mobile/src/features/swap/swap.tsx
+++ b/apps/mobile/src/features/swap/swap.tsx
@@ -8,7 +8,7 @@ import { analytics } from '@/utils/analytics';
 
 import { stxAsset } from '@leather.io/constants';
 import { SwappableFungibleCryptoAsset } from '@leather.io/models';
-import { useLiveSwapEstimate, useSwapState } from '@leather.io/state/swap';
+import { SwapProvider, useLiveSwapEstimate, useSwapContext } from '@leather.io/state/swap';
 
 import { SwapFormScreen } from './screens/swap-form-screen';
 import { SwapReviewScreen } from './screens/swap-review-screen';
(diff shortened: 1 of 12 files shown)


apps/mobile/src/features/swap/swap.tsx
```diff
@@ -21,29 +21,37 @@ interface SwapProps {
...
   const liveEstimate = useLiveSwapEstimate({
-    quoteQuery: swapStateResult.quoteQuery,
-    networkFeeQuery: swapStateResult.networkFeeQuery,
-    baseMarketDataQuery: swapStateResult.baseMarketDataQuery,
-    nativeAssetMarketDataQuery: swapStateResult.networkFeeAssetMarkedDataQuery,
+    quoteQuery: swapContext.quoteQuery,
+    networkFeeQuery: swapContext.networkFeeQuery,
+    baseMarketDataQuery: swapContext.baseMarketDataQuery,
+    nativeAssetMarketDataQuery: swapContext.networkFeeAssetMarkedDataQuery,
   });
```
edgarkhanzadian, line 51: Should we consider adding liveEstimate to the swap context as well?

edgarkhanzadian (approved):

apps/mobile/src/features/swap/swap.tsx
tigranpetrossian: I thought about this before, but `useLiveEstimate` is really a glue hook between the main state machine and the UI, and it exclusively exists to cater for UI quirks: timers, merging data from queries in a very opinionated manner, etc. I'd avoid baking it into the core state at least for now.

PR #2114, opened 2026-02-16
Add useAmountField hook

tigranpetrossian:
Adds a generic hook for controlling amount fields in different features across platforms.

The core hook supports input gating, formatting (with optional grouping), and signaling input rejection. 
The actual exported hooks are thin wrappers over the core with platform-specific adapters.

This is meant to replace all our amount fields across features (Send/Swap, etc) across platforms with a single shared implementation and behavior.

I'll post a demo in the upcoming extension PR where it's being used.

### DIFF
--- a/packages/ui/src/hooks/use-amount-field/use-amount-field.native.ts
+++ b/packages/ui/src/hooks/use-amount-field/use-amount-field.native.ts
@@ -1,0 +1,44 @@
+import { type RefObject, useRef } from 'react';
+import { type TextInput, type TextInputProps } from 'react-native';
+
+import {
+  type CurrencySign,
+  type UseAmountFieldProps,
+  useBaseAmountField,
+} from './use-base-amount-field.shared';
+
+type NativeInputProps = Pick<TextInputProps, 'value' | 'onChangeText' | 'keyboardType'>;
+
+export interface UseAmountFieldResult {
+  ref: RefObject<TextInput | null>;
+  textInputProps: NativeInputProps;
+  touched: boolean;
+  currencySign?: CurrencySign;
+}
+
+export function useAmountField(options: UseAmountFieldProps): UseAmountFieldResult {
+  const inputRef = useRef<TextInput>(null);
+
+  const { displayValue, handleChange, touched, currencySign } = useBaseAmountField(options);
+
+  function onChangeText(text: string) {
+    const result = handleChange(text, text.length);
+
+    if (!result.accepted && inputRef.current) {
+      inputRef.current.setNativeProps({ text: result.displayValue });
+    }
+  }
+
+  return {
+    ref: inputRef,
+    touched,
+    currencySign,
+    textInputProps: {
+      value: displayValue,
+      onChangeText,
+      keyboardType: 'decimal-pad',
+    },
+  };
+}
+
+export { type CurrencySign, type UseAmountFieldProps };
--- a/packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
+++ b/packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
@@ -1,0 +1,64 @@
+import {
+  type ChangeEvent,
+  type InputHTMLAttributes,
+  type RefObject,
+  useLayoutEffect,
+  useRef,
+} from 'react';
+
+import {
+  type CurrencySign,
+  type UseAmountFieldProps,
+  useBaseAmountField,
+} from './use-base-amount-field.shared';
+
+type InputProps = Pick<
+  InputHTMLAttributes<HTMLInputElement>,
+  'value' | 'onChange' | 'type' | 'inputMode'
+>;
+
+export interface UseAmountFieldResult {
+  ref: RefObject<HTMLInputElement | null>;
+  inputProps: InputProps;
+  touched: boolean;
+  currencySign?: CurrencySign;
+}
+
+export function useAmountField(props: UseAmountFieldProps): UseAmountFieldResult {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const cursorRef = useRef<number | null>(null);
+
+  const { displayValue, handleChange, touched, currencySign } = useBaseAmountField(props);
+
+  useLayoutEffect(() => {
+    if (cursorRef.current !== null && inputRef.current) {
+      inputRef.current.setSelectionRange(cursorRef.current, cursorRef.current);
+      cursorRef.current = null;
+    }
+  });
+
+  function onChange(e: ChangeEvent<HTMLInputElement>) {
+    const result = handleChange(e.target.value, e.target.selectionStart ?? 0);
+
+    if (result.accepted) {
+      cursorRef.current = result.cursorPosition;
+    } else {
+      e.target.value = result.displayValue;
+      e.target.setSelectionRange(result.cursorPosition, result.cursorPosition);
+    }
+  }
+
+  return {
+    ref: inputRef,
+    touched,
+    currencySign,
+    inputProps: {
+      type: 'text',
+      inputMode: 'decimal',
+      value: displayValue,
+      onChange,
+    },
+  };
+}
+
+export { type CurrencySign, type UseAmountFieldProps };
--- a/packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
+++ b/packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
@@ -1,0 +1,242 @@
+import { useMemo, useState } from 'react';
+
+export interface UseAmountFieldProps {
+  value: string;
+  onChange(rawValue: string): void;
+  onReject?(): void;
+  maxDecimals: number;
+  locale?: string;
+  enableGrouping?: boolean;
+  currency?: string;
+}
+
+export interface CurrencySign {
+  symbol: string;
+  placement: 'prefix' | 'suffix';
+}
+
+export interface ChangeResult {
+  accepted: boolean;
+  displayValue: string;
+  cursorPosition: number;
+}
+
+export interface UseBaseAmountFieldResult {
+  displayValue: string;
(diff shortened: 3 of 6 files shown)


packages/ui/src/hooks/use-amount-field/use-amount-field.native.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const onChangeText = useCallback(
+    (text: string) => {
+      const result = handleChange(text, text.length);
+
+      if (!result.accepted && inputRef.current) {
+        inputRef.current.setNativeProps({ text: result.displayValue });
+      }
+    },
+    [handleChange]
```
edgarkhanzadian: handleChange is not wrapped in useCallback, so onChangeText is going to rerun on every render. I'd suggest either removing useCallback from onChangeText or adding one to handleChange

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
```diff
@@ -0,0 +1,62 @@
...
+  touched: boolean;
+}
+
+export function useAmountField(props: UseAmountFieldProps): UseAmountFieldResult {
+  const inputRef = useRef<HTMLInputElement>(null);
+  const cursorRef = useRef<number | null>(null);
+
+  const { displayValue, handleChange, touched } = useBaseAmountField(props);
+
+  useLayoutEffect(() => {
```
edgarkhanzadian, line 33: Just curious, why useLayoutEffect instead of useEffect here?

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
```diff
@@ -0,0 +1,62 @@
...
+      const result = handleChange(e.target.value, e.target.selectionStart ?? 0);
+
+      if (result.accepted) {
+        cursorRef.current = result.cursorPosition;
+      } else {
+        e.target.value = result.displayValue;
+        e.target.setSelectionRange(result.cursorPosition, result.cursorPosition);
+      }
+    },
+    [handleChange]
```
edgarkhanzadian: same thing regarding handleChange

packages/ui/src/hooks/use-amount-field/use-amount-field.native.ts
```diff
@@ -0,0 +1,41 @@
...
+    touched,
+    textInputProps: {
+      value: displayValue,
+      onChangeText,
+      keyboardType: 'decimal-pad',
+    },
+  };
+}
+
+export { type UseAmountFieldProps };
```
edgarkhanzadian: file name is `use-amount-field.native.tsx` even though it doesn't contain any tsx. Suggest renaming to `use-amount-field.native.ts`

packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
```diff
@@ -0,0 +1,225 @@
...
+  maxDecimals,
+  locale = 'en',
+  enableGrouping = true,
+}: UseAmountFieldProps): UseBaseAmountFieldResult {
+  const formatter = useMemo(() => new Intl.NumberFormat(locale), [locale]);
+
+  const decimalSeparator = useMemo(() => extractDecimalSeparator(formatter), [formatter]);
+
+  const groupSeparator = useMemo(
+    () => (enableGrouping ? extractGroupSeparator(formatter) : ''),
```
edgarkhanzadian: Why not set groupSeparator to null? Code checks groupSeparator on almost every use anyway
```suggestion
    () => (enableGrouping ? extractGroupSeparator(formatter) : null),
```

packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    if (exceedsMaxDecimals(sanitized, decimalSeparator, maxDecimals)) {
+      onReject?.();
+      return rejected;
+    }
+
+    if (sanitized === value) {
+      return rejected;
+    }
```
edgarkhanzadian, line 107: Code doesn't call onReject here, is that intentional?
```suggestion
    if (sanitized === value) {
      onReject?.()
      return rejected;
    }
```

packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
```diff
@@ -0,0 +1,225 @@
...
+    () => escapeForRegex(decimalSeparator),
+    [decimalSeparator]
+  );
+
+  const [touched, setTouched] = useState(false);
+
+  const displayValue = formatRawForDisplay(value, formatter, decimalSeparator, enableGrouping);
+
+  function handleChange(newText: string, cursorPosition: number): ChangeResult {
+    setTouched(true);
```
edgarkhanzadian, line 56: setTouched is never called with a false value, is that expected?

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
```diff
@@ -0,0 +1,62 @@
+import {
+  type ChangeEvent,
+  type InputHTMLAttributes,
+  type RefObject,
+  useCallback,
+  useLayoutEffect,
+  useRef,
+} from 'react';
+
+import { UseAmountFieldProps, useBaseAmountField } from './use-base-amount-field.shared';
```
edgarkhanzadian: nit of all nits
```suggestion
import { type UseAmountFieldProps, useBaseAmountField } from './use-base-amount-field.shared';
```

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
tigranpetrossian: There's a subtle distinction here I always trip on too, render !== commit to DOM, `useEffect` runs after former, `useLayoutEffect` - the latter, so the position intermittently positions against the old string with `useEffect`.

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
tigranpetrossian: ✅
There's a fixable rule we could enable for this, I'm just hesitant because of how many files it will touch initially.

packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
tigranpetrossian: Expected, this is a one-off flag for the lifecycle of the input.

packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
tigranpetrossian: It is intentional, but it is admittedly confusing. Basically, if the user deletes or types anything that results in the same value, e.g. deleting a group separator, the hook doesn't reject-reject, just no-ops (tests cover this).

I think the name `rejected` is what's misleading here, because the actual payload `accepted: false` with the same value. I'll see if I can express this better or as last resort add a comment.

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
```diff
@@ -0,0 +1,62 @@
...
+type InputProps = Pick<
+  InputHTMLAttributes<HTMLInputElement>,
+  'value' | 'onChange' | 'type' | 'inputMode'
+>;
+
+export interface UseAmountFieldResult {
+  ref: RefObject<HTMLInputElement | null>;
+  inputProps: InputProps;
+  touched: boolean;
+}
```
kyranjamie, line 25: Can this go in a `*.shared.ts` file?

kyranjamie (approved):

kyranjamie:
Also, if this is for currency amounts, can the hook name specify this?

packages/ui/src/hooks/use-amount-field/use-base-amount-field.shared.ts
tigranpetrossian: groupSeparator is used as a `split` argument down the line, and no-ops if it's empty.

packages/ui/src/hooks/use-amount-field/use-amount-field.native.ts
tigranpetrossian: ✅

packages/ui/src/hooks/use-amount-field/use-amount-field.native.ts
tigranpetrossian: Claude loves those and I keep missing it. 
Removed both. I'm all for stable references for hooks that are called at the top level of large features, but this isn't one.

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
tigranpetrossian: ✅

packages/ui/src/hooks/use-amount-field/use-amount-field.web.ts
tigranpetrossian: This is a web-specific signature. The web and native have just one field in common, so probably not worth the abstraction.

tigranpetrossian:
> Also, if this is for currency amounts, can the hook name specify this?

The use cases happen to be currency amounts but it's not inherently currency-only--no baked in opinions or currency symbols, so it can be used for any amounts.

tigranpetrossian:
@edgarkhanzadian I think I addressed all of your feedback, thanks 🙏 
I missed bunch of `act` warnings in tests, need to add & verify these before merging.

tigranpetrossian:
Added an optional currency sign support, since the hook internally deals with formatting anyway. Still avoiding to name it currency amount in case we ever need to use for any other amount inputs.

PR #2116, opened 2026-02-16
Branch out old Swap in the extension

tigranpetrossian:
This separates the existing Swap, sets up the basic structure for the new version under a flag.

The routing structure in the old version is rather flawed, but preserving it for backwards compat until completely phased out.

@kyranjamie Had to think about this twice given how you feel about flags, but I think this is a classic legitimate use-case. Thankfully (not to jinx it), the work on the extension version is moving pretty fast.


apps/extension/src/app/pages/swap/swap-review.tsx
```diff
@@ -0,0 +1,5 @@
+import { Card } from '@app/components/layout';
+
+export function SwapReview() {
+  return <Card>{null}</Card>;
+}
```
edgarkhanzadian, line 5: Do we need this?

apps/extension/src/app/pages/swap-legacy/swap.tsx
```diff
@@ -0,0 +1,56 @@
...
+          fullWidth
+        >
+          Review and swap
+        </Button>
+      }
+    >
+      <SwapAssetSelectBase />
+      <SwapAssetSelectQuote />
+
+      {isBtcToSbtcSwap(values.swapAssetBase, values.swapAssetQuote ?? { tokenId: 'unknown' }) && (
```
edgarkhanzadian, line 49: why tokenId: 'unknown' instead of tokenId: null?

edgarkhanzadian (approved):

apps/extension/src/app/pages/swap-legacy/swap.tsx
tigranpetrossian: No idea, old code--soon to be gone.

apps/extension/src/app/pages/swap/swap-review.tsx
tigranpetrossian: Yep, just a placeholder for now for routes to work. This is under a flag.

PR #2118, opened 2026-02-16
chore: update use-enroll-transaction.ts

314159265359879:
users are still calling the old contract to "enroll" but that doesn't work they have to use the new dual-stacking contract.

when the user clicks enroll here
image

the new contract needs to be called.


kyranjamie:
opening another pr

PR #2122, opened 2026-02-17
fix(services): failing rate limiter unit test

alexp3y:
Fixes failing `RateLimiterService` unit test.

Adds timeout between test to ensure the rate limiter interval is reset in the underlying PQueue instance.

### DIFF
--- a/packages/services/src/infrastructure/rate-limiter/rate-limiter.service.spec.ts
+++ b/packages/services/src/infrastructure/rate-limiter/rate-limiter.service.spec.ts
@@ -1,3 +1,7 @@
+import { beforeEach, describe, expect, it } from 'vitest';
+
+import { delay } from '@leather.io/utils';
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):

packages/services/src/infrastructure/rate-limiter/rate-limiter.service.spec.ts
```diff
@@ -17,6 +19,10 @@ describe('RateLimiterService', () => {
       }) as unknown as UserSettings,
   };
 
+  beforeEach(async () => {
+    await new Promise(resolve => setTimeout(resolve, 100));
```
kyranjamie: we have a delay utility for this you can use fyi

kyranjamie:
merged in another pr, thanks alex

PR #2124, opened 2026-02-17
chore: update send-form-confirmation.layout.tsx

camerow:
## Summary
- Improve warning text when sending to a smart contract

## Notes
This PR takes over the work from #2104 (by @314159265359879) to ensure CI runs properly with repo permissions.

Closes #2104

### DIFF
--- a/apps/extension/src/app/pages/send/send-crypto-asset-form/form/send-form-confirmation.layout.tsx
+++ b/apps/extension/src/app/pages/send/send-crypto-asset-form/form/send-form-confirmation.layout.tsx
@@ -79,8 +79,7 @@ export function SendFormConfirmationLayout({
       )}
       {principalType === 'contract' && (
         <Callout variant="warning" title="Sending to a smart contract" px="space.03" mb="space.05">
-          The recipient is a stacks contract principal. Make sure you trust the contract before
-          sending funds.
+          {`The recipient is a Stacks contract address. Make sure you understand how the contract works before sending funds.`}
         </Callout>
       )}
 


apps/extension/src/app/pages/send/send-crypto-asset-form/form/send-form-confirmation.layout.tsx
```diff
@@ -79,8 +79,7 @@ export function SendFormConfirmationLayout({
       )}
       {principalType === 'contract' && (
         <Callout variant="warning" title="Sending to a smart contract" px="space.03" mb="space.05">
-          The recipient is a stacks contract principal. Make sure you trust the contract before
-          sending funds.
+          {`The recipient is a Stacks contract address. Make sure you understand how it works and that fund recovery is possible.`}
```
kyranjamie: >  and that fund recovery is possible

This assumes the use case of the transaction is to later get funds back. That isn't always the case, perhaps you're sending funds expecting to not get them back, in which case this doesn't make sense.

314159265359879: Ah you are right. Lets change to this instead:

The recipient is a Stacks contract address. Make sure you understand how the contract works before sending funds.

camerow: Updated

kyranjamie (approved):

PR #2134, opened 2026-02-19
feat(extension): use one fee service to rule them all

edgarkhanzadian:
Use fee services to calculate fees for both stx and btc transactions.


camerow:
cc @314159265359879 could you do a bit of sanity testing for the changes to fees here?

edgarkhanzadian:
@camerow @314159265359879 this is not yet production ready, give me just a little bit of time to finish up testing on my local and sync changes with the branch 🙏

apps/extension/src/app/components/fees-row/fees-row.tsx
```diff
@@ -17,8 +17,20 @@ import { FeeEstimateSelect } from './components/fee-estimate-select';
...
+  return {
+    fee: quote.value,
+    feeRate: quote.type === 'stacksFeeRate' ? quote.rate : 0,
+  };
+}
+
+function transactionFeesToEstimates(txFees: StacksTransactionFees): StacksFeeEstimate[] {
+  const { low, standard, high } = txFees.options;
+  return [toEstimate(low), toEstimate(standard), toEstimate(high)];
+}
```
edgarkhanzadian, line 30: this smells, we need to remove StacksFeeEstimate type entirely in favor of fee quotes, but i'm keeping that outside of this pr's scope as it would need more work on stacks sends

kyranjamie:
**Not to merge until extension release**

kyranjamie (approved):

apps/extension/src/app/components/bitcoin-custom-fee/hooks/use-bitcoin-custom-fee.tsx
```diff
@@ -16,6 +16,12 @@ interface UseBitcoinCustomFeeArgs {
   recipients: TransferRecipient[];
 }
 
+/**
```
pete-watters, line 19: Could we get rid of this now if it's deprecated? Or is something relying on it

apps/extension/src/app/components/bitcoin-fees-list/use-bitcoin-fees-list.ts
```diff
@@ -1,121 +1,78 @@
...
-    };
-
-    const determineUtxosForLowFeeArgs = {
-      ...determineUtxosDefaultArgs,
-      feeRate: feeRates.hourFee.toNumber(),
-    };
+    if (!feeRates) return [];
 
-    const feesArr = [];
+    const { low, standard, high } = feeRates.options;
```
pete-watters, line 61: I think this can throw a type error if `feeRates` is truthy but `feeRates.options` is `undefined` or `null`. Maybe we can add an extra guard here or use optional chaining instead?

apps/extension/src/app/features/dialogs/transaction-action-dialog/hooks/use-btc-increase-fee.ts
```diff
@@ -186,8 +178,8 @@ export function useBtcIncreaseFee(btcTx: BitcoinTx) {
 
           // check if fee is higher than 50 times the highest fee
           if (
-            feesList.length > 0 &&
-            bnValue.isGreaterThan(feesList[0].feeRate * MAX_FEE_RATE_MULTIPLIER)
+            feeRates &&
+            bnValue.isGreaterThan(feeRates.high.rate * sizeInfo.txVBytes * MAX_FEE_RATE_MULTIPLIER)
```
pete-watters: I think it would be nice to use a `const` for `feeRates.high.rate * sizeInfo.txVBytes * MAX_FEE_RATE_MULTIPLIER` 

This is already a bit crazy: `if( something and call function( do multiplication in args ) )`

packages/state/src/swap/strategies/protocol/protocol.ts
pete-watters: I think this needs to be fixed as it has a lint error?

pete-watters (approved):
Nice work @edgarkhanzadian 👍  I flagged a couple of things we could improve but overall its really nice work. 

Perhaps we could add a few more tests for peace of mind, maybe testing out that new skill on this branch to find gaps in our tests

314159265359879:
@edgarkhanzadian I have tested this version for several days and it looks fine, I am missing one thing.

✅ I have send the different kind of transactions we have fee ranges for: SIP10 sends, STX send, contract calls and contract deployments. I have not seen fees that were not within the expected ranges according to our profile. 

✅ I see that big transactions (byte size) get the proper minimum fee when their size is outside the range it will cap to the minimum instead. 

❌ I am missing the guard that should kick in when a fee is being set above 5 STX. I am wondering if we should even decrease that amount to something like 3 STX. This should trigger a dialogue to check if the user really intended to set such a high fee, and then they have to check a box to proceed. 



What I do see is that fee suggestions are generally higher than they need to be. I believe the fee estimation API still leaves room for improvement too. Because I often see the wallet reverting to the maximum of the default ranges we allow. When I lower it it still gets confirmed in seconds. I think we are protecting our users from overspending a lot already. We may be able to tweak the ranges further to do even more.

apps/extension/src/app/components/bitcoin-custom-fee/hooks/use-bitcoin-custom-fee.tsx
edgarkhanzadian: deprecated is probably a wrong word here🤔 i want this tagged as a bad implementation but refactor it later down the road. I'll just change this to TODO

apps/extension/src/app/components/bitcoin-fees-list/use-bitcoin-fees-list.ts
edgarkhanzadian: in this case, if feeRates is truthy, then feeRates.options is considered defined and not null

apps/extension/src/app/features/dialogs/transaction-action-dialog/hooks/use-btc-increase-fee.ts
edgarkhanzadian: good point!

PR #2135, opened 2026-02-19
feat(extension): trending tokens

edgarkhanzadian:
15 items:

https://github.com/user-attachments/assets/f3eaabfb-54d4-4374-b5c9-d92d8ff6a6f2

https://github.com/user-attachments/assets/152bc57e-0b2b-4f0d-8350-996d3e8e4029

5 items:

https://github.com/user-attachments/assets/a5518e59-3330-451c-afb1-85028284f0ae

https://github.com/user-attachments/assets/5ac9bd2c-38a6-4603-b3e9-be2b8ba1c666

NOTE:
Updated arrow icons later:

https://github.com/user-attachments/assets/ddde89ef-3a0a-4d8a-90cb-5c1d2d7043de

### DIFF
--- a/apps/extension/src/app/features/trending-tokens/trending-token-card.tsx
+++ b/apps/extension/src/app/features/trending-tokens/trending-token-card.tsx
@@ -1,0 +1,49 @@
+import { Flex, styled } from 'leather-styles/jsx';
+
+import { type TrendingToken, getPriceChangeColor } from '@leather.io/features';
+import { ArrowTriangleTopIcon, Sip10AvatarIcon } from '@leather.io/ui';
+
+interface TrendingTokenCardProps {
+  item: TrendingToken;
+}
+
+export function TrendingTokenCard({ item }: TrendingTokenCardProps) {
+  const { symbol, name, contractId, imageCanonicalUri } = item.asset;
+  const changePercent = item.marketStats.priceChange['1d'] ?? 0;
+  const color = getPriceChangeColor(changePercent);
+
+  return (
+    <Flex
+      alignItems="center"
+      borderRadius="round"
+      border="1px solid"
+      borderColor="ink.border-default"
+      px="space.03"
+      py="space.02"
+      gap="space.02"
+      whiteSpace="nowrap"
+    >
+      <Sip10AvatarIcon
+        size="md"
+        contractId={contractId}
+        imageCanonicalUri={imageCanonicalUri}
+        name={name}
+      />
+      <styled.span textStyle="label.03">{symbol}</styled.span>
+      {changePercent !== 0 && (
+        <ArrowTriangleTopIcon
+          color={color}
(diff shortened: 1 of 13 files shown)


edgarkhanzadian:
playwright tests failing, probably requires requests mocking

apps/extension/src/app/features/trending-tokens/trending-token-card.tsx
```diff
@@ -0,0 +1,49 @@
+import { Flex, styled } from 'leather-styles/jsx';
+
+import { type TrendingToken, getPriceChangeColor } from '@leather.io/features';
+import { ArrowTriangleTopIcon, Sip10AvatarIcon } from '@leather.io/ui';
+
+interface TrendingTokenCardProps {
```
pete-watters, line 6: Maybe we don't need an interface here for just one prop?

apps/extension/src/app/features/trending-tokens/trending-tokens.tsx
```diff
@@ -0,0 +1,99 @@
...
+  scrollbarWidth: 'none',
+  '&::-webkit-scrollbar': { display: 'none' },
+});
+
+export function TrendingTokens() {
+  const { data: trendingTokenData } = useTrendingTokensQuery();
+  const scrollRef = useRef<HTMLDivElement>(null);
+  const isLargeScreen = useViewportMinWidth('md');
+  const [canScrollLeft, setCanScrollLeft] = useState(false);
+  const [canScrollRight, setCanScrollRight] = useState(true);
```
pete-watters, line 27: This scrolling stuff seems a bit complicated and I wonder if it's best to check with the designers to see if we can do it a better way? 

Having stackable flex's seems like a more simple solution we could use that would look a bit better too

pete-watters (approved):
Nice work @edgarkhanzadian 👍  This code LGTM. I think the scrolling behaviour is a bit overcomplicated but I think thats part of the design?

I think you need to fix the large screen view as when I tested the tokens didn't fill the space available

Image

In the action popup some longer name ones force a scroll too which I think is just weird and my vote would be to stack them / have ellipsis on the name

Image

apps/extension/src/app/features/trending-tokens/trending-tokens.tsx
edgarkhanzadian: sure, we can discuss it during design office hours

apps/extension/src/app/features/trending-tokens/trending-token-card.tsx
edgarkhanzadian: imo it's ok, it is quite common in our codebase to do this

pete-watters: Yeah its no big deal at all 👍

PR #2141, opened 2026-02-20
feat(extension): integrate collectibles service and update UI 

pete-watters:
This PR updates the `extension` collectibles to:
- use the collectibles service
    - this includes use of Gamma metadata
    - exposure of new collectible types and formats (audio, video etc.)
    - this helps fix issues rendering previews of HTML collectibles
- deprecate legacy code
- add UI components for rendering of collectibles in the gallery view
- increase test coverage for collectibles

**NOTE**: 
- To keep the PR size down work on empty states has been separated - more to come in https://github.com/leather-io/mono/pull/2146
- For 'Send' ordinal I improvised and added a 'Send' option to the old dropdown. This will likely be deprecated soon once we have collectible details in 

**Demo**

**Populated wallet**

https://github.com/user-attachments/assets/b0c4713a-d0bb-477d-aa9b-de2bfa8d5d39

**Empty wallet**

https://github.com/user-attachments/assets/91f76c8a-cfd1-4a1a-8c27-b85675a98684

**Send Ordinal**

https://github.com/user-attachments/assets/dd1975a6-5c65-4a96-8724-d9abe4d64572


packages/utils/src/assets/inscription-helpers.ts
```diff
@@ -106,7 +106,6 @@ export function createInscriptionAsset(data: CreateInscriptionData): Inscription
       ...sharedInfo,
       mimeType: 'html',
       name: 'inscription',
-      thumbnailSrc: ordIoSrc,
```
pete-watters, line 109: This is important as we were over-writing the thumbnail with the source when it should fall back to an image preview instead of trying to render HTML on the extension. 

Using the thumbnail preview allows us to display HTML ordinals correctly e.g.: 

Image

apps/extension/src/app/features/collectibles/collectibles.tsx
```diff
@@ -1,45 +1,59 @@
...
-      onDiscardAllInscriptions={() => discardedInscriptions.discardAllInscriptions()}
-      onRecoverAllInscriptions={() => discardedInscriptions.recoverAllInscriptions()}
+      isLoading={isPending}
+      isError={isError}
+      amount={collectibles.length}
+      hasCollectibles={collectibles.length > 0}
+      onRefresh={() => void refetch()}
     >
-      <CurrentBitcoinSignerLoader>{() => <AddCollectible />}</CurrentBitcoinSignerLoader>
-      {isNftMetadataEnabled && (
```
pete-watters, line 30: I'm deprecating this in favour of the service and I don't think we need it. 

Similarly in the latest design we don't have an `<AddCollectible />` section anymore. I will check if this was missed but I think it's not really needed as it was just opening the receive modal and we have a new empty state now that helps first timers instead.

apps/extension/src/app/query/common/remote-config/remote-config.query.ts
```diff
@@ -147,11 +147,6 @@ export function useConfigFeeEstimationsMinValues() {
   return config.feeEstimationsMinMax.minValues.map(value => createMoney(value, 'STX'));
 }
 
-export function useConfigNftMetadataEnabled() {
```
pete-watters, line 150: I'm getting rid of this in favour of the service and to match mobile as I don't think it's necessary now.

fabric-8:
Good stuff, just some minor observerations:

Actions menu (list view): There's a rectangular white-bordered square visible behind the action button — this shouldn't be there.

CleanShot 2026-02-23 at 23 17 51@2x

---

Send collectible dialog: gap on the left and right edges of the dialog that should be removed. Otherwise the send dialog looks good.

CleanShot 2026-02-23 at 23 18 15@2x

---

Gallery view: Individual collectible items currently have slightly rounded corners. Only the outer container should have rounded corners — individual items should have square edges so they sit flush against each other.

CleanShot 2026-02-23 at 23 21 57@2x

pete-watters:
> Good stuff, just some minor observerations:
> 
> Actions menu (list view): There's a rectangular white-bordered square visible behind the action button — this shouldn't be there.
> 
> CleanShot 2026-02-23 at 23 17 51@2x
> Send collectible dialog: gap on the left and right edges of the dialog that should be removed. Otherwise the send dialog looks good.
> 
> CleanShot 2026-02-23 at 23 18 15@2x
> Gallery view: Individual collectible items currently have slightly rounded corners. Only the outer container should have rounded corners — individual items should have square edges so they sit flush against each other.
> 
> CleanShot 2026-02-23 at 23 21 57@2x

Thanks @fabric-8 👍  I have the next PR coming with [some UI improvements](https://github.com/leather-io/mono/pull/2146) so will fix those issues in there.

FYI that weird stuff with 'Send' or the three dot button is actually like that in production now:
Screenshot 2026-02-24 at 07 01 25
Screenshot 2026-02-24 at 07 01 16

I will fix it so it looks better and then likely we will remove that three dot button when we are adding collectible details

apps/extension/src/app/features/collectibles/components/collectible-card.web.tsx
```diff
@@ -0,0 +1,30 @@
+import { type ComponentProps, type ReactNode } from 'react';
+
+import { Box } from 'leather-styles/jsx';
+
+type BoxProps = ComponentProps<typeof Box>;
```
kyranjamie: `BoxProps` can be imported directly from leather-styles

apps/extension/src/app/features/collectibles/collectibles.tsx
```diff
@@ -1,45 +1,59 @@
...
+  switch (view.asset.protocol) {
+    case 'stamp':
+      return <StampCard item={view.asset} />;
+    case 'sip9':
+      return <Sip9Card item={view.asset} isBns={view.isBns} />;
+    case 'inscription':
+      return <InscriptionCardActions item={view.asset} />;
+    default:
+      return null;
+  }
```
kyranjamie: this should be a component

apps/extension/src/app/features/collectibles/components/collectible-image.web.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }, [src]);
+
+  if (!src || hasError) {
+    return <ImageUnavailable />;
+  }
+
+  if (onPress) {
+    return (
+      <CollectibleCard>
+        <styled.button
```
tigranpetrossian: This pattern is used a lot, it could probably moved into the `CollectibleCard` itself

packages/features/src/token/token-url.ts
```diff
@@ -0,0 +1,38 @@
...
+  const slashIndex = urlPath.indexOf('/');
+  if (slashIndex === -1) {
+    throw new Error(`Invalid asset URL path: ${urlPath}`);
+  }
+
+  const protocol = urlPath.slice(0, slashIndex);
+  const id = decodeURIComponent(urlPath.slice(slashIndex + 1));
+
+  return serializeAssetId({ protocol, id } as CryptoAssetId);
+}
```
kyranjamie, line 38: I hesitate with any code that bakes in logic like _if bitcoin this, if stacks that_. Yet another place in code we'l have to change if we support another chain (even if unlikely, I think this is a design consideration we should make).

How could this code be structured such that, we just have two separate functions, one for stacks one for bitcoin?

How about

```tsx
function BitcoinTokenFeature() {
  return <GenericTokenFeature context={{ urlPathToAssetId: bitcoinUrlPathToId }} />
}
function StacksTokenFeature() {
  return <GenericTokenFeature context={{ urlPathToAssetId: stacksUrlPathToId }} />
}
// future EthereumTokenFeature, configured in scoped place rather than deeply nested in files throughout app
```

in this example we avoid conditional switching, and feed in chain-specific logic in a single designated place

kyranjamie (approved):
Thanks Pete. Approved, but would love to see alternative approach to handling chain specific logic than if statements in many places

packages/features/src/token/token-url.ts
pete-watters: Thanks @kyranjamie 🙏 I will refactor this in a follow up

pete-watters: I looked into this and I think your suggestion is correct generally but in this case its just a simple URL mapping function to help us avoid having ugly URLs for BTC and STX. 

I will keep the design feedback in mind for future implementations but in this case I believe a simple early return here is better than a lot of context and prop drilling. If we do decide to add another chain we can refactor this and the other areas we have

PR #2145, opened 2026-02-23
feat(models,services): unify SwapDex and YieldProvider into StacksProtocol model, ref LEA-3457

alexp3y:
This PR:
- Introduces a unified `StacksProtocol` domain model replacing the separate `SwapDex` and `YieldProvider` types
- Adds `StacksProtocolService` for protocol lookup by ID/address and contract action resolution via the new `/v1/protocols` API
- Migrates all swap provider and yield provider services to use `StacksProtocol`


alexp3y:
Adopted naming migration here from PR https://github.com/leather-io/mono/pull/2136

pete-watters (approved):
Nice work 👍

PR #2146, opened 2026-02-23
feat(extension): collectibles empty states and UX improvements

pete-watters:
This PR builds on work on https://github.com/leather-io/mono/pull/2141 and adds some new UI for empty states for Token list and collectibles. 

It adds: 
- empty states for collectibles on `extension` + `mobile` 
- improved UX for loading states on `extension` - filling available height
- updated UX for `Send Inscription` applying @fabric-8's feedback [here](https://github.com/leather-io/mono/pull/2141#issuecomment-3949678063)
- Adds a tooltip beside 'Amount' in populated collectibles view 
- Adds 'Learn' section below 'Collectibles' + 'Tokens' on extension 
- Fixes bugs showing the loading state in a flash if there are actually no collectibles 
- Moves some files incorrectly called `.web.tsx` after migration from `packages/ui`

**DEMO - Extension empty Collectibles**

https://github.com/user-attachments/assets/921193e4-8d7b-49e5-8f0c-aab09a7170f9

**DEMO - Extension Collectibles**

https://github.com/user-attachments/assets/fdb13d70-d6f8-4cb7-a3c6-74f6688f1410

**DEMO - Mobile empty Collectibles**

Simulator Screenshot - iPhone 16 Pro - 2026-02-25 at 14 18 28


packages/constants/src/index.ts
```diff
@@ -19,7 +19,7 @@ export const ORD_IO_URL = 'https://ord.io';
 export const GAMMA_URL = 'https://gamma.io';
 export const GAMMA_API_URL: string = `${GAMMA_URL}/api`;
 export const MAGIC_EDEN_URL = 'https://magiceden.io/bitcoin';
-export const BNS_REGISTRATION_URL = 'https://www.bnsv2.com/';
+export const BNS_REGISTRATION_URL = 'https://bns.one/';
```
pete-watters, line 22: Updating the BNS_REGISTRATION_URL to match the design but there is ? there

Image

@fabric-8 @mica000 - can you confirm where you want to go for this?

camerow: I'm not sure which is better to use, who maintains each of those?

pete-watters: I'm not sure either 🤷  I'll ask Rena. Could be that https://btc.us/ is best

pete-watters: OK so I checked and [the consensus](https://trustmachines.slack.com/archives/C05LRS7G44R/p1771884609615749?thread_ts=1771856479.339919&cid=C05LRS7G44R) is https://bns.one/ is best

apps/mobile/src/assets/btc-domain.png
pete-watters, line 1: We are doubling up on these image assets in `extension` and `mobile` which is not great. 

I tried a few things but every single image in the mobile app uses local `require('@/assets/...')`. No images are   loaded from external packages. Metro's asset pipeline only works reliably with local requires.                                                                                   
                                               
We could try using our image hosting service but that would be a network call per image. For now I think this is good enough and I will accept some duplication

fabric-8:
CleanShot 2026-02-27 at 10 53 55

Let's use the same header style (formatting, spacing, look of icon button) in collectibles as in tokens

pete-watters:
> Let's use the same header style (formatting, spacing, look of icon button) in collectibles as in tokens

Thanks @fabric-8! I was too focused on the code I missed this obvious mistake 🫨  If you noticed anything else let me know!

I've fixed it now 
Screenshot 2026-02-27 at 10 43 51


https://github.com/user-attachments/assets/0e9a8de3-6f74-47ae-96de-924aff6fbb46

apps/extension/src/app/features/collectibles/components/collectibles-empty.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { openInNewTab } from '@app/common/utils/open-in-new-tab';
+import { useZeroIndexTaprootAddress } from '@app/store/accounts/blockchain/bitcoin/bitcoin.hooks';
+
+interface NftItem {
+  title: string;
+  caption: string;
+  image: string;
+  action: string;
+  onAction(): void;
+}
```
kyranjamie, line 19: Should this model be defined in the empty collectible component? We only use it here?

apps/extension/src/app/features/collectibles/components/collectibles-loading.tsx
```diff
@@ -26,14 +28,27 @@ function CollectibleCardSkeleton({ index }: { index: number }) {
...
+function getSkeletonCount(containerWidth: number) {
+  const columns = Math.max(2, Math.floor(containerWidth / 180));
+  const cardSize = containerWidth / columns;
+  const rows = Math.max(2, Math.ceil(window.innerHeight / cardSize));
+  return rows * columns;
+}
+
 export function CollectiblesLoading() {
+  const ref = useRef<HTMLDivElement>(null);
+  const width = ref.current?.clientWidth ?? window.innerWidth;
```
kyranjamie, line 40: does this need to use ResizeObserver? What if window size changes? Already have a hook of this I think

apps/extension/src/app/features/collectibles/components/collectibles-marketplaces.tsx
```diff
@@ -0,0 +1,72 @@
...
+          />
+
+          <Stack gap="space.01" flex="1" minWidth={0}>
+            <styled.span textStyle="label.01">{item.name}</styled.span>
+            <styled.span textStyle="caption.01" color="ink.text-subdued">
+              {item.description}
+            </styled.span>
+          </Stack>
+        </styled.button>
+      ))}
```
kyranjamie, line 69: could be a component in own right

apps/extension/src/app/pages/home/home.tsx
```diff
@@ -34,6 +37,9 @@ interface HomeProps {
 
 export function Home({ isBackground }: HomeProps) {
   const { activityRevamp } = useFlags();
+  const accountIndex = useCurrentAccountIndex();
+  const account = useAccountAddresses(accountIndex);
+  useAccountCollectibles(account);
```
kyranjamie, line 42: what does this do by itself? seems side-effectyu

apps/mobile/src/features/token/components/get-first-nft-section.tsx
```diff
@@ -26,44 +18,68 @@ export function GetFirstNftSection() {
...
-        caption={t`Bitcoin Blockchain`}
-        asideComponent={
+        </Cell.Aside>
+      </Cell.Root>
+      <Cell.Root pressable={false}>
+        <Cell.Icon borderRadius="round">
+          <Image
+            style={{ height: 40, width: 40 }}
+            contentFit="cover"
+            source={require('@/assets/ord-inscription.png')}
```
kyranjamie, line 66: Require is commonjs and we shouldn't use anywhere

apps/extension/src/app/pages/home/home.tsx
pete-watters: I just moved this call up the chain so we can run the query sooner and determine if users have collectibles or not. It helps stopping the page be so janky - showing a collectibles loading state if you actually don't have any and it makes the UI snappier and more responsive. 

I was thinking to do the same thing for activity also to pre-load some data there

apps/mobile/src/features/token/components/get-first-nft-section.tsx
pete-watters: We need to use `require()` in mobile or else the image loading doesn't work as we are using `expo-image`'s `<Image>`. We are using it for all other images in mobile. 

Metro bundler resolves `require()` at build time for images, providing the correct asset reference for each platform. You can't use a string path or an import — `require()` for static images is the documented Expo/RN pattern.

apps/extension/src/app/features/collectibles/components/collectibles-empty.tsx
pete-watters: Yes. It's only used in this file which is the component for the collectibles empty state 

Screenshot 2026-02-27 at 13 47 50

apps/extension/src/app/features/collectibles/components/collectibles-marketplaces.tsx
pete-watters: I kept it simple here as it's a static list of 2 items, rendered in one place. I think even using a map could be overkill here 
Screenshot 2026-02-27 at 13 50 06

Given the fact Magic Eden is [to shutter Bitcoin and EVM](https://blockspace.media/insight/magic-eden-to-shutter-bitcoin-and-evm-marketplaces-sunset-multi-chain-wallet/) I'd say we have spent enough time here already and could be removing this section soon anyway.

apps/extension/src/app/features/collectibles/components/collectibles-loading.tsx
pete-watters: I don't think thats necessary here. This loader shows for a brief period until collectibles load and as I am pre-loading them a level above in `home.tsx` it's even faster. 

`useElementHeightListener` exists but only tracks height. I can add a width observer if we want pixel perfect skeletons on resize if you think it's worth it? 

Based on our delivery priorities I would defer that to a new task for the upcoming cooldown

pete-watters:
Thanks for the review @kyranjamie 👍  I've answered all of your questions. At this point I don't think we need to change anything here

camerow (approved):

apps/extension/src/app/pages/home/home.tsx
kyranjamie: What about a loader component or something? `<PrefetchAccountCollectibleData>{children}</>`.

A hook loose like this is asking for someone to come along and think "not used, removing", as nothing really communicates the intent you describe in your comment here

kyranjamie (approved):

apps/extension/src/app/pages/home/home.tsx
pete-watters: Thanks 👍 I didn't consider that. I will use that approach for working on activity and can update this call then while I am at it. I [made this task](https://linear.app/leather-io/issue/LEA-3484/improve-pre-loading-of-activity-collectibles ) so I don't forget

PR #2152, opened 2026-02-26
Add Swap form page layout and the asset selector

tigranpetrossian:
* Wire up the swap state in the extension
* Add basic swap form layout (with some placeholders)
* Add asset selector

https://github.com/user-attachments/assets/9ed36d00-7284-404a-9fa1-c3694e39861d

### DIFF
--- a/apps/extension/src/app/pages/swap/components/amount-field.tsx
+++ b/apps/extension/src/app/pages/swap/components/amount-field.tsx
@@ -1,0 +1,35 @@
+import { Flex, styled } from 'leather-styles/jsx';
+
+import { InputCurrencyMode, SwappableFungibleCryptoAsset } from '@leather.io/models';
+import { SecondaryAmount } from '@leather.io/state/swap';
+import { ArrowTopBottomIcon } from '@leather.io/ui';
+
+interface AmountFieldProps {
+  asset: SwappableFungibleCryptoAsset | undefined;
+  value: string;
+  secondaryAmount: SecondaryAmount;
+  inputCurrencyMode: InputCurrencyMode;
+}
+
+export function AmountField({ value }: AmountFieldProps) {
+  return (
+    <Flex direction="column" gap="space.03">
+      <styled.input
+        caretColor="red" // TODO: Temp
+        textStyle="heading.03"
+        fontSize={26}
+        lineHeight="32px"
+        value={value}
+        outline="none"
+        autoFocus
+      />
+
+      <Flex gap="space.01" alignItems="center">
+        <styled.div textStyle="label.03" color="ink.text-subdued">
+          $2867.74
+        </styled.div>
+        <ArrowTopBottomIcon variant="small" color="ink.text-subdued" />
+      </Flex>
+    </Flex>
+  );
+}
--- a/apps/extension/src/app/pages/swap/hooks/use-swap-dependencies.ts
+++ b/apps/extension/src/app/pages/swap/hooks/use-swap-dependencies.ts
@@ -1,0 +1,165 @@
+import { useMemo } from 'react';
+
+import { type StacksNetwork } from '@stacks/network';
+import {
+  type StacksTransactionWire,
+  type TxBroadcastResultOk,
+  broadcastTransaction,
+} from '@stacks/transactions';
+import { SbtcApiClientMainnet, SbtcApiClientTestnet } from 'sbtc';
+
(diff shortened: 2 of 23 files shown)


apps/extension/src/app/queries/match-query-result.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { assertUnreachable } from '@leather.io/utils';
+
+interface Matchers<TData, TError> {
+  pending(): ReactElement;
+  error(error: TError): ReactElement;
+  success(data: TData): ReactElement;
+}
+
+export function matchQueryResult<TData, TError>(
```
pete-watters, line 13: This is cool! 😎

apps/extension/src/app/pages/swap/components/amount-field.tsx
```diff
@@ -0,0 +1,35 @@
...
+  value: string;
+  secondaryAmount: SecondaryAmount;
+  inputCurrencyMode: InputCurrencyMode;
+}
+
+export function AmountField({ value }: AmountFieldProps) {
+  return (
+    <Flex direction="column" gap="space.03">
+      <styled.input
+        caretColor="red" // TODO: Temp
```
pete-watters, line 18: Can we do this now? Not really a big deal though

apps/extension/src/app/pages/swap/components/amount-field.tsx
```diff
@@ -0,0 +1,35 @@
...
+        fontSize={26}
+        lineHeight="32px"
+        value={value}
+        outline="none"
+        autoFocus
+      />
+
+      <Flex gap="space.01" alignItems="center">
+        <styled.div textStyle="label.03" color="ink.text-subdued">
+          $2867.74
```
pete-watters, line 29: This `$2867.74` shouldn't be hardcoded?

apps/extension/src/app/pages/swap/hooks/use-swap-dependencies.ts
```diff
@@ -0,0 +1,165 @@
...
+import { useSignBitcoinTx } from '@app/store/accounts/blockchain/bitcoin/bitcoin.hooks';
+import { useCurrentAccountNativeSegwitIndexZeroSignerNullable } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useWalletFingerprint } from '@app/store/in-memory-key/in-memory-key.hooks';
+import { useCurrentStacksNetworkState } from '@app/store/networks/networks.hooks';
+import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
+import { useSignStacksTransaction } from '@app/store/transactions/transaction.hooks';
+
+// This file is mostly vibed to be able to plug useSwapState and move forward with the UI work
+// TODO: Needs a careful rewrite.
```
pete-watters, line 38: Is it worth logging this in linear?

apps/extension/src/app/pages/swap/swap-container.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useSwapDependencies } from './hooks/use-swap-dependencies';
+
+export function SwapContainer({ children }: HasChildren) {
+  const dependencies = useSwapDependencies();
+
+  return (
+    <SwapProvider
+      baseAsset={stxAsset}
+      dependencies={dependencies}
+      quoteCurrencyPreference="USD"
```
pete-watters, line 17: Maybe `quoteCurrencyPreference` should be dynamic? Then it can default to `'USD'` in `renderUseSwapState`?

pete-watters (approved):
Great work here @tigranpetrossian 👍  I flagged a couple of minor things but happy to merge this and have them fixed later as its flagged anyway. 

For `getFungibleAssetDisplayName` and `getAssetIndicator` I thought we have some similar code elsewhere but couldn't find it. Maybe they could live somewhere shared like `features/token` as they are re-usable. `getAssetIndicator` could even be a UI component / baked into avatar. 

As they are only used in Swap anyway thats not really relevant to this PR but just mentioning

tigranpetrossian:
Thanks @pete-watters, I can do a pass afterwards and see if any more helpers qualify to be lifted somewhere. 
I don't think feature-specific area like `features/token` signals reusability enough for this kind of thing, those can be domain specific utilities perhaps.

PR #2153, opened 2026-02-26
feat(extension): add swap amount field

tigranpetrossian:
This adds the amount of field to extension's new Swap, building on top of the headless hook I added earlier in https://github.com/leather-io/mono/pull/2114

https://github.com/user-attachments/assets/9ede2a25-161c-49bd-bca4-57342517d578


apps/extension/src/app/pages/swap/components/amount-field/amount-field-colors.ts
```diff
@@ -0,0 +1,46 @@
+import { ColorToken, token } from 'leather-styles/tokens';
+
+import { cryptoAssetColors } from '@leather.io/constants';
+import { SwappableFungibleCryptoAsset } from '@leather.io/models';
+
+type Variant = 'initial' | 'active' | 'invalid';
+
+const defaultCaretColor = '#2cb5c1';
```
pete-watters, line 8: Do we have a design token for this?

pete-watters (approved):
This looks great, nice work! Hopefully we can use this in the send form soon too. 

Maybe the amount field component could live at base level - `components` instead

apps/extension/src/app/pages/swap/components/amount-field/amount-field-colors.ts
tigranpetrossian: Actually we might, good catch. I ported this from mobile where it's not trivial to retrieve theme values outside react. Will take a look.

tigranpetrossian: Turns out this is a custom color roughly replicating the blue caret color. We didn't have anything of quite a matching contrast because our raw color ramps aren't in the tokens package for some reason. No big deal either way I guess.

tigranpetrossian:
> This looks great, nice work! Hopefully we can use this in the send form soon too.
> 
> Maybe the amount field component could live at base level - `components` instead

Thanks @pete-watters, this will definitely end up in the root level components at some point, but there's not much point in doing it now without knowing just how much of it needs to abstracted.

PR #2159, opened 2026-02-26
fix(extension): update storybook version (audit dependencies issue)

edgarkhanzadian:


pete-watters (approved):

pete-watters:
Thanks @edgarkhanzadian ! I think this is the most work we do with storybook now 😄

PR #2167, opened 2026-03-02
fix(extension): show taproot send data in activity

pete-watters:
This PR attempts to fix the problem we have reported in LEA-3481 where taproot activity data is not showing in the activity list.

### DIFF
--- a/apps/extension/src/app/common/transactions/bitcoin/utils.ts
+++ b/apps/extension/src/app/common/transactions/bitcoin/utils.ts
@@ -43,29 +43,31 @@ export function getBitcoinTxCaption(transaction?: BitcoinTx) {
 // If vin array contains a prevout with a scriptpubkey_address equal to
 // the address, then that is the current address making a `Sent` tx (-)
 // and the value of the prevout is the tx amount
-function transactionsSentByAddress(address: string, transaction: BitcoinTx) {
-  return transaction.vin.filter(input => input.prevout.scriptpubkey_address === address);
+function transactionsSentByAddresses(addresses: string[], transaction: BitcoinTx) {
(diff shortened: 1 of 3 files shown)


apps/extension/src/app/common/transactions/bitcoin/utils.ts
```diff
@@ -43,29 +43,31 @@ export function getBitcoinTxCaption(transaction?: BitcoinTx) {
...
-  return transaction.vout.filter(output => output.scriptpubkey_address === address);
+function transactionsReceivedByAddresses(addresses: string[], transaction: BitcoinTx) {
+  return transaction.vout.filter(output => addresses.includes(output.scriptpubkey_address));
 }
 
-export function isBitcoinTxInbound(address: string, transaction: BitcoinTx) {
-  const inputs = transactionsSentByAddress(address, transaction);
-  const outputs = transactionsReceivedByAddress(address, transaction);
+export function isBitcoinTxInbound(addresses: string | string[], transaction: BitcoinTx) {
+  const addressList = Array.isArray(addresses) ? addresses : [addresses];
```
kyranjamie, line 57: can use `ensureArray` here

pete-watters:
@edgarkhanzadian suggested that by looking at the PR code, the fix only covers 0 index taproot addresses, to fix this one properly we’d need to check:

- `m/86'/0'/X’/Y/Z`
- `m/84'/0'/X’/Y/Z`

Where X is the current account index and Y is change index that can be 0 and 1, and Z being address index that we might want to go quite high with. Technically Z index can be a very large number depending on how person used the wallet.

As he implemented this he has more context than me and will take over here.

pete-watters:
Closing this to reduce noise in the PR queue

PR #2170, opened 2026-03-03
fix(extension): activity txs [LEA-3481]

edgarkhanzadian:
Use both taproot and native segwit 0 index addresses to check bitcoin tx for user-related inputs and outputs.

Also update utils to receive a `checkAddress` function instead of an address. We can later use `checkAddress` to match tx input/output addresses to user account's derived set of addresses

### DIFF
--- a/apps/extension/src/app/common/transactions/bitcoin/utils.ts
+++ b/apps/extension/src/app/common/transactions/bitcoin/utils.ts
@@ -40,43 +40,54 @@ export function getBitcoinTxCaption(transaction?: BitcoinTx) {
   return transaction ? truncateMiddle(transaction.txid, 4) : '';
 }
 
+type IsCorrespondingAddressFn = (address: string) => boolean;
+
 // If vin array contains a prevout with a scriptpubkey_address equal to
 // the address, then that is the current address making a `Sent` tx (-)
 // and the value of the prevout is the tx amount
-function transactionsSentByAddress(address: string, transaction: BitcoinTx) {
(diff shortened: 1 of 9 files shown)


apps/extension/src/app/common/transactions/bitcoin/utils.ts
```diff
@@ -40,43 +40,40 @@ export function getBitcoinTxCaption(transaction?: BitcoinTx) {
...
+  const inputs = transactionsSentByAddress(checkAddress, transaction);
+  const outputs = transactionsReceivedByAddress(checkAddress, transaction);
 
   if (inputs.length && outputs.length) return false;
   if (inputs.length) return false;
   return true;
 }
 
-export function getBitcoinTxValue(address: string, transaction?: BitcoinTx) {
+export function getBitcoinTxValue(checkAddress: CheckAddressFn, transaction?: BitcoinTx) {
```
kyranjamie: `checkAddress` isn't super clear imo. `isCorrespondingAddressFn`, `isAddressInTxFn` or something?

kyranjamie (approved):

pete-watters (approved):
Thanks @edgarkhanzadian

apps/extension/src/app/common/transactions/bitcoin/utils.ts
edgarkhanzadian: good point! I'll update it

edgarkhanzadian:
also added e2e tests here

PR #2172, opened 2026-03-03
feat: add token name override capability dawg ref LEA-3483

pete-watters:
This PR comes from token owner via a request from @314159265359879 to display a different name for a particular token. I've added a new overrides const so we can easily do this for other tokens as needed. 

The contract address is `SPQYMRAKZPQPJAADX5JBEFT0FHE3RZZK9F8TYBQ3.dawgpool-stxcity` but they want their token name to display as `Dawgcoin`

image (4)


packages/query/src/stacks/sip10/sip10-tokens.utils.ts
```diff
@@ -15,7 +16,8 @@ export function isTransferableSip10Token(token: Partial<FtMetadataResponse>) {
 
 export function createSip10Asset(key: string, ftAsset: FtMetadataResponse): Sip10Asset {
   const { contractAssetName } = getStacksAssetStringParts(key);
-  const name = ftAsset.name || contractAssetName;
+  const contractPrincipal = key.split('::')[0];
+  const name = SIP10_TOKEN_NAME_OVERRIDES[contractPrincipal] ?? (ftAsset.name || contractAssetName);
```
kyranjamie: Rather than doing the look up of the object key directly, can we expose a function? 

e.g.?
```
getSip10TokenNameWithOverrides(contractPrincipal, ftAsset.name ?? contractAssetName)
```

kyranjamie (approved):

packages/query/src/stacks/sip10/sip10-tokens.utils.ts
pete-watters: Thanks @kyranjamie 👍  I've updated this and also improved the quality by adding tests etc. 

I'm going to assume the code is GTG and if not ask for forgiveness after rather than permission now.

Let me know if you want me to change anything else

PR #2178, opened 2026-03-04
fix: add tr related unit tests + fix bug in max send calculation

edgarkhanzadian:
Add more unit tests for taproot changes

### DIFF
--- a/packages/bitcoin/src/coin-selection/coin-selection.utils.ts
+++ b/packages/bitcoin/src/coin-selection/coin-selection.utils.ts
@@ -95,12 +95,14 @@ export function getSpendableAmount<T extends InputData>({
   utxos,
   feeRate,
   recipients,
+  isSendMax,
 }: GetSpendableAmountArgs<T>) {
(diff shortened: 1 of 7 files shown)


packages/bitcoin/src/coin-selection/coin-selection.utils.ts
```diff
@@ -95,12 +95,14 @@ export function getSpendableAmount<T extends InputData>({
...
   feeRate,
   recipients,
+  isSendMax,
 }: GetSpendableAmountArgs<T>) {
   const balance = utxos.map(utxo => utxo.value).reduce((prevVal, curVal) => prevVal + curVal, 0);
 
   const size = getSizeInfo({
     utxos,
     recipients,
+    isSendMax,
```
edgarkhanzadian, line 105: This is one of the bugs caught during running unit tests. When calling getSpendableAmount with isSendMax, we didn't pass that value to getSizeInfo

pete-watters (approved):

PR #2179, opened 2026-03-04
chore: add e2e tests for extension (1/n)

edgarkhanzadian:
Implement e2e tests for signing psbt with mixed utxos, sends with mixed utxos and max sends with mixed utxos

This PR is the first of few for creating e2e tests in extension

### DIFF
--- a/apps/extension/tests/mocks/mock-utxos.ts
+++ b/apps/extension/tests/mocks/mock-utxos.ts
@@ -188,3 +188,24 @@ export async function mockUtxoRequestsWithInscriptions(page: Page, utxos: OwnedU
     return route.fulfill({ json: [] });
   });
 }
+
+interface LeatherUtxoMock {
+  txid: string;
+  vout: number;
+  value: string;
+  height: number;
+  address: string;
+  path: string;
+}
+
+export async function mockMixedUtxoRequests(page: Page, utxos: LeatherUtxoMock[]) {
+  await page.route('**/v1/utxos/**', route => {
+    const url = route.request().url();
+    const trUtxos = utxos.filter(u => u.path.startsWith("m/86'"));
+    const nsUtxos = utxos.filter(u => u.path.startsWith("m/84'"));
+    if (url.includes(encodeURIComponent('tr('))) {
+      return route.fulfill({ json: trUtxos });
+    }
+    return route.fulfill({ json: nsUtxos });
+  });
+}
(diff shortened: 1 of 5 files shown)


apps/extension/tests/mocks/mock-utxos.ts
```diff
@@ -188,3 +188,24 @@ export async function mockUtxoRequestsWithInscriptions(page: Page, utxos: OwnedU
...
+  address: string;
+  path: string;
+}
+
+export async function mockMixedUtxoRequests(page: Page, utxos: LeatherUtxoMock[]) {
+  await page.route('**/v1/utxos/**', route => {
+    const url = route.request().url();
+    const trUtxos = utxos.filter(u => u.path.startsWith("m/86'"));
+    const nsUtxos = utxos.filter(u => u.path.startsWith("m/84'"));
+    if (url.includes('tr%28') || url.includes('tr(')) {
```
kyranjamie: should you use encodeURIComponent here?

apps/extension/tests/specs/rpc-sign-psbt/sign-psbt.spec.ts
```diff
@@ -277,4 +293,241 @@ test.describe('Sign PSBT', () => {
...
+    function createTaprootTestPsbt() {
+      const psbt = new btc.Transaction();
+      psbt.addInput({
+        txid: '3f6c95b9895cd22943c8f8bcac899c10997b99651f0a7d9bd8c0c6f3536cbbc7',
+        index: 2,
+        witnessUtxo: {
+          amount: 5000n,
+          script: taprootPayment.script,
+        },
+        tapInternalKey: schnorrPubKey,
```
kyranjamie: shouldn't this be `taprootPayment.tapInternalKey`?

edgarkhanzadian: from what i understand

```
    const schnorrPubKey = ecdsaPublicKeyToSchnorr(taprootKeychain.publicKey);
    const taprootPayment = btc.p2tr(schnorrPubKey, undefined, bitcoinTestnet);
```

In this case `taprootPayment.tapInternalKey` returns the same `schnorrPubKey`. Technically this should be ok as we don't particularly need to test p2tr function

apps/extension/tests/mocks/mock-utxos.ts
edgarkhanzadian: definitely, updating that part now

apps/extension/tests/specs/rpc-sign-psbt/sign-psbt.spec.ts
edgarkhanzadian: but probably would be better to use tapprotPayment just for the consistency here

kyranjamie (approved):

PR #2183, opened 2026-03-04
chore: add collectibles + token analytics ref LEA-3095 

pete-watters:
This PR adds some improved analytics to areas we are lacking.

It also adds improvements to naming suggested here https://github.com/leather-io/mono/pull/1999/changes#r2693454534

### DIFF
--- a/apps/extension/src/app/common/app-analytics.ts
+++ b/apps/extension/src/app/common/app-analytics.ts
@@ -1,9 +1,14 @@
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 
 import { z } from 'zod';
 
 import { makeAccountIdentifer } from '@leather.io/crypto';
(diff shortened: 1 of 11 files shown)


packages/analytics/src/events.ts
```diff
@@ -124,10 +130,6 @@ interface HistoricalEvents {
   create_new_account: undefined;
   change_network: { id: string };
   click_open_in_new_tab_menu_item: undefined;
-  click_change_network_menu_item: undefined;
```
pete-watters, line 127: @camerow is it OK to deprecate these?

apps/extension/src/app/common/app-analytics.ts
```diff
@@ -143,3 +153,94 @@ export function useAccountScaledBalanceAnalytics({ accountIndex }: { accountInde
     scaledUsdBalance,
   ]);
 }
+
```
pete-watters, line 156: Adding these here but perhaps we should move these hooks to the analytics package for better cross platform sharing?

PR #2184, opened 2026-03-04
fix: taproot rbf

edgarkhanzadian:
increasing fee on txs with taproot inputs works correctly.

https://github.com/user-attachments/assets/0b541f01-5512-48da-ad78-53f1fe5c2033

NOTE:
Found another bug though:

1. Broadcast transaction
2. bump fee from 1sat/vbyte to 2sat/vbyte
3. open the increase fee dialog again, it defaults to 1sat/vbyte for some reason.

It is unrelated to taproot changes but something that needs to be fixed later on

### DIFF
--- a/apps/extension/src/app/common/transactions/bitcoin/utils.ts
+++ b/apps/extension/src/app/common/transactions/bitcoin/utils.ts
@@ -4,9 +4,6 @@ import { BtcSizeFeeEstimator } from '@leather.io/bitcoin';
 import type { BitcoinTransactionVectorOutput, BitcoinTx } from '@leather.io/models';
 import { satToBtc, sumNumbers, truncateMiddle } from '@leather.io/utils';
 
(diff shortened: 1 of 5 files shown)


pete-watters (approved):
This LGTM nice work. 

In the video though I see another visual bug - the background page behind the dialog is blank. This is another instance where we have a routable dialog and we need to set the `backgroundLocation` page to render content behind it properly. 

This isn't a release blocker but if we can fix it that would be great. We can consider this pattern of routable dialogs overall though as I think it offers little value apart from adding complexity apart from in Ledger flows where it is useful but in those cases it doesn't have this background bug

kyranjamie (approved):

kyranjamie:
I encourage it to move towards the pattern of react-call for modals, that can be called practically anywhere and have no routes. 

No suggesting we do this here, but think this is a good example

PR #2186, opened 2026-03-04
fix: rpc utxo selection + use useCurrentUtxos instead of outdated useCurrentNativeSegwitUtxos

edgarkhanzadian:
- Replaces useCurrentNativeSegwitUtxos with useCurrentUtxos across the extension so UTXO selection includes all address types (native segwit + taproot), not just native segwit
- Adds E2E tests for sendTransfer with taproot-only UTXOs

### DIFF
--- a/apps/extension/src/app/pages/swap-legacy/containers/bitcoin-swap-container.tsx
+++ b/apps/extension/src/app/pages/swap-legacy/containers/bitcoin-swap-container.tsx
@@ -1,5 +1,5 @@
 import { BitcoinNativeSegwitAccountLoader } from '@app/components/loaders/bitcoin-account-loader';
-import { BitcoinUtxosLoader } from '@app/components/loaders/bitcoin-utxos-loader';
+import { BitcoinNativeSegwitUtxosLoader } from '@app/components/loaders/bitcoin-utxos-loader';
 
 import { BitcoinSwapProvider } from '../providers/bitcoin-swap-provider';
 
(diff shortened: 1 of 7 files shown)


apps/extension/src/app/pages/rpc-send-transfer/rpc-send-transfer-container.tsx
```diff
@@ -14,34 +14,31 @@ export function RpcSendTransferContainer() {
...
+  const { utxos } = useCurrentUtxos();
 
   const navigate = useNavigate();
 
   const { recipients, amount } = sendTransferState;
 
   return (
-    <BitcoinUtxosLoader>
-      {utxos => (
-        <BitcoinFeeEditorProvider
```
kyranjamie, line 25: Why loader removed? what happens if utxos are still loading?

edgarkhanzadian: ah good catch, updating

apps/extension/src/app/pages/swap-legacy/containers/bitcoin-swap-container.tsx
```diff
@@ -1,17 +1,17 @@
...
   return (
     <BitcoinNativeSegwitAccountLoader current>
       {signer => (
-        <BitcoinUtxosLoader>
+        <BitcoinNativeSegwitUtxosLoader>
           {utxos => {
             return <BitcoinSwapProvider signer={signer} utxos={utxos} />;
           }}
-        </BitcoinUtxosLoader>
+        </BitcoinNativeSegwitUtxosLoader>
```
edgarkhanzadian, line 14: i'm a little unsure here if we need to keep native segwit only or we need to migrate this to all utxos @kyranjamie

pete-watters (approved):
Nice work @edgarkhanzadian  👍  I'm wondering - should we be also making these changes to support this on mobile while we are at it?

edgarkhanzadian:
@pete-watters mobile should already support taproot utxos fully, but we can run one more round of checks on it

PR #2188, opened 2026-03-05
feat(extension): add swap review summary section

tigranpetrossian:
Adds summary section to the swap review, and the details building blocks for implementing detail rows. 
Adding only couple basic rows in this PR, the more complex ones like the slippage editor coming in the next PR.

The account row is pretty compared to mobile, but will be improved once multi-wallet & icon support is in the extension.

https://github.com/user-attachments/assets/3d6a8ce0-376f-44c4-b198-51afc2f3da1e

### DIFF
--- a/apps/extension/src/app/pages/swap/components/review/swap-review-details.tsx
+++ b/apps/extension/src/app/pages/swap/components/review/swap-review-details.tsx
@@ -1,0 +1,86 @@
+import { ReactNode } from 'react';
+
+import { AnimatePresence, motion } from 'framer-motion';
+import { Flex, Stack, styled } from 'leather-styles/jsx';
+import { isString } from 'remeda';
+
+import { Button, Hr, HrProps, SettingsGearIcon } from '@leather.io/ui';
+
+import { HasChildren } from '@app/common/has-children';
+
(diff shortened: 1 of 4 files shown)


edgarkhanzadian (approved):

apps/extension/src/app/pages/swap/components/review/swap-review-details.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+function renderTextOrNode(node: ReactNode, color?: string) {
+  return isString(node) ? (
+    <styled.span textStyle="label.02" color={color}>
+      {node}
+    </styled.span>
+  ) : (
+    node
+  );
+}
```
kyranjamie, line 86: this is a component, no?

```tsx
export function TextOrNode({ children, color }: TextOrNodeProps) {
  if (isString(children)) {
    return (
      <styled.span textStyle="label.02" color={color}>
        {children}
      </styled.span>
    );
  }

  return <>{children}</>;
}
```

tigranpetrossian: Agreed. I hit this pattern very often, especially in RN where `<Text>` is required for strings. Replaced for now, but it's worth implementing a slightly more customizable `ui` version of this at some point.

PR #2189, opened 2026-03-05
feat(services): activity v2, closes LEA-3464

alexp3y:
This PR introduces a new activity data model `BlockchainActivity`, built from the ground up as a pure, normalized representation of on-chain activity.

### Purpose
The existing `Activity` model (detailed in LEA-3464) attempted to model a complex, hierarchical catalogue of activity items based on user-facing categories. This pushes presentation concerns into the data layer, obfuscating the information carried in activity list output.

### `BlockchainActivity` Details
The model is now much more flexible as a flat, chain-agnostic representation of on-chain transactions. Each `BlockchainActivity` item carries an `events[]` array of relevant transfers / actions, contract metadata, protocol identification and normalized action classification, and core Tx data like fee, blockHeight, status, chain.

The presentation layer now becomes responsible for further user-facing categorization and data views.

It is scoped as "blockchain" activity as the model is solely concerned with representing on-chain data. Off-chain activity (app connections, wallet events, etc.) can be introduced separately in the future.

### Migration
`BlockchainActivity` will coexists initially with the original Activity. New feature work (e.g. extension activity redesign) should use `BlockchainActivityService`. Mobile can follow, after which the original `Activity` model and service can be safely removed.

### Changes
Model: `BlockchainActivity`, `BlockchainActivityEvent` - (@leather.io/models)
Service: `BlockchainActivityService` with paginated request/response, filtering (asset/protocol/chain) -(@leather.io/services)
Queries: `createBlockchainActivityQueryConfig` for React Query integration (@leather.io/queries)
Performance: Hiro rate limit increase, removal of contract args from Stacks API requests
Scripts: New pnpm test:activity CLI script, asset-list script cleanup, unified scripts/README.md


packages/services/src/activity/stacks-activity-event.utils.ts
```diff
@@ -0,0 +1,191 @@
...
+
+function extractStxAssetEvent(
+  event: TransactionEventStxAsset,
+  stxAddress: string,
+  events: StacksRawActivityEvent[]
+) {
+  const { asset } = event;
+  if (!asset.asset_event_type || !asset.amount) return;
+
+  if (asset.asset_event_type === 'transfer') {
```
pete-watters, line 103: Maybe a `switch(true)` would be easier to read here than the `else if`s? Then you can assert conditions in an easier to read way

packages/services/src/activity/bitcoin-blockchain-activity.utils.ts
```diff
@@ -0,0 +1,83 @@
+import BigNumber from 'bignumber.js';
+
```
pete-watters, line 2: It would be great to smother these utils with some more unit tests if we can, more as documentation of expected data in and data out.  That will help us for the FE E2E test mocks too

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+        action: raw.action,
+        asset: sip10Asset,
+        counterparty: raw.counterparty,
+        amount: {
+          crypto: createMoney(initBigNumber(raw.rawAmount), sip10Asset.symbol, sip10Asset.decimals),
+          quote: createMoney(0, 'USD'),
+        },
+      };
+    } catch {
+      return undefined;
```
pete-watters, line 249: Maybe we should log this somehow or do something here like returning a typed result?

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+    } else if (request.filter?.protocol) {
+      allActivity = allActivity.filter(
+        a => a.contract?.type === 'call' && a.contract.protocol === request.filter!.protocol
+      );
+    } else if (request.filter?.chain) {
+      allActivity = allActivity.filter(a => a.chain === request.filter!.chain);
+    }
+
+    const total = allActivity.length;
+    const pagination = request.pagination;
```
pete-watters, line 75: Nice work with the pagination 🚀

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+      default:
+        return [];
+    }
+  }
+
+  private async fetchAllActivity(
+    account: AccountAddresses,
+    signal?: AbortSignal
+  ): Promise<BlockchainActivity[]> {
+    const [btcActivity, stacksActivity] = await Promise.all([
```
pete-watters, line 96: Does stacksActivity include sBTC data? Also do you know if we include any activity about NFTs?

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+    private readonly sip10AssetService: Sip10AssetService,
+    private readonly sip9AssetService: Sip9AssetService,
+    private readonly stacksProtocolService: StacksProtocolService
+  ) {}
+
+  public async getActivity(
+    request: ActivityRequest,
+    signal?: AbortSignal
+  ): Promise<ActivityResponse> {
+    let allActivity = await this.fetchAllActivity(request.account, signal);
```
pete-watters, line 62: Do we need to do things this way - in memory filtering after fetching all activity? I know it's complicated so I am assuming so. 

Doing it like this means we fetch everything and then if the filter is chain specific we discard a lot of it immediately. 

> When request.filter.chain === 'bitcoin', all Stacks transactions are still fetched and discarded. The filter is applied after fetchAllActivity returns. For chain and asset filters the early-exit could happen before hitting the API. Given the Stacks API fetches up to 20 pages (stopAfter: 20) this is a meaningful wasted cost.
> **Suggestion**: pass the filter into fetchAllActivity and skip getStacksActivity / getBtcActivity accordingly, or at least short-circuit when filter.chain is set.

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+    account: AccountAddresses,
+    signal?: AbortSignal
+  ): Promise<BlockchainActivity[]> {
+    if (!hasStacksAddress(account)) return [];
+
+    const [pendingTxs, txs, txEvents] = await Promise.all([
+      this.stacksTransactionsService.getPendingTransactions(account.stacks.stxAddress, signal),
+      this.hiroStacksApiClient.getAddressTransactions(
+        account.stacks.stxAddress,
+        { allPages: true, stopAfter: 20 },
```
pete-watters, line 129: Do we want to fetch all pages up front even if we have pagination? 🤔

packages/services/scripts/activity-test.mjs
```diff
@@ -0,0 +1,401 @@
+import 'reflect-metadata';
```
pete-watters, line 1: This reminds me of a script we had at BAML called 'do the hustle' that was doing a lot of things. Maybe it could be split up a bit or have some comments. What are the functions like - `formatAmount`? Are they ones we have elsewhere in the repo or throwaway test code.

packages/services/scripts/activity-test.mjs
```diff
@@ -0,0 +1,401 @@
...
+    const recipient = asset.recipient;
+    if (sender === stxAddress || recipient === stxAddress || asset.locked_address === stxAddress) {
+      matched++;
+    }
+  }
+  return { total, matched };
+}
+
+async function fetchHiroAllPages(url, limit, stopAfter) {
+  const axios = (await import('axios')).default;
```
pete-watters, line 141: Maybe we can change this to just use `fetch` or else add `axios` as a dev dependancy to the services package? This works if axios is already a transitive dep, but it's brittle.

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+      case 'nativeBtc':
+        return activity.filter(a => a.chain === 'bitcoin');
+      case 'nativeStx':
+        return activity.filter(a => a.events.some(e => e.asset.protocol === 'nativeStx'));
+      case 'sip10':
+        return activity.filter(a =>
+          a.events.some(e => e.asset.protocol === 'sip10' && e.asset.assetId === asset.assetId)
+        );
+      default:
+        return [];
```
pete-watters: Could we handle this better rather than returning empty for undefined protocols? Maybe we can log something or return a type  even just have a comment. 

Returning `[]` could look like "no transactions" when it's actually "unsupported filter type".

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -0,0 +1,385 @@
...
+          this.resolveProtocolInfo(
+            tx.contract_call.contract_id,
+            tx.contract_call.function_name,
+            signal
+          ),
+        ]);
+        return mapStacksContractCall(tx, account, events, protocolInfo);
+      }
+      default:
+        return;
```
pete-watters, line 174: This returns `undefined` for unrecognised `tx_type` values (e.g. _coinbase_, _tenure_change_). 

Maybe this is intentional but should be documented. These tx types will silently disappear from the activity list.

Maybe we could have a test / declare all types

packages/services/scripts/activity-test.mjs
```diff
@@ -0,0 +1,401 @@
...
+
+function loadEnv() {
+  try {
+    const envPath = resolve(new URL('.', import.meta.url).pathname, '..', '.env');
+    const content = readFileSync(envPath, 'utf-8');
+    for (const line of content.split('\n')) {
+      const match = line.match(/^(\w+)=['"](.*)['"]\s*$/);
+      if (match) process.env[match[1]] ??= match[2];
+    }
+  } catch {}
```
pete-watters, line 18: Maybe we can add a `console.warn` here in case `.env` resolution fails so it will help us debug later

pete-watters (approved):
Nice work here @alexp3y  👍  I mentioned a few things to improve but as we plan to integrate this to the new `extension` activity UI first we can add those as a follow up if needed. I'm happy to get this in and then improve on it further as we go. 

One thing that stands out is we could do with much more unit test coverage especially as this is a service layer with significant async logic and branching (filter paths, tx type dispatch, asset resolution). 

It would be great to get some unit tests on: 

- **filterByAsset** (all protocol branches)
- **getStacksTxActivity** (all tx_type cases)
- **mapBitcoinTxToActivity** (self-transfer, send-with-change, receive paths)

alexp3y:
@pete-watters I added quite a few unit tests to increase coverage. Some of the service method tests would require more complex mocks so I'll table that for now. 

To answer the paging question - the complexity of page translations from combined data sources makes virtual pagination our best bet right now. It will grab up to 1000 Stacks Txs and up to 750 Bitcoin Txs. So, a max of up to 1750 txs to work with seems reasonable to cover most users. Currently benchmarking service responses at ~2-3 seconds.

Agreed - Goal here is to get a working implementation behind the updated model / interface so UI integration / build out can start. Service improvements def can/will be added subsequently.

PR #2193, opened 2026-03-05
feat(ui): add web version of numeric input

tigranpetrossian:
Adds a web version of the numeric input, to be used in the slippage selector for extension swaps.

Extracted the logic as is from the component into a shared `useNumericInput` hook so that it's reusable, and added some tests while at it.

I would love for this to be an actual HTML input in the future, but it's a considerable effort to build one, so I'd rather look into adopting and existing one after this Swaps are out.

https://github.com/user-attachments/assets/237d7aca-b4cb-430c-be09-c7db7869382d


pete-watters (approved):
Nice work @tigranpetrossian

packages/ui/src/components/numeric-input/use-numeric-input.shared.ts
```diff
@@ -0,0 +1,132 @@
...
+  return decimals !== undefined ? value.toFixed(decimals) : String(value);
+}
+
+export function countDecimals(value: number): number {
+  return (String(value).split('.')[1] || '').length;
+}
+
+export function roundToPrecision(value: number, decimals: number): number {
+  const factor = Math.pow(10, decimals);
+  return Math.round(value * factor) / factor;
```
kyranjamie, line 127: Should we use BigNumber here to avoid floating point imprecision issues?

kyranjamie (approved):
Tidy component @tigranpetrossian

packages/ui/src/components/numeric-input/use-numeric-input.shared.ts
tigranpetrossian: The precision would definitely be an issue, although not in our current use-cases (mostly 1-2 decimal points with 0.1 step)
I think BigNumber is an overkill for the UI library, but I can build a minimal snapping algorithm to catch and fix drifts on the fly, I've done this before, and some decent numeric input libraries do the variations of that.

Will do this in a followup to avoid a context switch: https://linear.app/leather-io/issue/LEA-3488/improve-precision-tolerance-in-numeric-input

PR #2194, opened 2026-03-06
feat(extension): first token

edgarkhanzadian:
https://github.com/user-attachments/assets/ddf1a7a8-06db-4943-886b-91e96ce28d16


apps/extension/src/app/pages/home/components/tokens.tsx
```diff
@@ -9,22 +9,31 @@ import type { SerializedCryptoAssetId } from '@leather.io/utils';
...
 import { TokensLearn } from './tokens-learn';
 import { TokensTabHeader } from './tokens-tab-header';
 
 export function Tokens() {
   const navigate = useNavigate();
   const location = useLocation();
   const { releaseTrendingTokens } = useFlags();
+  const accountIndex = useCurrentAccountIndex();
+  const account = useAccountAddresses(accountIndex);
+  const activityQuery = useActivity(account);
```
pete-watters, line 26: Maybe we can also call `useActivity(` at `home.tsx` level to pre-query and populate the cache in advance?

pete-watters (approved):
Nice work  @edgarkhanzadian 👍  

This code LGTM but you need to:
- add some E2E tests to protect this logic so it doesn't get broken
    - when and why to show BUY / empty state
- get the CI working as it's failing due to changes to the `/fund` route I think

@fabric-8 also flagged a design issue to fix https://trustmachines.slack.com/archives/C05LRS7G44R/p1772791104782539?thread_ts=1772641583.118149&cid=C05LRS7G44R

apps/extension/src/app/pages/home/components/tokens.tsx
edgarkhanzadian: well tokens component is used as a child of home component so imo it's not going to be that much beneficial

apps/extension/src/app/pages/home/components/tokens.tsx
```diff
@@ -9,22 +9,31 @@ import type { SerializedCryptoAssetId } from '@leather.io/utils';
...
 
+import { FirstTokenBanner } from './first-token-banner';
 import { TokensLearn } from './tokens-learn';
 import { TokensTabHeader } from './tokens-tab-header';
 
 export function Tokens() {
   const navigate = useNavigate();
   const location = useLocation();
   const { releaseTrendingTokens } = useFlags();
+  const accountIndex = useCurrentAccountIndex();
```
kyranjamie, line 24: Let's please not merge this until we release & merge multi-wallet

kyranjamie: or build the changes on top of the multi-wallet PR

edgarkhanzadian: sure, this is a small part of the onboarding flow that i've thought i've implemented on extension but actually missed it. I'm fine with pushing this back a little, we can double check with design.

PR #2200, opened 2026-03-09
fix(extension): trending tokens click

edgarkhanzadian:


pete-watters (approved):
Nice work @edgarkhanzadian 👍  I think we need to do a design review of this feature pre-release. 

The trending tokens list doesn't respond to increased screen width, I'm not sure if thats correct? 

Image

apps/extension/src/app/pages/home/components/tokens.tsx
```diff
@@ -35,7 +35,7 @@ export function Tokens() {
       {showFirstTokenBanner && <FirstTokenBanner />}
       {!showFirstTokenBanner && <TokensTabHeader />}
       <TokenList filter="enabled" onSelectAsset={handleSelectAsset} showDepositButtons />
-      {releaseTrendingTokens && <TrendingTokens />}
+      {releaseTrendingTokens && <TrendingTokens onSelectAsset={handleSelectAsset} />}
```
pete-watters, line 38: Lets deprecate the `releaseTrendingTokens` flag now too and just go live with it?

edgarkhanzadian:
Good point, i'll ping design

tigranpetrossian (approved):

fabric-8:
Great stuff!

I guess we can make Bitcoin and Stacks clickable as well, the same way USDCX works. This way people can also learn more about the token (buy button should stay intact and keep working the same). The pattern is consistent, but they can also buy from the details screen. 

Regarding zero token balances (slightly off topic, but mentioning here since it stands out particularlyhere): is there a way that we could just display a single zero instead of fiat plus crypto plus decimals? Would express the same thing and be less noisy. WDYT @mica000 ?

edgarkhanzadian:
@fabric-8 sounds good! Making bitcoin and stacks clickable would be a separate pr not connected to this one, this one only enables pressing on trending tokens. I'll try to get that one today as well

mica000:
> Regarding zero token balances (slightly off topic, but mentioning here since it stands out particularlyhere): is there a way that we could just display a single zero instead of fiat plus crypto plus decimals? Would express the same thing and be less noisy. WDYT @mica000 ?

@fabric-8 like this?
CleanShot 2026-03-09 at 10 31 58@2x

apps/extension/src/app/pages/home/components/tokens.tsx
edgarkhanzadian: i'd rather have it just in case until next release, so that we'd be able to remove it at any point in prod

PR #2201, opened 2026-03-09
feat(extension): collectible details screens

pete-watters:
This PR adds the collectible details screens to the `extension` . I've stuck a balance between keeping exiting behaviour but also adding the new design via a settings menu option to unprotect inscriptions.

It's not clear from the design / plan if we want to change existing behaviour a lot but opening this PR now for design/ code review and we can discuss it more. 

Here's a demo of what its doing:

https://github.com/user-attachments/assets/eff067f8-3615-469d-b48e-c9f8a00f59b6


kyranjamie (approved):
Can haz loads of JSX

PR #2202, opened 2026-03-09
fix(extension): make deposits pressable

edgarkhanzadian:
https://github.com/user-attachments/assets/5fc47fe3-9ef1-4cd7-bb1b-f051b898e690


apps/extension/src/app/components/deposit-item/deposit-item.tsx
```diff
@@ -33,13 +44,25 @@ export function DepositItem({
       img={icon}
       titleLeft={spamFilter(titleLeft)}
       captionLeft={spamFilter(captionLeft)}
-      titleRight={titleRight}
+      titleRight={null}
```
pete-watters: Can this be optional instead and have it handled in that component instead of passing `null`?

pete-watters (approved):

PR #2211, opened 2026-03-11
fix(ui): improve switch on/off contrast and add focus ring

mica000:
CleanShot 2026-03-10 at 21 56 33@2x

## Summary

Improves the switch component's on/off state differentiation and adds focus ring for accessibility.

**Changes:**
- **Selected (on) state**: Track background changes to `ink.action-primary-default` (dark) for clear contrast
- **Unselected (off) state**: Light track with border (unchanged)
- **Focus ring**: Added blue outline (`lightModeBlue.500`) for keyboard navigation accessibility

## Problem

Previously, the switch had nearly identical on/off states - the track background stayed the same and only the thumb shifted between two similar grays, making it hard to differentiate.

## Test plan

- [ ] Load extension and verify switch shows dark track when selected/on
- [ ] Verify switch shows light track with border when unselected/off
- [ ] Test keyboard navigation - verify blue focus ring appears on Tab
- [ ] Check other switch usages across the app for regressions

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/ui/src/components/switch/switch.web.tsx
+++ b/packages/ui/src/components/switch/switch.web.tsx
@@ -11,6 +11,15 @@ const switchRootStyles = css({
   borderColor: 'ink.border-default',
   borderWidth: 1,
   position: 'relative',
+  '&[data-state="checked"]': {
+    backgroundColor: 'ink.action-primary-default',
+    borderColor: 'ink.action-primary-default',
+  },
+  '&:focus-visible': {
+    outline: '2px solid',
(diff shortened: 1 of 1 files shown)


tigranpetrossian:
Thanks @mica000, looks great. We don't have `lightModeBlue.500` anywhere in the codebase, I've noticed it's used in couple other places as well, but that's a mistake -- should be `blue.border`.

tigranpetrossian (approved):

kyranjamie (approved):

PR #2221, opened 2026-03-13
fix(web): resolve extension detection race condition

kyranjamie:
## Summary
- The extension provider (`window.LeatherProvider`) is injected asynchronously by the content script (~20ms after page scripts start), but the web app's Jotai `extensionStateAtom` checked it synchronously once during initial render and never re-evaluated
- Added a reactive `providerDetectedAtom` with a polling hook (`useDetectLeatherProvider`) that checks every 50ms for up to 2s, updating the atom when the provider arrives
- Called the hook from the app root so the button correctly transitions from "Install" to "Connect" once the provider is detected

## Test plan
- [x] Verified with Playwright: 10/10 rapid page loads now detect the provider within the poll window
- [ ] Manual test: install extension, visit web app, confirm "Connect" button appears (not "Install")
- [x] Verify no regressions on the connected state (address display)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ tmp
 **/.claude/settings.local.json
 .cursorrules
 test-results
+.claude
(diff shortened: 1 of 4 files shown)


314159265359879:
It is pretty consistent here showing install when not available and connect when it is.

314159265359879:
should the popup also change the order when it is not yet installed, or not worth the effort? having "connect" in the top does feel like the best default. 
image

fabric-8:
Nice! @314159265359879 Yeah good callout, we also consider "greying" out the install so things stay consistent but the action to take is clearer

pete-watters (approved):
`jotai` lives on in web?! 🫨

PR #2223, opened 2026-03-15
feat(extension): add info tooltips to swap review screen

tigranpetrossian:
Adds info tooltips to the swap review screen — min. receive, slippage, price impact, and estimated fees.

https://github.com/user-attachments/assets/b692eab4-1afa-474d-82c9-c5c7605eae0b

**+ Few minor fixes:**
* Added few useful input attributes to the amount field.
* Fixes auto-focus on the amount field — now only focuses after selecting the base asset, and maintains focus after tapping max.

### DIFF
--- a/apps/extension/src/app/pages/swap/components/review/fees-tooltip-content.tsx
+++ b/apps/extension/src/app/pages/swap/components/review/fees-tooltip-content.tsx
@@ -1,0 +1,63 @@
+import { Flex, styled } from 'leather-styles/jsx';
+
+import { SwapProviderId } from '@leather.io/models';
+import { LiveSwapEstimate } from '@leather.io/state/swap';
+
+import { formatCurrency } from '@app/common/currency-formatter';
+
+type SuccessLiveSwapEstimate = Extract<LiveSwapEstimate, { status: 'success' }>;
+
+const providerDisplayNames: Record<SwapProviderId, string> = {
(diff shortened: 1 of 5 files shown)


edgarkhanzadian (approved):

apps/extension/src/app/pages/swap/components/review/fees-tooltip-content.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+type SuccessLiveSwapEstimate = Extract<LiveSwapEstimate, { status: 'success' }>;
+
+const providerDisplayNames: Record<SwapProviderId, string> = {
+  'bitflow-sdk': 'Bitflow',
+  'alex-sdk': 'ALEX',
+  'velar-sdk': 'Velar',
+  'sbtc-bridge': 'sBTC Bridge',
+  'bitflow-bff-api': 'Bitflow',
+};
```
kyranjamie, line 16: Shouldn't mapping data like this live higher up, in a context for example? Aim being to lower surface area of changes when adding/removing providers

tigranpetrossian: I agree, this is silly. Alex and I discussed including the name in the service payload so this wouldn't be necessary, I can no longer remember why we didn't do it, but we probably should. I'll log this so we can address it in both apps in a followup when he's back.

PR #2226, opened 2026-03-16
fix(extension): new account becomes active account

kyranjamie:
Makes new account the active one, and improves UX slightly we simulated delay and loading spinner.

### DIFF
--- a/apps/extension/src/app/features/dialogs/switch-account-sheet/switch-account-sheet.tsx
+++ b/apps/extension/src/app/features/dialogs/switch-account-sheet/switch-account-sheet.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, useState } from 'react';
 import { Virtuoso } from 'react-virtuoso';
 
 import { Box, Flex } from 'leather-styles/jsx';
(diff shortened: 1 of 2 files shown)


apps/extension/src/app/features/dialogs/switch-account-sheet/switch-account-sheet.tsx
```diff
@@ -31,6 +36,8 @@ export function SwitchAccountSheet({ isShowing, onClose }: SwitchAccountSheetPro
   );
 
   async function onCreateAccount() {
+    setIsCreatingAccount(true);
+    await bufferDelayBeforeThreadBusy();
```
edgarkhanzadian: I assume you use a delay here to make sure the state is updated before proceeding to some heavy computation? If that's the case maybe we could utilize flashSync here or requestIdleCallback

kyranjamie: I attempted requestAnimationFrame before landing on this. Will try requestIdleCallback, but the gist is that, without a delay, the animation won't start because before it's had the chance, we're hammering the thread with account generation logic

Using a worker and doing async would be the correct solution, this is a hack

edgarkhanzadian: I think this should do the trick. Agree, would be nice to have a worker
function onCreateAccount() {
    setIsCreatingAccount(true);
    requestIdleCallback(async () => {
      await createAccount();
      onClose();
    });
  }

edgarkhanzadian (approved):

PR #2228, opened 2026-03-16
Visual test: use text-subdued-primary everywhere

mica000:
This is a visual test to use text-subdued-primary everywhere instead of text-subdued-secondary.

I created this branch to see how the darker subdued color looks across the app. It swaps all `ink.text-subdued-secondary` usages to `ink.text-subdued-primary`.

Before we update this with dev and merge, can you review the color direction? You can load the extension from this branch to compare. Let me know if this is the right path or if we should stick with secondary.

This is how the color scale will be. The new color being added in Fab's PR is the `text-sudued-primary`. It will replace the old `text-subdued` which now will be called `text-subdued-secondary`

CleanShot 2026-03-16 at 16 55 19@2x

### DIFF
--- a/apps/extension/src/app/components/account/account-list-item.layout.tsx
+++ b/apps/extension/src/app/components/account/account-list-item.layout.tsx
@@ -40,7 +40,7 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
       titleRight={
         isLoading ? (
           <Spinner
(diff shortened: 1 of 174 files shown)


fabric-8:
I browsed a bunch of views in the extension & compared with the update. I think this looks good 👍

fabric-8 (approved):

PR #2233, opened 2026-03-18
fix: bis pagination

edgarkhanzadian:


packages/services/src/infrastructure/api/best-in-slot/best-in-slot-api.client.ts
```diff
@@ -34,6 +36,36 @@ export class BestInSlotApiClient {
...
+  private async fetchPaginated<T>(
+    url: string,
+    baseParams: URLSearchParams,
+    schema: z.ZodType<T>,
+    signal?: AbortSignal
+  ): Promise<T[]> {
+    const results: T[] = [];
+    let offset = 0;
+
+    while (true) {
```
kyranjamie, line 48: Would suggest adding the actual condition here over using break, even if this means having state outside

PR #2237, opened 2026-03-25
refactor(web): redesign pooled stacking position details view

fabric-8:
CleanShot 2026-03-25 at 18 07 54@2x

## Summary

- Moved pool name and action buttons into the info grid as a unified header row (desktop), removing the separate header section
- Consolidated pool STX address and pool's BTC pox address together under "Pool address"
- Renamed "Reward address" to "My reward address" and now displays the user's own BTC address instead of the pool's
- Made action buttons smaller (`size="md"`) and changed "Stop pooling" to a ghost button with a stop icon

## Test plan

- [ ] Navigate to an active pooled stacking position page
- [ ] Verify pool name + action buttons appear inside the info grid on desktop
- [ ] Verify action buttons appear above the grid on mobile
- [ ] Verify "Pool address" shows both the STX and BTC addresses stacked
- [ ] Verify "My reward address" shows the user's own BTC address
- [ ] Verify "Stop pooling" renders as a ghost button with the stop icon
- [ ] Verify no vertical divider between pool name and buttons in the header row

### DIFF
--- a/apps/web/app/features/stacking/components/address.tsx
+++ b/apps/web/app/features/stacking/components/address.tsx
@@ -1,7 +1,6 @@
 import { Box, Flex, styled } from 'leather-styles/jsx';
 
 import { CopyIcon, ExternalLinkIcon, useClipboard } from '@leather.io/ui';
-import { truncateMiddle } from '@leather.io/utils';
 
 interface AddressProps {
   address: string;
@@ -17,7 +16,7 @@ export function Address({ address }: AddressProps) {
       overflow="hidden"
       textOverflow="ellipsis"
     >
-      {truncateMiddle(address)}
+      {address}
     </styled.p>
   );
 }
@@ -31,7 +30,8 @@ export function CopyAddress({ address }: AddressProps) {
       gap="space.01"
       onClick={onCopy}
       cursor="pointer"
-      maxWidth={['250px', 'none', '220px', 'none']}
+      minWidth={0}
+      overflow="hidden"
     >
       <Box display="inline-block">
         <CopyIcon variant="small" />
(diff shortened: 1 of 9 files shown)


314159265359879:
Thanks for creating this. I will take a closer look tomorrow. A question: could we show the whole address instead of the shortened version? It seems like we have ample space here.

fabric-8:
@314159265359879 Good point, just pushed a fix for that

edgarkhanzadian:
ui breaks on smaller screens

bug1

apps/web/app/features/stacking/components/address.tsx
```diff
@@ -17,7 +16,7 @@ export function Address({ address }: AddressProps) {
       overflow="hidden"
       textOverflow="ellipsis"
     >
-      {truncateMiddle(address)}
```
edgarkhanzadian, line 20: we should probably keep this to not break ui

fabric-8:
https://github.com/user-attachments/assets/0f430820-2dc5-4976-b872-5baf6b681448

Thx @edgarkhanzadian !

I think we have two options: Either truncate via CSS (which will always truncate the end) or always use our truncator function. Theoretically, we could have a function that dynamically truncates the middle. However, that would be unreasonably complex/heavy, so I don't think it's a realistic option. 

@314159265359879 Do you have any strong objections regarding truncating the end? If yes, we should stick to the current approach.

314159265359879:
@fabric-8 I believe the truncating as shown in the attached video is okay. 

I think you could avoid having to truncate most of the time if you display these cells below eachother instead of next to eachother? Also gets rid of the alignment issue with the cel border in the middle. 

I would display "My reward address" first then "pool address" below. 

image

314159265359879 (approved):
This looks good to me too, lets merge.

Image

fab-stacks:
Superseded by #2348 — recreated with a verified signature under the current `fab-stacks` identity (the original commit was authored under `hello@fabric8.de`, which fails the new signed-commit rule). #2348 also addresses the open feedback from @314159265359879 about stacking "My reward address" above "Pool address" vertically.

PR #2238, opened 2026-03-25
feat(web): Clarify position actions

fabric-8:
https://github.com/user-attachments/assets/54f8868b-f6de-4f43-9628-a399be28808b

## Summary
- Position rows are now fully clickable and show a hover highlight on mouse over
- Replaced the chevron dropdown next to pool name with an ellipsis icon button that appears flush next to the provider icon on hover
- The ellipsis button visually extends the provider icon (shared border radius flattened between them) and uses a transparent hover token (`ink.component-background-hover`)
- ProviderIcon refactored to accept a `size` prop, keeping 24px default for tables while allowing 32px in positions

## Test plan
- [ ] Hover a stacking position row — all cells highlight with `ink.background-secondary`
- [ ] Ellipsis button appears attached to the provider icon on hover
- [ ] Click the ellipsis to open dropdown with position actions
- [ ] Click anywhere else on the row to navigate to position details
- [ ] Provider icons in pool/liquid stacking tables remain 24px
- [ ] Dropdown stays open if mouse leaves the row while menu is active

### DIFF
--- a/apps/web/app/components/icons/provider-icon.tsx
+++ b/apps/web/app/components/icons/provider-icon.tsx
@@ -1,20 +1,30 @@
-import { ReactElement } from 'react';
-
 import { ProviderId, providerIdSchema } from '~/data/data';
 
 import { ImgFillLoader } from '../img-loader';
 
-const stackingProviderIcons: Record<ProviderId, ReactElement> = {
-  fastPool: <ImgFillLoader src="/icons/fastpool.webp" width="24" fill="black" />,
-  fastPoolV2: <ImgFillLoader src="/icons/fastpool.webp" width="24" fill="black" />,
-  planbetter: <ImgFillLoader src="/icons/planbetter.webp" width="24" fill="black" />,
-  restake: <ImgFillLoader src="/icons/restake.webp" width="24" fill="#124044" />,
-  xversePool: <ImgFillLoader src="/icons/xverse.webp" width="24" fill="black" />,
-  stackingDao: <ImgFillLoader src="/icons/stacking-dao.webp" width="24" fill="#1C3830" />,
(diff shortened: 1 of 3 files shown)


edgarkhanzadian:
clicking on menu items also triggers the container's click function.
In case of clicking "stop pooling", it also moves user to the pool position screen. (but it only should show the approver flow)
In case of clicking "increase pooling amount", most of the time it only moves user to the pool position screen and not increase pooling amount screen

https://github.com/user-attachments/assets/a56ce40a-42b0-46e2-83cd-eb30d1014b41

314159265359879:
I am not sure this improves visibility because now the user has to hover over the item before they see there is a menu button. 

Before, there was an arrow that was missed as a drop-down menu, but it was always displayed and didn't require a hover. 

What about a self-explanatory button that is always visible?
"Manage Pool"

In an extra column or row, perhaps?


I am not sure we should display the same options for the liquid stacking pool; we do not currently support "increase/stop" there, although we could direct users to app.stackingdao.com instead.

fabric-8:
CleanShot 2026-03-26 at 11 27 29@2x @314159265359879 how about this? It feels a bit "old school", however, it can make sense here as rows are generally not clickable in the zu which is why I think people might indeed miss the hover/click

fabric-8:
@314159265359879 Went ahead and pushed my suggestion, hope that makes things super clear ;)

314159265359879 (approved):
Thanks fab, this looks good to me!

fab-stacks:
Superseded by #2339 — rebased onto current `dev` and re-authored under the current `fab-stacks` identity so the commit passes the repo's verified-signature push rule. Closing this one.

PR #2239, opened 2026-03-26
fix(extension): remove inscriptions buttons

edgarkhanzadian:
https://github.com/user-attachments/assets/0dfe7be8-8062-4b6c-b700-115a3c543746


kyranjamie:
Don't these want to be put back to their original position? Doesn't this kill the feature entirely?

edgarkhanzadian:
it only removes it from the settings menu, but i need to change this up a little based on this convo: https://trustmachines.slack.com/archives/C05LRS7G44R/p1774528987952639

kyranjamie (approved):

PR #2265, opened 2026-04-13
feat: implement auth service, ref LEA-3526

alexp3y:
This PR adds a new auth infrastructure layer to enable backend JWT authentication.

Adds a new authenticated backend API client `LeatherAuthApiClient` that handles bearer token auth, 401 token refresh flow, and auth failure events. All authenticated domain endpoints (e.g. `/v1/multisig/*`) are accessed using this client.

The auth API client makes use of an app-specific infrastructure service `TokenAuthService` that can be optionally injected by implementing applications (e.g. `web`) in order to retrieve data from authenticated services (e.g. `MultisigService`).

The `TokenAuthService` implementation provides simple token storage and lifecycle callbacks, while token lifecycle is managed by the API client. Once tokens are stored via `setTokens()`, all subsequent service calls are automatically authenticated, with no need for token passing or refreshing.

```typescript
// app handles initial authentication and token persistence
const { accessToken, refreshToken } = await getLeatherAuthApiClient().authenticate(signature, publicKey, timestamp);
webTokenAuthService.setTokens(accessToken, refreshToken);

// service calls then use token from the injected infra (no token passing)
await getMultisigService.doSomethingAuthenticated();
```

A placeholder `WebTokenAuthService` has been added to `web` + injected into the DI container to demo token state usage.

A simple `MultisigService` is added with a single authenticated call to allow e2e tests. This will subsequently be replaced by a full domain-specific implementation during mutisig build-out.

Finally, an e2e test script `pnpm test:auth` was added to `@leather.io/services` which validates the full flow (authentication, authenticated endpoint access, token refresh) against the Staging API.


apps/web/app/services/web-token-auth.service.ts
```diff
@@ -0,0 +1,28 @@
+import type { TokenAuthService } from '@leather.io/services';
+
+let accessToken: string | null = null;
+let refreshToken: string | null = null;
```
edgarkhanzadian, line 4: Should these be stored in localStorage?

packages/services/src/infrastructure/api/leather/leather-auth-api.client.ts
```diff
@@ -0,0 +1,105 @@
...
+export class LeatherAuthApiClient {
+  private readonly client;
+  private isRefreshing = false;
+
+  constructor(
+    @inject(Types.TokenAuthService) private readonly tokenProvider: TokenAuthService,
+    @inject(Types.Environment) env: Environment,
+    private readonly rateLimiter: RateLimiterService
+  ) {
+    const clientId = uuidv4();
```
edgarkhanzadian, line 24: Should the clientId be persisted?

edgarkhanzadian (approved):

apps/web/app/services/web-token-auth.service.ts
alexp3y: Totally up to you. I spun this up just as a simple placeholder to demo the touch points, but you can implement and integrate however makes the most sense when wiring up the app auth flow.

packages/services/src/infrastructure/api/leather/leather-auth-api.client.ts
alexp3y: This is ok as-is. It's not used for persistent identity, just request correlation and rate-limiting so its safe to regenerated when the session / context refreshes. We use the same pattern in the existing non-authenticated Leather API client.

PR #2272, opened 2026-04-16
chore: claude code reviews

edgarkhanzadian:


alexp3y (approved):
Nice addition!

Looks like the also have a security review action:
https://github.com/anthropics/claude-code-security-review

Wonder if its worth having both?

edgarkhanzadian:
good call! I'll add it as well

PR #2286, opened 2026-04-21
chore: bump claude version

edgarkhanzadian:
With 4.6 i'm not really reaching even 10% usage, so let's bump the version to 4.7


andresgalante-stacks (approved):
🤖 🚀

PR #2288, opened 2026-04-22
fix: remove mobile runes/stamps

edgarkhanzadian:
1. Fix security audit for xmldom https://github.com/leather-io/mono/actions/runs/24822841403/job/72651398988
2. Remove runes from mobile

As claude correctly pointed out, there are a few shortcuts that i took regarding runes. Mostly because we already have a PR lined up that would remove all runes entirely from the monorepo codebase, this one is a hotfix to unblock mobile users


apps/mobile/src/features/token/collectibles-list.tsx
```diff
@@ -26,12 +25,12 @@ interface RenderCollectibleProps {
 function renderCollectible({ item, height, onPress }: RenderCollectibleProps) {
   const collectible = (() => {
     switch (item.protocol) {
-      case 'stamp':
-        return <Stamp item={item} height={height} onPress={onPress} />;
       case 'sip9':
         return <Sip9 item={item} height={height} onPress={onPress} />;
       case 'inscription':
         return <Inscription item={item} height={height} onPress={onPress} />;
+      case 'stamp':
```
jannik-stacks, line 32: Do we keep these because of old state or any particular reason?

apps/mobile/src/features/token/collectibles-list.tsx
```diff
@@ -57,7 +56,10 @@ interface CollectiblesListProps {
 }
 
 export function CollectiblesList({ collectiblesState, header }: CollectiblesListProps) {
-  const collectibles = collectiblesState.state === 'success' ? collectiblesState.value : [];
+  const collectibles =
+    collectiblesState.state === 'success'
+      ? collectiblesState.value.filter(c => c.protocol !== 'stamp')
```
jannik-stacks, line 61: Same, are these stamps still returned?

alexp3y (dismissed):

jannik-stacks (approved):
Looks good on my end, logic is fine if the `available` to `total` switch is what we want

alexp3y (approved):

apps/mobile/src/features/token/collectibles-list.tsx
edgarkhanzadian: just in case we accidentally return stamp from the api

apps/mobile/src/features/token/collectibles-list.tsx
edgarkhanzadian: just in case we accidentally get stamps. it will fully be removed after we fix mobile

apps/mobile/src/features/token/collectibles-list.tsx
jannik-stacks: 🙏🏻

PR #2292, opened 2026-04-24
fix: sticky comment

edgarkhanzadian:


.github/workflows/extension:pr-build.yml
```diff
@@ -37,15 +37,15 @@ jobs:
     name: set-in-progress-message
     if: github.repository == 'leather-io/mono' && github.actor != 'dependabot[bot]'
     runs-on: ubuntu-24.04
+    permissions:
+      pull-requests: write # post sticky build-status comment on the PR
```
jannik-stacks, line 41: Will this work, I remember pr write is off org wide, maybe I'm wrong

jannik-stacks (approved):

.github/workflows/extension:pr-build.yml
edgarkhanzadian: nop it can only work in this repo

alexp3y (approved):

PR #2306, opened 2026-04-29
feat(mobile): migrate off eas

edgar-stacks:
Tested with dummy android keystore and dummy ios certificates, seems to be working correctly.
Final test is going to be once we merge this into dev and actually use `ios:production` and `android:production` to publish to appstore. It might need a follow-up PR in case submission fails due to some configuration

This PR:
- removes firebase notifications from mobile (keeping notificationPreference in store for now to not accidentally cause issues with the store during migration)
- separates android build and signing into separate jobs. Only signing requires environment `android:production`
- separates ios build and signing into separate jobs. Only signing requires environment `ios:production`
- creates workflow dispatch. Only possible to run this CI on dev branch with workflow dispatch. (all | android | ios)


edgar-stacks:
Optional points mentioned by Claude review can be addressed in the follow-up PRs. Let's get this in and actually test if the flow fully works first.

aperry-stacks (approved):

andresgalante-stacks (approved):

.github/zizmor.yml
```diff
@@ -8,3 +8,4 @@ rules:
       allow:
         - CODECOV_TOKEN
         - CLAUDE_CODE_OAUTH_TOKEN
+        - SENTRY_AUTH_TOKEN
```
adriano-stacks, line 11: nit, but I think we said we'd use OIDC for Sentry as well?

apps/mobile/fastlane/android/fastlane/Fastfile
```diff
@@ -1,3 +1,65 @@
 default_platform(:android)
 
-# stubbed out but not used for now
\ No newline at end of file
+desc "Look up next versionCode from Play and build an unsigned release AAB."
+lane :build_android_aab do
+  next_code = (google_play_track_version_codes(
+    package_name: "io.leather.mobilewallet",
+    track: "internal",
+    json_key: ENV.fetch("GOOGLE_APPLICATION_CREDENTIALS"),
```
adriano-stacks, line 8: I don't remember where this come from, since we decided we'd use OIDC in CI, so is this to try local builds?

.github/zizmor.yml
edgar-stacks: After doing some research i haven't found oidc support for sentry ci

apps/mobile/fastlane/android/fastlane/Fastfile
edgar-stacks: fastfile receives GOOGLE_APPLICATION_CREDENTIALS from ci's gcp oidc https://github.com/leather-io/mono/pull/2306/changes#diff-0e3ba282d9deb511bbac622ae0a315e4961a821a7736159c2d25aaa8296e1674R236-R241

PR #2308, opened 2026-05-04
fix: mobile release

edgar-stacks:
- Use ANDROID_KEYSTORE_PASSWORD as both keystore and key password on android
- Make marketing and build version the same for simplicity of ios uploads


edgar-stacks:
Concerns mentioned by claude review are OK as buildNumber === version would improve the ci simplicity. we can revisit it later if it blocks any of our flows.

adriano-stacks (approved):

aperry-stacks (approved):

PR #2320, opened 2026-05-06
feat(mobile): ordinals feature flag

edgar-stacks:
https://github.com/user-attachments/assets/e0ba795d-433c-4548-8a64-890636e4dd79


edgar-stacks:
Addressing claude's response

> 1. account-collectibles.query.ts:30,48 — flag gates all bitcoin collectibles, not just inscriptions

Yes we want that at this point.

> 2. Pre-existing param mutation, now broader

That's OK, we are going to rip all of this out of the codebase after May 18.

aperry-stacks:
@edgar-stacks so the `release_ordinals_feature` flag is currently `true` and we'll be flipping to `false` when we deprecate, right?

edgar-stacks:
@aperry-stacks yeap yeap

aperry-stacks (approved):
LGTM

jannik-stacks (approved):
Lgtm

PR #2324, opened 2026-05-08
feat(extension): support originator post-condition mode

### LINKED ISSUE #2287: stx_callContract: accept "originator" as a valid postConditionMode
## Summary

`@stacks/transactions@7.4.0` added `PostConditionMode.Originator` (string form `"originator"`), and the mode has been live on-chain since Stacks epoch 3.4. `@stacks/connect@8.2.6` passes the string through to the wallet without validation. However Leather rejects it with:

```
Validation error: Invalid input at "postConditionMode"
```

…thrown before the sign prompt, so there is no way to dispatch an Originator-mode tx from a Leather-connected dApp.

## Repro

```ts
await request("stx_callContract", {
  contract: "SP….CONTRACT.name",
  functionName: "settle",
  functionArgs: [],
  postConditionMode: "originator",
  postConditions: [Pc.principal(userAddress).willSendLte(10_000).ustx()],
});
```

Result: `Validation error: Invalid input at "postConditionMode"` (no sign prompt shown).

Changing the mode to `"allow"` or `"deny"` works as expected.

## Expected

Leather accepts `"originator"` as a valid `postConditionMode` value (alongside `"allow"` and `"deny"`) and forwards the tx to the signer for user approval. The Leather UI may want to surface the mode to the user explicitly so they understand that only their own outflows are being checked.

## Why this matters

Originator mode is the clean pattern for dApps where a single user tx causes the contract to move assets between many principals (DEX settlements, batch auctions, payout distributions, multi-depositor refunds). Without it, the app has to choose:

- `"allow"` — works today but loses strictness on undeclared caller outflows;
- `"deny"` — requires enumerating every inter-principal transfer inside the contract, which is impossible when the set of recipients is dynamic.

With Originator mode the caller's wallet is still fully guarded (their declared PCs are the only ones enforced against them) while all internal contract transfers flow freely.

## Context

- Type shipped: `@stacks/transactions/postcondition-types.d.ts` — `export type PostConditionModeName = 'allow' | 'deny' | 'originator';`
- Passed through: `@stacks/connect@8.2.6` `_e()` helper returns strings as-is (no client-side rejection)
- On-chain support: live since Stacks epoch 3.4

Reported from Jing Swap (jingswap.com). Happy to provide a minimal repro URL if useful.

Thanks!

aperry-stacks:
Updates post-condition support to add "Originator" mode and "May Send" condition.

Bumps `@stacks/transactions` to `7.4.0`

### DIFF
--- a/apps/extension/tests/transation-test-utils.ts
+++ b/apps/extension/tests/transation-test-utils.ts
@@ -53,7 +53,7 @@ export async function generateContractCallToken({
     contractName: 'hello-world',
     functionArgs: [],
     functionName: 'print',
-    postConditionMode: PostConditionMode.Allow,
+    postConditionMode: PostConditionMode.Allow as ContractCallOptions['postConditionMode'],
     network: network as any,
     postConditions: [serializePostConditionWire(postConditionToWire(pc))],
     ...txOptions,
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -82,7 +82,7 @@
     "@stacks/network": "7.0.2",
     "@stacks/network-v6": "npm:@stacks/network@6.13.0",
     "@stacks/rpc-client": "1.0.3",
-    "@stacks/transactions": "7.0.5",
+    "@stacks/transactions": "7.4.0",
     "@stacks/wallet-sdk": "7.1.2",
     "@tanstack/query-async-storage-persister": "5.85.5",
     "@tanstack/react-query": "5.85.5",
--- a/apps/extension/src/app/common/transactions/stacks/post-condition.utils.ts
+++ b/apps/extension/src/app/common/transactions/stacks/post-condition.utils.ts
@@ -3,6 +3,8 @@ import {
   type FungiblePostConditionWire,
   NonFungibleConditionCode,
   type NonFungiblePostConditionWire,
+  PostConditionMode,
+  type PostConditionModeName,
   PostConditionType,
   type PostConditionWire,
   type STXPostConditionWire,
@@ -19,6 +21,12 @@ import { stacksValue } from '@app/common/stacks-utils';
 //
 // These were used with legacy tx requests, so we have to keep them
 // until we are able to remove that code entirely.
+
+export function rpcPostConditionModeToEnum(mode?: PostConditionModeName) {
+  if (mode === 'allow') return PostConditionMode.Allow;
+  if (mode === 'originator') return PostConditionMode.Originator;
+  return PostConditionMode.Deny;
+}
 
 export function getIconStringFromPostCondition(
   pc: STXPostConditionWire | FungiblePostConditionWire | NonFungiblePostConditionWire
@@ -78,6 +86,8 @@ export function getPostConditionCodeMessage(
       return `${sender} will transfer`;
     case NonFungibleConditionCode.DoesNotSend:
       return `${sender} will keep or receive`;
+    case NonFungibleConditionCode.MaybeSent:
+      return `${sender} may transfer`;
     default:
       assertUnreachable(code);
   }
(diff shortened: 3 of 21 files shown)


apps/extension/tests/transation-test-utils.ts
```diff
@@ -53,7 +53,7 @@ export async function generateContractCallToken({
     contractName: 'hello-world',
     functionArgs: [],
     functionName: 'print',
-    postConditionMode: PostConditionMode.Allow,
+    postConditionMode: PostConditionMode.Allow as ContractCallOptions['postConditionMode'],
```
jannik-stacks, line 56: why does this need a cast?

jannik-stacks (dismissed):
looks good

edgar-stacks (dismissed):

aperry-stacks:
Added Originator callout UI.

edgar-stacks:
PR for fixing audit packages: https://github.com/leather-io/mono/pull/2325

apps/extension/tests/transation-test-utils.ts
aperry-stacks: Overrides conflicting `@stracks/transactions` transitive dep from `@stacks/connect-jwt`

aperry-stacks:
Fixed a pre-existing NFT post-conditions display bug in mobile preventing display of "may transfer" PCs.

edgar-stacks (dismissed):

apps/extension/src/app/features/rpc-stacks-transaction-request/stacks/post-conditions/post-condition-mode-warning.tsx
```diff
@@ -1,6 +1,18 @@
...
+interface PostConditionModeWarningProps {
+  mode: 'allow' | 'originator';
+}
+
+export function PostConditionModeWarning({ mode }: PostConditionModeWarningProps) {
+  if (mode === 'originator') {
+    return (
+      <Callout variant="info" mb="space.03" title="Originator mode">
+        Only these assets may move from your wallet. Other transfers in this transaction are not
+        checked.
```
jannik-stacks: 'originator' is "keep origin safe" right? so might be something like 

- "these assets will/may be transferred" (show only if there is additional post conditions that relate to accounts assets)
- "no other assets from your account can be transferred by this transaction" (or similar)

jannik-stacks: right now this shows in general right? above the asset transfers of the user account but also other post-conditions from e.g. a smart contract. maybe somehow we can

aperry-stacks: cc @314159265359879 @fabric-8 @fab-stacks thoughts?

apps/mobile/src/features/approver/components/post-conditions/post-conditions-warning.tsx
```diff
@@ -15,5 +15,12 @@ export function AllowModePostConditionWarning({ txHex }: { txHex: string }) {
       </Callout>
     );
   }
+  if (tx.postConditionMode === PostConditionMode.Originator) {
+    return (
+      <Callout title={t`Originator mode`} variant="info">
+        {t`Only these assets may move from your wallet. Other transfers in this transaction are not checked.`}
```
jannik-stacks: yeah i think the "other transfers" wording doesn't seem very clear.  "other transfers" could read as what deny mode essentially is. as a user i would read this as: "these assets <below> may move from my wallet. other transfers <my own included> in this tx are not checked <may move or may not move, mode doesn't care about them>".

where it's more like:
- "the asset transfers below will happen, fact" (list of may/will items below)
- originator mode: "nothing else from your account will be transferred out <but that doesn't mean other smart contract can't send stuff>"

`<subtext>`

jannik-stacks (changes requested):
Minor wording change request I'm realizing

jannik-stacks (approved):

edgar-stacks (approved):

PR #2335, opened 2026-05-18
refactor(services): stacks nft paging and increase sip9 page size

aperry-stacks:
Aligns NFT paging requests with established Stacks API client pattern.

Increases default page size to account for inflated NFT holding lists via SFT contracts.

### DIFF
--- a/packages/services/src/collectibles/sip9s.service.spec.ts
+++ b/packages/services/src/collectibles/sip9s.service.spec.ts
@@ -53,9 +53,11 @@ describe(Sip9sService.name, () => {
 
       const sip9s = await sip9sService.getAccountSip9s({ account });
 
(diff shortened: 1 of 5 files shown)


edgar-stacks (approved):
we would need to implement paging on frontend as well right?

aperry-stacks:
> we would need to implement paging on frontend as well right?

No. Paging here only refers to the API client level, which is abstracted from the FE via the `CollectiblesService`, which itself doesn't have a paging concept.

That said, now that Ordinals are deprecated and Collectibles == Stacks NFTs, FE paging becomes much easier to support. V2 Activity is also inching towards pageability, so we should consider looking into a FE paging pattern / design.

jannik-stacks (approved):

PR #2336, opened 2026-05-18
feat(extension): add aeUSDC retirement callout

aperry-stacks:
Removes ordinals deprecation warning and replaces with a new dynamic `aeUSDC` retirement callout. 

Callout visibility is dependent on user having `aeUSDC` balance in wallet account.

image


fab-stacks (approved):

edgar-stacks (approved):

edgar-stacks:
we probably need to roll this out asap

PR #2337, opened 2026-05-19
chore: add zest token to default list

314159265359879:
add zest token to default list

and remove runes token from displaying by default now that runes support has been dropped.


aperry-stacks:
Default token list additions are now handled asynchronously via backend app-config update.

PR #2343, opened 2026-05-25
feat(extension): remove runes and ordinals

edgar-stacks:
In extension only, remove:
runes
ordinals
brc-20
stamps
src-20

I'll get to mobile and packages in other PRs to at least keep these PRs a little bit smaller


edgar-stacks:
Regarding "Dead Redux state" mentioned by Claude, i am going to handle it as part of a separate PR to keep state migrations out of this big PR

apps/extension/src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
```diff
@@ -57,8 +57,7 @@ export function useRpcSignPsbt() {
 
     return await broadcastTx({
       tx,
-      // skip utxos check for psbt txs
-      skipSpendableCheckUtxoIds: 'all',
+      skipTaprootWarning: true,
```
jannik-stacks, line 60: if this always safe, or should user get an additional confirmation dialog/something?

apps/extension/src/app/query/bitcoin/transaction/use-check-taproot-utxos.ts
```diff
@@ -0,0 +1,50 @@
...
+          if (!input.txid) return false;
+          const txid = bytesToHex(input.txid);
+          const match = allWalletUtxos.find(u => u.txid === txid && u.vout === input.index);
+          if (!match) return false;
+          return isTaprootAddress(match.address);
+        });
+
+        if (!hasTaprootUtxos) return false;
+
+        const { userAcceptedRisk } = await TaprootUtxoWarningDialog.call();
```
jannik-stacks, line 40: ah looks like there is a dialog,

apps/extension/src/app/services/extension-settings.service.ts
```diff
@@ -11,13 +11,15 @@ export class ExtensionSettingsService implements SettingsService {
...
-            id: tokenSetting.id,
-          }),
-          tokenSetting.enabled,
-        ])
+        Object.values(selectTokenState(store.getState()).entities)
+          .filter(tokenSetting => tokenSetting.id.includes('::'))
+          .map(tokenSetting => [
+            serializeAssetId({
+              protocol: 'sip10',
+              id: tokenSetting.id,
```
jannik-stacks, line 19: duplicate block of code, but probably still fine. doesn't need to be too DRY

CLAUDE.md
```diff
@@ -114,6 +114,7 @@ You MUST run these after any code changes. Do not consider a task complete until
 pnpm format
 pnpm lint
 pnpm typecheck
+pnpm --filter @leather.io/extension lint:unused-exports
```
jannik-stacks, line 117: intended to commit?

jannik-stacks (dismissed):
a lot of deletions, nice!

CLAUDE.md
edgar-stacks: yeap yeap, claude constantly forgets to run this check

apps/extension/src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
edgar-stacks: for some specific reason we didn't want to prompt users when they do sign psbt rpc call. I'll review that behavior in the future, maybe we might want to display the dialog for all taproot cases

edgar-stacks: cc @314159265359879

apps/extension/src/app/services/extension-settings.service.ts
edgar-stacks: is it? i think this part only moves the protocol filtering logic to a separate `filter` array method

aperry-stacks (approved):
🧹🧹🧹

aperry-stacks (dismissed):

aperry-stacks (approved):

jannik-stacks (approved):

PR #2352, opened 2026-05-26
fix: remove runes/ordinals pages from queries

edgar-stacks:
Remove runes/ordinals from query/queries package


jannik-stacks (approved):

aperry-stacks (approved):
🪓🪓🪓

PR #2356, opened 2026-05-27
feat: multiwallet UI

edgar-stacks:
# feat: multiwallet UI

Turns the extension from a single-wallet app into a **multiwallet** one: users can hold several wallets at once (each software wallet from its own secret key, plus any number of Ledger wallets), with multiple accounts per wallet, all unlocked by a single device password.

The central new surface is a redesigned **"Select account / Manage wallets"** sheet: switch accounts across wallets, add wallets (create / restore / connect Ledger), add accounts per wallet, and — in **Manage** mode — rename wallets/accounts, hide/show accounts, view a wallet's secret key, and remove wallets.

## What changed

**Data model**
- New persisted `accounts` entity slice — per-account metadata (custom name, hidden status) keyed by `fingerprint/accountIndex`.
- Reworked `active` slice/thunks: single active `(fingerprint, accountIndex)` pointer; `changeActiveAccount` syncs the background; wallet removal cascades and re-points to the first *visible* account of a survivor.
- `getAddWalletError` blocks duplicate / cross-type seeds; `stx-chain` purges on removal (no account "resurrection" on re-import).

**Key management & security**
- Decrypted secret keys **moved out of Redux** into a private, non-persisted in-memory `Map` (one per unlocked wallet), exposed via `useSyncExternalStore` — no leakage through DevTools/persistence/logs.
- **One shared password/salt/encryption key** unlocks all software wallets (verified before adding a wallet).
- Cross-frame **wallet-lock broadcast** so locking one window locks them all.

**Migration & persistence**
- `v3 → v4` store migration backfills the `accounts` table for existing users — no re-onboarding, no data loss.

**UI & flows**
- New switch-account sheet (grouped by wallet) + add-wallet sheet, with per-wallet/per-account action menus and rename/remove/hide dialogs.
- Onboarding reworked: `/set-password` and `/forgot-password` routes removed; `SetPasswordPage` is now inline and switches between "set" (first wallet) and "re-enter existing" (additional wallets). Adding a wallet no longer wipes existing ones.
- Receive flow converted from route-based modals to an in-page dialog (Outlet context); `AccountGate` rewritten to be wallet-count aware.
- One-time **feature introducer** announcing multiwallet (persisted "seen" state).
- New `WalletPlusIcon` in `@leather.io/ui` (+ native icon bug fix); `fibonacci` util made O(n) for account-restoration index scanning.

## Tests
- Unit (Vitest): in-memory key store, stx-chain removal, v3→v4 migration, cross-frame lock, `canHideAccount`, account gate.
- E2E (Playwright): new `switch-account` page object + specs for switching, management, **data-integrity/key-isolation invariants**, wallet lock, and sign-out/recreate.

## Notes
- Ships **GA-on with no LaunchDarkly kill-switch flag** (product decision); the only flag change is a test-only `streaming: false` fix.
- Some mobile/extension store deltas (`upsertOne` vs `updateOne`, action-type namespacing) are intentional, to reconcile when slices move to shared `@leather.io/state`.


314159265359879:
1. The font for the title doesn't look like the right one for titles in Leather, right?
image

2. If the user clicks "try it out", do you want them to be directed to the page where they can see the option to add an extra key/wallet? 

It is not very obvious that the feature is in "account selection". We a similar options from the top right menu because the user can simply click the account. Are we still happy with that or should we add a "manage accounts/wallets" option in the top right corner menu too?
The benefit of this would be that it is in line with expectations users have based on older versions of the wallet and it is easier (less ambiguity) to refer to in our guide. 

image

--> 

image

314159265359879:
3. Going through this flow to add a wallet
A. Click account 1
B. Click add wallet
C. See pop-over to select "connect hardware wallet"
D. after the wallet is added the pop-over remains, I think this should automatically close and not show above the newly created wallet with the new wallet greyed out. 

image

314159265359879:
4. When I click "select account" in the (bitcoin or stacks) send flow, the accounts are all bunched up instead of split between wallets. 

image

the user sees this now:
image

the user expects this (same as "select account" page)
image



5. For the third wallet it is a little bit stranger. You see with wallet 1 and wallet 2 it does start counting 1, 2, 3 etc in the wallet avatars. With the third account it looks like they belong to wallet 2 (in this example wallet 2 and 3 are both ledger wallets, both networks are connected but no addresses are displayed or filled when selecting. 
Balances aren't displayed correctly either, looks like wallet can't get the info to display here. 

image


What I expect to see is numbering 1, 2, 3 in front, and I expect to see the short bitcoin and stacks address displayed below the account name. 

Looks relevant: wallet 2 and 3 here are Ledger wallets.
When I added three software wallets with each 1 account I did see 1, 1, 1 and addresses for each
image

leah-stacks:
This looks fantastic. I do agree with Werner here:

> It is not very obvious that the feature is in "account selection". We a similar options from the top right menu because the user can simply click the account. Are we still happy with that or should we add a "manage accounts/wallets" option in the top right corner menu too?

I found it fairly intuitive to add the new wallet in account selection - however, I agree it would be clearer if there were a top-level menu item for managing wallets.

Otherwise, a small nitpick (not a blocker):

Screenshot 2026-06-03 at 2 13 25 PM

In this onboarding screen, there is no space under the bottom button, the extension window cuts it off in my browser. Still totally usable, just an optional polish thing

314159265359879:
Wallet numbering in titles isn't updated after adding a wallet, and then when adding new ones you can get the same number again. Go through this flow:
1. Create a wallet (wallet 1)
2. Add a wallet (wallet 2)
3. Add another wallet (wallet 3)
4. Now manage and remove wallet 2

You will see 
Wallet 1
Wallet 3

image


5. Now add another wallet

You will see
Wallet 1
Wallet 3 (retained old number)
Wallet 3 (just created)

edgar-stacks:
Thank you @314159265359879 @leah-stacks for reviews! I'm going to start addressing the bugs today.

A few notes:

> It is not very obvious that the feature is in "account selection". We a similar options from the top right menu because the user can simply click the account. Are we still happy with that or should we add a "manage accounts/wallets" option in the top right corner menu too?
The benefit of this would be that it is in line with expectations users have based on older versions of the wallet and it is easier (less ambiguity) to refer to in our guide.

> When I click "select account" in the (bitcoin or stacks) send flow, the accounts are all bunched up instead of split between wallets.

I agree with both, let's raise these with design to discuss further.

> For the third wallet it is a little bit stranger. You see with wallet 1 and wallet 2 it does start counting 1, 2, 3 etc in the wallet avatars. With the third account it looks like they belong to wallet 2 (in this example wallet 2 and 3 are both ledger wallets, both networks are connected but no addresses are displayed or filled when selecting.
Balances aren't displayed correctly either, looks like wallet can't get the info to display here.

yea i think that whole thing needs rework, missed during the multiwallet implementation.

> You will see
Wallet 1
Wallet 3 (retained old number)
Wallet 3 (just created)

iirc that is the behavior that we currently have in mobile as well. When creating a wallet we assign name to it right away, i'm not sure renaming wallets (and filtering which ones are autogenerated names and which ones were already updated by a user) once one is deleted is going to be trivial to accomplish here 🤔 Maybe we could add a page after adding a wallet that would prompt user to add name to the wallet?

314159265359879:
> Maybe we could add a page after adding a wallet that would prompt user to add name to the wallet?
Prompting for the wallet name I think is an elegant solution here, we can still prefill "Wallet 3" allow users to edit this. 

When connecting to a dapp I would expect to see the wallet name too:

image

aperry-stacks:
Looks great @edgar-stacks ! Few issues / nits / questions:
* BNS names become unrecoverable once an account gets renamed. Might need to persist a flag per account to prefer BNS name (+ checkbox in account rename dialog?).
* What's the plan for avatar support? Would argue for adding for a few reasons - 1) The multi-wallet illustration shows them so would make consistent there; 2) greater consistency with mobile; 3) just general visual superiority to existing random gradients
* Having collapse / expand support could be really useful here for wallets with many accounts:
image

* The stack of vertical kebab icons is a bit hard to parse visually. Horizontal kebab might work better here?
image

* Should be key icon:
image

* Should we move towards using these framed / bordered icons in the acocunt/wallet settings dropdowns? IMO have a much nicer / more premium feel:
image

314159265359879:
When I restart the computer and chrome with this build I see this message after restarting the browser
image

The extension appears to be removed
image

The first time I reinstalled the wallet (same version of the dev wallet build from source) I was asked for the password and I can unlock it, seeing the state saved. The second time around no saved state was displayed and I was presented with the onboarding page. 

I have never seen this chrome error with a dev build before not sure if it is related to this build in particular. There is also no data why it was blocked by chrome.

fab-stacks:
Very happy to see this becoming real! Here's the design review:

## Onboarding / Feature Announcement

Force line-break after Wallet so we have 2 lines (It's what we should strive for whenever possible for headlines with this font)

**Intro animation (optional, low priority)**
Might provide a lottie animation for the visual, depending on the public release timeline (let's discuss)

---

## Wallet List & Manage Mode

- Use the horizontal "..." icon
- Align the "..." and close (X) on one axis
CleanShot 2026-06-05 at 13 07 59@2x

- Hide the balance in Manage mode? Mostly aesthetic, but it reads cleaner without it
- Move "Add account" into Manage mode. Adding accounts is part of the management workflow. Only allow adding accounts inside Manage mode, not in the default view. I'd go as far as removing "Add account" from the default view unless there are strong reasons speaking against this cc @314159265359879 😅

---

## Add Wallet Screen

- Give the top area the same background as the graphic. Looks nicer.
- Add spacing between the graphic and the buttons: No gap currently between the graphic and the hover buttons below it.
- Consistent hover containers: Hover effects are inconsistent: Here they go edge-to-edge, elsewhere the list items (+ body) are containerized and the hover effects has some gap / space to the outer container. I actually think if we can and should do the hover effects everywhere in the account picker like here :) (It'll also help with the spacing issue we have regarding "..." and "x" in the top bar)

---

## Rename Flow

More breathing room via an inner container: Match the design → wrap the content in an inner container so there's consistent padding around it / to the top bar.

---

## Header

Center the close (X) in the title bar: The X is misaligned relative to the title label. Fix the title bar so the X is vertically centered against the label. Seems to be a global issue

---

## Popover Button Behavior

Treat action buttons as a footer in extension popover: Buttons like options in the rename popover or "try it out" in the onboarding screen should behave like footer buttons.

---

## Other

## Additional entry point to "Manage wallets"

To make wallet management more discoverable, how do we feel about adding an additional menu item to the settings dropdown, which then launches the manage wallet mode of the account picker right away? 

CleanShot 2026-06-05 at 11 56 27@2x

**Account selection state (post-MVP?)**
The selected-account state from the design isn't reflected, and the current selection is hard to spot. Nice-to-have, can ship as a follow-up.

leah-stacks (dismissed):
Works great for me, no issues with testing adding existing wallet, creating new wallet, switching between wallets 🎉

314159265359879 (approved):
I tested with hardware wallet because Leah hadn't gotten around to that yet. Adding wallet, creating new wallet, switching between wallets, connecting on a dapp, and signing a transaction on a dapp. Sending STX and Bitcoin from one wallet account to another wallet+account. Worked smoothly. 

I ran into one non-reproduceable issue twice that showed this error: TransportError: Invalid channel
Adding a hardware wallet failed to initialize account lookup (started with Stacks connection) I saw this error on two occasions (the first two wallets for the multiwallet I set up), then when I hit "try again" it is fine and adds the wallet/accounts right away.
I tried some things to reproduce but that failed. When I added a software wallet, I didn't see it, and I didn't see it when I added a fourth hardware wallet either. Removing and adding one didn't result in the error and neither did resetting/reinstalling the wallet and restoring it the same way. It doesn't look reproduceable and it easy to work around so you can disregard this for now.

image

edgar-stacks:
@314159265359879 i think i've noticed that issue on a dev branch as well, looking into it

fab-stacks (approved):
Lfg 🚀 As discussed I'll do a follow-up w/ minor visual tweaks tomorrow

aperry-stacks (approved):
LGTM 🚀

PR #2357, opened 2026-05-27
fix(web): fastpool v2 ledger stacking

edgar-stacks:
Make btc rewards address optional for stx-only pools


314159265359879 (approved):
LGTM, I can now sign with Leather/Ledger even when only the Stacks network is connected. 

Thanks for the quick fix lets merge.

leah-stacks (approved):

adriano-stacks (approved):

PR #2358, opened 2026-05-27
fix: remove runes/ordinals pages from services/models

edgar-stacks:
Remove runes/ordinals from services and models packages, also remove a little more code from other apps/packages as the interfaces changed


edgar-stacks:
redux state is going to be handled in a separate PR to handle potential migrations

leah-stacks (approved):

aperry-stacks (approved):

PR #2360, opened 2026-05-28
fix(ui): truncate long token titles in item layout with ellipsis

leah-stacks:
Long token names in the asset list (e.g. "Ethereum USDC via Allbridge") were overflowing their column and pushing the fiat amount off-screen. This change makes the title display truncated with an ellipsis and the right-hand price stay put.

ItemLayout and Flag are used across the extension (asset list, activity list, account lists, PSBT rows, fee editor, etc.). The flex changes only have visible effect when content would otherwise overflow its column. For the vast majority of rows in the standard popup width, nothing changes visually.

Before: 
Screenshot 2026-05-28 at 10 05 12 AM

After:
Screenshot 2026-05-28 at 9 47 25 AM


aperry-stacks (approved):
LGTM 🚀

andresgalante-stacks (approved):

PR #2363, opened 2026-06-01
chore(multisig): scaffold proposed auth shape for review

leah-stacks:
Proposal scaffold — not for merge. Visualizes the V2 multisig auth shape against the spec so the Services owner can review the contract before implementing.

Models:
- ChainNetworkId / AuthIdentity / AuthSession in @leather.io/models

Services:
- V2 TokenAuthService interface (per-network methods)
- WalletSignInPayload + SignInInput transport types
- AuthService.signIn stub (returns rejected Promise with TODO)
- LeatherAuthApiClient adapted to V2; network bound at construction with stx:mainnet default to preserve V1 behavior
- getAuthService() resolver

Dapp:
- WebTokenAuthService rewritten as a Jotai-backed V2 implementation
- sessionsAtom (atomWithStorage — covers session bootstrap)
- useSession / useSignIn / useSignOut hooks under features/multisig/auth

Knip flags the three hook files as unused — expected; no UI consumes them yet. Pre-existing typecheck failures in @leather.io/utils (runes/ ordinals migration leftover) are present on plain dev, unrelated.


packages/models/src/auth.model.ts
```diff
@@ -0,0 +1,12 @@
+export type ChainNetworkId = `${'stx' | 'btc'}:${'mainnet' | 'testnet'}`;
+
+export interface AuthIdentity {
+  address: string;
+  publicKey: string;
```
aperry-stacks, line 8: We should add `network` here.

packages/services/src/infrastructure/token-auth.service.ts
```diff
@@ -1,6 +1,12 @@
+import type { AuthSession, ChainNetworkId } from '@leather.io/models';
+
 export interface TokenAuthService {
```
aperry-stacks: Now that the type is more formally defined as `AuthSession`, we could rename this to `AuthSessionService` as it's the mechanism by which the session information is provided to services.

packages/services/src/infrastructure/token-auth.service.ts
```diff
@@ -1,6 +1,12 @@
...
-  getAccessToken(): string | null;
-  getRefreshToken(): string | null;
-  onTokenRefreshed(accessToken: string): void;
-  onAuthFailure(): void;
+  getSession(network: ChainNetworkId): AuthSession | null;
+  getActiveNetworks(): ChainNetworkId[];
+
+  setSession(network: ChainNetworkId, session: AuthSession): void;
+  clearSession(network: ChainNetworkId): void;
```
aperry-stacks: We should drop `setSession`/`clearSession`. The services layer should read state but not write it, meaning session persistence stays entirely an app-side concern. Services execute sign-in exchange and refresh/failure mechanics but are unopinionated about how / when any of that gets persisted.

E.g. Apps can:
- call `signIn()`, return the session and store it.
- Listen to `onTokenRefreshed` / `onAuthFailure` for session-related events, and own the reaction (update the token or clear session).

packages/services/src/multisig/auth.service.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import type { AuthSession } from '@leather.io/models';
+
+import { LeatherAuthApiClient } from '../infrastructure/api/leather/leather-auth-api.client';
+import type { TokenAuthService } from '../infrastructure/token-auth.service';
+import { Types } from '../inversify.types';
+import type { SignInInput } from './auth.types';
+
+@injectable()
+export class AuthService {
```
aperry-stacks: few thoughts here:
1) Could this be named more specifically - Maybe even simply `SignInService`?
2) Our auth infra should stay domain-agnostic, so suggesting we move out of multisig. I wouldn't say this is technically infrastructure, but we could co-locate in `/infrastructure/auth` alongside the `AuthSessionService` to drive home that these are related and cross-domain.
3) I like the idea of the `walletSignIn` callback keeping message construction centralized, but I think it mixes too many app-level concerns into services. If possible we should keep the wallet interaction external and just pass the payload. For message construction we should make sure that goes into a non-app-specific auth utility.
4) We probably can get rid of the token/session service injection (see above message about session persistence updates).

packages/services/src/infrastructure/api/leather/leather-auth-api.client.ts
aperry-stacks, line 1: This needs to be reworked quite to move away from the network-specific clients concept. Instead of persisting a network on the service instance, we can just accept network as part of a client request (e.g. `fetchVaults(network: NetworkId, ...)` and then just look-up the matching token dynamically from the `AuthSessionService` registry.

I can do the rework here when updating the Leather API types.

packages/services/src/infrastructure/auth/auth.types.ts
```diff
@@ -0,0 +1,15 @@
...
+  publicKey: string;
+  address: string;
+  message: string;
+  timestamp: number;
+  xpub?: string;
+  xpubOriginFingerprint?: string;
+  xpubOriginPath?: string;
+}
+
+export interface SignInInput {
```
aperry-stacks, line 14: We need to add a new "application" param here. Auth is domain agnostic but adding this hint here allows some backend to make use of the sign-in payload for some conditional domain-specific side-effects (like registering a multisig user). This is needed b/c we only have access to things like `xpub` here at this moment.

``` typescript
export interface SignInInput {
  application: AuthApplication[],
  ...
}

// auth.model.ts
const authApplications = ['multisig', ...];
export type AuthApplication = (typeof authApplications)[number];
```

packages/services/src/multisig/auth.service.ts
leah-stacks: Yeah I like "Sign In" better than Auth, it better distinguishes it from the Connect flow.

PR #2365, opened 2026-06-03
feat(web): Leather Multisig UI (initial import, behind multisigEnabled flag)

fab-stacks:
## Leather Multisig UI (initial import)

This is the **initial import of the Leather Multisig UI** into the web app, prototyped and built with Claude Code. It's behind the `multisigEnabled` feature flag, so it's fully isolated from the rest of the app.

The goal of this PR is to get the **first engineer eyes on it and give us real UI to work with**. I've ironed out the biggest issues from the raw import, but this is very much a work in progress: there are still **design adjustments and plenty of follow-up work to be done.** Treat it as a foundation to iterate on rather than a final state.

### Prototype / dev tools menu

There's a **floating dev tools panel in the bottom right** — a little prototype menu from which you can **switch between the different app states** (empty, loading, populated, etc.) to preview the UI without real data.

A couple of notes on it:
- How long we keep it around is **TBD** — it depends on how we end up wiring real data and whether we still need a manual state switcher.
- It's also a natural place to **expand preview functionality** over time (more states, more scenarios, toggles for edge cases) if we find it useful during development.

### Scope

- New multisig UI lives under `apps/web`, gated by `multisigEnabled`.
- Some shared bits touched already in `packages/ui` (badge component) to support the new UI.

### Status

🚧 **Draft / WIP** — design and implementation still in flux. Opening early for visibility and feedback.


leah-stacks:
This looks really awesome. After testing the UI and thinking on it a bit, I'm in favor of merging this into the codebase as a starting point, what do you think?

Just a small CI issue that I can help with fixing - I can add a commit right onto this branch to help with that

leah-stacks (approved):
In favor of merging this as a starting point for the webapp UI work!

edgar-stacks (approved):

PR #2368, opened 2026-06-05
feat: auth infrastructure

aperry-stacks:
PR adds full domain/network-agnostic signature-based JWT session authentication infrastructure (auth models, `leather-auth-api.client.ts`, `sign-in.service.ts`) and wires to the Multisig dapp via `/web`.

Folds in and builds on work from @leah-stacks on https://github.com/leather-io/mono/pull/2363

Keeps the web app sign-in hooks but temporarily adds to `knip.json` ignore list assuming a next-up full sign-in flow integration in the Multisig dapp.


leah-stacks (approved):
FYI - the full sign-in flow integration is in progress on branch `feat/multisig-auth` (currently stacked on Fab's `feat/multisig-ui`). That branch has the onboarding UI wired to these hooks.

I think we should move forward with merging this first and then I'll plan to rebase `feat/multisig-auth` on top (drop my scaffolding and use your services layer, adapt to AuthNetworkId, carry over the UI wiring + bootstrap, and remove the walletSignIn stub and knip ignores in this PR)

apps/web/app/features/multisig/auth/sessions.atom.ts
```diff
@@ -0,0 +1,17 @@
...
+type SessionsRecord = Record<AuthNetworkId, AuthSession | null>;
+
+const defaultSessions: SessionsRecord = {
+  'stx:mainnet': null,
+  'btc:mainnet': null,
+  'stx:testnet': null,
+  'btc:testnet': null,
+};
+
+export const sessionsAtom = atomWithStorage<SessionsRecord>(
```
leah-stacks, line 14: (This can be added later in my UI branch) but fyi this needs `getOnInit: true`, otherwise the first client render reads empty sessions even when records exist.

aperry-stacks: Just to confirm, this is okay to merge as-is and you'll add the `getOnInit: true` in the `feat/multisig-auth` follow-up?

edgar-stacks (approved):

PR #2372, opened 2026-06-08
feat(web): wire multisig auth sign-in flow and session bootstrap (ref LEA-3577)

leah-stacks:
Wires the multisig dapp's per-chain sign-in flow and session bootstrap on top of the auth infrastructure from #2368 and the UI prototype from #2365 (both now on dev). Gated behind `multisigEnabled` — hidden in production, visible on dev/staging/previews.

## What this does

- **Per-chain sign-in** — "Connect Bitcoin" / "Connect Stacks" drive an independent one-step flow: read addresses (reusing the shared store, no redundant prompt) → sign the timestamped ownership message (`stx_signMessage` / `signMessage`) → `POST /v1/auth` → persist the session. Partial auth (one chain signed in) is a steady state per spec §2.3.
- **Network-bound sign-in message** — the signed message is structured and binds the network: `Sign in to Leather` / `Network: {mainnet|testnet}` / `Issued: {timestamp}`, matching the backend's reconstruction.
- **`walletSignIn` seam** (`wallet-sign-in.ts`) — the interim two-step `getAddresses` + `signMessage` path the spec sanctions until the extension ships `wallet_signIn`. Isolated behind one function so the future swap is a one-liner.
- **Session bootstrap** (§3.5) — on entering `/multisig/*`, expired access tokens are silently refreshed against `/v1/auth/refresh`; dead sessions are dropped so "signed in" is never a lie. Per-chain "Restoring…" state while in flight, guarded against mid-refresh session changes.
- **Onboarding rewired** to real session state (signed-in address, sign-out, pending/error/restoring states); `/multisig` redirects to onboarding when no session exists.
- **Services**: `refreshSession` on `SignInService`; `getJwtExpiry`/`isJwtExpired` folded into `sign-in.utils`, with tests.

## Status

- **STX sign-in works end-to-end against staging** — connect → wallet signature → JWT → session → dashboard, verified.
- **BTC sign-in is blocked upstream** on `xpubOriginFingerprint` at connect (the backend BTC `/v1/auth` branch requires the xpub triple; the two-step seam can't supply the fingerprint). Needs an extension `getAddresses` enhancement. BTC fails gracefully (in-row error) until then. STX is unaffected.

Safe to merge behind the flag — production never renders this; staging/previews exercise the working STX path.

## Testing

1. `/multisig` with no session → redirects to `/multisig/onboarding`
2. Connect Stacks → single `stx_signMessage` prompt → `/v1/auth` (staging) → "Connected to Stacks", redirect to dashboard
3. Reload with a stored session → no false bounce to onboarding (`getOnInit`)
4. Sign out → returns to Connect

## Also in this PR

- `fix: audit shell-quote` — a critical advisory (`shell-quote` ≤1.8.3) was published against a transitive dep, failing `audit-dependencies` dev-wide. Added the standard override (`shell-quote@<1.8.4` → `>=1.8.4`); the fix lands in dev on merge and clears the failure for other branches too.

## Verification

`pnpm format`, `lint`, `typecheck`, `knip`, `lint:unused-exports` all green; 331 services tests pass (incl. JWT-expiry + structured-message specs).


leah-stacks:
1. restoreSession race — fixed, Applied the token-identity guard to both the success and catch paths — the success path had the same latent issue (its if (!prev[network]) guard only caught sign-out mid-refresh, not a fresh sign-in), so both now bail if the stored token no longer matches the one being refreshed.

2. Testnet ignored in onboarding/redirect — intentional for now. V1 is mainnet-pinned (FE spec §2.2; testnet is dev-only via manual localStorage editing), and the proper fix — threading an active network through — is the §2.2 network-setting generalization that's explicitly out of V1 scope. Keeping the testnet keys in sessionNetworks is forward-looking; the only current effect is bootstrap refreshing a testnet token the UI doesn't surface, which is harmless. Will revisit when the network setting is generalized.
  
3. btcSignIn pubkey/address from different sources — deferring. This whole branch is the interim two-step seam that gets replaced once wallet_signIn (or the getAddresses keyOrigin enhancement) lands — BTC sign-in is blocked on xpubOriginFingerprint until then, so the branch isn't exercised. Will be reworked wholesale at that point rather than patched here.

edgar-stacks (dismissed):

apps/web/app/features/multisig/auth/use-sign-in.ts
```diff
@@ -28,7 +14,7 @@ export function useSignIn(network: AuthNetworkId, application: AuthApplication[]
     mutationKey: ['multisig-auth-sign-in', network, application],
     mutationFn: async () => {
       const { message, timestamp } = buildSignInMessage();
```
aperry-stacks: Can we pass the network and use it in the message like this? Gives the message a bit more structure and binds network to the signature. I'll update this on the backend and get it deployed to staging so you can test against.
```typescript
export function buildSignInMessage(network: AuthNetworkId): SignInMessage {
  const timestamp = Math.floor(Date.now() / 1000);
  const networkMode = network.split(':')[1];
  return {
    message: ['Sign in to Leather', `Network: ${networkMode}`, `Issued: ${timestamp}`].join('\n'),
    timestamp,
  };
}
```

Image

leah-stacks: Thanks, just pushed this change and confirmed it works

aperry-stacks (dismissed):
LGTM nice work! glad to see e2e auth finally working 🦾

edgar-stacks (dismissed):

aperry-stacks (dismissed):

aperry-stacks (approved):

edgar-stacks (approved):

PR #2374, opened 2026-06-09
feat: implement wsh signing in signPsbt

edgar-stacks:
This PR implements signing of custom wsh descriptors on mainnet and custom networks, using software and ledger wallets


apps/extension/src/app/features/psbt-signer/hooks/use-descriptor-psbt-details.tsx
```diff
@@ -0,0 +1,88 @@
...
+        policyAddress,
+        destinations: outputs
+          .filter(output => output.address !== policyAddress)
+          .map(output => ({ address: output.address, value: createMoney(output.value, 'BTC') })),
+        amountLeavingPolicy: createMoney(Math.max(0, policyInputTotal - changeTotal), 'BTC'),
+        fee: createMoney(Math.max(0, allInputsTotal - allOutputsTotal), 'BTC'),
+        hasDisallowedSighash,
+      };
+    } catch {
+      return null;
```
leah-stacks, line 85: Is it worth a log here so the error can be viewed if needed? I get that we want to fall back if the threshold is not met so the path makes sense

leah-stacks (approved):

aperry-stacks (commented):
Great work! This is legendary.

One issue I am seeing: During signing / key match it looks like we assume a pubkey at address 0, but for Multisig that often won't be the case.

In a Multisig vault, the Leather account xpub is imported (`m/84/0/n`) and each vault account is given a different fixed address index `i` (`m/84/0/n/0/i`) and so we'll need the ability to match and sign at arbitrary indices.

The supplied descriptor itself shouldn't be ranged, but it can def be a non-zero index: e.g. `wsh(sortedmulti(k, xpubA/0/i, xpubB/0/i, ...))`

apps/extension/src/app/features/psbt-signer/hooks/use-descriptor-psbt-details.tsx
edgar-stacks: yeap good point! I'll add that in a follow-up pr

edgar-stacks:
@alexp3y nice catch, thank you! Bitcoin staking only needs index 0, so we can merge this one to unblock them and then  I can add that capability in the follow up PR, wdyt?

aperry-stacks (approved):

PR #2375, opened 2026-06-09
fix: audit shell-quote

leah-stacks:


aperry-stacks (approved):

aperry-stacks:
@leah-stacks reopened this so we can get it merged, but let me know if there is a reason it should not be.

edgar-stacks (approved):

PR #2376, opened 2026-06-09
feat(web): multisig per-chain connection-status dropdown

leah-stacks:
Adds a multisig-only connection-status dropdown to the page header, showing per-chain sign-in (JWT session) status (shows on the Multisig section) — distinct from the global wallet-connection dropdown (shows everywhere else).

Screenshot 2026-06-09 at 2 35 26 PM

## What this does

- New `MultisigConnectDropdown` driven entirely by the per-chain `sessionsAtom` (the multisig JWT sessions), independent of the global `useLeatherConnect`. "Disconnect Stacks/Bitcoin" drops that chain's session; it does not disconnect the wallet.
- Swapped into the header via a small `SignInSlotProvider` context: `multisig.layout` provides the multisig dropdown, `PageHeader` reads it and falls back to the global `SignInButton` everywhere else. No route-sniffing, and every current/future `/multisig/*` page gets it automatically.

## States

- **Neither signed in** → "Connect wallet" pill → "Choose a chain to connect" with Bitcoin / Stacks rows.
- **One chain** → pill with that chain's status dot + address → signed-in row (address + copy + Disconnect) and a "Sign in to {other}" row + Open extension.
- **Both** → pill with stacked dots (orange = Stacks, yellow = Bitcoin) → both signed-in rows with Disconnect + Open extension.

## Open question — BNS names, priority for V1?

The dropdown shows truncated addresses, not BNS names (e.g. `carey.btc`), as shown in the designs. Reverse lookup is feasible — `BnsService.getAccountPrimaryBnsProfile` exists and resolves in the web container (only needs `CacheService`, already provided) — but wiring it (AccountRequest/UserSettings construction, async loading + address fallback, STX-only) is a medium task. Is BNS-name display a priority for multisig V1, or is address-only fine to ship?

## Note for reviewers

Stacked on `feat/multisig-auth-wiring` (#2372) — this PR targets that branch so the diff is just the dropdown. Retarget to `dev` once #2372 merges.

## Verification

`pnpm format`, `lint`, `typecheck`, `knip`, `lint:unused-exports` all green.


aperry-stacks:
Nice! Looks great.

Couple of small tweaks: 
* When logged-in to BTC/STX simultaneously you get a double-bottem border line.
* Left padding looks a bit small compared to y-padding
* Content is confined to left 50% of the dropdown box, seems like it could extend a bit further
image

apps/web/app/pages/multisig/components/multisig-connect-dropdown.tsx
```diff
@@ -0,0 +1,285 @@
...
+        fill="currentColor"
+      />
+    </svg>
+  );
+}
```
edgar-stacks: let's use packages/ui for svg icons

edgar-stacks (commented):
there are multiple components in one apps/web/app/pages/multisig/components/multisig-connect-dropdown.tsx file and that makes it hard on readability. Can you please separate those into multiple files?

apps/web/app/pages/multisig/components/multisig-connect-dropdown.tsx
leah-stacks: oops yes, thanks for the reminder! done!

leah-stacks:
Screenshot 2026-06-10 at 12 47 53 PM

Thanks for the UI feedback @aperry-stacks - I decided to display more of the addresses since width allowed, what do you think?

Also what you (and @edgar-stacks) think about the BNS question - the designs show BNS name instead of just the "Stacks" label, but I did not include that in this PR - do you think it's a priority for v1?

fab-stacks:
@leah-stacks figured that the menu's design didn't get a lot of attention during the claude design → codebase migration, so figured it would be easier if I open a PR (#2380) based on your work here containing the tweaks & improvements  - lmk what you think, hope that that works!

aperry-stacks (commented):
Seeing a couple issues with the `addresses` cached in localStorage. 

1) If you sign-out it doesn't clear, locking you to a specific account.
2) The single addresses value for all sign-ins bind all chains / networks to those addresses. So, if I sign-in to BTC on account 1, then change my wallet to account 2 and try to sign-in to STX there it fails.

Wondering do we need to store addresses? I'm guessing yes for identity info (you can also retrieve identity information on demand the backend via `multisigService.getMe()`) but maybe they should be cached by `AuthNetworkId`?

leah-stacks:
@aperry-stacks Oh, good catch, ty. I'm thinking multisig sign-in needs to be decoupled from the persisted addressesAtom entirely, and fetched fresh at sign time so it always matches the active account. I don't think we need to store addresses for multisig - multisigService.getMe() can be the source of truth here. Will make those changes next

edgar-stacks (approved):

aperry-stacks (approved):

PR #2377, opened 2026-06-10
feat(web): populate btc sign-in xpub fields

aperry-stacks:
Enables BTC sign-in. Dummy fingerprint can be replaced once updated RPC support is ready.


aperry-stacks:
Feedback item 2 from Claude is valid. We'll have to enhance this utility once we enable testnet in order to support tpubs.

edgar-stacks:
@aperry-stacks i've addressed that in my PR

edgar-stacks (approved):

PR #2380, opened 2026-06-11
feat(web): polish multisig connect dropdown visuals

fab-stacks:
CleanShot 2026-06-11 at 11 21 39

Design pass on the multisig connection dropdown from #2376, shared as changes rather than review comments. Rebased onto the latest `feat/multisig-connection-status` — it builds on top of the `AccountSection` refactor, the divider fix, the "Restoring…" state, and the longer `truncateMiddle` addresses, and re-applies the visual system below onto that structure.
## Visual tweaks

**Trigger button (connected states)**
- The colored chain dots are replaced with the actual token icons: both chains always shown as 20px avatars inset in 24×24 containers (safe space like regular icons), overlapping by 12px, Bitcoin stacked on top of Stacks
- Connected chains render in full color; disconnected ones in grayscale — so the button itself communicates per-chain connection status
- The logged-out state is unchanged: solid "Connect wallet" CTA with the wallet icon
- Note: this removes the only consumer of the new `chain.btc` / `chain.stx` color tokens — left the tokens in place in case they're wanted for upcoming chain UI, but they can be dropped from this stack if not

**Open extension icon**
- Replaced the `wallet-sparkle-20-20.svg` glyph (drew small inside its viewBox, rendered tiny at 16px) with the icon already used for "Open extension" in the account switcher popover, as a full-bleed 24×24 `currentColor` asset
- The account switcher (`sign-in-button.layout.tsx`) now uses the shared `WalletSparkleIcon` instead of its hardcoded inline SVG — one source of truth, and it picks up theming

**Icon sizing system**
- Token/chain avatars: 32×32 everywhere in the menus
- New `IconSlot` (32px wide, height hugs content): smaller icons (16px action glyphs) center on the same axis as the avatars, so all text labels share one left axis while icon sizes can vary
- Rows with small icons collapse to 40px; avatar rows 56px; the chain identity row uses `space.02` vertical padding to tighten the gap to the action below

**Menu scale — matched to the account switcher dropdown**
- Text styles `label.02` → `label.03` throughout
- Menu width 340px → 300px with the account switcher's `space.02` inset, so hover pills and dividers sit inside the menu edges
- Chain-picker rows: 40px avatars → 32px

**Address line (connected state)**
- The copy chip is now a 20px-tall flex row (caption line-height): the title→address spacing matches the title→caption rhythm of the picker rows, and the hover pill can't overlap the line above
- Tight hover padding (`space.01`) with matching negative offset so the address text stays on the text axis
- Copy/checkmark icons 16px → 12px

**Dividers**
- Removed the divider's horizontal margin so it spans exactly the hover-pill extent (the pressable pseudo-element outsets `space.03`, landing on the container width)


edgar-stacks (approved):

leah-stacks (approved):
This is great, thank you!

PR #2385, opened 2026-06-13
feat(web): multisig vaults UI (create, view, invites)

leah-stacks:
- Fully functional: Create and view vaults; send and accept invites
  - Create Vault form has validation + UI polish
  - Invitation acceptance modal has UI polish
- Vault list & details page: both wired with basic skeleton, no UI polish yet

## Gating (no production impact)
- `multisigEnabled` is `false` in production (`CLOUDFLARE_ENV === 'production'`)
- All `/multisig/*` routes 404 via the layout loader, and the nav entry is hidden
- Changes outside the multisig tree are small and low-risk: additive `orange` color tokens (unused elsewhere); a shared `ToastsProvider` now mounted at the app root (replaces the multisig-scoped provider — inert until a toast is triggered, and only multisig triggers them); a barrel export of `LeatherApiError` from `@leather.io/services`; and a `pnpm audit` ignore for a Deno-only esbuild advisory.

## Note
- Vault theme & vault member names aren't persisted server-side yet. TODO (discussed with @aperry-stacks)


leah-stacks:
Closes LEA-3579 and LEA-3581, partially LEA-3580

leah-stacks:
Claude review response:

1. Dashboard isResolvingVaults when no session. Correct in theory, but yeah unreachable: the layout redirects to /onboarding whenever neither session exists, so the dashboard never renders session-less.

2. member.address === me.data?.address (BTC case). Leaving as-is. We canonicalize BTC to lowercase at submit, and the wallet session address is canonical lowercase, so both sides are lowercase and strict equality holds.

3. Retry on AbortError. Leaving as-is. React Query short-circuits its retryer on cancellation, so a query aborted by navigation is not run through the predicate and won't retry. The failureCount < 3 default applies only to genuine non-cancellation errors (network blips, 5xx), which is the intended behavior.

edgar-stacks (dismissed):

edgar-stacks (dismissed):

aperry-stacks (dismissed):
Looking great 🦾 A few thoughts below. We could drill into some more design tweaks here but I'll leave that to @fabric-8 's eye. Also, happy to just push forward with functionality now and do a deeper design pass later.

Resource not found screens - I don't think we need to include the long resource GUID in the not found messages here.

When a user logs out of one side (BTC or STX) on a resource screen to which they lose access but they're still logged in with the other side account, should we route them back to the vaults home screen instead of showing them the resource 404?

Can we add cursor:pointer to button hover states?

We may want to add a filter dropdown to the vaults list so a user can see declined vaults (could also use to filter only invted/pending/active). We should probably build out a flow for users to recover invitations and join a vault that they've declined, possibly mistakenly. cc @fabric-8 

Invitation dialog:
- vault icon missing chain indicator
- member list doesn't show addresses / copy button

Vault screen:
- No themed header banner / balance yet?
- Account list should be before vault member list?
- Vault icon missing chain indicator
- Member list missing addresses / copy button

leah-stacks:
@aperry-stacks Thanks for testing and for the helpful feedback - yes, the vault list & details page still need UI work, they're just basic skeletons right now. I will proceed with the merge on this, and then work through those, as well as your other feedback items about the loading screens, invitation modal, etc. 👍

andresgalante-stacks (approved):

aperry-stacks (approved):

PR #2404, opened 2026-06-18
feat: multisig service transactions

aperry-stacks:
Adds Multisig transaction calls to `multisig.service.ts`:
* `listVaultAccountTransactions()`
* `getTransaction()`
* `addTransactionSignatures()`
* `cancelTransaction()`
* `broadcastTransaction()`
* `proposeTransaction()`

Also, implements the shared machinery needed to successfully propose a transaction using "sig-as-auth" against the `/v1/multisig-ext/propose` endpoint:
* `@leather.io/utils`:
  * `computeProposalHash()`
* `web`:
  *  `walletPropose()` -> constructs proposal hash and triggers extension signature prompts
  * `useProposeTransaction()`

Also adds a small utility to the multisig dev tools to test transaction proposals for a requested address pre-UI implementation.

### DIFF
--- a/packages/crypto/src/multisig/proposal-hash.ts
+++ b/packages/crypto/src/multisig/proposal-hash.ts
@@ -1,0 +1,51 @@
+import { sha256 } from '@noble/hashes/sha256';
+import { bytesToHex, concatBytes, utf8ToBytes } from '@noble/hashes/utils';
+import { base64, hex } from '@scure/base';
+
+export const multisigProposalDomainTag = 'leather-multisig-proposal-v1';
+
+export type ProposalChain = 'btc' | 'stx';
+
+// Decodes the transport-encoded `rawPayload` into the bytes the proposal hash
+// commits to. The encoding is per-chain and part of the protocol — BTC PSBTs are
+// base64, STX raw transactions are hex — and both ends MUST decode identically
+// before hashing (the length prefix is over the byte length, not the string).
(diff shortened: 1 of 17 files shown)


packages/crypto/src/multisig/proposal-hash.ts
```diff
@@ -0,0 +1,51 @@
...
+  proposalTimestamp,
+}: ComputeProposalHashArgs): string {
+  const preimage = concatBytes(
+    lengthPrefixed(utf8ToBytes(multisigProposalDomainTag)),
+    lengthPrefixed(utf8ToBytes(multisigAddress)),
+    lengthPrefixed(rawPayload),
+    u64BigEndian(proposalTimestamp)
+  );
+  return bytesToHex(sha256(preimage));
+}
```
edgar-stacks, line 51: nice! i'll reuse this one for extension

edgar-stacks (dismissed):

packages/crypto/src/multisig/proposal-hash.ts
edgar-stacks: is this stacks specific? Or this works for both chains?

fab-stacks (dismissed):

packages/crypto/src/multisig/proposal-hash.ts
aperry-stacks: both chains.

fab-stacks (approved):

leah-stacks (approved):

PR #2411, opened 2026-06-19
fix(extension): multi-wallet account picker, header & logo visual polish

fab-stacks:
Visual polish for the extension's multi-wallet account picker (the surfaces from #2356), plus a couple of adjacent header/logo fixes spotted while testing. All verified live in the extension popup.

## Changes

**Popover side gaps (LEA-3591)** — `packages/ui/.../sheet/sheet.web.tsx`
At the popup (`base`) breakpoint the sheet used `width: 100vw; left: 50%`, which resolved ~15px narrow and centered, exposing the dark overlay as ~8px gutters each side. Pinned the content flush to both edges (`left: 0; right: 0; width: auto`) and dropped the horizontal translate at `base`; the `md` floating-dialog behavior is unchanged. Fixes all extension sheets at popup size, not just these two.

**Account picker mode stability (LEA-3589)** — `switch-account-sheet.tsx`, `wallet-header.tsx`
Default and manage rows now share identical geometry: the account row is full-width in both modes and the manage `•••` menu is an absolute overlay (so toggling never reshapes the row), and the row's hover/selection background fills the full width again. The wallet header has a stable `min-height` in both modes so introducing the `•••` button no longer makes the header grow taller and shove content down.

**Sticky wallet header (LEA-3594)** — `switch-account-sheet.tsx`
Made the sticky group-header top padding uniform so every pinned wallet label lands at the same offset (previously the first wallet pinned ~4px from the top vs ~24px for the rest, reading as "wallet B sits lower than A"). The full iOS-style **push-out** (incoming header displacing the pinned one) is a structural react-virtuoso change tracked as a follow-up: **[LEA-3595](https://linear.app/stackslabs/issue/LEA-3595)**.

**Adjacent fixes spotted while testing** (pre-existing, from #2356):
- **Sheet close button** — `packages/ui/.../sheet/sheet-header.web.tsx`: was rectangular (default Button height ≠ width); set `height: headerContainerHeight` so it's a 40×40 square, matching the home header buttons.
- **Header full-width + settings alignment** — `headers/header.tsx`: the header's `maxWidth: 100vw` resolved ~15px short, so the settings gear didn't align with the content edge; switched to `100%` so the header spans full width and the gear lands on the content's right edge.
- **Logo left-alignment** — `headers/logo-box.tsx`: `margin: auto` was centering the logo within the stretched header column (~40px right of the content edge); changed to `my: auto` (vertical only) so it left-aligns with the page content like the heading/input.


docs/plans/2026-06-19-001-fix-extension-multiwallet-visual-polish-plan.md
```diff
@@ -0,0 +1,268 @@
+---
```
edgar-stacks: probably no need for this plan file, let's remove it

edgar-stacks (commented):
Nice work! Let's just remove the plan file and we should be good to go 🚀

docs/plans/2026-06-19-001-fix-extension-multiwallet-visual-polish-plan.md
fab-stacks: Done — removed the plan file (dropped the docs commit, force-pushed). Thanks! 🚀

edgar-stacks (approved):

aperry-stacks (approved):

PR #2413, opened 2026-06-19
refactor(extension): redesign Add Wallet as a compact popover menu

fab-stacks:
The Add Wallet view was a full sheet with a big illustration on top — way too heavy for what's really just a three-option picker. This swaps it for a compact popover that hangs off the **Add wallet** button, same pattern as settings. The illustration is gone, and the three options (create / restore / connect Ledger) stay put and wire to the exact same flows.

Under the hood it reuses the dropdown we already run *inside* the account switcher (`account-action-menu`), so it plays nicely with the sheet behind it — opens upward off the button, closes on select, no weirdness with focus or dismissal.

**What changed**
- New `add-wallet-menu` (compact `DropdownMenu`) replaces `add-wallet-sheet`
- Switcher footer renders the menu in place of the old sheet + its open-state plumbing
- E2E page-object retargeted to the menu (labels kept identical, so the helper chain still works)

**Demo**

CleanShot 2026-06-19 at 15 10 35


fab-stacks:
Renamed `openAddWalletSheet` → `openAddWalletMenu` across the page-object and the switch-account / multiwallet-integrity specs, since it opens a menu now, not a sheet. Also fixed the underlying e2e break

edgar-stacks (approved):

aperry-stacks (approved):

PR #2415, opened 2026-06-22
feat: multisig fixed address account

aperry-stacks:
Multisig vault accounts use a single fixed on-chain address, not the HD descriptor structure the services layer assumes. This PR expands the account model and data services to represent and serve them, so the Multisig dapp can pull balances/activity for a vault account through existing `@leather.io/services`.

* Models: `account.model.ts` converted from zod to interfaces; `BitcoinAddressInfo` is now a discriminated union — `'hd'` (descriptors) vs `'fixedAddress'` (`{ address, paymentType: 'p2wsh' }`).
* Services: new `fetchUtxosByAddress` / `fetchBitcoinTransactionsByAddress` client methods + cache keys; `UtxosService` and `BitcoinTransactionsService` branch on `bitcoin.type` (address fetch vs descriptor fan-out). Balances/activity ride this unchanged.
* Dapp: `useMultisigAccountAddresses` maps a `VaultAccount` → services `AccountAddresses` (synthesized `AccountId` from the vault account id); `useVaultAccountBalance` wired through per-chain balance configs, demonstrated on the account detail page.
* Migration: all `AccountAddresses.bitcoin` construct/read sites updated for the union (models, services, queries, extension, mobile, web).
Screenshot 2026-06-22 at 12 33 31 PM

Screenshot 2026-06-22 at 12 33 42 PM

### DIFF
--- a/packages/utils/src/accounts/account-addresses.ts
+++ b/packages/utils/src/accounts/account-addresses.ts
@@ -2,6 +2,7 @@ import {
   AccountAddresses,
   AccountId,
   BitcoinAddressInfo,
+  HdBitcoinAddressInfo,
   StacksAddressInfo,
 } from '@leather.io/models';
 
@@ -15,6 +16,7 @@ export function createAccountAddresses(
   const nativeSegwitDescriptor = btcDescriptors.find(desc => desc.startsWith('wpkh('));
   if (taprootDescriptor && nativeSegwitDescriptor) {
     accountAddresses.bitcoin = {
+      type: 'hd',
       taprootDescriptor,
       nativeSegwitDescriptor,
     };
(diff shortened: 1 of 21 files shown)


packages/utils/src/accounts/account-addresses.ts
```diff
@@ -31,6 +33,12 @@ export function hasBitcoinAddress(
   return account.bitcoin !== undefined;
 }
 
+export function hasHdBitcoinAddress(
```
edgar-stacks: this is unused, should we remove it?

edgar-stacks (dismissed):

packages/utils/src/accounts/account-addresses.ts
aperry-stacks: Good catch. Removed ✅

edgar-stacks (approved):

leah-stacks (approved):

PR #2416, opened 2026-06-22
fix: account lookup

edgar-stacks:
It fixes account discovery in the extension wallet — making sure all of a user's funded accounts actually show up after they restore or unlock their wallet.

Previously, the wallet only scanned for active accounts once, at the moment a wallet was first added. If a long-time user had accounts at higher indexes, or created accounts on another device, those accounts could silently fail to appear.

Key improvements for users

- Accounts are now rediscovered on every unlock, not just at wallet creation — so accounts added on another device (or missed previously) get picked up automatically.
- Discovery resumes from where it left off instead of re-scanning from the beginning, making it faster and avoiding repeated work for users with many accounts.
- Accounts appear progressively as they're found, rather than all at once at the end — the user isn't left waiting on a blank screen, and discovery runs in the background without blocking them from reaching the home screen.
- Detection is more thorough — an account now counts as "active" based on Bitcoin and Stacks balances, transaction history, and BNS names (across more address types than before), so funded accounts are far less likely to be missed.

Reliability improvements

- More resilient to flaky network conditions — if some checks time out or fail, discovery still succeeds on whatever data does come back, instead of giving up.
- Safer against duplicates and regressions — the discovered account count can never accidentally shrink, and the wallet won't run the same scan twice at once.


314159265359879 (approved):
I tested a couple wallets with more than 100 accounts. Lookup worked excellently!

I like how you are also showing the lookup in the console, I can see what it is doing, this will be great if I ever need to troubleshoot an issue with the feature. 

By the way, I don't see the debug command fixed yet. My sense is that it may have broke when Kyran was working on prepping for multiwallet. Now you may also need to add a target to the wallet (wallet 1, wallet 2). To keep it simple just setHighestAccountIndex for the active wallet, that way the command doesn't need an extra argument?

Here is the failure that is shown when the command is tried:
Image

I am approving this because I wasn't sure you'd want this to be part of the scope.

aperry-stacks (changes requested):
This isn't rebased on latest `dev` yet but we did just introduce address-based BTC calls to the `LeatherApiClient` and have equivalent STX and BNS clients available via `@leather.io/services` that should cover account lookup requirements: 

``` typescript
  dispatch(keyActions.probeNextAccountAndDiscoverAccounts({
    leatherApi: getLeatherApiClient(),   
    hiro:       getHiroStacksApiClient(), 
    bns:        getBnsV2ApiClient(), 
  }));
```

We really should not be building new things with the `@leather.io/query` package. Even if there are gaps, let's address them rather than reach for a package that's being deprecated.

edgar-stacks:
@aperry-stacks yeap that's a very good point, i'll update it now

aperry-stacks (dismissed):
Nice. LGTM 🚀 Claude caught an interesting regression that looks it could be an easy patch on this PR if you're up for it. Suggests migrating `use-account-names.ts` to also use services clients so it benefits from the pre-warmed cache during account recovery (i.e. instead of duplicating BNS lookups calls).

### `BNS caching regression`

Discovery now reads BNS names via the services client (cached in the services HttpCacheService), but use-account-names.ts still reads them through @leather.io/query + React Query — a
  separate cache. The old code primed React Query directly; that priming was dropped here. Result: names discovery already fetched get refetched by the UI — a redundant request per
  restored account.

  Fix: point use-account-names.ts at createAccountBnsNamesQueryConfig (@leather.io/queries) so both flows share the services cache. Hook signature stays the same, so consumers don't
  change — ~1 file + spec.

  Precedent: mobile already does this (use-account-display-address.ts → useAccountBnsNames → getBnsService().getAccountBnsNames).

314159265359879 (approved):
This is working great. I also like the new Ledger label for the wallet and account labels over the account avatar.

edgar-stacks:
@aperry-stacks updated the account names requests, now it uses services package
@314159265359879 removed 1000 account cap

aperry-stacks (approved):

fab-stacks (approved):

PR #2418, opened 2026-06-22
feat: build multisig transfer transactions

aperry-stacks:
Adds tooling to build and propose unsigned multisig transactions for both BTC and STX, with correct multisig-aware fee estimation. This is the proposal half of the multisig signing flow — it produces a valid transaction and POSTs it to `/propose`; signature collection and broadcast come later.

### Multisig-aware coin-selection fee sizing (`@leather.io/bitcoin`, `@leather.io/services`)

* The fee estimator already knew how to size a P2WSH M-of-N input — it just wasn't wired up. `getSizeInfo` now takes an optional `InputSizing` spec and routes P2WSH inputs (and change) through `calcTxSize`; threaded through coin-selection and max-spend. Default single-sig behaviour is untouched.
* `FixedBitcoinAddressInfo` is self-describing (`multisig: { threshold, signerCount }`), so `performCoinSelection`, the fees service, and max-spend all produce correct numbers for multisig accounts with no signature changes.

### Transaction builders (dapp-local)

* STX — `buildUnsignedMultisigStxTransfer`: non-sequential multisig transfer with placeholder nonce 0 (the backend assigns the real nonce on queue promotion); fee is committed at propose time.
* BTC — `buildUnsignedMultisigBtcTransfer`: reconstructs `wsh(sortedmulti(…))`, runs coin selection, assembles the PSBT. Byte-level PSBT assembly (`assembleWshMultisigPsbt`) lives in `@leather.io/bitcoin`, mirroring `generateBitcoinUnsignedTransaction`.
* Both guard that the derived address equals the vault's `multisigAddress` before proceeding.


fab-stacks (approved):

apps/web/app/features/multisig/transactions/build-btc-transfer.ts
```diff
@@ -0,0 +1,57 @@
...
+function getBitcoinNetworkMode(network: AuthNetworkId): BitcoinNetworkModes {
+  return network === 'btc:mainnet' ? 'mainnet' : 'testnet';
+}
+
+export function getMultisigDescriptor(account: VaultAccount): string {
+  const publicKeys = account.signers.map(signer => signer.signingPubkey);
+  return `wsh(sortedmulti(${account.threshold},${publicKeys.join(',')}))`;
+}
+
+export async function buildUnsignedMultisigBtcTransfer({
```
edgar-stacks, line 28: i think i'd need to extract these functions into a shared package for extension as well

edgar-stacks (approved):

apps/web/app/features/multisig/transactions/build-btc-transfer.ts
aperry-stacks: Sounds good. cc @leah-stacks to coordinate when building out the feature in `web` as well.

PR #2420, opened 2026-06-22
fix: signTransaction fee/nonce bug

edgar-stacks:
Currently the app blocks any fee/nonce updating by the user during signTransaction flow. This PR fixes that and only blocks it if the transaction is multisig.

more info on this here: https://stackslabs.slack.com/archives/C09U43UADGU/p1781689885322689


aperry-stacks (approved):
Good call on the Multisig carveout 🦾

leah-stacks (approved):

PR #2421, opened 2026-06-22
feat: add btc/stx_addAccount rpc approver flow

edgar-stacks:
First PR in the line up of adding the btc/stx_addAccount rpc request.

This only implements receiving the rpc request, allowing user select a correct account, blocking account that is not associated with the multisig, and responding with a placeholder data. In future PRs I would also add multisig address derivation for both chains, adding that account to extension's state and proposing transaction on behalf of that multisig account

It is also blocked by a feature flag for now


packages/rpc/src/methods/stacks/stx-add-account.ts
```diff
@@ -0,0 +1,33 @@
...
+import { policyAccountRoleSchema } from '../bitcoin/btc-add-account';
+
+const compressedSecp256k1PublicKeySchema = z.string().regex(/^0[23][0-9a-fA-F]{64}$/);
+
+const stxAddAccountRequestParamsSchema = z
+  .object({
+    publicKeys: z.array(compressedSecp256k1PublicKeySchema).min(2),
+    threshold: z.number().int().min(1),
+    name: z.string(),
+    vaultAccountId: z.string(),
```
aperry-stacks: Do we still need these `accountId` params? The propose transaction endpoint works with just a Multisig address now. Wonder if we would need this for anything else?

edgar-stacks: yes you're right, this is redundant, removing

edgar-stacks: updated 👍

leah-stacks (approved):

aperry-stacks (approved):

PR #2428, opened 2026-06-24
fix(extension): label Bitcoin API dropdown as Custom when values diverge from a preset

fab-stacks:
On the extension's **Add Network** form, the **Bitcoin API** dropdown picks a preset (Mainnet, Testnet3, Testnet4, Signet, Regtest) that fills in the Bitcoin API URL and Stacks API URL. Once you manually edit those URLs, the dropdown kept showing the old preset name (e.g. still "Mainnet" after you've pointed it at a different node). That contradiction can be confusing.

This makes the dropdown label honest:

- When the URL fields no longer match the selected preset, the trigger reads **"Custom"**.
- As soon as the values match the preset again — or you pick a different preset — the preset name returns.
- **"Custom" is display-only** and is never added to the selectable option list.

CleanShot 2026-06-24 at 14 14 09


edgar-stacks:
btc address changes if it is a mainnet/testnet/regtest network, i wonder if extension derives that from the url alone or if we use this network label to decide how we should derive an address. We'll need to double check this

fab-stacks:
Pushed `0a12914` addressing both comments:

**@claude's controlled-Select nit (revert-by-reclick):** Fixed. The `Select` is now driven by the *matched* preset value (empty string when "Custom"), so re-selecting the already-active preset to revert is a real value change and fires `onValueChange` → URLs reset. Added a unit test (`getSelectedNetworkValue`) and an E2E (`reselecting the active preset after a custom edit resets its URLs`) for exactly this path.

**@edgar-stacks on address derivation:** Good question — checked it. Address type is derived from the stored `bitcoinNetwork` / `mode` (`mode: bitcoinNetworkToNetworkMode(bitcoinNetwork)` in `use-add-network.tsx`), **not** from `bitcoinUrl` — the URL is persisted only as the API endpoint. This PR is display-only and never mutates `bitcoinNetwork`, so address derivation is unchanged. The mainnet-mode-with-a-custom-testnet-URL mismatch you're describing was already possible before this PR (the dropdown just silently said "Mainnet"); this change makes that divergence *visible* by showing "Custom" instead. Whether a "Custom" config should force an explicit network-mode choice for derivation is a separate product decision — happy to spin that out as a follow-up if we want it.

aperry-stacks (approved):

leah-stacks (approved):

edgar-stacks (approved):

PR #2430, opened 2026-06-24
feat: policy account state

edgar-stacks:
- policy id now consists of `${parentAccountId}/${policyAddress}/${networkKey}`
- If policy id exists for a particular network key, that network can't be deleted or edited.
- if network is not passed by dApp, default to mainnet


aperry-stacks (commented):
The STX address mismatched looks like its worth addressing IMO, not sure why claude review downplayed it but here's what my local review takeaway: 
_"STX address is derived from the wallet's active network, not the policy's network — register-stx-policy-account.ts (network.chain.stacks.chainId). Needs discussion.
  The schema accepts an optional params.network but it's ignored; derivation uses whatever network the extension is currently on. Stacks public keys don't encode a network, so if the
  user's active network ≠ the vault's network (e.g. extension on testnet, vault on mainnet), the derived address — stored in state and returned to the dApp as accountId/address — silently
  becomes the wrong-network address and won't match the vault.

  Note the asymmetry: BTC sidesteps this entirely because getWshDescriptorAddress auto-detects the network from the descriptor's own key prefixes (self-contained, robust). STX has no such
  anchor. Recommend deriving the STX network from params.network (pin it to the policy), or validating that the active network matches before allowing approval. Nothing upstream guards
  this — the approver's match check compares pubkeys only, which is network-independent."_

edgar-stacks:
@aperry-stacks thank you for catching this, it turned out to become quite a rabbit hole of custom network related issues. The PR is now ready for re-review

leah-stacks (approved):
Great work!

fab-stacks (approved):

PR #2431, opened 2026-06-24
feat: multisig tx signing

aperry-stacks:
# Multisig transaction signing, broadcast, cancel + pre-signing verification
Completes the multisig transaction lifecycle on top of propose: a vault member can now sign a proposed transaction with the extension, broadcast it once threshold is met, and cancel it — for both BTC and STX. Every sign path is gated by §7 pre-signing verification, which re-derives and cryptographically checks everything the (untrusted) backend serves before the wallet is ever invoked.

Proven end-to-end on mainnet, both chains: propose → sign-to-threshold → broadcast → confirmed on-chain.

### Signing (`features/multisig/transactions/signing/`)

* `signBtcTransaction` — converts the proposed PSBT to hex, signs via the extension's `wsh(sortedmulti)` descriptor `signPsbt`, extracts the per-input `partialSig`s.
* `signStxTransaction` — substitutes the backend-assigned nonce for the placeholder (byte-precise), signs the whole tx via `stxSignTransaction`, extracts the single VRS signature.
* Both call `preSignVerification` as their first line — non-bypassable by construction.

### Pre-signing verification (`pre-sign-verification.ts`) — a legible step sequence, each throwing on failure:

1. re-derive the multisig address from the served signer set; assert it matches.
2. proposer is a signer on the account (userId).
3. (BTC) bind the proposer's identity key ↔ signing key via the xpub.
4. recompute the proposal hash and verify the proposer's commitment against a vault-anchored key — BIP-322 over the p2wpkh of the bound publicKey (BTC); SIP-018 against the bound signingPubkey (STX).

### Crypto primitives

* `@leather.io/bitcoin`: `verifyBip322Signature` (new `bip322-js` dep), `deriveXpubChildPublicKey`, `getP2wpkhAddressFromPublicKey`, `extractWshMultisigSignatures`, `psbtBase64ToHex`.
* `@leather.io/stacks`: `verifySip018Signature`, `substituteStxNonce`, `extractStxMultisigSignature`.
* `derive-multisig-address.ts` — single shared derivation used by both the builders and verification, so propose-time and sign-time can't drift.

### Model / cleanup — VaultAccountSigner gains userId (proposer resolution) and drops the redundant derivationIndex (it was a non-stored view on account.accountIndex; coordinated with the backend).


packages/bitcoin/src/utils/derive-xpub-child-public-key.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { deriveKeychainFromXpub } from '@leather.io/crypto';
+
+import { deriveAddressIndexKeychainFromAccount } from './bitcoin.utils';
+
+export function deriveXpubChildPublicKey(
+  xpub: string,
+  changeIndex: number,
+  addressIndex: number
+): string {
```
edgar-stacks: nit: can we do object here? Otherwise we get function calls like this `deriveXpubChildPublicKey(xpub, 0, 0)`

packages/bitcoin/src/bip322/verify-message-bip322.ts
```diff
@@ -0,0 +1,14 @@
+import { Verifier } from 'bip322-js';
```
edgar-stacks: Any chance we could replace bip322-js's Verifier with existing packages? or write it ourselves if it's small? Just to have one less critical security dependency

packages/stacks/src/transactions/multisig-nonce.ts
```diff
@@ -0,0 +1,44 @@
...
+const signerHashOffset = 7;
+const nonceByteLength = 8;
+
+// Signer-hash size per supported multisig hash mode;
+const signerHashSizeByHashMode: Record<number, number> = {
+  [AddressHashMode.P2SH]: 20,
+  [AddressHashMode.P2WSH]: 32,
+  [AddressHashMode.P2SHNonSequential]: 20,
+  [AddressHashMode.P2WSHNonSequential]: 32,
+};
```
edgar-stacks: flagged by ai

substituteStxNonce uses wrong signer-hash size (32) for P2WSH multisig modes

The map sets P2WSH → 32 and P2WSHNonSequential → 32, but @stacks/transactions serializes the multisig signer as a 20-byte hash160 for all four modes (verified: authorization.js:123 reads readBytes(20) unconditionally). For a P2WSH-family payload the code computes nonceOffset = 7+32 = 39 instead of 27, overwriting the fee field 12 bytes too late — silently corrupting the tx (the length guard doesn't catch it on realistic payloads).

packages/bitcoin/src/transactions/wsh-multisig-transaction.ts
```diff
@@ -49,3 +49,26 @@ export function assembleWshMultisigPsbt({
...
+export function extractWshMultisigSignatures(signedPsbtHex: string): WshMultisigSignature[] {
+  const tx = btc.Transaction.fromPSBT(hex.decode(signedPsbtHex));
+  const signatures: WshMultisigSignature[] = [];
+  for (let inputIndex = 0; inputIndex < tx.inputsLength; inputIndex++) {
+    const signature = tx.getInput(inputIndex).partialSig?.[0]?.[1];
+    if (!signature) continue;
+    signatures.push({ inputIndex, signature: hex.encode(signature) });
+  }
+  return signatures;
+}
```
edgar-stacks, line 80: flagged by ai:

These are not bugs on any reachable path today, but they live in reusable domain packages with assumptions encoded only in comments:

- extractWshMultisigSignatures takes partialSig[0] positionally — packages/bitcoin/.../wsh-multisig-transaction.ts:72. @scure/btc-signer sorts partialSig by pubkey on serialization, so [0] is "smallest pubkey," not "current signer." Safe only because each signer signs a fresh single-key proposal. If a PSBT ever carries a co-signer's sig first, the wrong signature is attributed. Recommend: thread the signer's signingPubkey and select via partialSig.find(([pub]) => equalBytes(pub, signerPubkey)).
- Multi-key wallet → extra signatures silently dropped (same [0] issue) — only the first of N partial sigs per input is returned.
- Signature submitted without its pubkey — WshMultisigSignature = {inputIndex, signature} discards partialSig[i][0], forcing the backend to ECDSA-recover which pubkey signed for the sortedmulti witness. Out-of-repo, but cheap to include the pubkey.
- extractStxMultisigSignature "exactly one signature" precondition is non-sequential-multisig-specific and undocumented as such — would throw if ever reused for sequential multisig.

packages/bitcoin/src/transactions/wsh-multisig-transaction.spec.ts
```diff
@@ -1,10 +1,16 @@
-import { base64 } from '@scure/base';
```
edgar-stacks, line 1: Flagged by ai:

- No multiple-input PSBT test for extractWshMultisigSignatures (wsh-multisig-transaction.spec.ts) — only single-input is exercised, so the per-input loop, inputIndex assignment, and the if (!signature) continue skip branch are uncovered. Multi-UTXO is the common real case → medium.
- 32-byte signer-hash path of substituteStxNonce untested (ties to bug #1).
- extractStxMultisigSignature >1 signature throw untested; extractWshMultisigSignatures 2+ partial-sigs case untested.
- Descriptor order-independence/BIP-67 only proven for raw-pubkey form; mixed xpub+raw signer set untested (the comment's core invariant).
- buildUnsignedMultisigBtcTransfer has only an error-path test (coin-selection DI singleton unmocked).
- assembleWshMultisigPsbt asserts change script but not the recipient address/values — an output.address ?? changeAddress swap would pass.

packages/stacks/src/transactions/multisig-nonce.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Overwrites the nonce of a serialized unsigned multisig Stacks transaction in
+// place, deriving the offset from the parsed spending-condition prefix (never a
+// hardcoded constant).
+export function substituteStxNonce(rawPayload: string, nonce: number): string {
```
edgar-stacks: same here regardin args

leah-stacks (dismissed):

packages/bitcoin/src/bip322/verify-message-bip322.ts
aperry-stacks: Done ✅ Dropped `bip322-js` and replaced with our own `verifyP2wpkhBip322Signature` function using existing `bitcionjs`.

packages/stacks/src/transactions/multisig-nonce.ts
aperry-stacks: I shouldn't have let this go in as a generic utility that supports multiple hash modes. This is a specific function related to multisig signing and only supports `P2SHNonSequential`. Scoped this down to the single hash mode, moved it out of `@leather.io/stacks` and into the `web` multisig dapp directory (`substitute-stx-nonce.ts`).

packages/bitcoin/src/transactions/wsh-multisig-transaction.ts
aperry-stacks: * added `signingPubkey` to the signature to select signer pubkey by match, bypasses positional issue
* function is now single-signer by definition
* signer pubkey is taken from jwt auth context, doesn't need to be passed to backend
* again here, `extractStxMultisigSignature` should have been a contextual helper that's specific to the multisig dapp signing flow (i.e. not a shared helper). Moved and re-contextualized there.

packages/bitcoin/src/transactions/wsh-multisig-transaction.spec.ts
aperry-stacks: Test gaps patched ✅

apps/web/app/features/multisig/transactions/signing/pre-sign-verification.ts
```diff
@@ -0,0 +1,99 @@
...
+  const proposer = account.signers.find(signer => signer.userId === transaction.proposerUserId);
+  if (!proposer) throw new Error('Proposer is not a signer on this account.');
+  return proposer;
+}
+
+// On BTC the proposer signs the commitment with their identity key while a
+// separate signing key sits in the multisig; bind the two so the authorizer is
+// provably a key-holder. STX uses one key for both, so there is nothing to bind.
+function assertProposerKeyBinding(proposer: VaultAccountSigner, account: VaultAccount): void {
+  if (!account.network.startsWith('btc')) return;
```
edgar-stacks, line 33: One more high i found during the review (ai)

🔴 Must-fix — Security (HIGH)

1. Network-confusion bypass of preSignVerification · pre-sign-verification.ts:33,55,65,75
The verifier picks the chain from transaction.network (line 55, and the STX domain on 75) but gates the BTC proposer-key binding and BIP-322 mode on account.network (lines 33, 65) — with no assertion the two agree. Both are independent attacker-served JSON fields.

Concrete bypass (independently confirmed): take a real STX vault (account.network=stx:mainnet, pinned because the address re-derives from it) and serve transaction.network=btc:mainnet. Then:
- chain='btc' → verification runs the BIP-322 branch against an address derived from proposer.publicKey.
- assertProposerKeyBinding returns early (gated on account.network.startsWith('btc')), so publicKey is never bound to the xpub. STX addresses derive from signingPubkey, never publicKey, so publicKey is fully attacker-controlled.
- Attacker controls both publicKey and proposalSignature → self-consistent BIP-322 → valid=true. The tamper-detection boundary is defeated.

▎ Fix: assert transaction.network === account.network at the top of preSignVerification, then derive a single chain+mode from one pinned source. One line collapses the three network sources to one.
▎ Mitigation today: only reachable via the dev-tools panel (gated out of prod by multisigEnabled), and the extension's own confirmation UI is a final human backstop — which is why it's HIGH, not CRITICAL. But preSignVerification is meant to be the programmatic guard a co-signer trusts, so fix before it's wired to real UI.

edgar-stacks (approved):

leah-stacks (approved):

PR #2433, opened 2026-06-25
fix(extension): tidy up the multi-wallet account row in approval views

fab-stacks:
In the multi-wallet world the account row in our approval views got crowded — it was trying to fit the **wallet name**, both the **Bitcoin** and **Stacks** addresses, and the **balance** on a single line. With both addresses present the address line ran into the balance, and a few smaller details looked off.

A few tweaks to tidy it up:

- Dropped the **wallet name** from the row — alongside the account name, two addresses and the balance it was just too much.
- The **separator dot** between the addresses now matches their subdued color instead of inheriting the darker default text color, so it no longer stands out when both addresses show.
- The **balance** is centered against the name and addresses again (removed an old offset that was nudging it down toward the address line).
- Brought back the **›** chevron on the row so it's clear you can tap to **switch accounts**
CleanShot 2026-06-25 at 19 18 24@2x

### DIFF
--- a/apps/extension/src/app/components/account/account-addresses.tsx
+++ b/apps/extension/src/app/components/account/account-addresses.tsx
@@ -14,7 +14,7 @@ export function AccountAddresses({ accountId }: AccountAddressesProps) {
   const account = useStacksAccountLoader({ accountId });
   const signer = useBitcoinNativeSegwitAccountLoader({ accountId });
   return (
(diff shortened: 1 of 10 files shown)


314159265359879 (approved):
LGTM

edgar-stacks (approved):

aperry-stacks (approved):

PR #2438, opened 2026-06-29
feat: policy accounts

edgar-stacks:
Introduces policy accounts: registered multisig accounts (Bitcoin wsh(...) P2WSH descriptors, Stacks pubkeys + threshold) where one of your single-sig accounts is a co-signer. They become first-class, switchable wallet entities plus an RPC surface for dapps to register/verify them. Gated behind the releaseAddAccount LaunchDarkly flag.

RPC protocol

- getAddresses gains opt-in allowPolicyAccounts — off by default (single-sig only); when true, also returns the account's multisig addresses. New BTC p2wsh type and STX kind: 'multisig' (threshold + publicKeys) response shapes.
- btc_addAccount / stx_addAccount responses gain `added: boolean` to distinguish "account registered" from "address only verified, nothing stored".

Dapp approval flows

- Add-account approver reworked into add vs verify mode by origin: whitelisted origins (app.leather.io) register the policy; all others get verify-only (shows the derived multisig address to double-check, stores nothing). Approver UI shows the address and mode-aware titles/buttons.
- Ledger on-device confirmation (BTC): registers an ephemeral wsh(...) policy on-device, compares device vs locally-derived address, and errors on mismatch.

In-wallet policy accounts

- Account switcher lists policies nested under their parent (current network) with name, address, and balance; tapping switches the active context.
- Manage mode adds per-policy Rename and Remove; rename logic now shared between accounts and policies, defaulting to Multisig <address>.
- Home screen is policy-aware (balances, receive address, token list resolve from the policy's addresses). Restricted while a policy is active: swaps disabled, Buy/Sell hidden, send collapsed to receive-only; Collectibles hidden for BTC multisigs.

Security / guards

- Old hard origin block removed in favor of the verify-mode gate; origin matching is now exact scheme+host+port. New guard rejects value-moving RPCs (e.g. sendTransfer) while a multisig account is active.

Supporting changes & fixes

- BIP322 attaches nonWitnessUtxo so Ledger accepts the message-signing PSBT; new getWshDescriptorThreshold decodes M-of-N from a descriptor; balance hooks keyed by addresses for policy accounts.
- signMessage fee-warning now defaults to mainnet (was testnet); multisig web demo "Add to wallet" wired to the RPC methods.
- Redux: active slice tracks activePolicyId; policy add/remove + selectors; cross-window replay sync. New E2E + unit coverage and policy mocks.


edgar-stacks:
i just noticed collectibles and activity on policy accounts are still singlesig, updating now

EDIT: fixed

aperry-stacks (dismissed):
🚀🚀🚀🚀

leah-stacks (dismissed):

aperry-stacks (approved):

leah-stacks (approved):

PR #2467, opened 2026-06-30
feat(web): multisig network switcher polish

fab-stacks:
Design pass on the network switcher, plus a mock harness so the multisig screens are reviewable without a backend.

Moves the network control out of the sidebar into the **multisig header** next to **Connect wallet**: a **globe** button with a **connection-status dot** (green when a vault chain is connected, grey when not) and a chevron. Off mainnet it picks up a subtle tint and a **Testnet** label, and a soft amber **glow** washes down from the top so it's always obvious when you're not on mainnet.

The **Dev tools** panel's **Preview data** control now drives mock mode — **Live / Populated / Empty**. Toggle it and MSW serves seeded, network-aware vaults (mainnet vs testnet show different data) or an empty state — no wallet or testnet backend needed — so these states are previewable right in the PR.

CleanShot 2026-06-30 at 11 36 58


fab-stacks:
Superseded by `feat/multisig-testnet-design` — same two commits, just renaming off the `claude/` prefix. Re-opening the PR there (stacked on #2465).

PR #2471, opened 2026-06-30
feat(web): multisig design QA — list-item consistency + UI polish

fab-stacks:
Building on #2435, this is a design-QA pass over the whole multisig web UI — the goal was to pull every list-style surface onto one shared primitive and tighten the small stuff so it all reads as one system instead of a dozen slightly-different rows.

- Every list row now comes from one shared **`ListItemBox`** — the **transaction feed**, **vault & account cards**, **members**, **signers**, the **create-vault** chain + member rows, and the **account / vault detail headers**. Boxed for standalone rows, "plain" inside cards, so avatar size, gap, and typography are consistent everywhere. The old local `VaultListItem` is gone.
- **Addresses** all share one **click-to-copy** style: truncated inline in rows, and a wrapped grouped block under the **Address** / **Recipient** labels — a notch smaller than the surrounding text and aligned to the text edge with a subtle hover.
- **Status chips** (Joined / Invited / Creator…) are sized to match the small buttons. A member row only calls out the **Creator** or a still-pending **invite** — _joined_ is the default, so no chip — and a signer who's signed gets a green ✓ chip.
- **Transaction rows** show the amount over its fiat value on the right for every state except the ones awaiting signatures, which keep a **Review** button instead.
- The **transaction detail** view now matches the account/vault balance heroes (big "Transfer" + "proposed … by"), its status reads as a dotted text label rather than a chip, and the **recipient** shows the full grouped address.
- Plus a round of small alignments: the **account icon picker** fits without a scrollbar (and slightly larger icons), the chain/token indicator badges match the transaction indicators, and the detail-card headers drop to the compact sidebar scale.

You can walk through every view + modal in-context at **`/multisig/pages-preview`**, and a component gallery at `/multisig/gallery-preview`. Those preview routes are dev-only — gated out of production builds (`import.meta.env.MODE`), so they're there for review but never reachable in the public app.


packages/ui/src/components/avatar/avatar.web.tsx
```diff
@@ -54,29 +54,35 @@ export const Avatar = forwardRef<AvatarElement, AvatarProps>((props, ref) => {
...
-          height="20px"
-          display="flex"
-          alignItems="center"
-          justifyContent="center"
-          overflow="hidden"
-        >
-          {indicator}
-        </styled.div>
-      ) : null}
+      {indicator ? <AvatarIndicatorBadge size={size}>{indicator}</AvatarIndicatorBadge> : null}
```
edgar-stacks, line 57: Do we need to implement changes to extension and mobile due to this file update?

edgar-stacks (dismissed):

leah-stacks (dismissed):
Not blocking, but I think making the clipboard change as the Claude review suggests would be a good idea.

packages/ui/src/components/avatar/avatar.web.tsx
fab-stacks: The indicator badge used to be hardcoded to 20px for every avatar size. The new size-scaled version maps both lg and xl to 20px, so should look the same on both extension + mobile

fab-stacks: No — this is backward-compatible, so no extension/mobile changes are needed. The indicator badge was previously hardcoded to 20px for every avatar size; the new size map keeps **lg and xl at 20px** (and `xl` is the default size), so every existing ext/mobile indicator usage — asset/token icons and the account avatar, all on `size="xl"` — renders identically. Only `md`/`sm`/`xs` now scale down, and those smaller sizes are used solely by the new multisig web rows.

leah-stacks (approved):

aperry-stacks (approved):

PR #2473, opened 2026-07-01
feat: blockchain activity paging and view mapping

aperry-stacks:
Re-builds the V2 activity service on top of new Stacks API `/principals/{p}/transactions` and `/balance-changes` endpoints that enable clean cross-chain activity paging.

It also fully integrates the protocol / action mapping concept into the view layer (`@leather.io/features`).

## Summary

Rebuilds the activity system around two ideas: **normalized protocol _actions_** that
deterministically drive how each row renders, and a **stateless, cursor-paged feed** that
merges Bitcoin and Stacks into one timeline. It replaces the skew-prone `getTransactionEvents`
approach and an abandoned integration attempt that flattened `events[] → events[0]` and
ignored action normalization.

**This is additive infrastructure with no user-facing change.** V2 (`BlockchainActivityService`
→ `BlockchainActivity`) runs in parallel with V1 (`ActivityService` → `OnChainActivity`), whose
logic stays live and untouched — though V1's underlying BTC fetch now goes through the reworked
`BitcoinTransactionsService` as a single explicit page request of the same depth as before.
Nothing in the apps consumes V2 yet — the query-configs and view
builder exist but are unwired. Adoption is incremental, per consumer, in follow-up PRs.

## Why

- V1 rendered rows from a flattened single event and re-derived meaning in the view layer
  (brittle, per-app heuristics). V2 interprets **once** in the service: every activity carries
  exactly one normalized `action`, and the view is a pure projection of it.
- V1's page sizes were variable/unbounded under a sparse-STX / dense-BTC window. V2 emits a
  fixed page every call.

## What's in it

**Domain (`@leather.io/models`)**
- `BlockchainActivity` with top-level required `action` (20-action vocabulary), optional
  `protocol` + `protocolName`, `balanceChanges[]` (`{ direction, asset, amount }`), `counterparty`,
  `contract`.
- `stacksProtocolActions` as the single shared action vocabulary.

**Service (`@leather.io/services`)**
- **Fixed-page cursor paginator** (`activity-paginator.ts`): merges wholesale BTC + cursor-paged
  STX on a strict `(timestamp, chain, txid)` total order; returns exactly `limit` items until the
  feed ends; cursor `{ afterKey, stxToken, stxDone }` resumes at an exact boundary (no gaps/dupes).
- **Mapping** (`bitcoin-activity.utils`, `stacks-activity.utils`): raw tx → `BlockchainActivity`
  (action classification, balance changes, SIP-10 metadata + fiat quotes), shared by confirmed
  and pending paths.
- **First-page pending stitch**: STX mempool + BTC unconfirmed, deduped by txid (confirmed wins).
- **By-asset** query path (`getActivityByAssetId`) as a bounded recent scan.
- Real **Hiro v3** schema + client (`/principals/{p}/transactions`, `/balance-changes`) and cache keys.
- **`BitcoinTransactionsService` takes an explicit page request** (single-page fetch, no internal
  fan-out). Activity fetches BTC wholesale in one `pageSize: 1000` request — the deliberate BTC
  visibility horizon (the server imposes no page-size max). V1 activity and UTXO call sites keep
  their prior 750-tx depth, now as one request instead of up to five.
- Removes the abandoned v2 utils (`*-blockchain-activity`, `stacks-activity-event`); the live
  `OnChainActivity` path is untouched.

**View (`@leather.io/features`)**
- `createBlockchainActivityView(activity, { formatMoney, translate? }) → BlockchainActivityView`
  — a deterministic `action → row shape` projection. `avatar` / `indicator` / `amount` stay
  structural (enums + data) for each app to map to its own icons/colors; `title` / `subtitle`
  are finished strings.
- i18n **by injection, not a key table**: conjugation lives once in features; `translate`
  defaults to English interpolation, mobile passes its i18n resolver in the query `select`. No
  i18n dependency added to web/extension.
- Per-slot **degradation** for pending / no-balance-change txs (avatar → contract-call icon,
  title → function name, amount omitted); blooms into full shape on confirmation.

**Queries (`@leather.io/queries`)**
- `createBlockchainActivityInfiniteQueryConfig` (`infiniteQueryOptions`) for the deep
  account-feed, and single-page `createBlockchainActivityQueryConfig` for bounded previews. Both
  return raw `ActivityResponse`; view mapping happens in each app's `select`.

## Testing

Unit tests across the pure mappers/builder, the paginator (size / no-gaps-or-dupes / dense-BTC /
tail), the service (by-asset, pending ordering, SIP-10 reclassify, STX fee handling incl.
sponsored), and the view builder (each action × state, degraded path, translate injection).
`format` / `lint` / `typecheck` / `knip` / extension unused-exports all green.

## Out of scope (follow-ups)

- App list item + wiring V2 into a consumer (first: web multisig — bounded previews + deep feed).
- By-asset query-config, protocol-level filtering.
- NFT balance changes (count-based, discriminated member).
- V1 retirement once all consumers move off `OnChainActivity`.


packages/services/src/activity/activity-paginator.ts
```diff
@@ -0,0 +1,182 @@
...
+  }>;
+  fetchAllBtcTx(): Promise<ActivitySourceItem[]>;
+}
+
+const timestampKeyWidth = 16;
+const maxStacksTxPageFetches = 20;
+
+// Strict total order so a cursor can resume at an exact boundary with no gaps or duplicates.
+function activitySourceItemKey(item: ActivitySourceItem): string {
+  return `${String(item.timestamp).padStart(timestampKeyWidth, '0')}:${item.chain}:${item.txid}`;
```
edgar-stacks, line 46: packages/services/src/activity/activity-paginator.ts:46: Stacks txs with the same block.time can be dropped across Hiro cursor pages. The cursor key orders ties by txid, but Hiro paging is not guaranteed to be in that same tie order; the next page filters with activitySourceItemKey(item) < afterKey at line 120.

Scenario: page 1 emits same-timestamp stx-a, page 2 later contains stx-z; stx-z sorts above afterKey and is filtered out forever. Fix by keying Stacks order with Hiro’s actual order, e.g. block.height / tx_index / final tie-breaker, or drain same-timestamp groups before advancing. Add a regression where equal timestamps are split across pages in non-txid order.

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -51,319 +82,513 @@ export class BlockchainActivityService {
...
+    });
+    // Pending is stitched onto the first page only and sits above the cursor keyspace, so
+    // pages 2+ never re-emit it. Dedup by txid, confirmed wins: a tx that confirms between
+    // refetches transitions atomically (leaves pending, appears confirmed, exactly once).
+    const [confirmed, allPending] = await Promise.all([
+      this.mapSourceItems(request.account, page.items, signal),
+      request.cursor === undefined
+        ? this.getPendingActivity(request.account, btcTxs, signal)
+        : Promise.resolve([]),
+    ]);
```
edgar-stacks, line 107: packages/services/src/activity/blockchain-activity.service.ts:104: pending dedup only compares pending txids against confirmed items on that same first page. If a pending tx confirms below the page-1 boundary before page 2 is fetched, page 1 keeps the pending row and page 2 can return the confirmed row; packages/queries/src/activity/blockchain-activity.query-config.ts:59 carries only nextCursor, no seen txids. Fix by carrying pending txids in cursor/query state or deduping the flattened feed by chain:txid with confirmed-wins semantics.

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -51,319 +82,513 @@ export class BlockchainActivityService {
...
+    btcTxs: Promise<BitcoinTransaction[]>,
+    signal?: AbortSignal
+  ): ActivitySources {
+    return {
+      fetchAllBtcTx: async () => {
+        const txs = await btcTxs;
+        const items: ActivitySourceItem[] = [];
+        for (const tx of txs) {
+          // No time means unconfirmed: those belong to the pending stitch, not the merge.
+          if (tx.time === undefined) continue;
```
edgar-stacks: 3. packages/services/src/activity/blockchain-activity.service.ts:568: BTC pending classification is inconsistent. Pending BTC uses height === undefined at line 253, but the paginator source admits any tx with time; the model/API allow height? and time? independently. A no-height tx with time can be stitched as pending and later emitted from the paginated source. Fix fetchAllBtcTx to exclude tx.height === undefined; only confirmed BTC belongs in the merge keyspace.

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -51,319 +82,513 @@ export class BlockchainActivityService {
...
+    result: HiroPrincipalTransactionsResultItem,
+    ftChanges: BlockchainActivityBalanceChange[],
     signal?: AbortSignal
-  ): Promise<BlockchainActivityEvent | undefined> {
-    try {
-      const isSent = transfer.sender === stxAddress;
+  ): Promise<BlockchainActivity | null> {
+    if (!isStacksActivityResultItem(result)) return null;
+    const tx = result.transaction;
+    const initiatedByUser = result.involvement === 'sender';
```
edgar-stacks: packages/services/src/activity/blockchain-activity.service.ts:337: sponsored transactions are misclassified for the sponsor account. initiatedByUser is only involvement === 'sender', so line 341 never treats involvement === 'sponsor' as fee-paid-by-user; a fee-only sponsor activity becomes a sent STX balance change instead of a fee. Fix by computing fee payer separately and netting sponsor fees out too.

packages/services/src/activity/activity-paginator.ts
```diff
@@ -0,0 +1,182 @@
...
+      break;
+    }
+    const frontier = lowestKey(stacksTxPages);
+    const safeBtcTx =
+      frontier === null
+        ? 0
+        : btcTxRemaining.filter(item => activitySourceItemKey(item) >= frontier).length;
+    if (countItems(stacksTxPages) + safeBtcTx >= limit) break;
+    token = page.nextCursor;
+  }
```
edgar-stacks: I think this type of code is way too hard to read 🤔 can we split parts of it into utilities by intent? Maybe we could also use maxStacksTxPageFetches.forEach instead of the for loop

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -51,319 +82,513 @@ export class BlockchainActivityService {
...
-            tx.contract_call.function_name,
-            signal
-          ),
-        ]);
-        return mapStacksContractCall(tx, account, events, protocolInfo);
+  ): Promise<ActivitySourceItem[]> {
+    if (!hasStacksAddress(account)) return [];
+    const items: ActivitySourceItem[] = [];
+    let cursor: string | null = null;
+    for (let page = 0; page < activityByAssetScanPages; page++) {
```
edgar-stacks, line 182: Do i understand correctly that this sends 10 requests with a for loop? Maybe we could use a Promise.all here?
How can we make sure this is not killing app's performance? 🤔

packages/services/src/activity/blockchain-activity.service.ts
```diff
@@ -51,319 +82,513 @@ export class BlockchainActivityService {
...
+
+    const changesByTxId = new Map<string, BlockchainActivityBalanceChange[]>();
+    for (const entry of balanceChanges) {
+      if (entry === null) continue;
+      const list = changesByTxId.get(entry.txId) ?? [];
+      list.push(entry.change);
+      changesByTxId.set(entry.txId, list);
+    }
+    return changesByTxId;
+  }
```
edgar-stacks, line 343: There is a lot going on in this function, can we split this up into utilities?

edgar-stacks (commented):
Can we split the packages/features/src/activity/blockchain-activity-view.ts into smaller utility function files?
Can we split packages/services/src/activity/blockchain-activity.service.ts file? It is currently ~500 lines of important logic

packages/services/src/activity/blockchain-activity.service.ts
aperry-stacks: > Maybe we could use a Promise.all here?

Unfortunately not. The new Stacks API `/balance-changes` endpoints are cursor-paginated.

> How can we make sure this is not killing app's performance? 🤔

This is specifically a for bounded-preview (AKA "recent activity") asset-specific lists, so the blast-radius is small. Testing this locally with the `activity-test.mjs` script benchmarks runtimes around 2-4s. Not great but acceptable (imo) for first render of asset detail pages.

packages/services/src/activity/activity-paginator.ts
aperry-stacks: It's cursor pagination that requires `break` statements so it's by nature a little verbose.

packages/services/src/activity/activity-paginator.ts
aperry-stacks: Fixed.

packages/services/src/activity/blockchain-activity.service.ts
aperry-stacks: This can't occur under the content-anchored cursor: a tx that confirms after page 1 was fetched gets a block time of roughly now, which is newer than afterKey — and pages 2+ only emit keys strictly below page 1's last item. So the confirmed row can't re-surface down-feed; it appears only on a top refetch, where React Query recomputes pages sequentially and the pending row transitions atomically to its confirmed position.

packages/services/src/activity/blockchain-activity.service.ts
aperry-stacks: Fixed.

packages/services/src/activity/blockchain-activity.service.ts
aperry-stacks: Fixed.

packages/services/src/activity/blockchain-activity.service.ts
aperry-stacks: Extracted a few of the more weighty non-async orchestration blocks. Should read a lot cleaner now.

aperry-stacks:
@edgar-stacks Comments are addressed. The one thing I haven't done yet is the decomposition of the activity view file. It's colocated with a bunch of outdated utils / helper files that I want to remove first to not muddy the directory. Right now it's a clean separation. When implementing activity in the wallet apps I'll do a clean-up pass and break that file apart.

packages/services/src/activity/activity-v2-design.md
```diff
@@ -0,0 +1,216 @@
+# Activity v2 — Bitcoin + Stacks activity rework
```
edgar-stacks: Do we need this md?

packages/services/src/activity/activity-unmapped-actions.md
```diff
@@ -0,0 +1,134 @@
+# Activity v2 — unmapped Stacks contract calls
```
edgar-stacks: and this one

edgar-stacks (dismissed):
sounds great, thank you very much!

packages/services/src/activity/activity-unmapped-actions.md
aperry-stacks: Nope. Both are working documents I added accidentally. Removed both.

edgar-stacks (approved):

packages/services/src/infrastructure/api/leather/leather-api.types.ts
```diff
@@ -3225,19 +3225,26 @@ export interface paths {
...
                   | 'swap'
+                  | 'bridge'
+                  | 'deposit'
+                  | 'withdraw'
+                  | 'claim-rewards'
                   | 'add-liquidity'
                   | 'remove-liquidity'
-                  | 'stake'
-                  | 'unstake'
+                  | 'stake-lp'
```
jannik-stacks, line 3239: what's `-lp`?

jannik-stacks (approved):
all in all good, 
i saw a lot of spread operators, typically okay, but could be interesting to look at if they slow things down for large activity entries

packages/services/src/infrastructure/api/leather/leather-api.types.ts
aperry-stacks: "Liquidity Pool" - As in, staking LP tokens.

PR #2476, opened 2026-07-01
feat(web): support custom private regtest network for multisig internal testing

leah-stacks:
Internal only. UI shows this indicator, but testing state is set through env var
Screenshot 2026-07-02 at 12 28 46 PM

Adds support for pointing the multisig web app at a self-hosted "Private" regtest network for internal testing. It's activated entirely by build-time env vars and reuses the existing *:testnet slot, so there's no new AuthNetworkId and no changes to the core models. When the private-network env vars are unset (all public and production builds), behavior is byte-for-byte unchanged.

Pairs with the workers-side stack on leather-io/workers https://github.com/leather-io/workers/pull/106, which serves the multisig backend locally and proxies aux data to production.


edgar-stacks:
- Fix bitcoin fee stimation for multisig transactions (it's 1-2 sats difference, but these do fail on testnet networks)
- attach correct custom network key to all rpc requests
- use stx multisig address for checking fee balance, not singlesig account

edgar-stacks (dismissed):

edgar-stacks (dismissed):

edgar-stacks (dismissed):

apps/web/app/constants/custom-network.ts
```diff
@@ -0,0 +1,50 @@
+import { z } from 'zod';
+
+// Custom "private" network config for internal testing (null in production).
```
aperry-stacks: Seems like we're using both adjectives "private" and "custom" for the same concept. We should pick one and use it across the board. (imo `custom` everywhere, no `private`).

apps/web/app/constants/custom-network.ts
```diff
@@ -0,0 +1,50 @@
...
+export interface CustomNetworkConfig {
+  flavor: 'regtest' | 'testnet';
+  name: string;
+  key: string;
+  bitcoinApiUrl: string;
+  stacksApiUrl: string;
+  stacksChainId: number | undefined;
+}
+
+function resolveCustomNetwork(): CustomNetworkConfig | null {
```
aperry-stacks: This isn't really a "network" that's being resolved here, it's the custom network "config". I would change the noun for the file and return type to be `customNetworkConfig` to be consistent and not collide with the actual app network settings.

apps/web/app/constants/custom-network.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (!privateNetworkFlavor || !privateBitcoinApiUrl || !privateStacksApiUrl) return null;
+  return {
+    flavor: privateNetworkFlavor,
```
aperry-stacks: Same naming gripe here as on the backend - the "flavor" concept is still a bit too ambiguous. It seems here like it's cross-chain, but I think at this point it only touches BTC-specific concerns. So could be much more specific, like `bitcoinNetworkMode` - basically the Bitcoin version of `stacksChainId`...

apps/web/app/constants/custom-network.ts
edgar-stacks: Renamed 👍

apps/web/app/constants/custom-network.ts
edgar-stacks: Renamed 👍

apps/web/app/constants/custom-network.ts
edgar-stacks: Renamed 👍 except the env variables, that would require a change to both workers and monorepo. We can do that as a separate pr

aperry-stacks (approved):

edgar-stacks (approved):

PR #2482, opened 2026-07-08
feat: multisig activity

aperry-stacks:
Adds a unified activity feed for multisig vaults. Each list reconciles on-chain history (Bitcoin + Stacks) against off-chain multisig proposals into one deduplicated, sorted stream, surfaced across the dashboard, vault, and account pages.

* Harmonization — matches broadcast proposals to their on-chain "twin" by txid so a signed-then-mined transaction appears once; synthesizes a view from the decoded payload for proposals still collecting signatures. Active proposals sort to the top.
* Three surfaces — dashboard (all active vaults, flat), vault page (one vault, top-10 preview), account page (one account, paginated with Load more).
* Protocol-aware rendering — resolves contract calls to a protocol + action label (swap, deposit, …); maps SIP-10 transfers, including the receiver counterparty.
* Pagination — infinite feed at the hook level over a cursor-paged on-chain source that merges a whole Bitcoin fetch with a token-paged Stacks stream (activity-paginator).
* Refactor — extracts the input-gathering logic into small intent-named, unit-tested pure helpers (build-multisig-activity-inputs).

### DIFF
--- a/apps/web/app/features/multisig/activity/use-vault-activity.ts
+++ b/apps/web/app/features/multisig/activity/use-vault-activity.ts
@@ -1,0 +1,132 @@
+import { useQueries } from '@tanstack/react-query';
+import { useUserSettings } from '~/hooks/use-user-settings';
+import { createBlockchainActivityViewsQuery } from '~/queries/activity/blockchain-activity.query';
+import { useMarketDataQuery } from '~/queries/market-data/market-data.query';
+
+import { btcAsset, stxAsset } from '@leather.io/constants';
+import type { AuthNetworkId, VaultAccountSummary } from '@leather.io/models';
+import { getMultisigService } from '@leather.io/services';
+import { isDefined } from '@leather.io/utils';
+
+import { useMultisigNetworks } from '../auth/use-multisig-networks';
+import { useSession } from '../auth/use-session';
+import { createMultisigAccountAddresses } from '../vaults/multisig-account-addresses';
+import { retryMultisigQuery } from '../vaults/use-vaults';
+import { multisigVaultKeys } from '../vaults/vault-query-keys';
+import {
+  type HarmonizedVaultActivity,
+  type VaultMultisigTransaction,
+  harmonizeVaultActivity,
+  selectTransactionIdsNeedingPayload,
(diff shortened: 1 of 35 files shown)


apps/web/app/features/multisig/transactions/decode-proposal-summary.ts
```diff
@@ -11,39 +11,92 @@ interface ProposalSummary {
...
+    transaction.proposalRawPayload
+  );
+  if (!payload) return {};
+  switch (payload.type) {
+    case 'btcTransfer':
+    case 'stxTransfer':
+      return { recipient: payload.recipient, amount: payload.amount, fee: payload.fee };
+    case 'contractCall':
+    case 'contractDeploy':
+      return {};
```
leah-stacks, line 98: I'm thinking about a not-yet-broadcasted contract-call proposal. We have the fee and amount in the data, but this would return nothing. I think we should return at least the fee - maybe the amount, too.

apps/web/app/features/multisig/activity/use-vault-activity.ts
```diff
@@ -0,0 +1,230 @@
...
+    isLoading,
+  };
+}
+
+interface UseVaultActivityResult {
+  items: VaultActivityItem[];
+  isLoading: boolean;
+}
+
+export function useVaultActivity(
```
leah-stacks, line 196: I've traced the preview assembly path and I'm pretty sure the home screen is fetching 50 but only displaying 10. The service default page size (`defaultActivityPageSize = 50`) is 50 per account, while the list only renders ~10 (`limit={10}` on `VaultActivityList`).

I'm thinking `createBlockchainActivityViewsQuery` could take a `limit` (like the infinite feed query does), so previews just get the most recent needed items - it's still a request per account, but this would reduce the response size and reduce the per-account decoding work.

leah-stacks (approved):
My comments are not blocking, they could be addressed in future branches. Overall this looks great.

edgar-stacks (approved):

aperry-stacks:
@leah-stacks 100% agree with your comments. Would prefer for them to be follow-on enhancements.

PR #2484, opened 2026-07-09
feat: migrate compliance check call to new service

aperry-stacks:
Moves the Chainalysis compliance check request out of individual apps and into shared services infrastructure. Implements service usage points in both `extension` and `mobile`.

Analytics events are still handled by apps directly, although this could be moved into services as well in the future. Adds a new event `compliance_check_unavailable` to capture and potential alert on compliance API failures.

Removes obsolete `@leather.io/query` version.

Updates Chainalysis API token.


apps/extension/src/app/common/validation/forms/compliance-validators.ts
```diff
@@ -1,27 +1,28 @@
...
 
 export function complianceValidator(
-  shouldCheckCompliance: yup.StringSchema<string | undefined, yup.AnyObject>,
-  network: NetworkModes
+  shouldCheckCompliance: yup.StringSchema<string | undefined, yup.AnyObject>
 ) {
   return yup.string().test({
     message: 'Compliance check failed',
     async test(value) {
-      if (network !== 'mainnet') return true;
```
edgar-stacks, line 15: i think we should rather keep this, we don't need to run the check on non-mainnet addresses

aperry-stacks: The network check is now in the [shared service call](https://github.com/leather-io/mono/pull/2484/changes/BASE..f659e0084ddcd4987286a8c2199754c304592077#diff-bc7ffbfab22eebe929e1e661f77bdbe8c228b1c03f829ac764663445ec4793acR29).

edgar-stacks (approved):
Great work!

leah-stacks (approved):

edgar-stacks (approved):

edgar-stacks (approved):

jannik-stacks (approved):

PR #2486, opened 2026-07-10
feat: multisig account recovery

aperry-stacks:
Adds automatic account recovery for multisig vaults. 

When the last member joins and a vault activates, the app immediately calls the new backend `/recovery` endpoint to find any accounts the member set already used on-chain (i.e. any threshold and at any index using gap limit scanning).

Recovery scans in Leather-native derivation index order on both chains, plus Asigna's ordering scheme on Stacks (Asigna BTC derivation style is incompatible with Leather), and creates them as ordinary vault accounts. 

Account creation is also reworked to match the backend's per-threshold index contract.

### Account recovery

- New services plumbing for the recovery endpoint in `LeatherAuthApiClient` and `MultisigService`. The endpoint is idempotent and returns only the accounts the call created.
- The FE owns the trigger: `useJoinVault` fires the scan when the join response's vault status flips to `active`. A retry affordance appears on failure; there is no standing rescan button.
- Scan state derives from the shared React Query mutation cache (`useMutationState`), so it renders wherever the user happens to be:
  - Vault page: "Scanning for existing accounts…" row; on failure, "Account recovery failed" + Try again
  - Dashboard: vault card caption swaps to "Scanning for accounts…"
  - On success: toast `Recovered N accounts in "Vault"`; empty scans are silent
- Recovered accounts are ordinary rows by design — no origin column or badge. Once renamed, provenance is meaningless.

### Account creation

- Indexes are per-threshold sequences. Create targets the **lowest unused** index for the selected threshold (previously vault-wide `max + 1`), filling gaps below sparse recovered indexes (e.g. an Asigna account at index 4).
- Index picking moved into `useCreateVaultAccount`: computed from the cached account list; on a 400/409 the hook refetches, recomputes, and resubmits exactly once — and only if the recomputed index differs (same index means the failure wasn't staleness, so it surfaces instead).
- Limits enforced in the UI: 10 accounts per threshold on both chains; STX additionally capped at min(10, N!) per the permutation space. Full thresholds render disabled in the picker; the create tile disables with a reason once every threshold is full.
- Create and recover both invalidate the vaults list, so dashboard `accountCount` captions stay fresh without a remount.


apps/web/app/features/multisig/vaults/vault-account-index.ts
```diff
@@ -0,0 +1,36 @@
+import type { AuthNetworkId, VaultAccountSummary } from '@leather.io/models';
+
+const maxAccountsPerThreshold = 10;
+
+function factorial(n: number): number {
+  let result = 1;
+  for (let i = 2; i <= n; i += 1) result *= i;
+  return result;
+}
```
edgar-stacks, line 9: should we move this to packages/utils?

edgar-stacks (approved):

apps/web/app/features/multisig/vaults/vault-account-index.ts
aperry-stacks: I asked Claude the exact same question during development and this was the pushback, which I thought was fair enough...
`On placement: I'd keep it where it is rather than promote it to @leather.io/utils/math. It's a 5-line private implementation detail with exactly one caller (accountLimitForThreshold); exporting it from the foundation package for a single consumer is speculative generality, and this repo actively polices unused exports (knip, lint:unused-exports). `

leah-stacks (approved):

PR #2487, opened 2026-07-10
fix: implement frameId for rpc requests

edgar-stacks:
use frameId to protect against a potentially hostile popup in the connected dapp


leah-stacks (approved):
Nice!

aperry-stacks (approved):

PR #2493, opened 2026-07-10
feat: multisig asset list and token details

aperry-stacks:
Adds a new Asset list component to the Multisig vault account page.

`STX`, `USDCx`, and `sBTC` are hoisted to the top of every list (even if account balance is 0).

Clicking an asset opens a token details modal, the design of which mirrors existing wallet app versions. 

Token detail modals have a "Recent Activity" list utilizing the new `BlockchainActivityService.getActivityByAssetId`.

Also, adds a multisig-specific custom tab component (instead of altering shared `ui` package version used by wallets).

https://github.com/user-attachments/assets/d60270c3-2233-41cc-ab41-e8ec82d44082

### DIFF
--- a/apps/web/app/features/multisig/assets/use-vault-asset-activity.ts
+++ b/apps/web/app/features/multisig/assets/use-vault-asset-activity.ts
@@ -32,9 +32,9 @@ export function useVaultAssetActivity(
 
   const views = useMemo(
     () =>
-      (query.data ?? [])
-        .filter(item => item.status !== 'pending')
-        .map(item => createBlockchainActivityView(item, { formatMoney: formatActivityMoney })),
+      (query.data ?? []).map(item =>
+        createBlockchainActivityView(item, { formatMoney: formatActivityMoney })
+      ),
     [query.data]
   );
 
--- a/apps/web/app/features/multisig/assets/use-vault-asset-activity.ts
+++ b/apps/web/app/features/multisig/assets/use-vault-asset-activity.ts
@@ -1,0 +1,42 @@
+import { useMemo } from 'react';
+
+import { useQuery } from '@tanstack/react-query';
+import { useUserSettings } from '~/hooks/use-user-settings';
+import { formatActivityMoney } from '~/queries/activity/blockchain-activity.query';
+
+import { type BlockchainActivityView, createBlockchainActivityView } from '@leather.io/features';
+import type { FungibleCryptoAsset, VaultAccount } from '@leather.io/models';
+import { createBlockchainActivityByAssetIdQueryConfig } from '@leather.io/queries';
+import { getAssetId } from '@leather.io/utils';
+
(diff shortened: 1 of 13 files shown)


apps/web/app/features/multisig/assets/vault-asset-items.ts
```diff
@@ -0,0 +1,80 @@
...
+function hasPositiveBalance(item: AssetListItem) {
+  return item.balance !== undefined && Number(item.balance.crypto.totalBalance.amount) > 0;
+}
+
+function compareVaultAssetItems(mode: VaultNetworkMode) {
+  return (a: VaultAssetItem, b: VaultAssetItem) => {
+    const priorityDiff = pinPriority(a.asset, mode) - pinPriority(b.asset, mode);
+    if (priorityDiff !== 0) return priorityDiff;
+
+    const fiatDiff = Number(b.fiat.amount) - Number(a.fiat.amount);
```
leah-stacks, line 55: These crypto amounts use different decimal scales, so comparing their `.amount`s can produce incorrect ordering. Could we normalize each amount by decimals before comparing?

apps/web/app/features/multisig/assets/use-vault-asset-activity.ts
```diff
@@ -0,0 +1,42 @@
...
+      getMultisigAccountAddresses(account),
+      getAssetId(asset),
+      settings
+    )
+  );
+
+  const views = useMemo(
+    () =>
+      (query.data ?? [])
+        .filter(item => item.status !== 'pending')
```
leah-stacks: Hmm, why hide pending transactions? Any downside to showing them?

leah-stacks (dismissed):

apps/web/app/features/multisig/assets/vault-asset-items.ts
edgar-stacks: I'd even say that comparing by crypto amount is generally unnecessary as every token has their own utility per quantity ratio. Comparing priority and fiat amount should be enough imo

edgar-stacks (dismissed):

apps/web/app/features/multisig/assets/use-vault-asset-activity.ts
aperry-stacks: Ah yeah, I was thinking there would be a conflict with the harmonized vault-account list items but it should be essentially the same. Removed the filter.

apps/web/app/features/multisig/assets/vault-asset-items.ts
aperry-stacks: Removed the crypto amount comparison entirely.

andresgalante-stacks (approved):

edgar-stacks (approved):

PR #2494, opened 2026-07-10
Multisig: improvements to activity feed + transaction details page

leah-stacks:
Polish activity feed display rows and proposal details pages. Add a new transaction details page for txs that do not map to multisig proposals (for example, receives and txs from recovered accounts).

Activity feeds
- You can now click into plain on-chain activity (like money you received, or history from a recovered account) and see a full detail page, even when there was never a proposal for it.
- When a combined activity feed is trimmed to the 10 most recent, a note is displayed "Open an account to view full history"
- UI polish: clickable areas for links, using real links instead of onclick, loading state

More readable transactions
- Contract calls (like swaps) now show what they actually are: the contract, the protocol, and the function, instead of just a "TRANSFER" label.
- Titles are more descriptive - i.e. "Send STX", "Receive BTC", "Add liquidity", or "Stack" instead of just "STX", so you know at a glance what happened.
- Actions done through a protocol show a "via Bitflow" / "via Stacking DAO" style subtitle so you know where it went.
- Swaps show a friendly "Swapped via Bitflow" line.
- If a transfer had a memo, it now shows on the proposal detail.
- Amounts and fees display like "1.00 STX ≈ $0.17"

Detail layout
- The activity detail page matches the proposal page's bordered-table style.
- The vault and account are displayed in a row with clickable links.
- UI polish like font size/color and alignment for addresses and transaction IDs, tx status shows as a badge/pill, signature status shows with the timestamp, gradient background row for proposals waiting

Misc
- A multisig proposed transaction stuck in "queued" shows an inline message explaining this one isn't yet able to be signed because it's stuck behind an active one.
- Fix some placeholder misnomers (network), prevent empty error toast when closing wallet, other little things

--

Up for discussion / candidates for follow-up work:

- I chose to add a vault/account row to the tx details page, with links, because I felt like navigating it all was a bit confusing without that context. Thoughts?

Screenshot 2026-07-10 at 4 47 09 PM

- In that same vein: in the combined feed, should we consider adding vault/account labels to the actual rows in the combined activity feed? It may be nice, if a user has several vaults, to be able to tell where each item came from at a glance.

- For the new non-proposal (read-only) detail page, I chose to reuse the proposal page's table style. (See screenshot below.) Flagging this for @fab-stacks to potentially refine the design as future work.

Screenshot 2026-07-10 at 4 53 14 PM

- Edge case: When you send from one multisig vault to another, and you look on the combined multisig dashboard, there are two individual rows displayed in the activity feed (one sent, one received). I thought this made the most sense - the 2 rows represent the same tx, but they would display 2 different tx detail pages.

Screenshot 2026-07-10 at 4 56 11 PM


apps/web/app/pages/multisig/tx/tx.page.tsx
```diff
@@ -59,6 +65,24 @@ function toFiat(money: Money | undefined, marketData: MarketData | undefined): M
...
+
+function proposalHeroTitle(
+  kind: 'transfer' | 'contractCall' | 'contractDeploy' | undefined,
+  functionName: string | undefined,
+  action: StacksProtocolAction | undefined
+): string {
+  if (kind === 'contractDeploy') return 'Contract deploy';
+  if (kind !== 'contractCall') return 'Transfer';
+  if (action !== undefined && action !== 'contract-execution') {
+    return humanizeProtocolAction(action);
```
aperry-stacks: Could this be replaced by the new `buildBlockchainActivityActionTitle()`?

aperry-stacks:
This looks great, awesome work 🦾

> I chose to add a vault/account row to the tx details page, with links, because I felt like navigating it all was a bit confusing without that context. Thoughts?

Yes, good call. Gives much-needed context to the Tx details page imo.

> in the combined feed, should we consider adding vault/account labels to the actual rows in the combined activity feed?

Also agree. My guess is this would quickly be an ask from users anyway and would make the combined feed much more useful / informational.  Perhaps vault names on the home page and account names on vault page?

-------

* Do we need the duplicate status badge in the header - It's already the first item in the details list.
image
* Not sure we need this informational banner here? Perhaps we could even shorten the combined list to like latest 8 or even 5 items (instead of 10) to show more obviously a limited combined preview.
image
* On the read-only details page, should we enforce a max-width? Otherwise the truncation look unnecessary.
image

edgar-stacks (dismissed):

apps/web/app/pages/multisig/tx/tx.page.tsx
leah-stacks: Yes good iea, there are a few uncovered cases which I will add to `buildBlockchainActivityActionTitle`

edgar-stacks (approved):

aperry-stacks (approved):

andresgalante-stacks (approved):

PR #2499, opened 2026-07-14
feat(extension): add wallet fingerprint to getAddresses

edgar-stacks:


aperry-stacks (approved):

aperry-stacks:
@edgar-stacks @leah-stacks Either here or in a follow-up we should make sure the now-available fingerprint gets mapped (instead of temp placeholder) during multisig sign-in (`wallet-sign-in.ts`).

edgar-stacks:
@aperry-stacks i'd suggest we merge this one and create a follow up for the web dapp (to not increase the testing surface of the pr)

leah-stacks (approved):

PR #2501, opened 2026-07-14
fix(extension): set correct app permissions for policy accounts

edgar-stacks:
Fixes a bug where signing a multisig policy transaction fails with an error instead of showing the approver popup.

When connecting to a dApp, we create an entry in the appPermissions slice containing the account data used for that connection. This entry never stored the policyId, so the extension had no way to correctly resolve which account the user had connected with — it could only fall back to the globally active account.

Repro:

1. Connect to multisig.
2. Add a Stacks policy via stx_addAccount.
3. Go to the Zest dApp.
4. Connect using the Stacks policy.
5. Propose a transaction using the Stacks policy.
6. Switch to multisig and refresh.
7. Try to sign → error messages instead of the approver popup.

Fix

Store policyId on the appPermissions entry so the extension can correctly select the account the user actually connected with, instead of relying on the active account selection.

### DIFF
--- a/apps/extension/src/app/pages/rpc-get-addresses/use-get-addresses.ts
+++ b/apps/extension/src/app/pages/rpc-get-addresses/use-get-addresses.ts
@@ -20,6 +20,7 @@ import { focusTabAndWindow } from '@app/common/focus-tab';
 import { useRpcRequestParams } from '@app/common/hooks/use-rpc-request-params';
 import { initialSearchParams } from '@app/common/initial-search-params';
 import { useFlags } from '@app/features/feature-flags';
+import { persistor } from '@app/store';
 import { useCurrentAccountId } from '@app/store/accounts/account';
 import {
   useCurrentAccountNativeSegwitPayer,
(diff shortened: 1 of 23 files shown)


leah-stacks (dismissed):

aperry-stacks (approved):

apps/extension/src/app/pages/rpc-get-addresses/use-get-addresses.ts
```diff
@@ -86,34 +87,36 @@ export function useGetAddresses() {
...
+    async onUserApproveGetAddresses() {
       if (!tabId || !origin) {
         logger.error('Cannot give app accounts: missing tabId, origin');
         return;
       }
 
       analytics.track('user_approved_get_addresses', { origin });
 
-      permissions.hasRequestedAccounts(origin);
+      const connectedPolicy = allowPolicyAccounts && currentPolicy ? currentPolicy : null;
```
jannik-stacks, line 98: ?? Operator is also useful

jannik-stacks (approved):

apps/extension/src/app/pages/rpc-get-addresses/use-get-addresses.ts
edgar-stacks: agree! i'll implement it in a follow up pr to unblock the release 🙏

PR #2502, opened 2026-07-14
feat: multisig sip10 transfers

aperry-stacks:
Enables SIP-10 token sends from multisig vault accounts. Sends utilize the existing transaction proposal modal with a new asset-selection dialog to pick from asset-list tokens.

Enhances `@leather.io/stacks` STX transaction decoding to identify and parse SIP-10 transfer contract calls. SIP10 details are decoded and details extracted, allowing for token amounts (crypto + quote) and asset to be visible in UI while pending / before broadcast.

Also adds send + receive buttons to the asset details sheet. Send branches to transaction propose with the asset pre-selected. Receive follows same pattern as extension and displays a QR code of the vault account address.

------
### Additional details: 

**Sending.** The send dialog gains an asset selector (toggle in the amount field opens a sheet listing the vault's sendable assets — positive balances plus STX). Amount validation is decimals-aware per token, fee estimation runs against the real contract-call draft, and self-sends are rejected. The new builder produces a `transfer` contract call wrapped in the existing non-sequential multisig envelope, always carrying a deny-mode fungible-token post-condition for the exact amount.

**Cosigner review.** Proposals are decoded and verified before anything is displayed. A proposal renders as "Send 100 USDh to SP2X…" only when the payload carries a deny-mode post-condition binding the vault to exactly the displayed amount on the called contract, the sender is the vault itself, and the token resolves in the token registry (with the resolved asset cross-checked against the post-condition's asset name). Anything that fails any of these checks renders as a generic contract call — the wallet never vouches for what it can't verify from the bytes. This applies to pending activity feed rows and the transaction details page, both with fiat values from market data.

**Account page.** Send and Receive buttons on the vault account page and on each asset's detail sheet — Receive opens a QR modal, and launching Send from a token sheet preselects that asset.

Package changes: `@leather.io/stacks` gains a verified SIP-10 transfer extractor and a `sip10Transfer` decode variant (fully covered by rejection tests for each verification gate); `@leather.io/queries` gains a SIP-10 asset-by-principal query config.


leah-stacks (approved):
Yesss, great work on this!

edgar-stacks (approved):

PR #2503, opened 2026-07-14
fix(extension): redux store sync between tabs

edgar-stacks:
## Summary

Each extension context—popup, full-page tab, and request window—has its own Redux store. The previous action-replay allowlist synchronized only selected mutations and allowed stale contexts to overwrite newer persisted state.

This PR replaces action replay with storage-backed, slice-aware synchronization so contexts converge without clobbering unrelated changes.

## What changed

- Track which persisted slices each context changes and merge only those slices into the latest stored state.
- Serialize cross-context persistence writes and preserve full writes for fresh storage and migrations.
- Hydrate clean slices from `chrome.storage.local` after rehydration, including a startup reconciliation for missed events.
- Preserve unflushed local edits while adopting changes to other slices.
- Flush persistence when a context closes or becomes hidden, and before BTC/STX add-account approval windows return success.
- Remove replay-action and account-change messaging in favor of direct Redux actions.
- Keep explicit lock, sign-out, and wallet-list messages for reloads and non-Redux cleanup.
- Make custom-network additions update existing IDs while rejecting duplicate keys.
- Add coverage for stale-context writes, concurrent edits, echo suppression, persistence migrations, sign-out propagation, and network updates.

## In other words, what got fixed:

- adding networks and/or app permissions won't mysteriously result in a lost network / app permission on page reload


leah-stacks (approved):

fab-stacks (approved):

314159265359879 (approved):
LGTM, overall this looks very stable, I have not seen it slip up and pop-up to sign with a different account/network then expected. 

tested switching account, dapp and active wallet in Leather to see if it every pops up the wallet to sign with anything other than the connected account. I have not seen this. 

When the wallet is switched to a custom network. Signing pop-up on a dapp that specifies mainnet, will pop-up a transaction for mainnet to connect (app.bitflow.finance for example) and it will also pop up a swap transaction sigining for the correct network. 

When a dapp does not specify the network the fallback is still whatever the active wallet is showing. I think that is okay and by design.

PR #2504, opened 2026-07-15
Multisig: fee selection, renaming vaults & account names, & other frontend improvements

leah-stacks:
A few improvements to the multisig experience:

- The activity feed now shows an "Awaiting your signature" flag on proposals you still need to sign.
- You can pick a network fee (Low / Standard / High) when proposing a transaction, with an info link to the fees guide.
- You can rename vaults and accounts from a small dialog. Vault names can only be changed by the vault creator; account names by any member.
- Removed a duplicate status badge on the proposed transaction detail page, to match the received tx detail page.
- Multisig transaction and activity data now auto-refreshes, so new proposals, signatures, and status changes show up without a manual reload.

Also fixes a shared UI bug where tooltips rendered behind sheets/dialogs (they now sit above), so it affects all three apps, not just web.


314159265359879:
We support a maximum of 15 accounts in a multisig. Lets make sure we stop allowing the user add more members after that:
image

leah-stacks:
@314159265359879 Good point, the cap at 15 has been added to this branch. Thank you!

aperry-stacks:
For the fee picker, maybe we can hide and/or disable the toggle until there are fee tier options loaded?
image

edgar-stacks (approved):

aperry-stacks (approved):

PR #2512, opened 2026-07-15
feat(web): edit multisig member names and show resolved bns names

leah-stacks:
Screenshot 2026-07-15 at 3 24 19 PM

Adds editing of vault member names, and shows a member's BNS name next to their name.

- Each member's name is now editable from the members list (creator can rename anyone in the vault; each member can rename their own). Uses the same rename dialog as vaults and accounts.
- The members list now shows each member's actual stored name (including yourself, marked with "(you)") instead of "Me".
- If a member's address resolves to a BNS name that differs from their stored name, it's shown in gray next to the name (Stacks mainnet only).

Also fixes a bug in the shared BNS service: `getAccountBnsNames` was returning nothing for addresses that own a name but haven't set a primary, because a slow/failed get-primary read and a cached failure could sink the result. It now fetches owned names independently and retries past a poisoned cache entry. This also benefits collectibles and vault creation.

### DIFF
--- a/apps/web/app/features/multisig/vaults/use-vault-mutations.ts
+++ b/apps/web/app/features/multisig/vaults/use-vault-mutations.ts
@@ -3,6 +3,7 @@ import { useMutation, useQueryClient } from '@tanstack/react-query';
 import type { AuthNetworkId, Vault, VaultMembershipResult } from '@leather.io/models';
 import {
   type CreateVaultRequest,
+  type UpdateVaultMemberRequest,
   type UpdateVaultRequest,
   getMultisigService,
 } from '@leather.io/services';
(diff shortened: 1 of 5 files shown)


edgar-stacks (dismissed):

andresgalante-stacks (dismissed):

andresgalante-stacks (dismissed):

fab-stacks:
@leah-stacks Nice! Only made a small change to the default icon color and pushed it right way, hope that works!

fab-stacks (dismissed):

edgar-stacks (approved):

andresgalante-stacks (approved):

PR #2515, opened 2026-07-16
Multisig: fix underestimated STX fees + clearer broadcast errors

leah-stacks:
Fixes multisig STX fees being estimated too low, and shows a clearer error when a broadcast is rejected.

- Fee estimation now accounts for the signatures a multisig transaction adds. `@stacks/transactions` only reserves signature space for sequential multisig, so our non-sequential multisig vaults were estimating a fee that was too low (roughly half), causing the node to reject broadcasts with `FeeTooLow`. Fees now scale with the number of required signatures.

- Broadcast/sign/cancel errors now show the reason from the backend, e.g. "Transaction rejected: FeeTooLow", instead of a generic "Leather API (...): 502". The API clients now attach the response body to the error so the message can be read.
Screenshot 2026-07-16 at 9 44 54 AM

- Error toasts now stay on screen a bit longer (6s instead of 2.6s). Success/info toasts are unchanged.

### DIFF
--- a/apps/web/app/features/multisig/transactions/use-vault-stx-transaction-fees.ts
+++ b/apps/web/app/features/multisig/transactions/use-vault-stx-transaction-fees.ts
@@ -40,7 +40,7 @@ export function useVaultStxTransactionFees({
 
   const draftTx = draftQuery.data;
   const feesConfig = draftTx
-    ? createStacksTransactionFeesQueryConfig(draftTx, settings)
+    ? createStacksTransactionFeesQueryConfig(draftTx, settings, account?.signers.length)
     : {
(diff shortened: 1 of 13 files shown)


werner-stacks (commented):
LGTM

werner-stacks:
my verdict was pre-mature. it is still suggesting fees that are to low from 0.000299 it was now suggesting 0.000318 and it is to low for the 3-of-12

werner-stacks (commented):
this latest fix resolved it for the 3 of 12 case!

edgar-stacks (approved):

andresgalante-stacks (approved):

PR #2524, opened 2026-07-17
fix: default to mainnet network and connected account

edgar-stacks:
Harden network and account resolution for dApp RPC requests

- Resolve the network for every RPC popup in one place with the precedence: explicit network param > network of the origin's bound policy > mainnet — the wallet's currently selected network is no longer used as a fallback, and per-handler defaults are removed.
- ⚠️ Breaking: reject unrecognized network values (including casing variants like 'Testnet') with INVALID_PARAMS across all RPC methods, instead of silently falling through to the wallet's current network. Valid values are the default network ids plus custom networks added to the wallet. dApps sending non-canonical network strings must update.
- Harden account binding: permissions without a valid integer accountIndex (pre-multiwallet installs) now fail cleanly with UNAUTHENTICATED instead of crashing the background handler; an explicit account: 0 is now honored as request-scoped; policyId is only forwarded alongside the stored account binding.
- Consolidate BTC/STX policy registration on the popup's resolved network (useCurrentNetwork) as the single source of truth, removing the duplicate params.network resolution, and add a network param to stx_signStructuredMessage.
- Add unit specs for rpc-request-utils, get-addresses, send-transfer, and sign-stacks-message (precedence order, unknown/empty/casing rejection, missing state, accountIndex edge cases) plus Playwright coverage for mainnet defaulting and bound-policy network proposals.


leah-stacks (approved):
Awesome work

werner-stacks (changes requested):
Lets not merge this yet. I tested a couple apps and it doesn't look like it is a default to specify the network currently for testing apps and then what you get is... Even when I am connected to the network I want to connect on a dapp, it doesn't work. (see example)
Reproduction steps with https://stx.fan/signer/03-deploy-manager.html and this custom stacks network API https://api.testnet-pox5.hiro.so
User clicks connect button, selects "connect Leather" and nothing happens. No error just returns to the dapp unconnected. ( see attached video)

I think we should do better. Instead of such an outright failure give them a choice, do you want to connect this app with the active wallet address/network or the default (mainnet accounts)?

Or leave this current behaviour (connecting to active network) atleast while a lot of people are testing for Epoch 4. We'd be breaking their apps in weird ways otherwise?

This change should also be accompanied by a developer doc that explains how dapp developer should specify the network in their dapp. 

Image

video:
https://github.com/user-attachments/assets/fe25f889-30d7-4f21-ae8d-65df19ac5798

PR #2525, opened 2026-07-17
fix(web): no testnet address in mainnet vaults

edgar-stacks:
Error out on testnet addresses in bitcoin mainnet vault creation flow:

testnet


leah-stacks (approved):

werner-stacks (approved):
Retested, LGTM

PR #2530, opened 2026-07-20
fix: clipboard in vault invites

edgar-stacks:
https://github.com/user-attachments/assets/54006ce5-8d15-470b-a740-6931eae49956


leah-stacks (approved):

werner-stacks (approved):
Tested, LGTM

Issue #2529, opened 2026-07-17
Design: Multisig invitations UI & consistent activity empty states


fab-stacks:
Two areas of the multisig web app that need proper redesign exploration (multiple approaches, not a tweak): the **invitation / pending-member UI** and the **activity empty states**. Grouped because both are about how incomplete states of the app look.

### Invitations / pending members

_Handled in #2532 — invite modal slimmed to a flat per-member list (copy message + copy link), pending-invite callout restyled._

### Activity / transactions empty states

- One consistent empty state across the dashboard, vault, and vault details — currently each looks different
- The empty state should always sit inside the same container the populated transaction list uses (as on the main dashboard)
- Refresh the empty-state visual — the current illustration dates back to very old Stacks days

**Refs:** sibling issues: #2527 (typography & value display), #2528 (vault details layout)

Issue #2528, opened 2026-07-17
Design: Multisig vault account details layout iteration


fab-stacks:
Next design iteration on the **vault account details** view in the multisig web app — a handful of contained layout changes to how actions and metadata are placed. Display-level refinements (type sizes, decimals, addresses) are tracked separately.

- Move the **Create Transaction** button into the **Transactions tab** instead of sitting above the tabs
- Renaming: replace the Edit button at the very top with an edit affordance on the details row (vault name + vault type) in the right-hand details panel
- Support changing the **vault icon** as part of renaming — currently only the name can be changed, the icon stays fixed
- Asset detail view: collapse the asset **description** to the first line + "Show more" — earlier feedback that the full description is boilerplate most people don't care about

_Status row iteration (look across vault states + pending-invite callout) handled in #2532._

**Refs:** sibling issues: #2527 (typography & value display), #2529 (invitations & empty states)

PR #2532, opened 2026-07-20
feat(web): playground design surface + multisig content refinements (#2527)

### LINKED ISSUE #2527: Design: Multisig typography, value & address display refinements
Next round of design refinements for the multisig web app, based on recent feedback and a review pass with a more active test vault. This issue collects everything about **how content renders** — type sizes, number formatting, and address display. Layout restructuring is split out into separate issues.

### Typography & sizing

The smaller sidebar scale is intentional (hierarchy: center column > sidebar), but the concrete sizes need another look.

- Revisit sidebar type sizes — stay on design-system font-size tokens (no custom values). One experiment: try the default size used for equivalent center-column items (e.g. vault members) and compare
- Sidebar account details: the address (code font) renders too large relative to surrounding text — tune down
- The center column can take more of its real estate: the create-transaction title/subtitle and the assets & transactions sections could be a step bigger
- Activity indicator icons in the sidebar are too small — size up
- Verify across multiple display settings (e.g. 43" 4K, 27" 4K with "larger text") — not just one setup

### BTC / STX value formatting

- Always show 8 decimals with padded zeros wherever a BTC balance is displayed (unless shown in sats)
- Define and apply an equivalent decimal convention for STX values
- ~~Show crypto amount first, fiat second~~ — **deferred**: the extension recently changed this ordering too, so hold off until we've settled it there

### Addresses

- Unify the address font: activity "Sent to" renders in the default sans while other surfaces use the code font — use the code font consistently
- Truncated addresses should still be copyable (copy the full address)
- Smart address sizing: addresses adapt to the space available — full address where there's room (preferable for security, e.g. signer rows with the flag layout and nothing on the right), sensible truncation when tight, while keeping reasonable spacing to neighboring text and the container edge

**Refs:** sibling issues: #2528 (vault details layout), #2529 (invitations & empty states)

fab-stacks:
**Activity feed**
- Crypto value leads, fiat beneath
- Row 2 always `vault account · threshold`
- Title carries the tx detail — sent-to / received-from counterparty, staking provider, inferred verb ("Swap via Zest")
- Review button on actionable rows
- Signature progress dropped in the account context (noise there)

**Value & digit display**
- Decimal precision unified in the shared currency formatter — every surface pads the same way
- BTC / BTC-pegged (sBTC, xBTC…) → fixed 8-decimal padding; other assets → adaptive cap
- Full precision in detail rows (no rounding surprises)

**Addresses**
- Adaptive mode — fills available width, truncates only when tight
- Activity counterparties shortened to first-3 / last-3

**Detail rows & hero**
- Body text primary color (was muted)
- Lighter inner dividers vs heavier outer container border
- Tx-detail hero synced to the real app + larger token avatar
- Squircle avatar radius now scales with tile size

**Invitations**
- Share / invite modal slimmed to a flat per-member list (copy message + copy link)
- Pending-invite callout restyled

**Playground**
- New dev-only `/playground` design surface — real components, not mockups
- Gated off prod; visible on PR-preview & staging for feedback
- New area = a folder + one registry entry

## Issues
- Resolves #2527
- Advances #2528 and #2529 — handled items removed from those tickets


apps/web/app/features/multisig/activity/build-multisig-activity-inputs.ts
```diff
@@ -95,6 +95,34 @@ export function buildContractActionTargets(
...
+// follow strong conventions, so a conservative keyword match still beats
+// rendering a verb-less "via {protocol}" row. Ordered most-specific first;
+// explicit registry data always wins.
+const inferredActionRules: readonly (readonly [string, StacksProtocolAction])[] = [
+  ['add-liquidity', 'add-liquidity'],
+  ['provide-liquidity', 'add-liquidity'],
+  ['remove-liquidity', 'remove-liquidity'],
+  ['withdraw-liquidity', 'remove-liquidity'],
+  ['stake-lp', 'stake-lp'],
+  ['unstake-lp', 'unstake-lp'],
```
leah-stacks, line 108: `unstake-lp` will never match here since `.find` will hit `stake-lp` first, so unstake txs will be labeled "Stake LP".

Suggested fix: Order ['unstake-lp', 'unstake-lp'] before this line (most-specific first)

leah-stacks (changes requested):
There's just the one small bug on the Unstake labeling I commented, but overall this looks great. I like the tiered activity feed 👍  And the decimal padding lgtm as well.

One more request: could you take a pass at trimming the comments here? A few feel unnecessary, and we lean toward minimal comments as a stylistic preference. Comments explaining "why" are worth keeping, but the ones that restate the code can be dropped.

leah-stacks (approved):

andresgalante-stacks (approved):

PR #2533, opened 2026-07-20
feat(web): content-rendering playground area (#2527)

fab-stacks:
## What

First playground area, stacked on #2532: an exploration surface for #2527 (multisig typography, value & address display refinements). One board per strand of the issue, each with independently linkable variants:

- **Sidebar typography scale** (`?type=`) — current scale vs the issue's named experiment (center-column parity), side by side with the unchanged center-column reference
- **BTC / STX value formatting** (`?btc=` / `?stx=`) — trimmed vs zero-padded decimals vs sats, on realistic balances with tabular numerals
- **Address display** (`?addr=`) — always-truncated vs always-full vs adaptive, one address rendered into three realistic container widths, using the real `CopyAddress` component

Boards render real components + design-system tokens with mock data. No production code is touched — when directions win here, they get promoted into the real multisig surfaces in follow-up PRs and this area is deleted (see playground README lifecycle).

## Try it

- `/playground/content-rendering` — defaults (current state)
- `/playground/content-rendering?type=parity&btc=padded&addr=adaptive` — all three experiments active

Base: `feat/web-playground` (#2532). Merges into the playground PR, not dev — GitHub retargets automatically when #2532 lands.

Refs #2527


fab-stacks:
Consolidated into #2532 — single branch/PR with clearly-labeled commits per track. The typography/values/addresses work now lives there as real component changes plus the content-refinements playground area.

PR #2534, opened 2026-07-20
fix: account name cap

edgar-stacks:
- Web: Implement account name cap (35 characters)
- Extension: do not reject if account name exceeds the limit, just truncate


leah-stacks (approved):

werner-stacks (approved):
Tested the web app feature: max length for account name 35, ✅ 
(do you want to show a countdown from 30 or so? something like (30/35)) I think it will indicate to the user this is working as intended and help them with precision if they need all the space we offer. 

I tried testing the wallet build too but that failed for another reason I will retest that once this is released.

edgar-stacks:
@werner-stacks good point! I will add it as a part of a separate PR as a nice to have. We need to merge this one to stop randomly failing rpc requests on long account names

PR #2535, opened 2026-07-22
feat: policy receive in send flow

edgar-stacks:
Add policy account (filtered by chain and network) in the send list

https://github.com/user-attachments/assets/238618ce-3852-43b1-910c-a7c8599c53be


leah-stacks (approved):

werner-stacks (approved):
Tested, love the way this works.

PR #2538, opened 2026-07-22
feat: extension proposal

edgar-stacks:
- Implement Send flow for multisig for both btc and stx chains using ledger/software wallets

https://github.com/user-attachments/assets/3e502634-5a8f-482a-9518-a86cdb73f7c7


fab-stacks:
This will create a transaction proposal in the multisig app, correct?

CleanShot 2026-07-23 at 15 11 03@2x

I'm inclined to suggest that on the screen here we add in a link next do "Done" to the transaction in the multisig web app. However, the annoying thing is that when the user would click through, it might get confusing, as the account the user is currently interacting from will not correspond with the account that needs to be connected in order to display everything in the multisig app as expected. wdyt?

edgar-stacks:
> I'm inclined to suggest that on the screen here we add in a link next do "Done" to the transaction in the multisig web app. However, the annoying thing is that when the user would click through, it might get confusing, as the account the user is currently interacting from will not correspond with the account that needs to be connected in order to display everything in the multisig app as expected. wdyt?

Yeap! They might be connected with an unrelated account.
I suggest we add this feature in the next iteration of the multisig, maybe we could think of a better integrated solution

leah-stacks (dismissed):

leah-stacks (approved):

andresgalante-stacks (approved):

PR #2539, opened 2026-07-22
feat: add pox-5 staking

### LINKED ISSUE #2550: PoX-5 (Bitcoin Staking) readiness
PoX-5 ("Bitcoin Staking") is a consensus hard fork that replaces pox-4 entirely — different contract, different function set, different reward model. At activation all STX locked in pox-4 unlocks and every stacker has to re-enroll through pox-5. This is the single place to see what that means for us, what is already in flight, and what is still open.

Most of the web work lands in #2539 (open, mergeable, pinned to the private pox-5 testnet). Individual items below link to the PR that covers them. Hosted preview: https://pr-2539-leather-web.wallet-6d1.workers.dev

The intent is one staking page: `/staking`, which is the URL going out in user-facing comms, with `/stacking` redirecting to it. That consolidation is **not done yet** and is bigger than it looks. `/` is not a page of its own — it redirects straight to `/stacking` through both a meta refresh and a client-side navigate — so redirecting `/stacking` creates a `/` → `/stacking` → `/staking` chain that races the meta refresh and breaks the fixture every e2e spec starts from. Consolidating means retargeting the homepage to `/staking` as well, and retiring the pox-4 specs that depend on landing on the old page. It was attempted in #2553 and backed out for that reason.

### Key dates

- Hard fork at Bitcoin block 960,230, expected 29 July around 03:00 UTC
- Re-lock deadline for existing stackers: block 962,050 (next prepare-phase start), roughly 12 August
- Whitelist opens 17 August
- Genesis bond 24 August, Bond 1 22 September, Bond 2 20 October
- First cycle after activation is STX-only staking

### Scope decisions

- STX-only staking keeps its page on the web app. Bitcoin staking is not replicated here — we link out to the staking app and to provider front ends instead, so Leather acts as a discoverability layer
- Providers running their own Bitcoin-staking front ends: Stacking DAO, Fast Pool, possibly Xverse
- STX-only staking stays a first-class priority alongside Bitcoin staking — the existing holder base participates through these pools and the first cycle is STX-only
- Passive-earning options consolidate under a single Earn entry point rather than separate top-level entry points
- Mobile responsiveness is de-prioritised for the staking and bond surfaces
- Moving Explore/Earn out of the wallet and onto the ecosystem site is a valid direction but parked until after activation

### Landing page and provider table

- Copy rewrite: the protocol renames "stacking" to "staking". Explainer and FAQ content describing reward slots, dynamic minimum thresholds and BTC streamed to a reward address is stale — #2539
- STX-only yield is now variable (the residual of the three-tranche waterfall after fixed-rate BTC bonds are paid) and denominated in sBTC — #2539
- Every pool integrates through its own signer-manager contract, so each one has a new contract ID. Pools without a migrated contract must not render — the list filters on presence of a contract ID — #2539
- A second listing criterion: only include pools that claim rewards on behalf of their members. Step 4 of the explainer ("rewards are claimed for you") is only true if the list is restricted that way, and restricting it looks like the right call
- Remove LISA (no longer actually providing liquid staking), Velar and ALEX
- Explainer step 3 copy: "Stake STX(-only) — stake your STX to the chosen pool"
- The "How long is my STX locked?" FAQ answer needs rewording. "Locks do not renew automatically" is technically true but reads as a warning, and at 96 cycles a lock that runs almost four years is closer to indefinite than to expiring. Framing it as how often you have to renew works better than framing it as how long you are locked
- Promo and education blocks referencing Dual Stacking need review — that is a pox-4-era rewards program and needs its own transition messaging. Those references should probably point at Bitcoin staking instead, though possibly not until it is publicly available
- Fixed 50,000 STX per-signer minimum (pool aggregate, not per staker) replaces the dynamic slot threshold — done in #2539

### Start staking flow

- The two-transaction delegate flow (`delegate-stx` plus `allow-contract-caller`) collapses into a single `stake` call carrying signer-manager, amount and number of cycles. Approval drops from three steps to two and pool membership persists across cycles — #2539
- Lock duration becomes a real user choice (1–96 cycles) and needs surfacing for the first time. Currently four presets plus a free numeric field — #2539
- Because a position can be unstaked at any point and unlocks at the end of the current cycle, duration is not really a lock — it is a renewal horizon. Working direction is to default to 96 cycles and demote the choice to an advanced option, which is the closest equivalent to pox-4's "indefinite". At 96 cycles a staker renews roughly once every four years, or whenever the PoX contract is upgraded
- BTC reward address becomes an optional payout preference (`{pox-addr, max-fee}`), default sBTC on Stacks. It has to be configured per pool, since only pools whose signer-manager supports L1 payout will honour it, and small positions are better off defaulting to sBTC to avoid L1 and UTXO consolidation fees — #2539
- Amount is locked directly, so the "delegate more than balance" semantics are gone — #2539
- Staking mutations are rejected during the prepare phase (last 100 blocks, roughly 17h, of every cycle). We notify rather than block, and locks take effect from the next cycle — #2539

### Active position and manage actions

- Position and status read from pox-5 staker state; the delegation-status model built on parsing delegate/revoke transactions no longer exists — #2539
- Claim rewards: rewards accrue as sBTC per cycle and are claimed through the pool's signer-manager. Needs a claimable balance aggregated across unclaimed cycles — done in #2539
- Claim rewards also has to be reachable from the homepage when the user is no longer staking with that pool, since rewards can be outstanding from an expired stake — #2539
- Unstake: STX unlocks at the end of the current cycle, all-or-nothing for the position. Not possible in pox-4 — #2539
- Update stake: one action covers extending, increasing and switching pools, with no cooldown cycle — #2539
- Pending-activity detection retargets from pox-4 function names to `stake` / `stake-update` / `unstake`. pox-5 emits topic-tagged events, which are a cleaner parse target than function names — #2539

### Liquid staking

- Entry flows call the protocols' own contracts so they likely survive, but Stacking DAO and LISA have to migrate their own protocols to pox-5 — their plans and any contract changes need confirming
- The increase flow breaks outright: it collects signer key and signature fields that have no pox-5 equivalent, because staker-side signer signatures are removed from the protocol
- Stacking DAO runs two liquid pools — one paying sBTC (claimable) and one accruing in STX. Only the STX one has ever been supported here. Current plan is to support non-liquid STX-only pools and link out to liquid staking sites initially

### Signer and pool tooling

- The signer key generation tool loses its purpose, since staker-side signer signatures no longer exist — remove it
- New signer management pages are needed for the hard fork, and possibly pool management. Existing throwaway apps and setup learnings have been shared as a starting point
- earn.leather.io stays up for at least two more weeks, then sunsets once signer and pool management tooling has migrated. A redirect needs the tools moved first

### Migration UX

- Around activation, detect STX unlocked from pox-4 and prompt re-staking
- Handle the transition cycle where pox-4 positions still display and pay out — stackers keep receiving rewards through the activation cycle and pox-5 becomes active the following cycle
- Dual Stacking keeps paying out through the gap and winds down coinciding with the Genesis bond on 24 August. Until then the link stays up, carrying a message that it is winding down and transitioning to Bitcoin Staking on that date, plus a path to the staking pools. The program itself continues afterwards in simplified form, as a budget allocated to apps that hit agreed on-chain activity metrics
- STX staked in Dual Stacking still unlocks at the fork like any other pox-4 position, so those users need to re-stake regardless of the August date — two separate things to communicate without conflating them

### Wallet-side readiness

- Staking post-conditions require the Ledger Stacks app 0.26.17 or higher, live since 23 July. Users signing staking transactions need to be told to update
- Clarity 6 contract deploys with Ledger work from Leather 6.106.3 upwards, which strips the Clarity-version payload as a workaround. From Stacks app 0.27.0 the payload works without it
- Optional on-device address verification for single-sig Bitcoin and Stacks — #2549. Multisig Stacks addresses cannot be verified until the Ledger Stacks app is publicly released and audited
- Network and account mismatch handling between dapp and wallet — #2524
- Custom private regtest network support for internal testing — #2476

### Cross-cutting

- #2539 is currently pinned to the private test network. Network constants, API URL and the production feature gate all need repointing to mainnet values before activation
- `@stacks/stacking` (we are on 7.0.5) is pox-4 semantics throughout — client methods, signature helpers and response types. If a pox-5-aware release lags activation we hand-roll the contract calls
- The pox-info shape changes (fixed minimum, no slot threshold). Claimable-rewards and weekly distribution-cycle clocks need to sit alongside the existing reward-cycle countdown
- Test fixtures and E2E flows are pinned to pox-4 contract IDs and response shapes and need rebuilding against pox-5
- Explorer surfaces need to show pox-5 state correctly — secondary to the staking pages
- Landing page assets for the bond app are needed by the Genesis bond

### Copy and UI follow-ups

Being picked up separately from #2539, as content-only changes PR'd into that branch. Deliberately excludes anything that changes what a transaction submits.

- Explainer step 3 reworded to "Stake STX (-only) — stake your STX to the chosen pool"
- Explainer step 4 softened so "claimed anytime" does not over-promise while the claim-on-behalf listing criterion is still open
- The lock framing appears in three separate strings that have to move together or they contradict each other: the duration helper, the first staking condition ("Your STX locks for the cycles you choose"), and the "How long is my STX locked?" FAQ answer. All three reframe around renewal horizon rather than lock length
- A "learn more" destination for users who want to read up on Bitcoin staking. https://www.stacks.co/bitcoin-staking is live and already serves as the hub — explainer, glossary, whitepaper, article feed — and links onward to docs and the forum. Worth confirming it is the canonical inbound target, and that the pox-4-era /learn/stacking page there gets updated or redirected
- Selected state on the duration presets — currently four identical outline buttons with nothing indicating which is active
- An explanation of why Dual Stacking went away, as a FAQ entry on `/staking` rather than a note on the old page, so it survives the redirect. Worded without dates so it holds regardless of how the gap question resolves
- Page title moves off "Bitcoin Staking" to plain "Staking". The page is the STX-only surface and Bitcoin staking links out, so the current title promises access that retail users do not have until community BTC pools exist. Needs sign-off from product and marketing, since "Bitcoin Staking" is the program brand

Also landed in #2553: the duration default moving to 96 cycles with the control relabelled as a renewal horizon, and the Dual Stacking wind-down block.

Attempted in #2553 and backed out: the `/stacking` redirect and the single sidebar entry, for the homepage-alias reason described at the top.

Still open: consolidating onto one staking page properly, filtering the pool list on claim-on-behalf, hiding duration behind an advanced disclosure (needs the e2e reworked), surfacing liquid token balances, removing the independent-stacking/signer-key link, repointing the network constants, and the custom-signer-contract row.

### Hydration bug in sanitizeContent

`sanitizeContent` in the UI package runs DOMPurify in the browser but falls back to manual escaping during SSR, and that fallback escapes `'`, `"` and `&` while DOMPurify leaves them alone. Any sanitized string containing one of those characters renders differently on server and client and throws React #418, which the e2e fixture treats as a failure. Every current FAQ answer and explainer step happens to avoid those characters, so it has never fired — writing "the pool's contract" in a FAQ answer broke five specs. Copy was reworded around it in #2553, but the trap remains for anyone writing sanitized copy containing an apostrophe, and it affects every consumer of that helper.

### Open questions

- Pool contract IDs can be collected before activation after all: a contract ID is just `address.contract-name`, and operators already know the address and name they will deploy from. So this is a coordination task rather than a hard post-activation blocker
- Unstake timing, derived from the pox-5 contract: unstaking is not immediate. It takes effect after the current cycle finishes, no stake mutations are possible during the last 100 blocks of a cycle, and the unlock lands at the start of the next cycle — the same timing as a stake update. Worth confirming against POOLING-05, which is P0 and still untested, but this is now a contract-level reading rather than an assumption
- Whether a BTC-only pool should make the reward-address field mandatory in our flow, since sBTC would never apply for it
- Whether the number of cycles affects anything beyond the renewal horizon — reward rate, weight, eligibility, or the ability to update the stake. If it does not, defaulting to 96 has no downside and the choice is genuinely advanced
- Liquid positions disappearing from the dashboard: if we list only non-liquid STX-only pools and link out for liquid staking, users who already hold stSTX will find their position gone. Showing the stSTX / stSTXbtc token balance is the minimal fix
- The site footer still links "Stack independently" at `earn.leather.io/choose-stacking-method`. Solo staking has no pox-5 equivalent and that host is being sunset, so the link dies twice over — it needs removing or repointing as part of the earn.leather.io wind-down
- Liquid "Start earning" buttons still route into the internal pox-4 liquid flow rather than out to the providers, and the increase flow is broken outright since it collects signer fields pox-5 removed
- Which pools support L1 BTC payout: Xverse confirmed, PlanBetter and Stacking DAO's non-custodial pool still to check
- Do we need to create and fund a wallet for the Ledger Stacks app multisig audit, or does an existing one cover it?
- Who registers the dust-amount positions in the Genesis bond for the wallets that cannot be tested on testnet, and when in the Genesis window does that need to happen?

### Reference

- Scope doc: [Update apps/web stacking pages for PoX-5](https://app.notion.com/p/Update-apps-web-stacking-pages-for-PoX-5-Bitcoin-Staking-3a4f958264358021a677eb393454789a)
- Test tracker: [Bitcoin Staking (PoX-5) Testing Plan](https://docs.google.com/spreadsheets/d/1eMBM1OqLMbz_Mxc_HCEjZnE9QGnBWY1ffSD9vj_24X4)
- Whitepaper: https://stx.is/bitcoin-staking-whitepaper
- SIP v1 draft: [forum thread](https://forum.stacks.org/t/introducing-the-bitcoin-staking-sip-v1-draft/18862)
- Reference implementation: `stacks-network/stacks-core` branch `pox-wf-integration` (`pox-5.clar`, `signer-manager.clar`)

edgar-stacks:
This PR implements a separate "staking" path, that allows users to stake in pox-5 pools. currently it is implement on "private" network, we will update it with prod values closer to the hard fork.

/stacking path is now removed, it is going to be redirected to /staking

Testing steps:
- Add testing network in extension
- - go to "Add network" tab in extension
- - then, in this order:
- - - name: (any name)
- - - bitcoin network "regtest"
- - - network key: "private"
- - - stacks url https://api.testnet-pox5.hiro.so
- Go to leather web build: https://github.com/leather-io/mono/pull/2539#issuecomment-5046098598
- Go to staking page
- Connect with any account that isn't staking right now. This is Account 12's signer-manager and only Account 12 is currently staking using that signer-manager. If you're going to unstake Account 12, make sure that you stake at least 50k to pass the staking threshold. (this is testing network specific, this isn't going to be an issue on mainnet)
- you should be able to see the staking position


fab-stacks (dismissed):

fab-stacks (dismissed):

aperry-stacks (commented):
@edgar-stacks 1 issue worth checking before merging:

**Blocker: amounts with >6 decimal places pass validation but crash on submit, with zero user feedback.**

`stxAmountSchema()` is only `z.coerce.number().positive()` (`start-staking/utils/staking-form-schema.ts:31`), so e.g. `0.0000001` validates; then `BigInt(stxToMicroStx(amount).toString())` at `start-staking/start-staking.tsx:152` throws `SyntaxError: Cannot convert 0.1 to a BigInt`. The throw escapes the submit handler and `Pox5SubmitError` only renders mutation errors, so nothing is shown.

Same bug in `update-staking/update-staking.tsx:271` — its regex `/^\d+(\.\d+)?$/` (schema at `:95`) allows unlimited decimals.

Fix: reject >6 decimals in both schemas (and add a spec case), or `.decimalPlaces(0)` before `BigInt`.

edgar-stacks:
@alexp3y nice catch! looking into it, should be a quick fix. Though i don't agree with it being a blocker, this is more of an edge case for someone to use more than 6 decimals

edgar-stacks:
fixed

aperry-stacks (approved):

fab-stacks (approved):

PR #2541, opened 2026-07-22
feat(web): multisig vault account details layout iteration (#2528)

### LINKED ISSUE #2528: Design: Multisig vault account details layout iteration

fab-stacks:
Layout pass on the multisig vault account detail view — where actions and metadata sit, plus how the asset description reads. Covers the four items from #2528.

**Create transaction**
- Moved the button out from above the tabs into the Transactions tab, where the action belongs

**Account rename & icon**
- Rename affordance moved off the page title onto the details-card row — a pencil next to the name
- Opens an `Edit account` sheet that changes the name *and* the icon; previously only the name could change, the icon stayed fixed from creation
- Icon grid renders inline in the sheet (not a floating popover) so it never clips in the short sheet
- Reuses the create-account picker via a shared `AccountIconNameField` + `GlyphButton` — no duplicated icon list

**Asset description**
- Collapsed to ~2 lines with a smooth fade over the second line — no ellipsis, no mid-glyph cut
- `Show more` / `Show less` toggle on the Description header row
- Only shown when the copy actually overflows

**Details row**
- Edit control is a proper square icon button (`size="sm"` → 32×32)

Resolves #2528


leah-stacks (dismissed):

leah-stacks:
Nice! Just one question on this - since account editing is now a modal, do we want to update vault renaming to a modal instead of an inline editable? I think having 2 different approaches is reasonable since the vault doesn't have an icon, but just wanted to check.

andresgalante-stacks (approved):

andresgalante-stacks (dismissed):

leah-stacks (dismissed):

leah-stacks (approved):

andresgalante-stacks (approved):

PR #2545, opened 2026-07-23
fix: clarity 6 contract deployment

edgar-stacks:
if contract version is unspecified or is version 6, strip away the contract version from contract deployment and let the node to default to 6:
https://explorer.hiro.so/txid/0x1dde1d34e6319c946ad11765160168a3bf13e0d865e1fb229accca2c0df82a90?chain=mainnet&api=https://api.testnet-pox5.hiro.so


leah-stacks (approved):

werner-stacks (approved):
tested on pox5 testnet, when clarity version payload is omitted wallet doesnt set arbitrary 0x03 for this contract anymore and it deploys 
https://explorer.hiro.so/txid/0x2fdb61916d397a71cc54d43f788ccd8da219aae109cc0f9081c3dc2769609e49?chain=mainnet&api=https://api.testnet-pox5.hiro.so 

If dapp tries to deploy with 0x06 payload to indicate clarity 6 the extension strips it? and the contract also deploys without issue
https://explorer.hiro.so/txid/0x3e51ae20328f1acd7ea829f0e39ef96b135333b5a227fe8869a1fd5d545bad03?chain=mainnet&api=https://api.testnet-pox5.hiro.so

PR #2549, opened 2026-07-28
feat: address verification

### LINKED ISSUE #2548: Allow user to verify address on Ledger device
Recent conversation covering this topic: https://stackslabs.slack.com/archives/C09U43UADGU/p1784883675647399?thread_ts=1784825551.296209&cid=C09U43UADGU

TLDR; implement non-blocking feature that allows verifying an address on device at the users request. 

For multisig stacks addresses this requires a change of the Stacks App (0.27.x or higher) we can test the draft once this feature is available. 

We can implement for
- Bitcoin (single sig)
- Stacks (single sig)
- Bitcoin (multisig)
- Stacks (multisig) feature flag to enable after the Stacks App feature is live?

edgar-stacks:
https://github.com/user-attachments/assets/39840350-44ce-4bf4-a502-09105c807a22


fab-stacks (approved):

werner-stacks (approved):
On a wallet where I have stacks and bitcoin connected in an account it looks great I hit the shield and can verify the address of a single sig, bitcoin and stacks. 

When adding a wallet policy (Bitcoin)
It shows the multisig address and says "confirm the address on your Ledger" but what it displays on the device first is the policy, the co-signing keys and I need to approve that first. the check on screen is "V Register Account", after that comes "Verify bitcoin address". 

I think we should display that in the pop-over
"Register account and verify address on your Ledger..."

here:
image


and here:
image


I understand that verifying stacks multisig address is not yet added here.
I have a ledger with 0.27.0, do you also want me to test verification of the multisig Stacks address, want to create it in a seperate PR so it can be merged when Stacks App for Ledger 0.27+ is live?

edgar-stacks:
> I think we should display that in the pop-over
"Register account and verify address on your Ledger..."

Good point!

PR #2554, opened 2026-07-28
feat(extension): new swaps flow with sBTC bridging

### LINKED ISSUE #2571: Finish updated swaps functionality

edgar-stacks:
Replaces the legacy swap page with the shared swap state (@leather.io/state/swap) and adds BTC⇄sBTC bridging through Emily. Roughly a third of the diff is legacy deletion; the rest is the new flow plus hardening from three ai review rounds.

Extension
- New swap form/review on the shared state; legacy swap-legacy/ and the swap-revamp flag removed (swapsEnabled remote config is the only kill switch).
- Ledger signing mounted under the swap review for both chains; a device deny returns to review, other device errors surface as a failure with the real message.
- Route assets resolved by asset id (duplicate tickers no longer match by symbol), BTC swap pill gated by the disabled-pair rules, compliance screening under a swap screening point.

Shared packages
- sBTC deposit: bitcoin broadcast/notify moved out of the abandoned sbtc client into services, retry + notified/uncertain result union, per-network 30s cache for Emily limits, native-segwit-only UTXO selection, supply-cap constraint.
- Shared SwapSigningCancelledError so cancellations aren't counted as failures in analytics; BTC pairs disabled off mainnet for both apps.

Mobile — only the changes the shared API forces (dependency wiring, constraint case, sbtc dep removal). The mobile hardening (review-screen state, sheet lock, biometric cancel) is in a follow-up PR based on this branch.

Testing — unit specs across extension/state/services; manual on hardware: Ledger BTC→sBTC and STX swaps, deny/unplug paths, taproot-holding account.


werner-stacks:
There are so many improvements in here I can't wait to see this live. 

A couple issues I ran into while testing:

swapping stx token to meme token:
image

with this pair it took an unusually long time to load the USDA amount. I'd say a couple minutes at least. After the first time though when I changed the input amount it would quickly calculate/load the opposing pairs value. 
image

When routes are not available should I be able to select the option?

If I start with LiSTX (from token), I see no options to select. I think that is correct. 
image


When I try this instead, I can select options to swap to but I get a similar message after that. 
image


Would it make sense to also mention something along this line:
Visit the issuers website/dapp for more info or other options.

LiSTX for example is a project that is winding down and the only thing users can do now is convert it back to STX. 

for stSTX to stSTXbtc i am not sure if there are changes due to PoX-5 but via the platform you could convert them, I think it may still be the case. 

I see the same issue on bitflow with stSTX stSTXbtc so perhaps routing changed and it is not yet implemented. Probably worth checking with the team. 

Since this is all minor and the implementation is has so many improvements I think it is good enough to get released.

werner-stacks (approved):
LGTM. I think swap could do better at helping users only select tokens that have available swap routes, I still see some issues with it. Something for the followup to this.

aperry-stacks:
I'm seeing the same random-ish performance issues as @werner-stacks but I will track those down separately and propose a follow-up fix.

I think overall the functionality is still a massive improvement as-is.

@edgar-stacks Here's a couple of quick UI nits, maybe it makes sense to knock these out now? I'll approve anyway so feel free to just tag these on a follow-up PR.

<hr/>

1. The "Continue" button needs x-padding:
image

<hr/>

2. There's a large bottom margin section that occludes a significant portion of the confirmation screen:
image

<hr/>

3. The BTC/STX chain indicators on tokens in the asset selectors have a black background instead of the yellow / orange we use elsewhere:
image

<hr/>

4. Weird / default font on the slippage editor (?):
image

aperry-stacks (approved):

leah-stacks (approved):

edgar-stacks:
@alexp3y thank you for these findings! I'm going to address those in a follow up PR to make the process faster

PR #2557, opened 2026-07-29
feat(web): pox-5 staking status route, cycle status and UI harmonization

fab-stacks:
Follow-up to #2553, picking up what came out of Tuesday's 4:30 staking session and Werner's notes in #stacks-leather. All pre-fork, based on `feat/pox-5-stacking` since that hasn't landed in dev yet.

Werner asked for two things and both are here. **Your position** now takes the top of the staking page instead of sitting under the marketing hero — if you have a stake, the "Earn Bitcoin yield with your STX" pitch and the hero card are replaced by a **Your position** section using the same header structure as **Staking pools** and **Liquid Stacking**. While the position scan is still resolving the discovery layout stays put, and progress shows as a **Looking for positions** pill beside the wallet control that fades out once it settles. He also wanted a link he could hand out without knowing someone's pool, so **app.leather.io/staking/status** resolves the pool from the position's signer principal and redirects there. A custom signer-manager we don't list lands on the overview, where the position now says it can't be managed here and shows the contract id — Werner hit exactly that case and had to guess why.

On duration, the default drops to **48 cycles** and the control stops reading like a lock: the renewal date moved into the input's label row, and the reassurance became an **Exit any time** block attached below it. Cycle state is also visible before it bites — **Next cycle** now says when staking closes, amber inside 48 hours and red once paused, and the cycle queries poll so the page stops insisting staking is paused after it has reopened.

- **Duration** accepts digits only, capped at two, keeping the inline range error
- Back arrow and wallet on all three `/staking/pool/*` routes, which had no header at all
- Info tooltips on each pool-overview cell, wording taken from the reviewed FAQ
- One 8px radius across hero, explainer steps, tables, callouts and position cards
- Tables boxed in, with headers aligned to their content regardless of the data
- Real sBTC and Bitcoin token icons instead of monochrome logos, and a duplicated "sBTC sBTC" gone
- Body background set app-wide, which was letting the browser canvas show through around the nav

Every state is previewable without a wallet at **/playground/staking-states** — each cycle stage, the start-staking form with live validation, and an active position. It's registered as an `exploration` area so it gets deleted when this lands. To see the real page with a position instead, set `leather-mock-mode` and `leather-mock-pox5-staked` to `true` in localStorage, reload, then Connect → Resolve.

48 cycles is per Daniel's follow-up: the tenure budget issue is addressed, 96 cycles works out to roughly 7-8% of the tenure budget (about 12-13 staking transactions per tenure), and he still recommends defaulting to ~2 years.

Two things I'd like eyes on. The **Fee** tooltip says each pool sets its own fee in its signer-manager contract and to check their terms — I cut a stronger line about operators being able to change fees, because my only source was the audit thread rather than anything approved for users. And the **Liquid Stacking** heading still says Stacking: it's a shared learn-article string the pox-4 page uses too, and the body copy says "liquid stacking tokens", so renaming it means moving all three together.

### DIFF
--- a/apps/web/app/components/basic-hover-card.tsx
+++ b/apps/web/app/components/basic-hover-card.tsx
@@ -1,22 +1,38 @@
 import { ReactNode } from 'react';
 
-import { styled } from 'leather-styles/jsx';
+import { Stack, styled } from 'leather-styles/jsx';
 
 import { HasChildren, HoverCard } from '@leather.io/ui';
 
 interface BasicHoverCardProps extends HasChildren {
+  title?: ReactNode;
   content: ReactNode;
   align?: 'start' | 'center' | 'end';
 }
-export function BasicHoverCard({ children, content, align = 'center' }: BasicHoverCardProps) {
+
+// HoverCard.Content already supplies the padding and max width, so nothing here
+// adds its own — a second layer is what made these read as over-padded.
+export function BasicHoverCard({
+  children,
+  title,
+  content,
+  align = 'center',
+}: BasicHoverCardProps) {
   return (
     <HoverCard.Root openDelay={220}>
       <HoverCard.Trigger>{children}</HoverCard.Trigger>
       <HoverCard.Portal>
         <HoverCard.Content side="top" align={align}>
-          <styled.p textAlign="left" textWrapStyle="pretty" color="ink.text-primary">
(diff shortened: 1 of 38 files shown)


edgar-stacks (approved):
Looks amazing!

PR #2558, opened 2026-07-30
fix(web): rework staking payout selection, hover cards and form radii

fab-stacks:
Design pass over the pox-5 staking form and the hover cards, picking up the leftovers from #2557 plus the review notes from the Slack thread. Now based on `dev`, since #2539 has merged.

**Payout selection**

- Replaces the checkbox with two radio cards — real sBTC/BTC token icons, sBTC first and default, a 2px action-primary outline for the selection (1px border + 1px inset ring, so the box doesn't shift on select), a **Recommended** badge, and the address / max-fee fields revealed under the group. `update-staking` reuses the same component, so it gets all of it too.
- Pools that can't pay in BTC now show the option **disabled with a reason** instead of hiding the section. Worth doing rather than cosmetic: StackingDAO's `validate-stake!` accepts a Bitcoin address and silently ignores it, so a hidden option meant a user could pick BTC and receive sBTC.
- Radios share a `name` inside a `role="radiogroup"` with a visible focus ring, so arrow-key navigation and screen-reader announcement work.
- The badge hides via a container query when the card is too narrow to fit it beside the token name.

**Cycle timing copy** — was one phrasing at every distance, now four tiers: `Staking closes in 11 days` → `Closing soon · 5h left` → `Closing within the hour`, plus `Staking paused · reopens within the hour`. The sub-hour cases drop the number because `humanizeSeconds` floors at `1h`, so a 4-minute window used to read "1h left". Added a matching stage to the staking-states playground.

**Layout and spacing**

- **Border radii aligned** on the form — 3 values instead of 4, and nesting now predicts radius. Form-flow elements (inputs, payout cards, inline callouts) at `sm`, page-level containers (pool matrix, confirmation panel) at `md`, pills `round`. Inputs were `xs`, so the largest boxes on the page had the smallest radius.
- **Exit any time** now reads as one block with the duration input: the input stays fully rounded and opaque, and the callout tucks 4px behind it so the grey fills the corners instead of a white notch. When a validation error appears the two separate cleanly with equal spacing and the callout regains its top corners.
- The input's focus ring grows inward (`inset: 0`) rather than 1px outward, so it no longer overshoots onto the callout.
- Fixed a dead band at **768–1023px** where there was neither a Review button nor a side-by-side layout, leaving the confirmation panel stranded in a narrow column. Both are now keyed to the `lg` switch the layout actually uses.

**Pool card** — logo left, name right, explainer pinned to the bottom, underline removed, and the updated Fast Pool mark (replacing the orphaned `.webp`).

**Copy** — `Stacking` → `Staking` across headings, tooltips and support copy, including `Liquid Stacking` → `Liquid Staking`. References to the *old* protocol keep their name (`successor to Stacking`, `What happened to Stacking (pox-4)?`, `Dual Stacking`), as do `Stacks` and `Stacking DAO`. Article `id`s and `slug`s are untouched since they route to `/support/<slug>`. Also took Werner's note: "claimed anytime" → "once the cycle concludes", plus "your pool usually claims for you".

**Hover cards** — the info icon sits inline beside its label at a consistent 4px everywhere. `LearnInfoHoverIcon` is deleted; it was the one component #2557 missed, double-padding its tooltip to 32px. `ValueHoverCard` now uses `LearnHoverCard`. "Learn more" no longer breaks across two lines — the culprit was a leading `&nbsp;` that made the gap inside the label the only available break point.

**Dark surfaces** — the hero card is marked `className="dark"` rather than each component being told what background it sits on, so every semantic token inside resolves to its dark variant. The info icon there goes from 3.4:1 to 14.5:1 against black. This let the unused `iconColor` prop go from three components.

One thing to flag: `Input`'s radius lives in `packages/ui`, so this moves every input in the extension from 2px to 4px as well.


edgar-stacks (dismissed):

fab-stacks:
Both review findings addressed in `dfe0931`:

1. **Radio group semantics** — both inputs now share `name="payoutPreference"` inside a `role="radiogroup"` with an `aria-label`, and the label gets a visible `_focusWithin` outline. Verified with a real ArrowRight keypress: selection moves, the ring applies, the dependent fields reveal and the helper copy swaps.
2. **Stale state in the update flow** — sBTC now reads as selected whenever BTC isn't actually available (`!value || !supportsBtcPayout`), matching what the schema and submit path already do. Confirmed exactly one option is selected on StackingDAO, where `supportsBtcPayout` is false.

edgar-stacks (approved):

aperry-stacks (approved):

PR #2568, opened 2026-08-03
feat(web): staking UI pass — position rows, provider table, shared address chips

fab-stacks:
https://github.com/user-attachments/assets/c94319d3-30f8-40f8-bea3-5b695fd791e8

UI pass over the pox-5 staking surfaces that landed in #2539–#2567, plus the design-system fix most of it kept tripping over: the square avatar radius was a hardcoded 10px at every size, so the same pool mark read as a different shape in the table (24px) than beside a balance (40px). It now scales at 20% of the tile in `Avatar` itself, and multisig's `AvatarSq` derives from the same ratio instead of carrying its own.

- **Your staked position** is one row for every position kind: listed pools show their brand mark, unlisted signer managers get a contract glyph with the **contract id as the identity line**, and the pending state sits on the same layout with a spinner in the avatar slot. **Claim rewards** lives here now (and only when there is something to claim) instead of breaking the provider table's layout — which also fixes claiming being silently unavailable for byosm positions.
- Provider table rows are **clickable** with a full-bleed hover, and only when they lead somewhere — while a position is open, other pools' rows stay inert, sharing one `useStakingPoolLink` decision with the button so the two can never disagree.
- Multisig's `CopyAddress` is promoted to `~/components` and now serves every address in staking too, with new **compact**, **underlined** (border-drawn, so it spans the truncation ellipsis) and **wide** grow/shrink variants plus an `ExternalAddress` sibling; the old stacking address module is deleted. **Contract Details** aligns both rows on a shared grid column.
- **Rewards history** toggles with a chevron left of Claim rewards and renders on a subdued background with dividers, cycle labels in regular text colour.
- Drive-bys: Stacks icon for the **Staking** nav item (was the sBTC glyph), `Page.Content` top spacing for the staking sub-views that sat flush against the header, and a nested-`<button>` fix in `Page.Header`.
- The **staking-states playground** now covers all 25 pox-5 surfaces with seeded query caches (`staleTime: Infinity` + offline mode, so no fetch can land on a seed) — the components that ship, only the data mocked.


fab-stacks:
Addressed in d57a9b9 (squashed into the existing commits):

- **2 — `useStakingPoolLink` during the position scan**: the hook now returns `{ isLoading, to: null }` while loading, so rows are inert and the button sits disabled/`aria-busy` until resolution — nothing is offered as a link that resolution could redirect away from, and the "can never disagree" comment now holds.
- **4 — nested interactives**: the `backTo` branch of `Page.Header` now renders the same plain `Button` (navigating via `onClick`) as `onBack`. Same fix applied to the other two instances of the `<Link><Button/></Link>` pattern in the staking surfaces (`StartStakingButton`, position row) — the staking pages now measure **zero** nested interactive elements.

Not changed, deliberately:

- **1 — multisig radius**: intentional and explicitly requested — unifying the square-avatar shape app-wide was the instruction, and the visual delta (14px → 8px on 40px tiles) is flagged in the PR body for exactly this design check.
- **3 — "Show history" naming**: the label wording was chosen deliberately; agreed the single-cycle case reads oddly, but that's a copy decision to revisit rather than gate on `byCycle.length > 1` silently.

edgar-stacks (dismissed):

aperry-stacks (changes requested):
We have an existing convention in `mono` to only add comments when absolutely necessary. This rule is enshrined in the repo's CLAUDE.md, which seems to be getting overridden / ignored here as the PR diff is ~15% new comments.

Either we should discuss whether we want to change a longstanding codebase convention or there should be a good rationale why these comments are necessary.

fab-stacks:
@aperry-stacks Sorry should have checked more carefully - Funnily I already course-corrected this earlier in a different PR, I'll make sure to make it more sticky

edgar-stacks (approved):

aperry-stacks (approved):

PR #2580, opened 2026-08-03
fix: add @stacks/connect

### LINKED ISSUE #2577: Make app.leather.io wallet agnostic. Add @stacks/connect

edgar-stacks:
Add @stacks/connect to app.leather.io. This should make the leather earn website wallet-agnostic.

https://github.com/user-attachments/assets/a91e6d83-d4f5-45ee-b781-b7e062500c18

### DIFF
--- a/apps/web/app/app.tsx
+++ b/apps/web/app/app.tsx
@@ -3,7 +3,6 @@ import { Outlet, useNavigate } from 'react-router';
 
 import { QueryClientProvider } from '@tanstack/react-query';
 import { queryClient } from '~/constants/query-client';
-import { ToastsProvider } from '~/features/toasts/toasts-provider';
 import { useDetectLeatherProvider } from '~/store/addresses';
 
 import type { LeatherProvider } from '@leather.io/rpc';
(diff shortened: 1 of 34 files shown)


aperry-stacks (dismissed):
Approving if you want to get this in, although #2 from the claude review looks like it might be worth fixing as it could lead to an annoying bug.

Another description here:
`revokeWalletPermissions()` returns `true` for any selected non-Leather provider — it returns `true` from both the success path and the `catch`, so it means "we attempted revocation", not "the session was revoked". Meanwhile `connect()` from `@stacks/connect` always sets `forceWalletSelect: true`, so "Switch account" reopens the picker every time, and closing it rejects with -31001. That lands in the catch → `disconnect()` + `setAddresses([])`.

werner-stacks:
I see the wallet agnostic pop-over come up on staking but not on multisig, is that missing?

leah-stacks (dismissed):
Looks good & works for me, but also echoing what Werner said - the wallet-agnostic popup does not appear for Multisig

leahjlou:
The wallet-agnostic popup intentionally only appears on the staking website.

aperry-stacks (approved):

andresgalante-stacks (approved):

PR #2606, opened 2026-08-06
feat(services): replace chainalysis with leather api compliance endpoint

aperry-stacks:
Moves address compliance screening from direct Chainalysis calls to the Leather API compliance endpoint, and makes the compliance analytics events precise about what was screened and where. Screening behavior is unchanged for users: compliant / non_compliant / unavailable, mainnet only, fail open when screening is unavailable.

This PR:

- points `ComplianceService` at the Leather API compliance endpoint and deletes the Chainalysis client along with its embedded API token
- normalizes every failure to `unavailable` with a `reason` derived from the endpoint's per-check results (`ofac`, `risk_screening_chainalysis_rate_limited`, `leather_api_http_500`, …), so the service never throws and the unavailable event always fires
- fixes `non_compliant_entity_detected` to name the address that actually failed screening — it previously reported the caller's input, which could be a different address than the flagged one
- adds `screeningPoint` to both compliance events identifying the emitting surface (send forms, RPC flows, PSBT signer) and tightens `address` to a single string


aperry-stacks:
re: claude review

1. fixed — specs now mock `/v1/compliance/addresses/**` instead of the dead chainalysis routes, and `mockLeatherApiRequests` gained a default compliant response so the rest of the suite stops making live compliance calls in CI.
2. fixed — service rethrows when `signal?.aborted` instead of normalizing to `unavailable`. fwiw the analytics-pollution claim didn't hold: only the RQ query path passes a signal, and RQ discards the post-abort resolution before it's committed, so no events were firing on navigation. fixed anyway — misclassifying a caller cancel as "api unavailable" is wrong regardless.
3. leaving as-is — reasons are always read alongside `status` and are for dashboard eyeballs, not parsing.

fab-stacks (approved):

edgar-stacks (approved):

PR #2609, opened 2026-08-06
fix: propagate backend error details to multisig app

aperry-stacks:
Shows the backend error message to the user when a multisig action fails, instead of a bare status code.

The API client already received error details from the backend but dropped them from the message, so every failure surfaced as an opaque status code — and several vault actions failed with no feedback at all.

This PR:
- includes the backend error detail in `LeatherApiError` and adds a `getErrorDetail` helper
- shows the detail in the web multisig surfaces (propose, create vault, create account, sign-in)
- adds error toasts to vault rename, member rename, vault cancel, and account edit, which failed silently
- shows the detail in the extension propose sheets
- says "propose" instead of "broadcast" in the error sheet when the transaction is a proposal


edgar-stacks (dismissed):

leah-stacks (approved):

werner-stacks (approved):
Thanks, Yes this is a more meaningful error "Transaction Rejected: BadNonce"

Would you mind also adding the nonce to the transaction details or saving that for later?

Image

PR #2610, opened 2026-08-06
feat(extension): add biometric unlock

stacks-alex:
## Summary

- add WebAuthn PRF-backed biometric unlock for software wallets in the Chrome extension
- support both password-backed enrollment and biometric-only first-wallet creation
- add biometric authentication to unlock, Secret Key reveal, and add-wallet flows
- add settings for enrollment, credential identification, password transition, and disablement
- serialize and verify authentication-state persistence across extension pages
- add a reusable real-provider qualification harness plus unit, persistence, and Playwright coverage

## Security and behavior

- require user verification and pin evaluation to the exact enrolled credential
- wrap the wallet encryption key with AES-256-GCM using the 32-byte WebAuthn PRF output
- validate persisted authentication state and durable postconditions before initializing a session
- fail closed for malformed, stale, missing, or mismatched credential state
- use a non-identifying registration tag so multiple Leather credentials can be distinguished in the platform provider

## Validation

- `pnpm format`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm knip`
- `pnpm --filter @leather.io/extension lint:unused-exports`
- extension unit suite: 106 files / 712 tests
- biometric production-flow Playwright suite: 3 scenarios
- U0 provider harness Playwright suite
- targeted persistence and cross-frame tests
- normal Chromium and Firefox extension builds
- wallet-lock regression Playwright suite
- real macOS qualification with Chrome, Apple Passwords/iCloud Keychain, and Touch ID

## Rollout note

The macOS development gate is complete. Windows provider qualification is intentionally deferred to external testing and compatibility is not claimed by this PR. The U0 harness is included to support that qualification.


edgar-stacks:
Thank you for the contribution! Being able to open the wallet using biometrics is a very nice _touch_ to the overall experience of the wallet.

This PR has a good chunk of seed security critical changes so i'd suggest we cut the scope to make sure we can focus on what's important.
A few things that caught my eye:
1. I believe we shouldn't allow both password and biometrics for the same wallet, for the scope of this PR i would suggest we simplify this and make users choose either one or the other.
2. I also think they shouldn't be able to change the biometrics in settings. That shouldn't be a popular flow for the users as they are going to have the biometrics stored locally on the mbp and they can only unlock the wallet using that mbp's touchId.

Dropping here my opus5 ultracode + fable xhigh review of the project. H1, M1, L1-L5 do need to be addressed + some of the top design&architectural notes. (L4 would close by itself if we just force biometrics-only flow)

## Findings

### Critical

None.

### High

**H1 — Spurious enrollment failure for every pre-PR wallet (merge blocker)**
`software-key.actions.ts:1053` + `software-key.slice.ts:85` + `use-wallet-authentication.ts:427`

Dev's slice never persisted `authenticationMode` (verified: zero occurrences on `dev`), so every existing wallet stores it as `undefined`. On first enrollment from settings:

1. `preparePlatformUnlockWithPassword` passes the raw persisted value (`sourceAuthenticationMode: undefined`); the pre-dispatch check passes.
2. The `platformUnlockConfigSaved` reducer normalizes: `state.authenticationMode = state.authenticationMode ?? 'password'`; flush persists `'password'` plus the wrapped key.
3. The postcondition builds `expected = { ...authoritative, platformUnlock }` from the **pre-dispatch** snapshot, so `softwareKeyStateSnapshotsMatch` strict-compares `undefined !== 'password'` → throws `'Biometric unlock configuration did not persist'`.
4. UI shows "Biometric unlock could not be completed. Try again." and fires `biometric_unlock_enrollment_failed` — while biometric unlock is genuinely enabled and the passkey can unlock the wallet, reveal the Secret Key, and authorize adding wallets. Retry succeeds (mode is now `'password'`) but mints a second passkey, orphaning the first in the user's password manager.

Security impact is low (fails toward a false negative, never toward granting access), but it hits 100% of upgraders on their first enrollment.

*Root cause:* five thunks derive `expected` by replaying the real reducers (`reduceWalletTransactionState`); two hand-spread it (`:1053`, `:1090`). One of the two drifted. `:1090` (`disablePlatformUnlock`) is correct only because `platformUnlockConfigRemoved` happens not to touch `authenticationMode`.

*Why tests missed it:* `software-key.actions.spec.ts` mocks `persistor.flush` via `persistStateOnNextFlush`, comparing the postcondition against hand-written fixtures instead of real reducer output; every `commitPlatformUnlockChange` test passes `authenticationMode: 'password'` explicitly. The E2E creates its wallet on this branch, which sets the mode.

*Fix:* derive `expected` by replaying the action through the real reducers (or normalize `authenticationMode` before comparing), plus a regression test that persists a legacy `authenticationMode: undefined` state and lets the real reducer drive the flush.

### Medium

**M1 — Biometric add-wallet permanently blocked when the seed's fingerprint exists as a Ledger wallet**
`software-key.actions.ts:870`

`addSoftwareWalletWithEncryptionKey` — the only add-wallet path available to a biometric-only wallet — rejects any fingerprint already present in the `wallets` slice, which includes Ledger entries keyed by the same root fingerprint. Touch ID succeeds, the PRF unwrap yields the correct key, then the guard throws `'Invalid wallet authentication state'`; `use-key-actions.ts:202` flattens it to `code: 'unavailable'`; the screen shows "Biometric unlock could not be completed. Try again." Because `capabilities.password` is false in biometric-only mode, no password field or Continue button renders — the only control is a retry that fails identically forever.

Asymmetry: the real password-mode add path (`setWalletEncryptionPassword`) has **no** `wallets.entities` guard, so password users can perform this same import. Fail-closed, no security impact, but a permanent, misdiagnosed dead end for a supported flow.

*Fix:* scope the guard to software keys (the preceding `authoritative.keys.some(...)` already does this) or emit a distinct error code.

### Low

**L1 — Disabling biometric unlock requires no re-authentication**
`biometric-unlock.tsx:191` → `use-key-actions.ts:159` → `software-key.actions.ts:1062`
Anyone at an unlocked wallet (password mode) disables biometric unlock with two clicks — a state-shape check only. No privilege is gained (the password still gates unlock/reveal; biometric-only mode cannot disable — reducer and thunk both refuse), but it is the one enrollment-state mutation not gated on a factor, while enroll / re-enroll / add-password all are. Optional hardening: require the password in the disable confirmation sheet.

**L2 — `restoreWalletSession` clears a valid session if it races store rehydration**
`session-restore.ts:44-50` + `software-keys/utils.ts:14`
`decryptAllSoftwareKeys` now throws on an empty key list, and the new catch actively clears the session key and in-memory keys. `renderApp` awaits `restoreWalletSession()` (`app/index.tsx:47`) against Redux state that redux-persist populates asynchronously; the ordering is unguaranteed. If restore ever wins the race, an unlocked wallet spuriously locks on popup open. Fails closed; doesn't reproduce in practice (E2E passes). Old code no-op'd here. Fix: gate the clear on a rehydrated store, or treat empty-keys as absent rather than corrupt — the emptiness policy belongs at the call sites, not in a leaf util.

**L3 — Post-flush verification can spuriously roll back legitimate wallet writes**
`software-key.actions.ts:421, 516, 729, 817, 923`
The six-slice `expectedTransaction` is compared to read-back storage with `isDeepEqual`. The write lock excludes concurrent wallet-auth writers, but any unrelated same-frame dispatch touching `accounts`/`active`/`chains`/`wallets` between dispatch and flush completion (a React effect reacting to the new wallet, a cross-frame hydrate) fails the check and rolls back a successful write with a generic error. H1 is the deterministic instance of this class; this is the residual probabilistic one. Retryable, fail-closed.

**L4 — Biometric-only onboarding removes the local password path with no disclosure at the point of choice**
`set-password.tsx:245-270`
The first-wallet "Use biometrics" button creates a wallet whose only local decryption path is a PRF unwrap of one platform credential (salt deleted — `software-key.slice.ts:73`; `capabilities.password` false forever). The backup screen one step earlier doesn't force revealing the Secret Key, and the passkey is labelled only `Leather biometric unlock · <TAG>` with no wallet identity. If the passkey is later deleted (Apple Passwords cleanup, profile keychain reset), the unlock screen renders no password field; the only remaining control is "Can't use biometrics?" → sign-out wipe. Not bricked — the mnemonic is a real recovery path and the settings copy explains it well — but the onboarding entry point doesn't, and passkey loss differs from password loss in ways users won't anticipate. Also verify on-device: enrollment requests `residentKey: 'discouraged'` (`platform-authenticator.ts:120`) while settings copy says the passkey "may sync through your provider" (`biometric-unlock.tsx:299`) — confirm where the credential actually lands before shipping the wording.

**L5 — Failure-code flattening destroys diagnosability**
`use-key-actions.ts` — seven `catch → { code: 'unavailable' }` blocks
The feature builds a good typed error vocabulary (8 distinct codes through `PlatformAuthenticatorResult` → `WalletAuthenticationResult`), then discards it at the last boundary: every thunk rejection becomes `'unavailable'` → "Try again." H1 and M1 — with nothing in common — present identically to users because of this. Fix: typed rejection codes from the thunks, or message mapping.

### Nits (not bugs)

- `platform-unlock.ts:146` — wrap-failure catch mislabeled `authentication-failed` (skews the analytics failure category only).
- `package.json:41` — `test:biometric` sets `BIOMETRIC_UNLOCK_E2E=true`, which nothing reads (verified).
- `software-key.actions.ts:698` — comment references `selectWalletSalt`, which this PR deletes.
- `addSoftwareWalletWithPassword` (`software-key.actions.ts:739`) — zero production callers; escapes knip/unused-exports via the `keyActions` object export.
- `useCheckPassword` (`software-key.hooks.ts:5-18`) — now fully unlocks the wallet (session init, mnemonics into memory, possible legacy migration) as a side effect of "checking" a password; name and `TODO` comment no longer describe it. Behaviorally stronger validation, but the add-wallet-with-password flow now runs argon2 twice.
- No timeout in the `deriveEncryptionKey` worker protocol — a silent worker hangs the promise forever. Pre-existing on `dev`; not a regression.
- `knip.json` — ~104 lines of unrelated Prettier reformatting in a security-sensitive PR; exactly one semantic change (the harness entry).

## Design and architecture notes

**Concerns, ranked:**

1. **The write-verify-rollback ceremony is copy-pasted ~8×** in `software-key.actions.ts` (8 lock blocks, 16 authoritative reads, 16 flushes, 8 rollback try/catches), and the expectation is computed three different ways — five sites replay real reducers, two hand-spread snapshots (one is H1), one compares fields ad hoc. Extract a single `runWalletStateTransaction({ precondition, actions, onRollback })` that always derives `expected` by replaying reducers; this makes H1's bug class unrepresentable and collapses ~320 lines.
2. **`WalletTransactionState` is a hand-maintained mirror of the store** (interface, empty-state builder, reducer-fold, shape validator — four sites listing the same six slices). A slice newly coupled to wallet writes fails silently (read-back just won't check it). Derive all four from one shared slice map.
3. **Typed failure codes are erased at the last boundary** (L5).
4. **Domain logic lives in a hook module**: `software-key.actions.ts` (store layer) imports `authenticateWithPassword` from `use-wallet-authentication.ts` (hooks layer) while that module imports store helpers — not a cycle, but the dependency direction flips mid-stack. Move the pure functions to a plain module.
5. **The rollback reducers are redundant**: `biometricOnlyToPasswordTransitionRolledBack` / `platformUnlockChangeRolledBack` (`software-key.slice.ts:121-147`) are the least-guarded reducers in the slice and duplicate what `adoptAuthoritativeSoftwareKeyState` (hydrate from the verified snapshot) already does. Delete them.
6. **The U0 harness validates a fork, not the shipped code**: `webauthn-prf.ts` near-duplicates `platform-authenticator.ts` (~100 lines byte-identical), so a green U0 run attests nothing about the production authenticator, and the two will drift. Either import the real module or retire the harness now that the CDP-virtual-authenticator E2E exists; 1,400 permanent lines in `src/` for a spike artifact is the wrong home either way.
7. **Auth state is optionals, not a discriminated union** — "password requires salt / biometric-only forbids it" lives in scattered guards instead of types.
8. **`software-key.actions.ts` is 1,111 lines with five responsibilities** (account discovery, transaction framework, auth state machine, locking, analytics). Split account discovery out; extract the transaction helper.
9. Smaller: `platformUnlockConfigsMatch(undefined, x) === true` is a fail-open wildcard hidden in a name (rename or force explicit wildcard); `/^[0-9a-f]{96}$/` hand-written in 9 places silently encoding `hashLen: 48` (export one guard next to the constant); duplicated `isWalletEncryptionKey`/`isRecord`; dead `hasSoftwareKeys` branch in `set-password.tsx:174-196` (the true-arm returns at :154); unused hook method `authenticateWithPassword`; module-level mutable singletons making specs order-dependent; `RequestWalletAuthentication` is controller+presenter with 10 props and renders a dead-end when no method is available (M1's screen); the auto-prompt intent machine spans three modules for one behavior.

stacks-alex:
@edgar-stacks how strongly do you feel about only doing password or biometrics but not both? I’m unsure and can see the argument both ways on it?

if we only allow one or the other then i definitely agree about removing from the settings page as there's not really any need for it there (you can just reset the wallet)

if we remove I'm also thinking about what the flow for existing wallets should be - do we leave a way for people to enable biometrics (and doing so just wipes out the password) or do they need to resetup the wallet and select biometrics instead of password. The second would be a bit simpler but the first seems more user friendly

edgar-stacks:
> how strongly do you feel about only doing password or biometrics but not both? I’m unsure and can see the argument both ways on it?

I want to think of a shortest way for us to get from no biometrics functionality to a secure basic version of one. We can make it biometrics-only macos-only, and then in separate implementations add to it:
- windows support
- linux support
- biometrics + password support
- support for switch from biometrics to password and vice versa (this might not be needed if we just allow biometrics + password support in the future)

So I think about password&biometrics as a separate feature by itself that can be created on top of the "only biometrics" and "only password" functionality.

> if we remove I'm also thinking about what the flow for existing wallets should be - do we leave a way for people to enable biometrics (and doing so just wipes out the password) or do they need to resetup the wallet and select biometrics instead of password. The second would be a bit simpler but the first seems more user friendly

I'd say for the v0 of this functionality, users should reset the wallet to use the biometrics.

stacks-alex:
@edgar-stacks pushed new commit (26b0375bc2672c9bee9fa820121622f2348bf7c7) that address the issues from your review as well as simplifies down to just biometrics OR password, but not both. This should still work on all platforms (and was previously tested on windows) since its just invoking the local passkey infra

PR #2615, opened 2026-08-07
fix: bring back ordinal banners and make those dismissable

### LINKED ISSUE #2574: Bring back the info banner for removing ordinals/runes support

edgar-stacks:
https://github.com/user-attachments/assets/c5649d7b-6909-46d2-a14d-f5290bbbf6f8


aperry-stacks:
Do we want the `aeUSDC` banner to be easily dismissible? It's already responsive to the wallet balance and will disappear after the user has migrated their balance.

edgar-stacks:
@aperry-stacks i was doing it for consistent looks but yea you're right, it's important to force users to swap aeusdc

edgar-stacks:
@alexp3y updated

aperry-stacks (dismissed):

fab-stacks:
image

Pushed one commit on top of this: the callout now links out to the [deprecation guide](https://app.leather.io/posts/deprecation-of-ordinals-inscriptions-support-in-leather) with a Learn more at the end of the body, and it only renders when the current account actually has UTXOs on its taproot path (checked via isTaprootDerivationPath from @leather.io/bitcoin, same pattern as the aeUSDC callout). Also gave the callout content a bit of right padding so the title no longer runs into the dismiss button.

Since dismissal ids live in the persisted settings store, dismissing is once per install across wallets, which together with the gating keeps this as quiet as possible: you only ever see it if taproot bitcoin is visible, and only until you dismiss it.

werner-stacks (approved):
LGTM

aperry-stacks (approved):

Issue #2622, opened 2026-08-08
Extension: "No app details found" crash connecting Ledger STX account via legacy auth (Brave)


wakqasahmed:
## Summary

Connecting a **Ledger Nano** wallet for **STX (Stacks)** via a dapp's legacy `@stacks/connect` auth flow crashes the Leather extension popup with `No app details found`, thrown from `LegacyAccountAuth`. On **Brave**, this reliably reproduces for me; I have not yet confirmed whether it also reproduces on Chrome.

## Environment
- Browser: Brave (Chromium-based)
- Extension: Leather (`ldinpeekobnhjjdofggfgjlcehhmanlj`)
- Hardware wallet: Ledger Nano, STX/Stacks app
- Flow: dapp-initiated legacy authentication request (`@stacks/connect` style), choosing to sign in with the Ledger-connected account

## Error / stack trace

```
Error: No app details found
    at LegacyAccountAuth (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/index.js:24370:19)
    at renderWithHooks (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:96965:21)
    at updateFunctionComponent (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:99702:15)
    at beginWork (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:100706:14)
    at performUnitOfWork (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:104318:14)
    at workLoopSync (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:104199:37)
    at renderRootSync (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:104180:7)
    at performWorkOnRoot (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:103785:40)
    at performSyncWorkOnRoot (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:105097:3)
    at flushSyncWorkAcrossRoots_impl (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:104968:17)
    at flushSpawnedWork (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/129.js:104739:5)
[...]

PR #2612, opened 2026-08-07
feat: delete legacy requests

edgar-stacks:
Removes support for legacy @stacks/connect (JWT token) requests from the extension and injected provider. All dApp interactions now go exclusively through the `LeatherProvider.request()` JSON-RPC API.

- Deletes the legacy message pipeline (inpage → content-script → background) and all JWT-based popup flows: transaction, signature, structured-data, PSBT, and auth/`ChooseAccount` (incl. the Ledger JWT-signing flow)
- Legacy provider methods (`transactionRequest`, `signatureRequest`, `psbtRequest`, `authenticationRequest`, …) now throw a deprecation error pointing to the RPC API — same behaviour mobile already had
- `window.StacksProvider` / `window.HiroWalletProvider` / `window.btc` remain as plain aliases of `LeatherProvider`
- Drops now-unused deps: `@stacks/connect(-jwt)`, `@stacks/connect-react`, `@stacks/auth`, `jsontokens`, `base64url`, `ecdsa-sig-formatter`, and test-app-only Stacks v6 packages
- Re-homes still-used transaction payload types locally and moves `useStacksBroadcastTransaction` out of the deleted feature dir; test-app reduced to RPC-only demos

> [!WARNING]
> Breaking for dApps still on old @stacks/connect: legacy calls now reject instead of opening approval UIs. Also removes the `legacy_request_initiated` / Sentry deprecation telemetry.

PR #2623, opened 2026-08-08
fix(extension): avoid crash when legacy account auth is missing app details

wakqasahmed:
## Related issue

#2622 — "No app details found" crash connecting a Ledger STX account via the legacy auth flow, filed against this repo.

## What this fixes

**Note on scope:** I have not been able to reproduce the exact trigger (why `origin`/`authRequest` end up missing on the popup url) in a live Brave + Ledger session — that part is left open in the linked issue for maintainer/reporter follow-up. This PR only fixes the *failure mode*: right now, when those values are missing for any reason, `LegacyAccountAuth` hard-crashes the popup:

```ts
if (!url) throw new Error('No app details found');
```

That throw propagates up to the top-level router error boundary, which renders the generic "Leather has crashed" screen — no way to retry, no diagnosis. This is inconsistent with how the same "auth details missing" condition is already handled a few call sites away: `useFinishAuthRequest` and `useCancelAuthRequest` both treat missing `authRequest`/`origin` as a non-fatal condition (log and no-op / early return), not an invariant violation.

This PR brings `LegacyAccountAuth` in line with that existing convention, and with the codebase's own precedent for this exact class of problem — `RouteUrls.RequestError` (`GenericError` component), already used elsewhere (`use-rpc-sign-psbt.tsx`, `use-psbt-request.tsx`, `psbt-signer.tsx`) to surface a recoverable error state with a close action instead of crashing:

```ts
if (!url) {
  logger.error('Legacy account auth request is missing app details from the popup url');
  return (
    <Navigate
      to={RouteUrls.RequestError}
      state={{
        title: 'Connection request failed',
        message: 'This connection request is missing information from the requesting app. Please close this window and try connecting again.',
      }}
    />
  );
}
```

## Test plan

- Added `legacy-account-auth.spec.tsx`:
  - Renders without throwing and navigates to `RouteUrls.RequestError` (with a title/message) when `useAppDetails()` returns no `url`.
  - Renders normally (no redirect) when `url` is present.
- `npx vitest run src/app/pages/legacy-account-auth/legacy-account-auth.spec.tsx` — 2/2 passing.
- `npx eslint` + `npx prettier --check` on both changed files — clean.
- `npx tsc --noEmit` — no new errors introduced (pre-existing unrelated `leather-styles` codegen errors elsewhere in the app are untouched by this change).

### DIFF
--- a/apps/extension/src/app/pages/legacy-account-auth/legacy-account-auth.spec.tsx
+++ b/apps/extension/src/app/pages/legacy-account-auth/legacy-account-auth.spec.tsx
@@ -1,0 +1,110 @@
+// @vitest-environment jsdom
+import { act, createElement } from 'react';
+import { createRoot } from 'react-dom/client';
+
+import { RouteUrls } from '@shared/route-urls';
+
+import { LegacyAccountAuth } from './legacy-account-auth';
+
+Reflect.set(globalThis, 'IS_REACT_ACT_ENVIRONMENT', true);
+
+const mocks = vi.hoisted(() => ({
+  useAppDetails: vi.fn(),
+  captureNavigate: vi.fn(),
+  errorLog: vi.fn(),
+}));
+
+vi.mock('react-router', async importOriginal => {
+  const actual = await importOriginal<typeof import('react-router')>();
+  return {
+    ...actual,
+    useLocation: () => ({ pathname: '/choose-account' }),
+    useNavigate: () => vi.fn(),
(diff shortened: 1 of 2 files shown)


edgar-stacks:
@wakqasahmed thank you for your contribution! Unfortunately, this change is redundant as we are currently in the process of removing the legacy requests from extension: https://github.com/leather-io/mono/pull/2612

PR #2633, opened 2026-08-12
fix: signPsbt-sighash

edgar-stacks:


edgar-stacks:
closing in favor of #2667

PR #2637, opened 2026-08-13
fix(web): explain multisig vault account limits

### LINKED ISSUE #2604: Web app: STX multisig vault account limits are not communicated well
Users didn't understand why there is a limit on stx account number, maybe a callout somewhere would give enough information to the user about it.

fab-stacks:
Closes #2604.

- The **Create vault account** sheet now opens with a quiet callout saying where the number comes from: "2 members allow 2 accounts per threshold, because a Stacks address is derived from its members and threshold. Members can’t change later, so more accounts means a new vault." It only shows where the ceiling really is a consequence of derivation, so Bitcoin vaults and Stacks vaults of four or more members — which sit on our own flat cap of 10 — don't get it.
- Members can’t be added to a vault after it’s created — `updateVaultMember` only takes a name, and there’s no add-member endpoint — so the callout points at a new vault rather than telling you to invite someone.
- Hovering or focusing a greyed **signing threshold** now says why that one in particular is out: "Threshold 2 already has its 2 accounts, so you can't pick it again."
- Those buttons carry `aria-disabled` instead of `disabled`. A disabled button receives no pointer events, so the tooltip would never have opened, and it couldn't be focused at all. They still announce as disabled and still can't be selected, but keyboard and screen reader users can now reach the explanation.
- **Vault accounts** carries a running count from the first account, so the ceiling is legible before it bites, and turns into a warning chip once it's reached.
- Removed three strings: "Thresholds at their account limit are disabled", "Pick a threshold to continue. Leather doesn't choose this for you", and the "This vault has reached its account limit" panel. The first was jargon, the second told the user nothing, and the third is redundant next to the count.
- Both surfaces now derive the limit from joined members. The vault page was counting every member including invited ones, which would have disagreed with the sheet if it were ever read before everyone had joined.

**The sheet** — the callout, and hovering a threshold with no room left

<p>
Create vault account sheet with the quiet callout above the signing threshold buttons
Tooltip on a greyed signing threshold reading: Threshold 2 already has its 2 accounts, so you can't pick it again
</p>

**Vault accounts** — room left, and at the ceiling

<p>
Vault accounts list with a neutral 3 of 18 accounts chip and the create new account tile
Vault accounts list with a warning 18 of 18 accounts chip and no create tile or blocked panel
</p>

### DIFF
--- a/apps/web/app/features/multisig/vaults/vault-account-index.ts
+++ b/apps/web/app/features/multisig/vaults/vault-account-index.ts
@@ -1,6 +1,6 @@
 import type { AuthNetworkId, VaultAccountSummary } from '@leather.io/models';
 
-const maxAccountsPerThreshold = 10;
+export const maxAccountsPerThreshold = 10;
 
 function factorial(n: number): number {
   let result = 1;
--- a/apps/web/app/pages/multisig/vault/components/accounts-section.tsx
+++ b/apps/web/app/pages/multisig/vault/components/accounts-section.tsx
@@ -120,7 +120,7 @@ function CreateAccountTile({ onClick }: { onClick(): void }) {
   );
 }
 
-function CreateAccountDisabled({ reason }: { reason?: string }) {
+function CreateAccountDisabled({ reason }: { reason: string }) {
   return (
     <Box
       width="100%"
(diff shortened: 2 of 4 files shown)


aperry-stacks (approved):
Doesn't look like it was added in this PR but this sentence in the callout is kind of a confusing: "Invite another member for more." 

More what? Also, you can't add a member to an active vault, so slightly misleading.

fab-stacks:
@aperry-stacks Yes agree, it was in there initially but I already removed it

edgar-stacks (approved):

PR #2644, opened 2026-08-14
fix: localhost tx proposal

edgar-stacks:
Add MULTISIG_API_URL env variable. It should be used to trigger tx proposals when running multisig on localhost

Add this to .env

```
MULTISIG_API_URL=http://localhost:8789
```


heidar-stacks (approved):
LGTM

aperry-stacks (approved):

leah-stacks (approved):

PR #2645, opened 2026-08-14
fix: activity view drift revert

aperry-stacks:
Reverts the activity view drift introduced in #2494: titles are bare asset symbols again, subtitles are standalone sentences, and web multisig composes its own action lines app-side.

#2494 moved multisig detail-page presentation into the shared view model, and the single-line vault rows from #2532 exposed it: bold rows reading a bare "via Fast Pool" with no action. #2555 then built on the regressed contract and cemented it in specs.

This PR:
- restores the original contract in @leather.io/features — symbol titles for all protocol-action shapes (including the newer one-legged swap/bridge shapes), verbs live in subtitles — and deletes the actionInTitle branch, with a spec guard asserting no subtitle can ever be a bare "via {protocol}" fragment
- gives bridge rows their own verbs ("Bridged via {protocol}") instead of reusing swap copy
- exposes protocolName on BlockchainActivityView so web multisig composes its bold line and detail heroes app-side ("Stack via Fast Pool", base verb) — this is the vault activity list bug fix
- updates extension e2e matchers to exact "BTC" matches, since the bare symbol substring-collides with the "BTC → sBTC" deposit overlay title

Before:
image

After:
image


edgar-stacks (approved):

fab-stacks (approved):
LGTM! Took me a second to reconcile the summary with the screenshots :)

PR #2647, opened 2026-08-17
fix(extension): cross-origin iframe callout

edgar-stacks:
Add a visible callout if the rpc call is coming from an iframe instead of the main website (example is using asigna)

https://github.com/user-attachments/assets/d81acd7c-c5df-4828-b32f-a1ac753291bf


werner-stacks (approved):
I like this solution.

leah-stacks (approved):

PR #2648, opened 2026-08-17
feat: bond exit

edgar-stacks:
Enable proposal and signing of bond template using signPsbt rpc request.

If user has connected with a multisig account, signPsbt is going to default to propose flow (same way sendTransfer works)

note: Mobile doesn't yet support propose param

Here's an example of me proposing, signing and broadcasting a psbt that spends a utxo of this descriptor:
`wsh(and_v(v:or_i(after(499999999),and_v(v:sha256(dbc1454203791389c7a6c24cbfcb4db42e1434fd0b6c0b682c90334507f47b99),pk(02b31fba61bbe4ee3753a72f4681c0d1ada9ecf23cff309a20a90af3201a37f5a0))),sortedmulti(1,<my tpubs>/0/0)))`

Afterwards, i've updated the PR to only accept xpubs/tpubs. Raw pubkey are not going to be accepted in the bond template as those are not signable by ledger

dApp, when calling signPsbt, already included the message and early exit signature into the psbt, so all that leather needed to do here is to sign and broadcast it

https://github.com/user-attachments/assets/c70660e7-1f32-4ca2-859c-eaf85dd2b3a5


aperry-stacks (dismissed):

packages/bitcoin/src/descriptors/bond-lock-script.ts
```diff
@@ -0,0 +1,125 @@
...
+// BIP-65/BIP-113 type boundary: lock values below are block heights, at or above are timestamps
+const minTimestampLockTime = 500_000_000;
+
+type ScriptOp = ReturnType<typeof Script.decode>[number];
+
+function isCompressedPubkey(op: ScriptOp | undefined): op is Uint8Array {
+  return op instanceof Uint8Array && op.length === 33 && (op[0] === 0x02 || op[0] === 0x03);
+}
+
+function decodeScriptInt(op: ScriptOp | undefined): number | null {
```
jannik-stacks, line 24: might be a helper fn in btc scure signer dependency for this

jannik-stacks (dismissed):
dont understand all of the reasoning but looks good

packages/bitcoin/src/descriptors/bond-lock-script.ts
edgar-stacks: ah i think there is! Thank you for catching it!

leah-stacks (approved):

aperry-stacks (approved):

PR #2650, opened 2026-08-17
fix: add private-1 network

edgar-stacks:
Add private-1 network

private

### DIFF
--- a/packages/models/src/network/network.model.ts
+++ b/packages/models/src/network/network.model.ts
@@ -35,6 +35,7 @@ export enum WalletDefaultNetworkConfigurationIds {
   sbtcTestnet = 'sbtcTestnet',
   sbtcDevenv = 'sbtcDevenv',
   devnet = 'devnet',
+  'private-1' = 'private-1',
 }
 
 export const defaultNetworkConfigurationsSchema = z.enum([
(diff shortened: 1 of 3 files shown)


packages/models/src/network/network.model.ts
```diff
@@ -230,6 +232,26 @@ const networkDevnet: NetworkConfiguration = {
   },
 };
 
+const privateNetworkStacksChainId = 256;
+
+const networkPrivate1: NetworkConfiguration = {
+  id: WalletDefaultNetworkConfigurationIds['private-1'],
+  name: 'Private',
```
aperry-stacks: We have a specific network for sBTC (`sbtcTestnet`), should we follow-suit here and give this a more meaningful name like `stakingTestnet` ?

edgar-stacks: sure, i'll update that as part of the separate PR, want to keep this small for a quick release

aperry-stacks (approved):

packages/models/src/network/network.model.ts
edgar-stacks: Done! 
network rename

jannik-stacks (approved):

PR #2664, opened 2026-08-20
fix(services): display all relevant Bitcoin multisig transactions in activity list

leah-stacks:
Multisig vaults were not showing all on-chain Bitcoin activity: faucet receives, maturity claims, early-exit sweeps and plain receives were failing to show. It was only showing proposals from the multisig backend.

Ownership is decided by comparing addresses against `tx.address`, which only `fetchDescriptorTransactions` attaches. A vault is a fixed-address account, so it goes through `getAddressTransactions`, leaving `tx.address` undefined — nothing was marked owned and `mapBitcoinActivity` dropped every transaction on its `isSend`/`isReceive` check.

Now the queried address is passed through when mapping. Ownership also requires an owned address to be present, so an addressless output (OP_RETURN) no longer matches an absent one.

### DIFF
--- a/packages/services/src/transactions/bitcoin-transactions.service.spec.ts
+++ b/packages/services/src/transactions/bitcoin-transactions.service.spec.ts
@@ -83,6 +83,44 @@ describe(BitcoinTransactionsService.name, () => {
       expect(result[0].txid).toEqual(tx.txid);
     });
 
+    it('marks the queried address as owned for a fixed-address account on regtest', async () => {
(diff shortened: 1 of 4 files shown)


aperry-stacks (approved):
lgtm! 

Fyi this issue only affects Regtest / Private-1. Mainnet will have the owned metadata populated on the BE Tx data response.

edgar-stacks (approved):

PR #2674, opened 2026-08-24
fix(bitcoin): resolve the bond co-signer as an extended key so ledger can sign

heidar-stacks:
resolve the bond co-signer as an extended key so ledger can sign

### DIFF
--- a/packages/bitcoin/src/descriptors/bond-covenant-keys.ts
+++ b/packages/bitcoin/src/descriptors/bond-covenant-keys.ts
@@ -1,0 +1,7 @@
+// The lock script only stores the co-signer's pubkey; Ledger needs the xpub (BIP-388).
+export const bondCovenantAccountKeys = [
+  // private-1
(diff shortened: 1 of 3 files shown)


leah-stacks (approved):

aperry-stacks (approved):

edgar-stacks (approved):

packages/bitcoin/src/descriptors/bond-covenant-keys.ts
```diff
@@ -0,0 +1,7 @@
+// The lock script only stores the co-signer's pubkey; Ledger needs the xpub (BIP-388).
+export const bondCovenantAccountKeys = [
+  // private-1
+  'tpubDF2dNLm8vgwf7xFNKn71bgcHWRwFN5AvTcWDiRYAUjTxibUNhEsmuVT7PP9GWaddDe8kfwN4yQ5gJPkSQ4Msf7oY9feFu5EFX9gy52zAmV7',
```
edgar-stacks, line 4: i think this is good as an initial solution but i think we need to have a service that dapps can use to receive xpubs. E.g. a hiro endpoint that would respond with all covenant account xpubs for all bonds. That way dapps won't need to hardcode this data in their code.

Issue #2676, opened 2026-08-25
for staking, max-fee of 1 sats or below 1000 sats should be rejected


friedger:
During staking, if the user sets max-fee to 1 sats, the reward payout will be rejected by the bridge. The UI should prevent entering a value for max fees that is too low.

Related to #2500

werner-stacks:
Thanks for reporting. I think it is a good idea to help prevent this problem. A failed claim via the bridge also means an additional transaction is needed to recover the funds. 
I think we can help the user by setting a default value, something like a 1000 sats. 

For the FastPool ([MAX500](https://github.com/stx-labs/signer-managers/issues/2)) contract we should support min-claim too that way a user can set a high amount there in case they prefer funds are not bridged automatically (when the claim is small). The user can still claim on their own.

PR #2679, opened 2026-08-25
feat(web): pending pool fees, max-fee floor and min-claim support

fab-stacks:
Fast Pool queued a fee increase (0% → 4.5% from cycle 142) and our staking page had no idea: we were reading the raw `fees-bips` data var, which on v2 signer-managers never reflects queued changes, and would have kept showing 0% even after activation. This reads `get-pending-fees` instead (falling back to the data var for v1 contracts) and rounds out the BTC payout config while at it, following up on #2676.

- The **Fee** column and pool overview now show the rate stakers will actually pay, with a **From cycle N** hint while a change is queued. The hint expires on its own once the cycle is reached. The switch-pool picker and its fee-increase warning use the same expected rate.
- Now als displays correct 10% fee for stackingdao liqiud staking table
- **Max withdrawal fee** now defaults to 1000 sats and rejects anything lower, so a 1-sat max-fee can no longer strand a reward at the bridge.
- New optional **Minimum claim** field on v2 signer-managers: rewards below it can only be claimed by the staker themselves, so small amounts are never force-bridged. Validation mirrors the contract's rule (above max fee + dust) and tells you the exact number that passes. v1 contracts keep getting the two-field calldata they expect; whether a contract is v2 is detected from the reads it exposes, so BYOSM gets this for free.
payoutfields
payoutvalidation
stakingtablefee
updateprefill

- The legacy liquid staking table now shows Stacking DAO's real **10%** fee (5% protocol + 5% signers, per their FAQ and the live contracts). The old `get-stack-fee` read only captured the protocol's half, so it and its supporting query code are gone.


werner-stacks:
this is what fastpool.org shows in the interface. Do you want to adopt any of the explaining from it?

image

I was being careful to not to go to high with the minimum fee 1000 sats seems good to me. I see fastpool has a higher default though. 3000 and 3546 sats respectively want to adopt something similar? I think it makes sense to have a higher minimum claim atleast. We want to prevent users from paying relatively high fees for small payouts on L1 setting a higher minimum will help in that case. I think we could even make it 10000. I would rather protect the user with a small stack here.

fab-stacks:
@werner-stacks Adopted both suggestions: **Minimum claim** now defaults to 10,000 sats (the 1,000 sats max-fee floor stays as is), and the helper copy now explains the tradeoffs the way fastpool.org does, including the smallest valid minimum computed live from the entered max fee ("At least 1,547 sats with your current max fee"). Kept our field labels for now; happy to align on "fee budget" / "smallest payout" naming in a follow-up if you prefer those terms.

image

werner-stacks:
the liquid stakingDAO pool fee is not yet accurate, it should be displayed as 10% fee. 
https://stackslabs.slack.com/archives/C09U43UADGU/p1787667301615249

edgar-stacks (approved):

werner-stacks (approved):
Tested LGTM, thanks for making the changes @fab-stacks

PR #2680, opened 2026-08-25
fix(extension): classify psbt inputs by script type when building signing config

aperry-stacks:
Derives a PSBT input's payment type from its output script instead of its address prefix, and only builds signing configs for p2wpkh and p2tr inputs.

This PR:
- `getInputPaymentType` reads the input's script via `OutScript.decode` rather than matching the bech32 prefix; the `network` param is dropped.
- `getAssumedZeroIndexSigningConfig` skips inputs of any other type instead of returning an empty derivation path for them.
- Corrects `paymentTypeMap.wsh` to `p2wsh`.


fab-stacks (approved):

edgar-stacks (approved):
Great work!

PR #2682, opened 2026-08-26
fix(services): keep sip10 balance when price lookup fails

leah-stacks:
A SIP-10 token disappeared from the balance list entirely if its market data lookup threw an error. The crypto balance is already known from Hiro's FT balances endpoint, so there was no need to hide the token.

`createSip10Balance` now catches the `getMarketData` rejection and returns the token with a zero quote instead of letting it reject into the `Promise.allSettled` filter in `getSip10AddressBalance`.

Uses try/catch rather than .catch() because getMarketData isn't guaranteed to return a promise.


werner-stacks:
wasn't there a good reason to show no balance. If it can't be checked the old value could be wrong? should we have some indication of that? No token at all is odd indeed though.

edgar-stacks (approved):

aperry-stacks (approved):

PR #2687, opened 2026-08-26
fix(extension): balance spacing and squareness of swap asset selector

fab-stacks:
The token selector on the legacy swap screen had a fully round hover surface much bigger than its own contents, so the **You pay** / **You receive** rows read as lopsided: 16px of padding to its left against 6px above and below, with the avatar, symbol and chevron all floating apart. Root cause was the `Avatar` having no explicit size, so it fell back to `xl` (48px) and pushed the ghost button to 64px tall inside a 76px field.

- The **hover surface** is now a rounded rectangle at the same 4px radius as the field it sits in, and 52px tall so it hugs its contents instead of ballooning past them.
- **Outer spacing** is even: 12px above, below and left of the hover surface. The right side keeps 16px so the right-aligned amount doesn't crowd the border.
- **Inner spacing** is even too, with the avatar pinned to `lg` (40px) and 6px on its left, top and bottom.
- Internals tightened: 8px between **avatar** and **symbol**, then symbol + **chevron** grouped at 4px so the chevron reads as part of the label rather than a third floating item.
- The hover background fades in over 100ms rather than snapping.

Only touches the legacy swap screen, so it's what renders while `swapRevamp` is off.


edgar-stacks (approved):

fab-stacks:
@edgar-stacks this will become obsolete with updated swaps, hence closing

PR #2690, opened 2026-08-27
fix(auth): bind sign-in message to requesting domain

aperry-stacks:
Binds the multisig sign-in message to the origin that requests it, so a signature obtained by one site cannot be redeemed for a session by another.

Pairs with an leather-api change that verifies the domain server-side against a per-environment allow-list; web and API deploy together, and the extension release must follow web since the wallet now refuses sign-in text whose domain doesn't match the requesting origin.

This PR:
- `buildSignInMessage` emits `Domain:` (the page's own origin) and `Application:` lines inside the signed text, and the same domain value travels in the `/v1/auth` body the backend rebuilds exact bytes from
- the background sign handlers (BIP-322 and STX plain-text) refuse a message starting with the reserved sign-in line whose `Domain:` differs from the port-reported origin, with `PERMISSION_DENIED` before any popup opens
- the reserved first line is a shared `@leather.io/constants` export so the services builder and the extension check cannot drift
- deletes the unused SIP-018 sign-in helpers and regenerates the Leather API types, which now require `domain` on `/v1/auth`

image


apps/extension/src/shared/rpc/sign-in-message-veto.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function shouldRefuseSignInMessageSigning(
+  message: string,
+  origin: string | undefined
+): boolean {
+  const lines = message.split(/\r?\n/);
+  if (lines[0] !== SIGN_IN_MESSAGE_FIRST_LINE) return false;
+  if (!origin) return true;
+  return lines[1] !== `Domain: ${origin}`;
+}
```
edgar-stacks, line 11: I'm a little unsure about this, basically that means that we reject any message that start with "Sign in to Leather". I think it is too broad, do you think we could do a more multisig specific gate here?

aperry-stacks: Can you give a reasonable example of how this is too broad? In what case would a legitimate 3p dapp prompt for a signature beginning with "Sign in to Leather"? IMO the strictness here actually removes an entire attack vector.

Also should be mentioned that this isn't multisig-specific. This is generic auth infa meant to be app-agnostic (e.g. there is an `application[]` array embedded in auth request that expresses this).

edgar-stacks (approved):

apps/extension/src/shared/rpc/sign-in-message-veto.ts
edgar-stacks: yea you're right, it is a low probability someone would use that for a message signing. My other thought is maybe we could test the whole string against regex instead of just the beginning? like this
```
const signInMessageRegex = /^Sign in to Leather\nNetwork: (mainnet|testnet|regtest)\nIssued: (\d+)$/;
```

fab-stacks (approved):

leah-stacks (approved):

PR #2691, opened 2026-08-27
feat: stacks multisig verification

edgar-stacks:


werner-stacks:
tested with 0.27.0 of the Ledger Stacks app, I can add the policy and check the address when adding the multisig account to the extension. 
And I can adhoc verify via transfer/stacks token. Tested with a 15-of-15 account and 2-of-3 account. Working as expected.

PR #2698, opened 2026-08-31
fix: sunset basic sbtc rewards

fab-stacks:
Enrollment for the Basic sBTC rewards program closed on August 27, so the **sBTC** page was still inviting people to enroll in a program that no longer accepts enrollments. This sunsets it so the page sets the right expectations:

- The **Basic sBTC Rewards** card and its **Enroll** button are removed from **Step 2: Choose sBTC rewards provider**, and the whole `sbtc-enroll` feature (enroll contract call + enrolled-status check) is deleted with it.
- The sunset callout stays with updated copy: **"Enrollment for basic sBTC rewards ended on August 27, 2026"**, noting the final cycle keeps paying yield until the program transitions to **Bitcoin Staking** (currently expected September 9, 2026) and pointing to the independent providers listed below it.
- The now-unused `sbtcBasicEnrollQuery` is dropped from `@leather.io/cms` and the page loader.

The FAQ copy about the rewards program lives in Sanity rather than code and is handled as drafts in the Studio alongside this PR.


edgar-stacks (dismissed):

werner-stacks (approved):
I tested the staging website, looks good to me.

aperry-stacks (approved):

PR #2704, opened 2026-09-02
feat: multisig production cutover

aperry-stacks:
Turns on Multisig in production `web` and redirects `extension` away from the staging multisig system toward production.

After merge, `web` and `extension` deployment (cutover) will be dependent on the production multisig backend being fully in place.

This PR:
- Removes the build-time gate that hides `/multisig` on the production web deploy, so the nav item, routes, and network switcher render at `app.leather.io`
- Scopes the staging web origin in the extension's policy-registration allowlist to non-production builds; the ext store build trusts `app.leather.io` alone
- Drops the extension's hardcoded staging fallback for multisig proposals so the store build defers to the API client's environment-aware base URL, while dev, feature, and testing builds keep their staging path and the `MULTISIG_API_URL` localhost override still works


apps/extension/src/app/features/multisig/use-propose-multisig-transaction.ts
```diff
@@ -11,7 +10,7 @@ export function useProposeMultisigTransaction() {
   const signProposalCommitment = useSignProposalCommitment();
   const mutation = useMutation(
     createProposeMultisigTransactionMutationConfig({
-      baseUrl: MULTISIG_API_URL || LEATHER_API_URL_STAGING,
+      baseUrl: MULTISIG_API_URL || undefined,
```
edgar-stacks, line 13: why pass undefined as baseUrl here? Or does passing undefined here means it just defaults to the mainnet link?

edgar-stacks (commented):
I assume we don't want to merge it until after the audit is done right? Otherwise it would impede our ability to deploy web builds to prod

apps/extension/src/app/features/multisig/use-propose-multisig-transaction.ts
aperry-stacks: Yeah exactly. `undefined` will route proposals to the default URL, which is either our staging or production url.

Keeping the `MULTISIG_API_URL` around at all allows for local dev set-ups.

edgar-stacks (dismissed):

andresgalante-stacks (dismissed):

edgar-stacks:
Let's not merge this one just yet until we decide on how we handle the already created staging multisig vaults

leah-stacks (dismissed):

fab-stacks (approved):

edgar-stacks (approved):

PR #2705, opened 2026-09-02
fix: bump broswerslist

aperry-stacks:
Bumps browserslist to patch GHSA-c83g-rgw3-j3cx


edgar-stacks (approved):
this overrides list is getting out of hand :D we need to schedule some time to update all of our dependencies to latest

andresgalante-stacks (approved):

PR #2706, opened 2026-09-03
fix(extension): render sip10 avatars in rpc approver post conditions

aperry-stacks:
Fixes issue with some SIP-10 token logos not rendering in the extension's RPC approver post-condition list.

The approver still read token metadata straight from Hiro. Everywhere else uses our backend's `Sip10AssetService`, which returns Leather-hosted logos. It also built the token contract ID from the wrong address, so the sBTC and USDCx overrides never matched.

NFT post conditions get the same contract ID fix, so their placeholder avatars are now seeded from the contract as well.

This PR:
- Loads post-condition token metadata through `Sip10AssetService`, same as the token list and mobile approver.
- Builds the contract ID from the asset address instead of the principal address, so sBTC and USDCx show their branded icons.
- Removes the unused `@hirosystems/token-metadata-api-client` dependency.
- Adds a spec for the contract ID fix.

image

image

image

### DIFF
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -43,7 +43,6 @@
     "@bitcoinerlab/descriptors": "3.1.7",
     "@bitcoinerlab/secp256k1": "1.0.2",
     "@bitflowlabs/core-sdk": "2.4.0",
-    "@hirosystems/token-metadata-api-client": "1.2.0",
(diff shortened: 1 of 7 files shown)


edgar-stacks (approved):
Nice!

werner-stacks (approved):

PR #2714, opened 2026-09-07
feat(services): bitcoin staking balances and positions service

aperry-stacks:
Adds locked / bonded BTC to the wallet's balances and exposes per-bond staking positions from the services layer.

Bonded outputs live in P2WSH lock scripts the wallet does not own, so they never appeared in an account's UTXO set or its balance. This PR fetches them from the new Leather staking positions endpoint and folds them in where every balance consumer already reads.

This PR:
- Adds `locked` to `UtxoTotals`, fetched fail-soft inside `getAccountUtxos` so a staking index outage never blocks a send.
- Adds `BtcBalance.lockedBalance`; total now includes bonded BTC, available does not.
- Adds `AccountBalancesService.getLockedBalance` (BTC + STX) for the cross-asset locked figure.
- Adds `BitcoinStakingService` composing positions from our registrations, Hiro v3 bond and principal data, and the burn tip, plus the next bond's enrollment window.
- Narrows the multisig vaults filter type: the API no longer accepts `declined` as a membership filter, while members keep it.


packages/services/src/multisig/multisig.service.ts
```diff
@@ -36,7 +36,7 @@ export interface UpdateVaultRequest {
 
 export interface ListVaultsFilters {
   status?: VaultStatus;
-  membershipStatus?: VaultMembershipStatus;
+  membershipStatus?: Exclude<VaultMembershipStatus, 'declined'>;
```
edgar-stacks, line 39: should this be part of this PR or an accidental change in multisig services?

packages/services/src/utxos/utxos.utils.ts
```diff
@@ -125,5 +146,6 @@ export function getUtxoTotals(
     dust: dustUtxos,
     unspendable: unspendableUtxos,
     available: availableUtxos,
+    locked: [],
```
edgar-stacks, line 149: why do you return an empty array here?

packages/utils/src/assets/balance-helpers.ts
```diff
@@ -24,19 +24,23 @@ export function createBtcBalance(
...
   const dustBalance = dustBal ?? zeroBalance;
   const unspendableBalance = unspendableBal ?? zeroBalance;
+  const lockedBalance = lockedBal ?? zeroBalance;
+  const unlockedBalance = subtractMoney(totalBalance, lockedBalance);
   const baseBalance = createBaseCryptoAssetBalance(totalBalance, inboundBalance, outboundBalance);
   return {
     ...baseBalance,
     dustBalance,
-    availableBalance: subtractMoney(totalBalance, unspendableBalance),
+    availableBalance: subtractMoney(unlockedBalance, unspendableBalance),
```
edgar-stacks: a note for our future selves - we'd need to double check in both extension and mobile that wherever we use availableBalance we actually intend it to be balance without the bond, and for totalBalance for it to be with bond.

packages/services/src/balances/btc-balances.service.ts
```diff
@@ -72,7 +72,8 @@ export class BtcBalancesService {
   ): Promise<AccountQuotedBtcBalance> {
     const utxos = await this.utxosService.getAccountUtxos(request, signal);
 
-    const totalBalance = createMoney(sumUtxoValues(utxos.confirmed), 'BTC');
+    const lockedBalance = createMoney(sumUtxoValues(utxos.locked), 'BTC');
+    const totalBalance = createMoney(sumUtxoValues([...utxos.confirmed, ...utxos.locked]), 'BTC');
```
edgar-stacks, line 76: **Critical.** `totalBalance` now includes locked bond outputs, but the consumers that treat it as spendable weren't updated. Point mobile `getAvailableBalance()` and web `hasPositiveBalance` / `filterSendableVaultAssets` at `availableBalance`.

packages/services/src/utxos/utxos.utils.ts
```diff
@@ -104,6 +108,23 @@ export function getKeyOrigin(fingerprint: string, path: string) {
...
+  if (!hasBitcoinAddress(account)) return null;
+  if (account.bitcoin.type === 'fixedAddress') return account.bitcoin.address;
+  if (exclusions?.nativeSegwitAddresses) return null;
+  return account.bitcoin.zeroIndexNativeSegwitPayerAddress ?? null;
+}
+
+export function createLockedUtxosFromBonds(bonds: LeatherApiStakingBond[]): Utxo[] {
+  return bonds
+    .flatMap(bond => bond.outputs)
+    .filter(output => !output.spent)
```
edgar-stacks, line 124: **Critical.** Nothing de-duplicates outpoints between `confirmed` and `locked`, and `spent` here is the indexer's cached flag, so a reclaimed bond is counted twice. Filter locked against the confirmed outpoint set, and add `leather-api-staking-bonds` to the refresh invalidation in `utxos.hooks.ts`.

packages/services/src/utxos/utxos.utils.ts
```diff
@@ -104,6 +108,23 @@ export function getKeyOrigin(fingerprint: string, path: string) {
   return `${fingerprint}/${path.replace('m/', '')}`;
 }
 
+export function getBondStakerAddress(
+  account: AccountAddresses,
+  exclusions?: AccountRequestAddressExclusionOptions
+): string | null {
+  if (!hasBitcoinAddress(account)) return null;
+  if (account.bitcoin.type === 'fixedAddress') return account.bitcoin.address;
```
edgar-stacks, line 116: **High.** This returns before the exclusions check, so `fixedAddress` accounts get the full bond in both scoped balance queries. Move the exclusions check above this line.

packages/services/src/utxos/utxos.utils.ts
```diff
@@ -104,6 +108,23 @@ export function getKeyOrigin(fingerprint: string, path: string) {
...
+  if (exclusions?.nativeSegwitAddresses) return null;
+  return account.bitcoin.zeroIndexNativeSegwitPayerAddress ?? null;
```
edgar-stacks, line 118: **Critical.** `zeroIndexNativeSegwitPayerAddress` is never populated in apps/web, so web users get no bonds and zero locked balance. Populate it in the web address store, or derive the staker from `nativeSegwitDescriptor` instead.

packages/services/src/utxos/utxos.service.ts
```diff
@@ -76,9 +88,24 @@ export class UtxosService {
...
+  private async getAccountLockedUtxos(
+    { account, exclusions }: AccountRequest,
+    signal?: AbortSignal
+  ): Promise<Utxo[]> {
+    const address = getBondStakerAddress(account, exclusions);
+    if (!address) return [];
+    try {
+      const bonds = await this.leatherApiClient.fetchStakingBonds(address, { signal });
+      return createLockedUtxosFromBonds(bonds);
+    } catch {
```
edgar-stacks: **High — blocking.** This catch has no error filter, so aborts, HTTP failures and programming errors all become a silently understated balance that React Query reports as success. It's already hiding one: `utxos.service.spec.ts:16-53` mocks `LeatherApiClient` without `fetchStakingBonds`, so the fixedAddress path throws `TypeError` and the test still passes.

packages/services/src/utxos/utxos.service.ts
```diff
@@ -47,7 +51,15 @@ export class UtxosService {
   /**
    * Retrieve categorized UTXO lists for given Bitcoin account.
    */
-  public async getAccountUtxos(
+  public async getAccountUtxos(request: AccountRequest, signal?: AbortSignal): Promise<UtxoTotals> {
+    const [ownedUtxos, locked] = await Promise.all([
```
edgar-stacks, line 55: **Medium.** Every caller now waits on the bonds request, including coin selection, max-spend and the send preload, which all discard `locked`. Make `locked` opt-in via the request so only the balance service pays for it.

packages/services/src/staking/bitcoin-staking.utils.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export function estimateBurnHeightDate(height: number, burnTip: number, now: Date): Date {
+  return new Date(now.getTime() + (height - burnTip) * averageBitcoinBlockTimeMs);
+}
+
+export function deriveBtcStakingPositionStatus(
+  registration: LeatherApiStakingBond,
+  burnTip: number
+): BtcStakingPositionStatus {
+  const announced = registration.exitAnnouncedAtBurnHeight !== null;
```
edgar-stacks, line 30: **High.** `!== null` treats a missing `exitAnnouncedAtBurnHeight` as announced, so every position derives as exiting/exited if the index omits the key. Use `!= null` or a `typeof` check.

packages/services/src/staking/bitcoin-staking.utils.ts
```diff
@@ -0,0 +1,132 @@
...
+export function toBtcStakingPosition({
+  registration,
+  stakerAddress,
+  bond,
+  principalBond,
+  burnTip,
+  now,
+}: ToBtcStakingPositionArgs): BtcStakingPosition {
+  const outputs = registration.outputs.map(toBtcBondOutput);
+  const unlockBurnHeight = outputs.reduce(
```
edgar-stacks, line 81: **Medium.** Seeded with 0 and run over spent outputs, so empty `outputs` gives a 2009 unlock date and the max can land on an already-spent output. Take the max over unspent outputs, and give the empty case its own status.

packages/services/src/staking/bitcoin-staking.utils.ts
```diff
@@ -0,0 +1,132 @@
...
+    0
+  );
+  return {
+    bondIndex: registration.bondIndex,
+    stxAddress: registration.stxAddress,
+    stakerAddress,
+    outputs,
+    exitAnnouncedAtBurnHeight: registration.exitAnnouncedAtBurnHeight,
+    status: deriveBtcStakingPositionStatus(registration, burnTip),
+    amount: sumBondOutputs(outputs),
```
edgar-stacks, line 92: **Medium.** Sums spent outputs while `lockedBalance` filters them out, so a withdrawn position reports its full historical size. Sum unspent, and add a separate field if the historical figure is wanted.

packages/services/src/staking/bitcoin-staking.utils.ts
```diff
@@ -0,0 +1,132 @@
...
+    .sort((a, b) => a.schedule.activation.bitcoin_height - b.schedule.activation.bitcoin_height)[0];
+}
+
+export function deriveBondEnrollmentWindow(
+  bond: HiroStakingBond,
+  poxInfo: HiroPoxInfoResponse,
+  now: Date
+): BtcBondEnrollmentWindow {
+  const activationBurnHeight = bond.schedule.activation.bitcoin_height;
+  const closesAtBurnHeight = activationBurnHeight - poxInfo.prepare_phase_block_length;
```
edgar-stacks, line 123: **Medium.** `closesAtBurnHeight` is never compared against the burn tip, so a closed enrollment window is still returned as upcoming. Return null once `burnTip > closesAtBurnHeight`.

packages/services/src/staking/bitcoin-staking.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  ): Promise<BtcStakingPosition> {
+    const [bond, principalBonds] = await Promise.all([
+      this.stacksApiClient.getStakingBond(registration.bondIndex, { signal }),
+      this.stacksApiClient.getPrincipalStakingBonds(registration.stxAddress, { signal }),
+    ]);
+    return toBtcStakingPosition({
+      registration,
+      stakerAddress,
+      bond,
+      principalBond: principalBonds.results.find(
```
edgar-stacks, line 87: **High.** Matched on bond index alone and never scoped to the staker, so principal-level STX and rewards get copied onto every registration in that bond. Match on the enrollment tx instead — `enrollmentTxId` / `enrollment.tx_id`.

packages/services/src/infrastructure/api/leather/leather-api.types.ts
```diff
@@ -3437,6 +3437,90 @@ export interface paths {
     patch?: never;
     trace?: never;
   };
+  '/v1/staking/addresses/{address}/bonds': {
```
edgar-stacks, line 3440: **High.** Hand-written into a codegen artifact, so the next `gen-api-types:leather` deletes the staking paths and reverts the multisig `membershipStatus` narrowing. Move the staking shapes to a hand-owned module, and the multisig change to its own branch.

edgar-stacks: i assume this one has a stale committed types file?

edgar-stacks (commented):
Great work! Left a few of mine and ai generated comments

packages/services/src/multisig/multisig.service.ts
aperry-stacks: Not accidental. Leather api types were regenerated against backend schema as part of this PR. 

There was previously a change to the multisig vault list filter types that got pulled in. 

This is a minimal change to accommodate. Small and imo doesn't necessarily warrant a separate blocking PR.

packages/services/src/utxos/utxos.utils.ts
aperry-stacks: To satisfy the type. Locked UTXOs are added to the object on a separate enhancement pass with staking service data.

packages/services/src/utxos/utxos.utils.ts
aperry-stacks: How exactly is this critical? We have no consumers of staking balance data in web.

packages/services/src/infrastructure/api/leather/leather-api.types.ts
aperry-stacks: What makes you think this is handwritten?

packages/services/src/balances/btc-balances.service.ts
aperry-stacks: Can you give a more clear example of where the issue is? Coin-selection runs against available balance, which does not include locked UTXOs. What exactly is the problem here?

packages/services/src/utxos/utxos.utils.ts
aperry-stacks: Users may continue to show a locked balance for a few minutes (avg 2-3min) while the backend updates the spent status after reclaiming a bond. Is that a critical issue? 🤔

packages/services/src/utxos/utxos.service.ts
aperry-stacks: It's a single call to a cached backend endpoint and it fails open. This is not an issue.

packages/services/src/staking/bitcoin-staking.utils.ts
aperry-stacks: This is a nullable field from the backend. Are you saying that we need to be defensive against our own backend schema for some reason?

packages/services/src/utxos/utxos.utils.ts
aperry-stacks: Not a bug. Exlcusions aren't relevant for fixed addresses.

edgar-stacks (approved):

leah-stacks (approved):

PR #2717, opened 2026-09-08
feat(extension): bond preview scenarios and playground

fab-stacks:
## What

A preview layer for the bond UI in #2715, so every state can be reviewed from the extension build without a funded wallet or a live staking index.

- `leather-mock-bond` localStorage key selects one of seven scenarios: `active`, `ending-soon`, `renewal-set`, `matured`, `exiting`, `with-history`, or `none`.
- The scenario replaces the two staking queries (positions, enrollment window) with fixtures typed as the real `BtcStakingPosition` / `BtcBondEnrollmentWindow`, and folds the bonded BTC into the balance queries with `select`, so Home, All balances, the token rows and the bond pages all see it the way the service would report it.
- `#/playground/bonds` (dev only) renders every scenario side by side with the real components, plus chips to set the live scenario and jump to the real pages.

## Gating

Everything is behind `WALLET_ENVIRONMENT !== 'production'` and an explicit localStorage key. With the key unset the hooks pass the real query configs through untouched, so this is a no-op for users and for CI.

## How to review

1. Download the extension zip from the build comment, load unpacked, open the wallet in a full tab.
2. Go to `#/playground/bonds`.
3. Pick a live scenario at the top, then use the "Open ..." chips to see Home, All balances and the bond page with that state applied.

Stacked on `feat/bitcoin-staking-ui`; only the preview files differ.

### DIFF
--- a/apps/extension/src/app/features/bonds/bond-fixtures.spec.ts
+++ b/apps/extension/src/app/features/bonds/bond-fixtures.spec.ts
@@ -1,0 +1,31 @@
+import { filterFixturePositions, getBondFixture } from './bond-fixtures';
+import { bondScenarios } from './bond-scenarios';
+
+describe('bond fixtures', () => {
+  it.each(bondScenarios)('%s reports locked sats equal to its unspent outputs', scenario => {
+    const fixture = getBondFixture(scenario);
+    const unspentSats = fixture.positions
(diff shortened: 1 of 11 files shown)


fab-stacks:
Folded into #2718 as a separate second commit, so the design follow-ups can be cherry-picked on their own.

PR #2718, opened 2026-09-08
feat(extension): bond UI design follow-ups

fab-stacks:
Design follow-ups on top of #2715, matching the Screens page in the Paper file [Bonds in the Wallet](https://app.paper.design/file/01M0Z1NTN6XS2P7BXM1V3QZ78J/3-0). Product changes only, ready to merge.

Two commits, kept separate because the second one is design system rather than bonds:

1. **`feat(extension): bond UI design follow-ups`** — everything below.
2. **`fix(ui): stop the pressable hover plate swallowing pointer events`** — one line in `packages/ui`. Independent of the bond work, but the tooltips in commit 1 are not hoverable without it. It affects every `Pressable` row in the wallet, so it may want a second pair of eyes.

## Bonds

- **Status badges follow the wallet's badge convention.** Grey is the resting state, blue a scheduled step, yellow a decision due, green completed. Active → default, Upcoming → info, Ends in N days → warning (new, derived from the unlock date), Ended → warning (it needs a withdrawal), Exiting → default with the mempool spinner, Withdrawn → success, Exited → default. Drops the `textColor="primary"` override so bond pills use the same coloured text as transaction and multisig badges.
- **Home Locked** becomes a bordered card with a chevron that opens All balances, and its tooltip no longer talks about stacking only.
- **Home callouts**: "Your bond unlocks in N days" (warning) within 7 days of unlock, "Your bond has unlocked" (success) when a position is matured. Dismissible per bond and state, so the second notice still shows after the first was dismissed. Both link to Bitcoin Staking.
- **STX side of the bond**: an "In a bond" row in the Stacks section of All balances, and a Balances section on the Stacks token page (available, in a bond with the unlock date, other locked, pending). Derived from `position.stxStacked`, so "STX locked" only reports what is locked for other reasons.
- **Renewal**: a registration for the next period renders as a compact row under the running bond rather than a second period card. Current positions sort by soonest unlock.
- **Periods are collapsible** on the bond page, expanded by default, with the amount shown in the header while collapsed.
- **All balances detail error state**: a failed UTXO fetch rendered as `$0.00 across 0 addresses`, indistinguishable from an empty wallet. It now says so.

## Locked balances on token rows

The part worth the closest look, because it changes a row every token uses.

A token row showed `2.00` next to a padlocked `2.00` with no way to tell whether that meant two total of which two are locked, or two spendable plus two locked. The first figure is `totalBalance`, so the locked figure is a subset — but nothing on the row said so.

- The locked share keeps its tinted plate and gains an info tooltip: *"Your total, and how much of it is locked."*
- The gap between the two figures tightens to 4px so they read as one amount line rather than two columns.
- The plate's padding is deliberately asymmetric (`pl 2px / pr 4px / py 3px`). The lock glyph carries ~3px of inset inside its own box, so uniform padding measures 7px on the left against 5px on the right. This lands all four sides within about a pixel.
- **STX now uses the same treatment.** It previously showed `2.00 · 2.00 locked` on the crypto line and `$X locked` on the fiat line. Both are replaced by the plate and tooltip, so BTC and STX read identically. Note this drops the fiat value of locked STX from that row, which BTC never had — say if you would rather keep it on both.
- With STX moved over, `titleRightBulletInfo` and `captionRightBulletInfo` on `CryptoAssetItemLayout` had no callers left and are removed, along with the `BulletSeparator` wrappers that existed only to host them.

## Tooltips

Two bugs meant most info tooltips on All balances did nothing:

- Section labels and detail heroes wrapped the whole `Flag` in `BasicTooltip` without `asChild`, so Radix rendered a button that stretched to the container width and anchored the tooltip at the row's right edge instead of at the icon.
- Rows passed `asChild` with a bare `<InfoCircleIcon />`. svgr runs without `ref: true`, so Radix never received a node and the trigger was inert — no hover, silently.

Both are replaced by one `InfoTooltip` that wraps the icon in a span, matching the pattern already used in `available-balance.tsx`. Anchoring is now the icon, so these went back to `side="top"`; that also fixes the shipped Total balance tooltip, which used `side="bottom"` and covered the number.

Three further fixes came out of reviewing them at size:

- **Vertical centring.** `inline-flex` on the trigger got baseline-aligned inside `Flag`'s block icon slot, so every section header's icon sat 2.25px high while row icons were correct. Now `flex`, and all of them measure dead centre.
- **Wasted width.** A tooltip box is shrink-to-fit from max-content, so once a label wraps the box sits at the full 250px and never shrinks to the wrapped lines. `text-wrap: balance` therefore shortened the lines inside a box that stayed 250px, leaving empty gutters — up to 89px on the shortest label. `text-wrap: pretty` fills the lines and only tidies the last one: average slack across all 23 tooltip strings drops from 38px to 14px, with no label orphaning its last word.
- **Copy.** `btcProtocol` restated its own label; it now says where sBTC counts instead. `stxAvailable` names its exclusions the way the BTC row does. `stxLocked` drops the stale "Stacking cycle" wording for "staking", matching the rest of the wallet.

### DIFF
--- a/apps/extension/src/app/components/balance/locked-balance-badge.tsx
+++ b/apps/extension/src/app/components/balance/locked-balance-badge.tsx
@@ -1,16 +1,38 @@
+import { Flex, styled } from 'leather-styles/jsx';
+
 import type { Money } from '@leather.io/models';
-import { Badge, LockIcon } from '@leather.io/ui';
+import { LockIcon } from '@leather.io/ui';
 
 import { formatCurrency } from '@app/common/currency-formatter';
+import { InfoTooltip } from '@app/ui/components/tooltip/info-tooltip';
 
 interface LockedBalanceBadgeProps {
   balance: Money;
 }
+
+const lockedBalanceTooltip = 'Your total, and how much of it is locked.';
+
+/** The locked share of a token row's balance: tinted plate, then the tooltip. */
 export function LockedBalanceBadge({ balance }: LockedBalanceBadgeProps) {
   return (
-    <Badge
-      icon={<LockIcon variant="small" />}
-      label={formatCurrency(balance, { showCurrency: false })}
-    />
+    <>
+      <Flex
+        alignItems="center"
+        gap="2px"
+        // Uneven to look even: the lock glyph carries ~3px of its own inset
+        pl="2px"
+        pr="space.01"
+        py="3px"
+        borderRadius="sm"
+        bg="ink.component-background-default"
+        color="ink.text-primary"
+      >
+        <LockIcon variant="small" color="ink.text-primary" width={12} height={12} />
+        <styled.span textStyle="label.03">
+          {formatCurrency(balance, { showCurrency: false })}
+        </styled.span>
+      </Flex>
+      <InfoTooltip label={lockedBalanceTooltip} size={14} />
+    </>
   );
 }
--- a/apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
+++ b/apps/extension/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
@@ -2,13 +2,7 @@ import DOMPurify from 'dompurify';
 import { Box, Flex, styled } from 'leather-styles/jsx';
 
 import type { Money } from '@leather.io/models';
-import {
-  BulletSeparator,
-  ItemLayout,
-  Pressable,
-  SkeletonLoader,
-  shimmerStyles,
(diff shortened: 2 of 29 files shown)


edgar-stacks (commented):
one thing that caught my eye, info-tooltip.tsx, tooltip.tsx, pressable.web.tsx have way too many comments attached to styles, i'd suggest we remove those to not clutter the files

aperry-stacks:
@fab-stacks looks like a couple of failing CI checks. Can you check those?

PR #2732, opened 2026-09-11
feat(web): add PlanBetter as a STX-only pool provider

### LINKED ISSUE #2728: Add PlanBetter as a STX-only pool provider (pox-5 signer manager)
## Summary

PlanBetter deployed a pox-5 signer manager and wants to be listed on
app.leather.io/staking. The contract is different from the Stacks Foundation reference. 
It claims all rewards to the pool operator and then does a (single) Bitcoin L1 transaction
to pay its members their shares taking an effective fee of 5% by the pool operator in the process. 
This is much like how they operated in the PoX-4 era. 

Some of the integration-visible differences: one extra required form
field, a mandatory validation, and several things the reward UI must NOT render.
This issue should have everything needed to wire it up correctly.

**Contract:** `SP3ZA8J49HPS7M3KD7EB01Y0ZAJS7VJS2NG87MDGN.planbetter-signer-manager` (mainnet)

Pin this full contract ID rather than resolving by name.

The stacking call itself is unchanged:
`pox-5.stake(signer-manager, amount-ustx, num-cycles, start-burn-ht, signer-calldata)`.

## How this pool differs from reference-manager pools

| | reference manager | PlanBetter |
| --- | --- | --- |
| `signer-calldata` | optional; omit to get paid in sBTC | **required**; omitting aborts the stake (`u1012`) |
| reward asset | sBTC on Stacks, or BTC if a pox-addr is given | native BTC only, paid off chain by the operator |
| member claim | permissionless on-chain `claim-staker-rewards` | **none, no on-chain claim path exists** |
| bonds (`register-for-bond`) | supported | **rejected on chain** (`u1014`), STX-only pool |
| `get-pox-addr` shape | `(some { pox-addr: {...}, max-fee: uint })` | `(some { version, hashbytes })`, flat |
| fee | readable on chain (`fees-bips`) | not on chain, operator policy |
| `max-fee` / `min-claim` | honored | ignored (no on-chain withdrawal) |

## 1. Required field: BTC payout address

The stake form needs one field the other pools do not require: a Bitcoin payout
address. It is mandatory and enforced by the contract, not by the app. Blocking
submit without it just avoids a guaranteed on-chain failure.

There is no sBTC fallback in this pool. Do not offer "pay me in sBTC" and do not
retry with `signer-calldata: none`.

## 2. Calldata encoding

Consensus-serialized Clarity tuple, passed as `(some (buff ...))`:

```ts
import { Cl } from '@stacks/transactions';
import { poxAddressToTuple } from '@stacks/stacking';

const calldata = Cl.serialize(
  Cl.tuple({ 'pox-addr': poxAddressToTuple(btcAddress) }),
);
// signer-calldata argument:
Cl.some(Cl.bufferFromHex(calldata))
```

- The key must be spelled exactly `pox-addr`. Renamed or missing gives `u1003`
  and the stake reverts.
- Extra keys are tolerated and ignored, so the Friedger-family formats
  (`{ pox-addr, max-fee }`, and `{ pox-addr, max-fee, min-claim }` from the newer
  max500 contract) also decode here. A single shared encoder across pool providers
  works, as long as `pox-addr` keeps that key name.
- Do not render `max-fee` or `min-claim` inputs for this pool. Neither is read.
- Well under the `(buff 500)` limit in all shapes.

## 3. Address validation

The contract's read-only `check-pox-addr` accepts pox-addr versions `0x00` through
`0x06`, requiring 20-byte hashbytes for `0x00`-`0x04` and 32-byte for `0x05`-`0x06`.

PlanBetter confirmed their payout batch handles whatever the user provides, so the
contract's range is the only constraint. No app-side narrowing needed.

They suggest defaulting to the user's own Bitcoin address from Leather, so prefill
from the connected BTC account rather than presenting an empty field. Keep it
editable, since some members will want payouts elsewhere.

Suggested pre-flight before broadcast: call `check-pox-addr` with the encoded
tuple, and check hashbytes length against version client-side.

## 4. What the reward UI must not do

This pool is custodial. The manager collects 100% of the signer's sBTC reward
stream, and members are paid native BTC off chain on the operator's schedule.

- **No claim button.** There is no function to call.
- **Do not render an unclaimed or claimable balance.** pox-5's read-only
  `get-earned-staker-rewards(manager, cycle, none, staker)` returns the member's
  gross accrual for a cycle, but this contract never calls
  `claim-staker-rewards-for-signer`, so that value is never zeroed. It does not
  decrease when the member is paid. Presenting it as outstanding or claimable will
  be wrong for every member who has already been paid. If shown at all, label it
  as protocol-attributed cumulative earnings and state that payment status lives
  with the operator.
- **Disclose the trust model on the pool card:** rewards in native BTC, paid off
  chain by the operator, custodial, no on-chain recourse.

## 5. Registered address: read and change

- `get-pox-addr(staker)` returns the flat `(some { version, hashbytes })`. If the
  staking UI shares a pox-addr reader across providers, it needs a per-provider
  shape (reference wraps it alongside `max-fee`).
- `stake-update` (extend or increase) also runs the same validation and
  **overwrites the stored address**. Prefill from `get-pox-addr` so an extend does
  not silently redirect a member's payouts, and make the consequence visible if
  they do change it.
- Successful registration emits `print { topic: "validate-stake", staker,
  pox-addr }`, usable for activity-feed confirmation.

## 6. Minimum stake: 1000 STX, enforced on chain

`validate-stake!` rejects anything below `u1000000000` micro-STX (1000 STX) with
`ERR_STAKE_BELOW_MINIMUM` (`u1015`). The comparison is `>=`, so exactly 1000 STX
is accepted.

It is a **constant, not an admin-settable variable**, and there is no read-only
getter for it. So pin the value against this contract ID rather than reading it
live: it cannot change for this contract, and a future change would be a new
contract with a new ID. An app-side minimum here is not an unenforced app rule,
it mirrors a hard contract rule, so blocking submit below 1000 STX is correct and
saves the user a guaranteed failed transaction.

The check runs on `stake-update` too, but it cannot lock anyone out: pox-5 passes
the post-increase total, which for an existing member is always at least the
amount they originally staked. Extending with a zero increase always passes.

## 7. Error mapping

From the manager, surfaced through the pox-5 `stake` / `stake-update` call:

| err | meaning | UI action |
| --- | --- | --- |
| `u1015` | stake below 1000 STX | form validation, see section 6 |
| `u1012` | no `signer-calldata` | require the BTC address (unreachable if the form is right) |
| `u1003` | calldata did not decode | encoding bug, do not retry |
| `u1004` | invalid pox-addr | address validation bug |
| `u1014` | bond registration attempted | this pool is STX-only; exclude it from any bond flow |
| `u1006` | manager called directly instead of via pox-5 | wiring bug |

From pox-5, the ones members will actually hit:

| err | meaning | UI action |
| --- | --- | --- |
| `u47` | stake attempted during the prepare phase (last 100 burn blocks of a cycle) | disable submit, show when the window reopens |
| `u19` | already staking | route to `stake-update` instead |
| `u8` | insufficient STX | standard balance validation |
| `u20` | invalid cycle count (must be 1-96) | form validation |
| `u24` | `start-burn-ht` does not resolve to the next cycle | do not let users post-date |
| `u17` | the pool's signer key grant is missing or revoked | pool is not accepting stake, hide or disable it |

## 8. Useful reads for the pool card

All on pox-5 unless noted:

- Pool is live and accepting stake: `get-signer-info(manager)` returns the signer
  key, and `verify-signer-key-grant(manager, key)` returns ok. A revoked grant
  means new stake is refused (`u17`) even though existing positions keep earning.
- Pool size for a cycle: `get-amount-delegated-for-signer(manager, cycle)`.
- Member's position: `get-staker-info(staker)`, and
  `get-signer-cycle-membership(staker, cycle)`.
- On the manager: `get-pox-addr(staker)`, `check-pox-addr(tuple)`,
  `get-earned-fees()` (operator's uncollected sBTC, not a member figure),
  `is-admin(principal)`.

## 9. Pool parameters, confirmed by PlanBetter

| | value | source |
| --- | --- | --- |
| fee | 5% | operator policy, published at planbetter.com |
| payout cadence | every 2 cycles, roughly monthly | operator policy |
| minimum payout | none | operator policy |
| minimum stake | 1000 STX | contract constant, `u1015` on breach |
| payout asset | native BTC to the registered address | operator policy |
| supported address types | anything `check-pox-addr` accepts | operator confirmed |

The 1000 STX minimum stake is what stands in for a minimum payout: it is the floor
that makes a monthly BTC payout worth sending, so there is no separate accrual
threshold and no carry-forward for small balances.

Only the minimum stake in that table is enforced on chain. The fee, the cadence and
the payout asset are operator policy with no on-chain equivalent to verify against,
so the pool card should attribute them rather than present them as protocol facts.
In particular, the 5% is applied during off-chain distribution; the contract itself
collects 100% of the reward stream, and `get-earned-fees()` returns the whole
uncollected amount, not the fee portion.

## Notes

The deployed contract has almost no comments; the documented rationale lives in
the PlanBetter repo (README plus `docs/PAYOUTS.md`), not in the on-chain source.
Anything the integration relies on should be confirmed against the deployed
source, which is the authority.

fab-stacks:
Closes #2728

Lists PlanBetter's pox-5 signer manager. Unlike the reference manager it collects the pool's sBTC and pays members native BTC off chain, so the integration needs a mandatory payout address, a 1,000 STX floor and no claim UI.

- Pins `SP3ZA8J49HPS7M3KD7EB01Y0ZAJS7VJS2NG87MDGN.planbetter-signer-manager` on mainnet, 5% fee fixed and labelled **PlanBetter policy**
- New `operatorBtcPayout` / `minStakeMicroStx` pool fields drive a `PoolPayoutMode` that replaces the `supportsBtcPayout` boolean in forms, summaries and labels
- Stake form: one required **BTC payout address** prefilled from the wallet, no sBTC option or fee inputs; amount enforces the 1,000 STX contract minimum
- Calldata sends `{ pox-addr }` only; decoder also reads the flat `{ version, hashbytes }` shape from `get-pox-addr`
- Active position: payout card with the registered address instead of **Claimable rewards**; earned-rewards reads skipped
- Update stake: registered address carried in, warning when edited; switching in checks the total against the minimum
- Pool table and overview show a **Custodial BTC payout** badge, BTC rewards token and the trust-model disclosure
- Fee and payout tooltips (and the **PlanBetter policy** caption under the fee) link to PlanBetter's FAQ, where the 5% fee and payout cadence are published

All states can be viewed in the staking playground on the preview: https://pr-2732-leather-web.wallet-6d1.workers.dev/playground/staking-states (boards "Pool that pays native BTC off chain", "Position paid in native BTC by the operator", "Operator-paid position: the registered address is carried in").

### DIFF
--- a/apps/web/app/components/info-tooltip-icon.tsx
+++ b/apps/web/app/components/info-tooltip-icon.tsx
@@ -1,6 +1,7 @@
 import { styled } from 'leather-styles/jsx';
 import type { ColorToken } from 'leather-styles/tokens';
 import { BasicHoverCard } from '~/components/basic-hover-card';
+import { LearnMoreLink } from '~/layouts/page/page';
 
 import { InfoCircleIcon } from '@leather.io/ui';
 
@@ -8,6 +9,7 @@ interface InfoTooltipIconProps {
   title: string;
   explanation: string;
   ariaLabel: string;
+  learnMoreUrl?: string;
   size?: number;
   color?: ColorToken;
 }
@@ -19,11 +21,21 @@ export function InfoTooltipIcon({
   title,
   explanation,
   ariaLabel,
+  learnMoreUrl,
   size = 16,
   color = 'ink.text-subdued',
 }: InfoTooltipIconProps) {
+  const content = learnMoreUrl ? (
+    <>
+      {explanation}
+      <LearnMoreLink destination={learnMoreUrl} />
+    </>
+  ) : (
+    explanation
+  );
+
   return (
-    <BasicHoverCard title={title} content={explanation}>
+    <BasicHoverCard title={title} content={content}>
       <styled.span
         display="inline-flex"
         alignItems="center"
--- a/apps/web/app/content/bitcoin-staking-content.ts
+++ b/apps/web/app/content/bitcoin-staking-content.ts
@@ -51,6 +51,39 @@ export const bitcoinStakingContent = {
   selfClaim: {
     explanation: `This pool takes no cut from your rewards. In exchange it never claims for you: rewards accrue each cycle and only you can claim them, through the pool's signer-manager contract.`,
   },
+  operatorPayout: {
+    badge: `Custodial BTC payout`,
+    explanation(poolName: string) {
+      return `${poolName} collects the pool's sBTC rewards and pays members in native BTC off chain, on its own schedule. There is no on-chain claim and no on-chain recourse; payment status lives with ${poolName}.`;
+    },
+    rewardsTokenCaption(poolName: string) {
+      return `Paid off chain by ${poolName}`;
+    },
+    rewardsTokenExplanation(poolName: string, cadence: string) {
+      return `Rewards reach you as native BTC at the payout address you register, sent by ${poolName} ${cadence}. The pool's sBTC rewards are collected by ${poolName} and distributed off chain, so there is no on-chain claim.`;
+    },
+    feeCaption(poolName: string) {
+      return `${poolName} policy`;
+    },
+    feeExplanation(poolName: string) {
+      return `The fee ${poolName} publishes on its own site and applies when it distributes rewards off chain. It is operator policy rather than a value enforced by the contract, so it can change without a contract change.`;
+    },
+    minStakeNote(poolName: string, minStx: string) {
+      return `${poolName} requires at least ${minStx} STX. The contract rejects anything smaller.`;
+    },
+    activeCard: {
+      title: `Rewards payout`,
+      paidBy(poolName: string) {
+        return `Native BTC, paid by ${poolName}`;
+      },
+      description(poolName: string, cadence: string) {
+        return `${poolName} pays your share ${cadence} to this address, after its fee. Payments are made off chain, so there is nothing to claim here and payment status lives with ${poolName}.`;
+      },
+      linkLabel(poolName: string) {
+        return `Visit ${poolName}`;
+      },
+    },
+  },
   poolOverviewInfo: {
     rewardsToken: `Rewards accrue as sBTC on Stacks each cycle and are claimed through the pool's signer-manager contract. Yield is variable: it depends on network-wide staking participation and the protocol reward waterfall.`,
     fee: `The share of your rewards this pool keeps. Each pool sets its own fee in its signer-manager contract, so check the pool's terms before staking.`,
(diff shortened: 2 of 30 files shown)


werner-stacks (approved):
Tested and it looks good to me, thanks for adding.

aperry-stacks (approved):

PR #2733, opened 2026-09-11
feat: add BNS testnet support

### LINKED ISSUE #2711: BNSv2 names support on the primary hiro testnet
Users would like to see this small update for easier tracking of accounts on testnet.

### What happens

On the primary Stacks testnet, Leather renders no BNS name for accounts that own testnet BNSv2 names. Mainnet rendering is unaffected.

### Why

`apps/extension/src/app/query/stacks/bns/bns-v2-client.ts` points the testnet basePath at the mainnet BNSv2 API:

```ts
const basePath = whenNetwork(bitcoinNetworkModeToCoreNetworkMode(network.chain.bitcoin.mode))({
  mainnet: BNS_V2_API_BASE_URL,
  // TODO: Add testnet support if there will be a testnet BNSv2 API
  testnet: BNS_V2_API_BASE_URL,
});
```

So on testnet the wallet asks the mainnet API for an ST address's names, gets `{"total":0,"names":[]}`, and stops there. The primary-name contract read (`fetchPrimaryName` in `packages/query/src/stacks/bns/bns.utils.ts`) never runs because it is gated on that list containing more than one name.

The TODO's condition is now met: the testnet BNSv2 API exists and is live at `https://api.bnsv2.com/testnet` (same shape as mainnet, for example `GET /testnet/namespaces`, `GET /testnet/names/address/{address}/valid`).

### Testnet chain state, for reference

BNSv2 was redeployed on the primary testnet after the chainstate reset, at the address `@leather.io/models` already pins (`bnsContractAddress.testnet`):

- BNS contract: `ST2QEZ06AGJ3RKJPBV14SY1V5BBFNAW33D9SZJQ0M.BNS-V2` (deployed 2026-08-25, `flip-migration-complete` on 2026-08-26)
- Zonefile resolver: `ST2QEZ06AGJ3RKJPBV14SY1V5BBFNAW33D9SZJQ0M.zonefile-resolver`
- 32 namespaces including `.stx`, `.btc`, `.id` with mainnet lifetimes; about 1,700 names airdropped

https://api.testnet.hiro.so/extended/v1/contract/ST2QEZ06AGJ3RKJPBV14SY1V5BBFNAW33D9SZJQ0M.BNS-V2

### Minimal fix

```ts
testnet: `${BNS_V2_API_BASE_URL}/testnet`,
```

Two related spots that keep testnet broken on other surfaces:

- `packages/services/src/infrastructure/api/bns-v2/bns-v2-api.client.ts` hardcodes the mainnet API URL with no network parameter.
- `packages/services/src/bns/bns.service.ts` `getPrimaryName` hardcodes `bnsContractAddress.mainnet` even though `bnsContractAddress.testnet` exists in the model.

### More robust alternative, if preferred

Resolve names from the Stacks node first (`get-primary` / `get-bns-from-id` read-only calls against the per-network pinned contract, as `bns.service.ts` already does for mainnet primary names) and use the BNSv2 API as the fallback and for bulk address lookups. On testnet the node-first path also removes any dependency on the testnet API's uptime.

aperry-stacks:
Adds BNSv2 testnet support to extension and resolves https://github.com/leather-io/mono/issues/2711 

This PR also includes a refactor to remove the remaining usage points of the deprecated `@leather.io/query` BNS implementation and replace with the consolidated `@leather.io/services` version.

The majority of the diff is the query version deprecation.
<hr/>
image
<hr/>
image
<hr/>
image


fab-stacks (approved):

edgar-stacks (approved):
Wow that's a nice amount of stuff deleted! LGTM

PR #2761, opened 2026-09-15
fix: swaps performance

aperry-stacks:
Stops a single unresponsive swap provider from hanging quote retrieval (>10 seconds), and disables the Velar provider.

The first commit fixed a bug in the Velar swap provider by adding a random / dummy Stacks address. This surfaced an issue with their current SDK version (`0.7.6`) where quotes would be extremely high (incorrectly), which would win the "best quote" selection and force using the Velar route.

The immediate fix is to remove Velar as a swap quote provider altogether. Longer-term we can improve quote selection logic to remove any obvious outliers such as this, and then reinstate Velar.

This PR:
- Wraps each provider quote call in a 10 second timeout, via a new `withTimeout` helper in `@leather.io/utils`, so a provider that never settles is dropped instead of blocking `Promise.allSettled` and the whole quote query.
- Removes Velar from the swap provider list. The provider classes stay wired into the service, so re-enabling is a one-line change once Velar ships a fix.
- Passes a random mainnet address as the Velar quote sender instead of an empty string, which Hiro rejected with a 400. Moot while Velar is disabled, but required if it is re-enabled.


packages/services/src/infrastructure/api/velar/velar-sdk.client.ts
```diff
@@ -33,7 +36,7 @@ export class VelarSdkClient {
     amount: number
   ): Promise<VelarSdkAmountOutResponse> {
     const swapInstance = await this.velarSdk.getSwapInstance({
-      account: '',
+      account: this.randomStacksAddress,
```
edgar-stacks, line 39: 😂

edgar-stacks (approved):

werner-stacks (approved):

PR #2763, opened 2026-09-15
feat: fully move to dmk btc/stx dmk packages

edgar-stacks:
Full migration to Ledger DMK on both stx and btc.

Tested using ledger on both stx and btc chains, signed singlesig and multisig transactions, called contracts and signed messages


fab-stacks (dismissed):

pnpm-workspace.yaml
```diff
@@ -21,6 +21,8 @@ minimumReleaseAgeExclude:
   - "@stacks/transactions"
   - "@stacks/common"
   - "@stacks/network"
+  - "@zondax/ledger-stacks"
+  - "@zondax/ledger-js"
```
aperry-stacks: We can probably remove these now from `minimumReleaseAgeExclude`

aperry-stacks (dismissed):

pnpm-workspace.yaml
edgar-stacks: good point!

aperry-stacks (approved):

fab-stacks (approved):

werner-stacks (commented):
I went through several flows, contract deployment, contract calls, transfers. LGTM.

